PR #128, opened 2014-09-25
More permissive regex for mimeType

claylo:
The regex used for mimeType is too restrictive. This change allows for values that meet the criteria described in [RFC 2045](http://tools.ietf.org/html/rfc2045) and [RFC 2046](http://tools.ietf.org/html/rfc2046).

Hat tip to @mnot for his regex, which he [posted here](http://lists.w3.org/Archives/Public/xml-dist-app/2003Jul/0064.html).

### DIFF
--- a/schemas/v2.0/schema.json
+++ b/schemas/v2.0/schema.json
@@ -177,7 +177,7 @@
     },
     "mimeType": {
       "type": "string",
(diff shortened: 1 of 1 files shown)


webron:
mimetype regex has been completely removed. Thanks for taking the time and providing the input everyone.

PR #133, opened 2014-09-29
Proposing adding SEARCH method

paulhill:
We started using the SEARCH method when we found we couldn't gracefully accommodate our document store backed (reductive find) GET /users and our elastic-search backed (additive search) SEARCH /users.
It's a draft spec but seems pretty handy. We use it in anger at any rate.


fehguy:
:+1:

webron:
I'm working on a new version on the 2.0_fixes branch. Would you like to resubmit your PR against that or should I just go ahead and make the modification myself?

ePaul:
This pull request comes too late for 2.0 (that is out for more than a year), and for 3.0 it should be a diff to the 3.0.md instead. I guess this can be closed for now, as #325 is collecting a list of methods to support?

webron:
Originally we considered adding it to 2.0 after release, but eventually didn't. As @ePaul mentioned, this can be closed now.

PR #255, opened 2015-01-29
Make clear that termsOfService should be an URL

MattiSG:
There is an ambiguity in 2.0: `info.termsOfService` is said to be a `string`, but it does not say whether this string should be the full ToS, a title, or an URL linking to it.

The example gives an URL, and `swagger-ui` now [makes this assumption](https://github.com/swagger-api/swagger-ui/pull/812/files) too.

This changeset thus states that `termsOfService` is expected to contain an URL. It does not add a normative format, for retrocompatibilty.

---

In a next release, I think `termsOfService` should expose an object similar to a `License Object` to remove any ambiguity.

### DIFF
--- a/versions/2.0.md
+++ b/versions/2.0.md
@@ -132,7 +132,7 @@ Field Name | Type | Description
 ---|:---:|---
 <a name="infoTitle"/>title | `string` | **Required.** The title of the application.
 <a name="infoDescription"/>description | `string` | A short description of the application. [GFM syntax](https://help.github.com/articles/github-flavored-markdown) can be used for rich text representation.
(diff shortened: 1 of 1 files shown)


earth2marsh:
This seems like a good fix to me. What has to happen now to resolve this PR?

webron:
The world needs to start spinning slower and complete a full circle in more than 24 hours per day ;) Extra benefit - we would feel lighter too!

ePaul:
I've created a new pull request #661 with the same change (to the 3.0 version) for OpenAPI.next. I guess this one can be closed now.

webron:
Closing this as we'll not be making the change for 2.0. We'll reiterate over #661.

PR #293, opened 2015-03-03
allowed response type of `file`

nickretallack:
Not sure if this is the best way to do it, but we should do something to support what the documentation says we do.  Comments welcome.

Not sure how to "test" this or try it out in swagger-editor.

### DIFF
--- a/schemas/v2.0/schema.json
+++ b/schemas/v2.0/schema.json
@@ -358,7 +358,7 @@
           "type": "string"
         },
         "schema": {
-          "$ref": "#/definitions/schema"
+          "$ref": "#/definitions/responseSchema"
         },
         "headers": {
           "$ref": "#/definitions/headers"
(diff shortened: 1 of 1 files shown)


webron:
If you fix it (the `responseSchema` is defined inside `schema` by mistake) and resubmit it against the 2.0_fixes branches, I'd be more than happy to merge it.

webron:
Please note that I just pushed a few changes to the schema object in that branch, so make sure to pull those.

ePaul:
If I understand right, when using `"type": "file"`, there is no much point in including anything else in this schema definition. So maybe something like

    "responseSchema": {
        "anyOf": [ {
                "$ref": "#/definitions/schema"
             }, {
                "type": "object"
                "properties": {
                    "type": {
                        "type": "string",
                        "enum": [ "file" ]
                    }
                }
            }
        ]
    }

would be shorter, easier maintainable and allow less nonsense? (I'm not an expert on JSON schema, I find the spec unreadable, and have no tools here to try if this works. I hope the intention is clear.)

webron:
Thanks, but due to code drift, had to implement it differently.

PR #320, opened 2015-04-04
Note that not all are compatible with Swagger V2.0

OwenBrotherwood:
The listing should be broken up with known Swagger V2.0 or unknown.
The listing gives an untrue idea of broad support for Swagger V2.0 causing a bit of lost time trying to find something one can use


webron:
Thank you for that. We're actually working on better organization of the list which would make it easier for the users to find what they need (including Swagger 2.0 support).

PR #324, opened 2015-04-10
Not sure if this is even possible

dilipkrish:
Having a hard time coming up with a way to generate this cleanly

### DIFF
--- a/fixtures/v2.0/json/models/modelWithEnumArray.json
+++ b/fixtures/v2.0/json/models/modelWithEnumArray.json
@@ -1,0 +1,18 @@
+{
+  "required": [
+    "id"
+  ],
+  "properties": {
+    "id": {
+      "type": "integer",
+      "format": "int64"
+    },
+    "childrensAges": {
+      "type": "array",
+      "items": {
+        "type": "string",
+        "enum": ["UNDER_10", "UNDER_15", "UNDER_18"]
+      }
+    }
+  }
+}


fehguy:
@webron ?

webron:
Other than the model missing `"type": "object"` at the top level, definition-wise it is fine. I'll merge it later since it requires a test alongside it, unless you feel like adding it to the tests as well, @dilipkrish.

dilipkrish:
I can add a test for it. Not really clear where type:object is needed. Its similar to the primitive array example

webron:
```json
{
  "type": "object",
  "required": [
    "id"
  ],
  "properties": {
    "id": {
      "type": "integer",
      "format": "int64"
    },
    "childrensAges": {
      "type": "array",
      "items": {
        "type": "string",
        "enum": ["UNDER_10", "UNDER_15", "UNDER_18"]
      }
    }
  }
}
```
Please be aware that when you generate models with springfox, you need to include that (that is, whenever you describe `properties` you should also note that it's of object type).

If you saw an example that doesn't have that, let me know and I'll fix it.

Do you need me to point me you to the test that needs to be modified to add this use case?

dilipkrish:
No that makes sense. What I meant was the [other example](https://github.com/swagger-api/swagger-spec/blob/master/fixtures/v2.0/json/models/modelWithPrimitiveArray.json) spent have it either.

webron:
Thanks, I'll fix those.

dilipkrish:
@webron is there a harness that uses this json to test? Cant find it or its not obv to me

webron:
Yup, it's https://github.com/swagger-api/swagger-spec/blob/master/src/test/scala/ModelTest.scala. Just copy one of the tests, change the text, file name and should be done.

webron:
@dilipkrish? :)

webron:
Closing for now - requires additional test and a fix to the sample itself. @dilipkrish - if you feel strongly about adding it, please resubmit.

PR #327, opened 2015-04-15
Swagger2.0: fix missing 'id' attribute in JSON schema

IvanGoncharov:


webron:
Thanks for the PR, but it's actually not missing. It's not part of the spec - https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#schema-object.

IvanGoncharov:
Yes your right, I missed this part of spec.
For me, it was 68c6777dc5eba98dc1eb791983693f089e9d3366 commit that change behavior.

But these bring very important problem, about compatibility with JSON schema.
And mean that you can't reference existing schemes inside your swagger file.
For example if make API that return swagger definition and want to reuse existing [schema](https://github.com/swagger-api/swagger-spec/blob/master/schemas/v2.0/schema.json), I can't simply use '$ref' or even copy-paste it.

Also it is problem for writing converters from other formats, for e.g. RAML and Google Discovery both support full json schema standard for payloads.
And finally it limits usage of existing json schema tooling, for e.g. you can't use json schema builders, generators or converters.

I understand, why swagger 2.0 extended json schema with additional attributes. But I don't understand the need for limiting backward compatibility.
I think adding couple additional attributes is safe change and could be done in scope of 2.0
If not I can create feature request for next version.

webron:
We try to limit changes to the spec, even if they are non-breaking. It's not really not-backwards compatible because previous versions did not use json schema at all. Here we explicitly chose a subset of json schema, can't really explain why `id` was left out.

IvanGoncharov:
By "backwards compatible" I meant: JSON Schema draft4 => Swagger 2.0

webron:
That's not really backwards compatibility :) As the spec states, we currently use an extended subset of json schema. We don't support all of it, and we have a few added attributes.

IvanGoncharov:
Because of additional attribute you can't reference Swagger inside JSON schema.
But referencing JSON schema inside swagger is really important feature and I don't see any reason why this impossible. Especially, given the fact that competing formats(RAML, Google Discovery) support these feature.
.Should I create separate feature request?

webron:
We *explicitly* don't support full JSON Schema and that's definitely not going to change for Swagger 2.0.

webron:
I'm afraid I can't merge this PR, sorry about that.

PR #352, opened 2015-05-05
Updated Uber.yaml Products endpoint.

kand617:
Also includes correct authentication info.
Updated to use correct referencing that is inline with the spec


cadesalaberry:
Thanks @kand617, I spent a lot of time tracking the `$ref: "#/definitions/Product"` that rendered as invalid on https://editor.swagger.io .

I would love this PR to be merged as well !

webron:
Sorry, missed this PR.

PR #364, opened 2015-05-14
adding Swagger2Postman to Scala section of README

josephpconley:
I've created a library called Swagger2Postman that converts a hosted Swagger doc to a Postman JSON collection.  Only supports 1.2 as of now, will try to get 2.0 done as time permits (PRs welcome too!)


webron:
Thanks! I believe it should be under the Community Driven Tools section though, no? It's not really an integration library with Scala. The development language is not relevant :)

josephpconley:
Ah good point, I didn't read that closely enough.  README updated.

PR #372, opened 2015-05-20
Guidelines

olensmar:
added example on how to external responses to the sample server app

### DIFF
--- a/guidelines/sample-server/files/models.json
+++ b/guidelines/sample-server/files/models.json
@@ -84,7 +84,7 @@
   },
   "ErrorMessage": {
     "description": "A shared model for error message",
-    "required" : ["error-code","reason"],
+    "required" : ["errorCode","reason"],
     "properties": {
       "errorCode": {
         "type": "integer",
--- a/guidelines/sample-server/files/models.json
+++ b/guidelines/sample-server/files/models.json
@@ -81,5 +81,23 @@
         "format": "double"
       }
     }
+  },
+  "ErrorMessage": {
+    "description": "A shared model for error message",
+    "required" : ["error-code","reason"],
+    "properties": {
+      "error-code": {
+        "type": "integer",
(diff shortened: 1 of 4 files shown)


guidelines/sample-server/files/models.json
```diff
@@ -81,5 +81,23 @@
         "format": "double"
       }
     }
+  },
+  "ErrorMessage": {
+    "description": "A shared model for error message",
+    "required" : ["error-code","reason"],
+    "properties": {
+      "error-code": {
```
webron: I'd avoid using dashes in property names in the examples.

olensmar: why?

webron: Which languages support dashes in variable names? The common practice is to use camel case or snake case.

olensmar: the purpose of this sample is to show how re-use works - not comply with best-practices for supporting different languages - just like other samples we have that show other features of Swagger without adhering to other "best practices"

That said - I'll happily change the name to something more to your liking :-)

guidelines/sample-server/files/models.json
```diff
@@ -81,5 +81,23 @@
         "format": "double"
       }
     }
+  },
+  "ErrorMessage": {
+    "description": "A shared model for error message",
+    "required" : ["error-code","reason"],
```
webron: This needs to be updated too ;)

guidelines/sample-server/files/models.json
fehguy: @webron who cares, it’s legal.  But because you asked…

scala> val `foo-bar` = "learn to love it”

guidelines/sample-server/files/models.json
olensmar: duh :-)

PR #378, opened 2015-05-26
Possilbe typo

sergeyromanov:
Feel free to discard if you feel that wording is fine.


webron:
Thanks for the suggestion, but the wording is as intended. Your input is much appreciated.

PR #384, opened 2015-06-01
Samples for swagger-js/#417: Separated files, yaml only for now

jasonh-n-austin:
YAML only for now, to start the discussion. Once we have a base set of examples, I can generate JSON and update the spec with these samples.


webron:
@jasonh-n-austin - couple of thoughts before merging:

- Since you refer to `petstore-separate-common.yaml` (for the `../...` ref), maybe all should be moved into another level of a sub-directory.
- Unfortunately, the YAML examples are not too good, and need to be modified. Would be great if you could fix it already in your generated example:
  - All model definitions should have `type: object` in them.
  - The composition sample is bad because the NewPet redefines the same property as adjacent Pet. You don't have to change this one, but it would be nice.
- The tests are failing :) Can't merge it without a fix to those.

jasonh-n-austin:
I was thinking a sub-dir was a better idea too.
I assumed this wasn't in mergeable state yet, wanted to get something started to give @fehguy some test data for swagger-js/#417.

Great feedback, I'll shape things up this afternoon.

jasonh-n-austin:
@webron moved separate files into their own directory. For now, added exclusion to YAML section of `validate.js` to skip tests for schema files in `petstore-separate`...not really positive how to best run schema validator against those (notably because `parameters.yaml` is a swagger fragment, and the schema files are really JSON Schema). Now tests are at least validating that `petstore-separate/spec/swagger.json`

jasonh-n-austin:
Will look at `NewPet`/`Pet`/`oneOf` fix now.

webron:
Ok, thanks.

jasonh-n-austin:
@webron  Just pushed fixed example for `petstore-expanded(.json&&.yaml)`. Now `NewPet` is the base(`id` only), extended by `Pet`.

webron:
Looks great, thanks! Could I just trouble you to add `"type": "object"` to Pet, NewPet and Error in the json variant?

jasonh-n-austin:
@webron OK added `type: object` to existing JSON examples.
Also:
* Generated JSON versions of `petstore-separate`
* Test fixes for `json/petstore-separate` exclusions

webron:
Great, thanks for all the work.

PR #390, opened 2015-06-08
Add api-spec-converter to tools list

bobby-brennan:
Would love to add our tool if you think it fits. Thanks!


webron:
Sorry for the delay. We've moved the list to https://github.com/swagger-api/swagger.io/blob/wordpress//integrations/open-source.md. Feel free to resubmit there at the appropriate section.

PR #391, opened 2015-06-08
link to swagger 2.4 play port.

matiwinnetou:


webron:
@matiwinnetou - thanks for the work (as @fehguy also mentioned in the email). For now, the list has moved to https://github.com/swagger-api/swagger.io/blob/wordpress/integrations/open-source.md so feel free to resubmit there (notice there's a change in structure, make sure you put it under the older implementations).

PR #399, opened 2015-06-24
Updated data type param defaultValue to default

EtienneBertou:
defaultValue is not accepted as a data type parameter, but default is.


webron:
In 1.2 it was `defaultValue`, not `default`. You can take a look at http://petstore.swagger.io/api/api-docs/pet.

Issue #214, opened 2014-11-20
Both HTTP and HTTPS ports cannot be specified


prantlf:
I have an API exposed on custom ports:

    http://server:8000/api
    https://server:8443/api

When I test the API with https://editor.swagger.io, I can switch the protocol from HTTP to HTTPS on the page, but because the port is entered only once with the host together and used for both schemes, I cannot actually send the request to the correct HTTPS endpoint:

    host: server:8000
    basePath: /api
    schemes:
      - http
      - https

Why not adding an additional property configuring the HTTP and HTTPS ports?

earth2marsh:
:+1:

mohsen1:
My opinion: 

We should make it possible to provide a list of hosts. Swagger should not be deterministic which host should be used for which scheme.

fehguy:
Is it unreasonable to expect that https is _aleays_ 443?

webron:
Yes. The OP.

casualjim:
I'm all for scheme: port type of mapping

while swagger assumes it's http there is also wss in there and soon there is server push and so on with http2. 
Now there is no reason why somebody couldn't expose the swagger api over a rabbitmq messsage bus or some zeromq interface which supports these messaging paradigms.

the host can be assumed to be the same but a scheme represents more often than not a port.

olensmar:
I agree - I think it would be insanely cool (almost) if we could open up for using swagger to describe APIs exposed over other protocols than HTTP(s) - (I'm primarily thinking CoAP, MQTT, WebSockets)

webron:
And pigeons.

fehguy:
@olensmar have you seen this project?

https://github.com/swagger-api/swagger-socket

Unbelievable potential with this.

ndhar-1:
Just out of curiosity, when you say you want to expose the APIs on both http and https, you intend to accept both secured and unsecured connections? I think it defeats the purpose of having security ... ?

prantlf:
Well, discussing security is out of the scope of this issue, isn't it? :-)

I can't know for what purpose is the API on the p[articular port used, how the connection is allowed the access to and how it is secured.  All I deal with is that the server offers both schemes.  I can either document and test both of them with Swagger, or I can use only one there.

webron:
Parent: #560.

fmvilas:
@olensmar you might be interested in the AsyncAPI specification: asyncapi.com.

SergioArrighi:
+1
[...]

PR #420, opened 2015-07-22
Fix 'host' regex and format: uri conflict

hornc:
I believe the `"format": "uri"` contradicts the regex pattern below in the `host` specification.

RFC 2396 URI require a scheme (e.g. 'http') but the swagger schema documentation https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#swaggerHost and the `pattern` regex specifically exclude schemes.

The description 'fully qualified URI' is therefore incorrect.

I've removed `"format": "uri"` and updated the description to better explain the field and gave an example. I didn't mention the optional `:<port number>` as I thought that might be too much detail for a description, but maybe that would be helpful too?

Before making this change I tested the json-schema `"format": "hostname"` in case that gave the desired validations, but that hostname does not accept port numbers, so this looks like it needs to be a custom format as it is NOT a rfc1034 compliant hostname as referenced from the json-schema docs, http://json-schema.org/latest/json-schema-validation.html#anchor114

### DIFF
--- a/schemas/v2.0/schema.json
+++ b/schemas/v2.0/schema.json
@@ -27,9 +27,8 @@
     },
     "host": {
       "type": "string",
-      "format": "uri",
       "pattern": "^[^{}/ :\\\\]+(?::\\d+)?$",
-      "description": "The fully qualified URI to the host of the API."
+      "description": "The host (name or ip) of the API. Example: 'swagger.io'"
     },
     "basePath": {
       "type": "string",


webron:
@hornc - thanks for the PR. 

I don't mind the description to include the optional port number as well, but since it's mentioned in the spec it's not too bad.

FWIW, back when creating the schema I checked with the people behind JSON Schema regarding the meaning of the `uri` format. It feels like it is not well defined, and some see it as a full URI and some see it as a partial URI (which seems to be the option agreed upon more). Based on that, I used the `uri` format. However, it looks like some validators don't respect that decision, so I'm fine with removing the format.

hornc:
Thanks @webron. I thought I linked to all the waypoints in my investigations above, but I seem to have missed the URI specific ones, which are a series of hops between documents.

The json-schema docs say of the `uri` format: "A string instance is valid against this attribute if it is a valid URI, according to [RFC3986]." [source] (http://json-schema.org/latest/json-schema-validation.html#anchor123)

As I interpreted [RFC3986](https://www.ietf.org/rfc/rfc3986.txt) a "URI" must have the scheme, although that document does mention "URI references" which seem to be the things that are more flexible and may be what you mean by a partial URI, which do sound like they are more commonly used, see these [examples](https://en.wikipedia.org/wiki/Uniform_resource_identifier#Examples_of_URI_references) on wikipedia.

The above is just to let you know my investigation process when I tried to make sense of it all, there were too many document sections to cross reference, I meant to include them all in the PR description!

webron:
@hornc - thanks for the details. It's easy to miss things when writing a specification (json schema, swagger), and some things are left open for interpretation to some extent. As far as I'm concerned, I'm not looking for the most strict schema, but rather something that helps and guides. Eventually, the source of truth is the spec itself and not the schema (especially since some parts of the spec cannot be covered by the schema). Again, thanks for taking the time to investigate this issue and pinpoint its source.

testn:
- host still does not allow port. Should port become its own field for clarity?
- ipv6 does not seem to be supported for host field either. How should it be specified? With bracket or not?

webron:
@testn - just tested it and it looks like the host does allow a port. How did you test it to see something else? As for ipv6 - you are correct. Can you open a separate issue on it? I'll try resolving it as soon as I can.

testn:
I used the regex pattern provided in the api. Is it correct?

webron:
@testn - yeah, the escaping got me too several times in the past, no worries. Thanks for opening the ipv6 issue.

testn:
Also the spec is not clear that host actually can contain port information as well even though it is confusing when you can do http/https at the same as mentioned in #214

webron:
From https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#swaggerHost - 

> The host (name or ip) serving the API. This MUST be the host only and does not include the scheme nor sub-paths. It MAY include a port.

And yes, as you mentioned #214 is a known issue and we'll hopefully deal with it better in the next version of the spec.

PR #421, opened 2015-07-22
Sharpen parameter "default" property description

earth2marsh:
As discussed on IRC


webron:
Thanks @earth2marsh!

The definition of `default` is repeated several times in the spec. Can you please update all descriptions?

earth2marsh:
@webron with pleasure. AFAICT, that means `itemsDefault` and `headerDefault`? Anything else?

webron:
Those two, yeah.

earth2marsh:
LMK if I missed anything!

webron:
Thanks a bunch!

tadhgpearson:
Thanks for this clarification.
What now replaces the 'other' value of default, where it's also used as the default pre-populated value for a field? I'm particularly thinking of Smartdocs here @earth2marsh  ;)

PR #424, opened 2015-07-26
Clarify operationId descriptions

earth2marsh:
Resolves swagger-api/swagger-spec#423


mohsen1:
` therefore, it is recommended to follow common programming naming conventions` is too loose and doesn't help anyone. I think if we can put a pattern that disallow spaces and dash (`-`) and other weird characters it would be safe to use `operationId` as an identifier in code.

earth2marsh:
@mohsen1 I sympathize with your statement, but @webron pointed out in swagger-api/swagger-spec#423, the spec says, `The text says MAY, not SHOULD`. Hopefully this at least gives more clarity and guidance about its purpose.

PR #426, opened 2015-07-27
Use correct schema reference

joelwurtz:
Actually the additionalProperties contain a reference to the original json schema, with this change additionalProperties should correctly contain the reference to the updated json schema used by Swagger.


webron:
Thanks!

PR #428, opened 2015-07-31
[WIP] Migrate unit-tests to JS

IvanGoncharov:


package.json
```diff
@@ -11,22 +11,22 @@
...
+    "jsonschema": "^1.0.2",
     "map-stream": "^0.1.0",
-    "mocha": "^1.21.3",
-    "q": "^1.0.1",
-    "request": "^2.39.0",
-    "z-schema": "^2.4.9"
+    "mocha": "^2.2.5",
+    "q": "^1.4.1",
+    "request": "^2.60.0",
+    "z-schema": "^3.12.0"
```
fmvilas: I would suggest to use fixed versions, just to avoid future problems. We have to be aware that not everybody is respecting SEMVER, sometimes because of a bug, sometimes because they just don't care :smile:

IvanGoncharov: Every lib deserve "presumption of innocence", so IMHO version should be fixed only if author explicitly ignore SEMVER. Because if you can't trust no one, what is the sense of SEMVER?

fmvilas: You're right. Just talking about my experience.

Issue #239, opened 2015-01-09
Links in @API Annotations


ryangyoung:
For documentation purposes, it would be wonderful if we could add href-type links in Swagger annotations. The purpose of these links would be to reference one API call from another for documentation completeness (not for documenting the flow, per se). Similar to the "transfers API" link here: https://stripe.com/docs/api#balance_object

In the annotations, it would be something like:

```@POST
    @Timed
    @ApiOperation(value = "Request an access token",
            notes = "This method exchanges the code returned in the [link to Auth resource].",

webron:
@ryangyoung - following our discussion in the google group, thanks for opening this issue.

I'm changing the title of the issue since it refers to swagger-core specifically, and we may need/want to address this in the spec itself (which is why the issue is here and not on swagger-core).

The idea is to add support for hyperlinking between operations within the description, to allow for references (not flow).

This may mandate the `operationId` becoming mandatory again, or otherwise, require it if hyperlinking is used. Of course, other suggestions would be appreciated.

ekampp:
It seems most likely that people who are using hyperlinking already implements the `operationId`. And I don't see a hinderance to listening the the hash change in the url, and opening the matching operation when it changes?

MattSANU:
I would also like to generate links to related URLs/resources solely from the part of the Swagger specification that describes the URL currently being accessed. I think the OP here and I share an intention to use Swagger and HATEOAS to create a truly RESTful API.

However rather than decorating paths and/or operations in the way that this issue proposes, I would prefer to permit an optional additional 'links' object inside paths and/or operations. See #440 for details.

MattSANU:
Just to be clear, I am not using Swagger to document an existing API - I'm using it to actually implement the API. I have Python WSGI middleware and applications which configure themselves according to a Swagger specification they load at runtime. This works well so far, but if I want my web service to be RESTful, in responses, I need to include links to other possibly-interesting URLs (so-called HATEOAS). To accomplish this, I either need to hard-code links into the application's responses (ugh), maintain information in an additional DSL outside Swagger which inter-relates the Swagger paths and operations (and keep the two DSL files in sync somehow, not fun), or (my preferred option) somehow teach Swagger to understand that paths and operations can have relationships to other paths and operations.

webron:
Parent: #560.

fehguy:
parent issue #586

tycho01:
[...]

PR #440, opened 2015-08-14
Permit a new top-level entry 'pathDefinitions', capable of containing…

MattSANU:
… a list of Path Items.

At the moment, one can refer to an external path definition using a reference object, eg:

    "paths": {
        "/icecream": {
        	"$ref": "example.com/#/pathDefinitions/icecream"
        }
    }

However, the reference must be to a definition located in an external document. It cannot refer to the referring document (using '#/' syntax), because there is nowehere in the Swagger API spec other than the 'paths' item where it is legal to place a Path Item object.

In other words, this change permits the previously-illegal:

    "pathDefinitions": {
        "IceCreamPath": {
            "get": {
                "summary": "Obtain ice cream"
            }
        }
    }
    "paths": {
        "/icecream": {
        	"$ref": "#/pathDefinitions/IceCreamPath"
        }
    }

This is consistent with other aspects of Swagger including parameters, definitions and schemas.

Once we can refer to existing path definitions by name like this, we can add high-level semantic information to the API specification by cross-linking related paths and/or operations - perhaps something like:

    "pathDefinitions": {
        "IceCreamPath": {
            "get": {
                "summary": "Obtain ice cream"
            },
            "links": [
                {
                    "rel": "self",
                    "href": {
                        "$ref": "#/pathDefinitions/IceCreamPath"
                    }
                },
                {
                    "rel": "topping",
                    "href": {
                        "$ref": "#/pathDefinitions/ToppingPath"
                    }
                }
            ]
        },
        "ToppingPath": {
            "get": {
                "summary": "Obtain chocolate topping"
            }
        }
    },
    "paths": {
        "/icecream": {
            "$ref": "#/pathDefinitions/IceCreamPath"
        },
        "/topping": {
            "$ref": "#/pathDefinitions/ToppingPath"
        }

### DIFF
--- a/schemas/v2.0/schema.json
+++ b/schemas/v2.0/schema.json
@@ -52,6 +52,9 @@
     "definitions": {
       "$ref": "#/definitions/definitions"
     },
+    "pathDefinitions": {
+      "$ref": "#/definitions/paths"
+    },
     "parameters": {
       "$ref": "#/definitions/parameterDefinitions"
     },


webron:
@MattSANU - thank you, but this is not how we operate. Spec changes are done once in a while in an organized process, and not just by submitting PRs to change the structure of the current spec. If you feel something needs to be changed, you're more than welcome to open an issue, describe your use case, what you're trying to solve and the necessity of it. If the feature is important enough, and gains enough traction from other users, it may get in a future version of the spec.

MattSANU:
Understood. I believe issue #239 already contains everything necessary - I share the same use case as the creator of that issue, which I said in that issue. Basically, we both want to implement HATEOAS.

webron:
@MattSANU - thanks for understanding. Your input is valuable, and it definitely helps us shape the future of the spec. Feel free to add more comments, open new feature requests and so on.

PR #450, opened 2015-08-28
Add type values to Security Scheme Object

lennybacon:
To support Digest and Negotiate (for Kerberos) it would be nice if the type values would be added to the specification.

### DIFF
--- a/versions/2.0.md
+++ b/versions/2.0.md
@@ -2124,7 +2124,7 @@ Allows the definition of a security scheme that can be used by the operations. S
 ##### Fixed Fields
 Field Name | Type | Validity | Description
 ---|:---:|---|---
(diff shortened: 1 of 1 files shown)


webron:
@lennybacon - thanks but there's a process we follow to updating the spec, and it's not by submitting a PR. 2.0 is what it is and will not change. Any proposals to be considered for a future version should be submitted as tickets with reasoning behind them (though it's worth checking whether there are already exiting tickets).

lennybacon:
Thanks so I'll move it to an issue.

lennybacon:
It would be nice to write down the *contribution guidelines* to the readme file. Should I open another Issue for that?

webron:
@lennybacon - yes please, that would be great. Thanks for understanding.

PR #462, opened 2015-09-14
Fixed missing types in JSON-Schema examples

RomanGotsiy:
Fixed missing `"type": "object"` in schemas examples. Without specifying `type` schema will validate to anything (string, number, bool, etc)


RomanGotsiy:
@webron, can your review my changes? They are pretty simple. People continue using examples with mistakes.

webron:
Thanks for that!

PR #464, opened 2015-09-17
Generate TOCs for specification docs

scottydawg:
This change adds a gulp task for generating a variant of the markdown specification with a table of contents (using the [DocToc](https://github.com/thlorenz/doctoc) library).

The swagger specification document is a bit unwieldy as a reference point because:

- It has no ToC to provide a high-level context of what is contained in the specification, and
- The styles used on http://swagger.io/specification/ visually deemphasize h4 elements; h5 elements are more pronounced. This makes the hierarchy difficult to navigate.

This change doesn't address styling issues on swagger.io, but does allow generation the specification document with a ToC.

### DIFF
--- a/docs/1.2.md
+++ b/docs/1.2.md
@@ -1,0 +1,1195 @@
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*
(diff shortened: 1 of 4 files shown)


ePaul:
Could you rebase your Pull request against current master (or maybe even better, current OpenAPI.next)? There the specs are in the `versions` folder.

webron:
We can consider this for 3.0, but we shouldn't make the changes to 1.2 and 2.0.

webron:
Closing as we'll be introducing only must-have updates to the specs (typos or 'bug' fixes).

PR #473, opened 2015-10-06
Fixed invalid urls in v1.2 specifications

georgthegreat:
swagger-api.github.io isn't here any more


webron:
Thanks. Will create a new PR to move to the newer URLs.

PR #477, opened 2015-10-09
2.0: stricter validation of additionalProperties

dolmen:
Following [clarification of `additionalProperties` for a schema](/swagger-api/swagger-codegen/issues/1318#issuecomment-146302879) by @webron:
- we now accept only `false` or a schema for `additionalProperties`: `true` is now rejected.
- `false` is the new default for `additionalProperties` (instead of `{}` which means "anything is accepted" in JSON Schema).

### DIFF
--- a/schemas/v2.0/schema.json
+++ b/schemas/v2.0/schema.json
@@ -991,10 +991,12 @@
               "$ref": "#/definitions/schema"
             },
             {
-              "type": "boolean"
+              "description": "false",
+              "type": "boolean",
+              "enum": [ false ]
             }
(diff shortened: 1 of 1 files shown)


dolmen:
After more thinking about [@webron comment](/swagger-api/swagger-codegen/issues/1318#issuecomment-146302879) I think I got it wrong and my patch is incorrect.

webron:
@dolmen - yeah, your second assertion is probably the better one.

dolmen:
Closing as this must not be merged as is.

PR #478, opened 2015-10-09
Add .mailmap to consolidate authorship (git shortlog -e)

dolmen:
Cc: @webron @fehguy @xhh @mohsen1 @silas @wting

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,12 @@
+Mohsen Azimi <me@azimi.me> Mohsen Azimi <mazimi@apigee.com>
+Mohsen Azimi <me@azimi.me> Mohsen Azimi <msn.azi@gmail.com>
+Ron Ratovsky <webron@gmail.com> Ron <webron@users.noreply.github.com>
+Ron Ratovsky <webron@gmail.com> webron <webron@users.noreply.github.com>
+Ron Ratovsky <webron@gmail.com> Ron <webron@gmail.com>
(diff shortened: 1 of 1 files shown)


wting:
Looks good to me, my emails are mapped correctly.

RobDolinMS:
I'm not familiar with mailmap, but I did a bit of digging and found documentation at: https://www.git-scm.com/docs/git-check-mailmap

RobDolinMS:
@dolmen @wting - It looks like this consolidates on people's personal email addresses rather than their work email addresses.  Is this intentional?  (Am I possibly misinterpreting the file?)

webron:
It's based on the commits. I don't really see a need for this.

RobDolinMS:
@namdeirf Do we have a Bitergia (sp?) account for OAI?

If so, does it use the .mailmap file or does it use something else?

RobDolinMS:
@webron I would be inclined to close this for now; and then re-open IF there is some additional need (such as my [comment to JeffEF](https://github.com/OAI/OpenAPI-Specification/pull/478#issuecomment-282844610)

.mailmap
```diff
@@ -0,0 +1,12 @@
+Mohsen Azimi <me@azimi.me> Mohsen Azimi <mazimi@apigee.com>
+Mohsen Azimi <me@azimi.me> Mohsen Azimi <msn.azi@gmail.com>
```
mohsen1: Please remove my name and email. I get enough emails for Swagger stuff already! 😁

webron:
Closing for now. Thanks for the work though.

PR #482, opened 2015-10-14
fix typo in spec 2.0

sebdiem:


webron:
Thanks for this, but that's actually not a typo :) It really means to say 'parameters is in path'. Your change modifies the meaning of the sentence suggesting that it applies only for parameters called "in".

sebdiem:
oups! ok my bad :) In fact I did not understand the sentence at first. Sorry about that

PR #486, opened 2015-10-16
schema 2.0: fix definitions of 'consumes'/'produces'

dolmen:
A JSON reference ($ref) is a JSON object containing ONLY the `$ref` property. The [specification (still a draft) for JSON references](https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03#section-3) says explicitely that any other property must be ignored:

> Any members other than "$ref" in a JSON Reference object SHALL be ignored.

In the schema for Swagger 2.0 the properties `consumes` and `produces` were not respecting this contraint. This is fixed by using an `allOf` schema that combines the `{ "$ref": ...}` and an other object containing the other properties.

### DIFF
--- a/schemas/v2.0/schema.json
+++ b/schemas/v2.0/schema.json
@@ -39,12 +39,24 @@
       "$ref": "#/definitions/schemesList"
     },
     "consumes": {
-      "description": "A list of MIME types accepted by the API.",
-      "$ref": "#/definitions/mediaTypeList"
+      "allOf": [
+        {
+          "$ref": "#/definitions/mediaTypeList"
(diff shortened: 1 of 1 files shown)


webron:
Thanks, but this is an unnecessary change. While the description will be 'ignored', it's there for internal documentation purposes only. This will just bloat the schema and make it more difficult to read. It has no effect on schema validation correctness.

fehguy:
let's move this into 3.0 but not apply to 2.0

webron:
Closing as won't fix for 2.0.

PR #492, opened 2015-10-22
Switch mention of "Vendor Extensions" to "Specification Extensions"

earth2marsh:
… includes rename the anchor as well.

### DIFF
--- a/versions/2.0.md
+++ b/versions/2.0.md
@@ -125,7 +125,7 @@ Field Name | Type | Description
 
 Field Pattern | Type | Description
 ---|:---:|---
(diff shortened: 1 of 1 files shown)


webron:
@earth2marsh - thanks, but we've been using the term vendor extensions since the work group on 2.0 and always referred to them as that. I think changing the terminology right now would be too confusing.

I'd suggest open a ticket instead so that we'll remember to possibly use your suggestion in the next version of the spec.

ePaul:
#657 is the new pull request for OpenAPI.next.

fehguy:
Not updating the 2.0 documentaiton

PR #500, opened 2015-10-30
use string keys for HTTP status codes in YAML

rutsky:
File that corresponds to Swagger specification must be correct JSON
file.
JSON doesn't allow integer keys.
YAML allows integer keys and parses "200: test" as {200: "test"} object.

### DIFF
--- a/versions/2.0.md
+++ b/versions/2.0.md
@@ -292,7 +292,7 @@ Field Pattern | Type | Description
     produces:
     - application/json
     responses:
-      200:
+      '200':
         description: A list of pets.
         schema:
           type: array
@@ -380,7 +380,7 @@ get:
   - application/json
   - text/html
   responses:
-    200:
+    '200':
       description: pet response
       schema:
         type: array
@@ -515,9 +515,9 @@ parameters:
   required: false
   type: string
 responses:
-  200:
+  '200':
     description: Pet updated.
-  405:
+  '405':
     description: Invalid input
 security:
 - petstore_auth:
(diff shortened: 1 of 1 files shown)


webron:
Thanks for this, but it's actually not necessary. Numeric keys will be converted to strings when converting to JSON. Try it out here - http://codebeautify.org/yaml-to-json-xml-csv.

rutsky:
> Numeric keys will be converted to strings when converting to JSON.

No they shouldn't (according to YAML specification).

They are converted to strings in JavaScript library because in JavaScript all keys are strings and access to non-string key leads to implicit conversion to string (e.g. `obj[0]` is equivalent to `obj["0"]`).

In other languages this may not be the case, e.g. in Python integer and string keys are different and loaded YAML document with integer keys will actually have integer keys:

```python
>>> import yaml
>>> obj = yaml.load('200: ok')
>>> obj
{200: 'ok'}
>>> obj[200]
'ok'
>>> obj['200']
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
KeyError: '200'
>>>
```

ingydotnet:
YAML processors generally convert unquoted scalars to what makes the most sense in their implementation (which is often language guided). Since YAML doesn't yet have a schema language (or at least one that all implementations follow) the schema for what data actually becomes is baked into the code of the YAML processor being used.

The smart move here would be to quote the numeric keys if this YAML is going to be loaded into a processor that supports numeric mapping keys.

webron:
@ingydotnet Thanks for the input, will follow with your suggestion.

PR #504, opened 2015-11-07
Correct type field description in items object

samskiter:


webron:
I'm afraid `"object"` is not allowed there.

samskiter:
Are you sure? An array of objects seems a pretty basic thing to support and is supported as far as I could tell from the parsers/generators I'd used.
In fact there are arrays with refs to items with type set to object in the swagger example on the editor http://editor.swagger.io/#/

webron:
The `Items Object` is the one used by non-Schema Object arrays, where objects are not allowed. For example, all non-body parameters. Check the spec carefully as to what points to the `Items Object`.

ePaul:
#654 will actually make a schema allowed for all parameters in 3.0, which makes this section go away.

I guess this change for the 2.0 version can be closed without merging.

webron:
This can be closed now as it's for 2.0.

PR #528, opened 2016-01-05
Fixing license in schema v2.0 package.json.

jsdevel:


webron:
If anything, it should be fixed to APL2.0 ;) but thanks.

jsdevel:
Just making it conform to `bower.json` and the `README`

jsdevel:
I can switch them all to `APL` if you prefer.

webron:
That would be great, thanks. That's the official license intended to be used. And to make sure, the intent is the Apache 2.0 License, not the Adaptive Public License.

jsdevel:
done

Issue #388, opened 2015-06-05
Improve explanation of Reference Object


jharmn:
With the addition of relative file references (notably in swagger-js, and samples), there should be an explanation of how that resolves (based on JSON Reference). `Samples/petstore-separate` provides a reference point that could be utilized.

jasonh-n-austin:
On further review, the REUSE guidelines should be updated too:
https://github.com/swagger-api/swagger-spec/blob/master/guidelines/REUSE.md

webron:
@jasonh-n-austin - are you referring specifically to how relative references are resolved?

jasonh-n-austin:
Right, so perhaps examples of what `$ref` links should/shouldn't look like, and what they should be relative to in pathing.

One example:
```
swagger.yaml
$ref: 'models/Thing.yaml'
```
```
Thing.yaml
$ref: 'models/AnotherThing.yaml' # note path is relative to swagger.yaml, not Thing.yaml
```

Also, deep links such as:
```
$ref: 'models/Common.yaml#Address' # an object within a schema file
```

All of this is really just how JSON Reference works, but some practical examples would probably help devs implementing the Swagger 2 spec get it right more consistently.

john-kurkowski:
:+1:

mission-liao:
@jasonh-n-austin  Hi, I'm refactoring my python client for external file reference, and feel confused with these:
```yaml
swagger.yaml
$ref: 'models/Thing.yaml'

Thing.yaml
$ref: 'models/AnotherThing.yaml' # note path is relative to swagger.yaml, not Thing.yaml
```

In another [discussion](https://github.com/swagger-api/swagger-js/issues/417), one of your comment:
> Each file resolves it's own refs, regardless where they're included from. In the swagger.yaml->Category->Person example, the $ref in Category is resolved from it's own location to find Person (as opposed to the location of swagger.yaml).

In the first comment, the base URI used to resolve 'models/AnotherThing.yaml' should be derived from __swagger.yaml__, although the $ref resides in Thing.yaml. But the comment in another issue states that the base URI should be derived from __Thing.yaml__.

Which one of my understanding is correct? Thanks for any feedback in advance. :+1:

whitlockjc:
I would expect relative references to be based on the location of the document containing the reference.

jasonh-n-austin:
[...]

PR #535, opened 2016-01-08
Refined guidelines for external file reuse

jasonh-n-austin:
Fairly big refactor of this section, to address #388 
* Renamed mentions of `Swagger` to `OpenAPI Specification` (I presumed we're doing this on any updates to docs now)
* Added overall guidance for external files in dedicated section
  * Separated external files by URL from 'relative', i.e. local file references

### DIFF
--- a/guidelines/REUSE.md
+++ b/guidelines/REUSE.md
@@ -7,17 +7,54 @@ We encourage reuse and patterns through references.
 The following types are reusable, as defined by the spec:
 
 * Parameters
-* Models (or Schema Objects in general)
+* Models (_or Schema Objects in general_)
 * Responses
 * Operations (_Operations can only be remote references_)
 
 ## Reuse strategy
 
-When reusing components in an API design, a reference is created from the definition to target entity. The references are maintained in the structure, and can be updated by modifying the source definitions. This is different from a "copy on design" approach where references are injected into the design itself.
+When authoring API design documents, common object definitions can be utilized to avoid duplication. For example, imagine multiple path definitions that each share a common path parameter, or a common response structure. The OpenAPI specification allows reuse of common object definitions through the use of "references".
 
-The reuse technique is transparent between JSON or YAML and is lossless when converting between the two.
+A reference a construct in your API design document that indicates "the content for this portion of the document is defined elsewhere". To create a reference, at the location in your document where you want to reuse some other definition, create an object that has a `$ref` property whose value is a URI pointing to where the definition is (more on this in later sections). 
 
-YAML anchors are technically allowed but break the general reuse strategy in OpenAPI Specification, since anchors are "injected" into a single document.  They are not recommended.
+OpenAPI's provides reference capabilities using the [JSON Reference](https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03) specification. 
+
+### JSON Example
+
+``` json
+{
+  // ... 
+  definitions: {
+    Person: {
+      type: 'object',
+      properties: {
+        friends: {
+          type: 'array',
+          items: {
+            $ref: '#/definitions/Person'
+          }
+        }
+      }
+    }
+  }
+}
+```
+
+### YAML Example
+
+``` yaml
+# ...
+definitions:
+  Person:
+    type: object
+    properties:
+      friends:
+        type: array
+        items:
+          $ref: '#/definitions/Person'
+```
+
+Note: YAML has a very similar feature, [YAML anchors](http://yaml.org/spec/1.2/spec.html#id2765878). Examples from this point will only be in JSON, using JSON References.
 
 ## Techniques
 
@@ -46,6 +83,7 @@ An example of a local definition reference:
 
 _Example from https://github.com/OAI/OpenAPI-Specification/blob/master/examples/v2.0/json/petstore.json_
 ``` json
+          // ... 
           "200": {
             "description": "pet response",
             "schema": {
@@ -65,6 +103,7 @@ Files can be referred to in relative paths to the current document.
 _Example from https://github.com/OAI/OpenAPI-Specification/tree/master/examples/v2.0/json/petstore-separate/spec/swagger.json_
 
 ``` json
+// ... 
 "responses": {
 	"default": {
 		"description": "unexpected error",
@@ -79,6 +118,7 @@ Remote references may also reference properties within the relative remote file.
 
 _Example from https://github.com/OAI/OpenAPI-Specification/tree/master/examples/v2.0/json/petstore-separate/spec/swagger.json_
 ``` json
+// ... 
 "parameters": [
 	{
 		"$ref": "parameters.json#/tagsParam"
@@ -233,7 +273,7 @@ To include these parameters, you would need to add them individually as such:
 
 ### Operations
 
-Again, Operations can be shared across files.  Although the reusability of operations will be less than with Parameters and models. For this example, we will share a common `health` resource so that all APIs can reference it:
+Again, Operations can be shared across files.  Although the reusability of operations will be less than with Parameters and Definitions. For this example, we will share a common `health` resource so that all APIs can reference it:
 
 Refer to [Guidelines for Referencing](#guidelines-for-referencing) for referencing strategies.
 
--- a/guidelines/REUSE.md
+++ b/guidelines/REUSE.md
@@ -1,38 +1,65 @@
 # Reuse Philosophy
+
 We encourage reuse and patterns through references.
 
 ## What is reusable
+
 The following types are reusable, as defined by the spec:
 
-* Operations
 * Parameters
-* Responses
 * Models (or Schema Objects in general)
+* Responses
+* Operations (_Operations can only be referenced externally_)
 
 ## Reuse strategy
+
 When reusing components in an API design, a pointer is created from the definition to target design.  The references are maintained in the structure, and can be updated by modifying the source definitions.  This is different from a "copy on design" approach where references are injected into the design itself.
 
 The reuse technique is transparent between JSON or YAML and is lossless when converting between the two.
 
-YAML anchors are technically allowed but break the general reuse strategy in Swagger, since anchors are "injected" into a single document.  They are not recommended.
+YAML anchors are technically allowed but break the general reuse strategy in OpenAPI Specification, since anchors are "injected" into a single document.  They are not recommended.
 
-Referenes can be made either inside the Swagger definition file or to external files. References are done using [JSON Reference](http://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03).
+Referenes can be made either internal to the OpenApi Specification file or to external files. 
 
 ## Techniques
 
 ### Guidelines for Referencing
 
+Whether you reference definitions internally or externally, you can never override or change their definitions from the referring location. The definitions can only be used as-is.
+
+#### Internal references
+
 When referencing internally, the target references have designated locations:
 
-* Parameters -> `parameters`
-* Responses -> `responses`
-* Models (and general Schema Objects) -> `definitions`
+* Parameters -> `#/parameters`
+* Responses -> `#/responses`
+* Models (and general Schema Objects) -> `#/definitions`
 
-Operations can only be referenced externally.
+All references are canonical and must be a qualified [JSON Pointer, per RFC6901](http://tools.ietf.org/html/rfc6901). For example, simply referencing a model `Pet` is not allowed, even if there are no other definitions of it in the file.
 
-An example for an internal reference - `#/definitions/MyModel`. All references are canonical and must be a qualified [JSON Pointer](http://tools.ietf.org/html/rfc6901). For example, simply referencing `MyModel` is not allowed, even if there are no other definitions of it in the file.
+_Example from https://github.com/OAI/OpenAPI-Specification/blob/master/examples/v2.0/json/petstore.json_
+``` json
+          "200": {
+            "description": "pet response",
+            "schema": {
+              "type": "array",
+              "items": {
+                "$ref": "#/definitions/Pet"
+              }
+            }
+```
 
-When referencing externally, use a valid URI as the reference value. If your referenced file contains only one definition that should be used, you can refer to the file directly. For example:
+#### External references
+
+If your referenced file contains only one root-level definition, you can refer to the file directly.
+
+To reference the example below: 
+
+`"$ref": "definitions/Model.json""`
+
+or
+
+`"$ref": "https://my.company.com/definitions/Model.json"`.
 
 _Assuming file https://my.company.com/definitions/Model.json_
 ```json
@@ -47,7 +74,16 @@ _Assuming file https://my.company.com/definitions/Model.json_
 }
 ```
 
-The reference would be `https://my.company.com/definitions/Model.json`.
+To reference `Model` in the example below:
+
+_Note this approach potentially combines URL, JSON Reference, and JSON Pointer_
+
+`"$ref": "definitions/models.json#/models/Model"`
+
+ or 
+ 
+`"$ref": "https://my.company.com/definitions/models.json#/models/Model"`
+
 
 _Assuming file https://my.company.com/definitions/models.json_
 ```json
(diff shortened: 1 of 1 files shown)


guidelines/REUSE.md
```diff
@@ -133,8 +209,20 @@ _Assuming file https://my.company.com/definitions/models.json_
 ```
 
 
+### Definitions
+
+Reuse schema definitions by creating a repository of definitions.  This is done by simply hosting a file or set of files for commonly used definitions across a company or organization.
+
+Refer to [External references](#external-references) for referencing strategies.
```
jasonh-n-austin: Intentionally left this section pretty empty, as `#external-references` has so many definition/schema examples.

guidelines/REUSE.md
```diff
@@ -1,38 +1,65 @@
...
-* Parameters -> `parameters`
-* Responses -> `responses`
-* Models (and general Schema Objects) -> `definitions`
+* Parameters -> `#/parameters`
+* Responses -> `#/responses`
+* Models (and general Schema Objects) -> `#/definitions`
+
+All references are canonical and must be a qualified [JSON Pointer, per RFC6901](http://tools.ietf.org/html/rfc6901). For example, simply referencing a model `Pet` is not allowed, even if there are no other definitions of it in the file.
+
+_Example from https://github.com/OAI/OpenAPI-Specification/blob/master/examples/v2.0/json/petstore.json_
```
jasonh-n-austin: Tried to use more examples from the repo, rather than contrived examples.

guidelines/REUSE.md
```diff
@@ -1,38 +1,65 @@
...
 
-An example for an internal reference - `#/definitions/MyModel`. All references are canonical and must be a qualified [JSON Pointer](http://tools.ietf.org/html/rfc6901). For example, simply referencing `MyModel` is not allowed, even if there are no other definitions of it in the file.
+To reference the example below: 
 
-When referencing externally, use a valid URI as the reference value. If your referenced file contains only one definition that should be used, you can refer to the file directly. For example:
+`"$ref": "definitions/Model.json""`
+
+or
+
+`"$ref": "https://my.company.com/definitions/Model.json"`.
```
jasonh-n-austin: Added examples of URL and relative in all cases.

guidelines/REUSE.md
whitlockjc: JSON References can reference within the same document so I think the _"creating a repository of definitions"_ is misleading.  In fact, Swagger already uses this for `parameters`, `responses`, `definitions`, ...

I think the point of _reuse_ is to suggest that reuse is enable through the use of JSON References.  If we need to explain more about JSON References, do it of course.

guidelines/REUSE.md
whitlockjc: This also isn't necessarily true.  While swagger has created some top level locations to put reusable objects of a specific type, most JSON Reference processors will resolve any JSON Reference.  There's no reason I can't have something like `#/info/title` as a JSON Reference within my Swagger document.

jasonh-n-austin: <This wording was here before my refactor>

I'd agree, and I've seen this before. However, there is a long-standing assumed convention on the aforementioned elements, and changing it could break existing implementations (probably not yours ;) ).

If this isn't already an existing issue, maybe it should be, tagged for proposal on next version?

guidelines/REUSE.md
jasonh-n-austin: <again, existing wording I tried to preserve>

You could say that `parameters`, `responses`, `definitions` are "repositories of definitions", which happen to reside in the same document (vs external files).

guidelines/REUSE.md
BigstickCarpet: I agree with @whitlockjc here.  It has always seemed weird to me that the Swagger 2.0 spec defined its own [Reference Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#reference-object) and then defined [specific](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields-9) [places](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#patterned-fields-3) [in](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields-4) [the](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields-5) [spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields-4) where that object was allowed.  Seems arbitrarily limiting, when JSON Reference is so much more powerful than that.  Parsers like mine and @whitlockjc's allow users to use JSON References anywhere they want, and in my experience, this is something that almost everyone ends up doing in the real world.

whitlockjc: I'm not 100% sure that's accurate.  By using JSON References, which Swagger knowingly did in 2.0, they knew that this would allow URIs to any place locally or remote.  The type-specific containers were just so you could do all of this in one document if you wanted to while still retaining the ability to validate the content of the containers.  @webron can set me straight if I'm wrong.

webron: Clarifying the intent of the original spec - while internally, you can theoretically reference anything using JSON Schema, we limited where you can define schema objects, responses and so on, so in effect you can only reference those internally. That was actually intentional. I've seen someone trying to abuse it (IMHO it's abuse) by making definitions in vendor extensions and referring to those. The spec does not explicitly mention if that's allowed or forbidden.

External references, on the other hand, are completely open and flexible, and can contain any structure that can be followed by a JSON Reference. 

Since the document is about how to do thing for Swagger/OpenAPI, I'd suggest keeping it now as-is, just to avoid potential confusion to some users. Those who know the capabilities of JSON Refs will use them either way.

whitlockjc: Also, `$ref` values are not guaranteed to be JSON Pointers.  JSON Pointers are only for local references, otherwise the `$ref` value is any URI.  This is based on the JSON References spec: http://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03#section-4

whitlockjc: Thanks @webron, I stand corrected.  I had no idea that creating these containers was to imply that you could only reference definitions within a container.  I can see why you'd want containers so you could do validation on their contents but I would had never assume that implied they were the only things you could officially reference.  Knowing that was the intent now, you should definitely document this.  But...like you said, knowing how JSON References works means many people will just use them as they were intended and reference whatever they need.  ;)

webron: Didn't really go into the details of the containers. There are two reasons for the containers (that I recall). 1 - allow for easier validation. Since you *know* that anything under `definitions` is a Schema Object you can easily validate them. To me, that's a major downside of external references as it makes validating a spec more difficult (not getting into the why now). 2 - All spec objects have no `additionalProperties` allowed except for vendor extensions. That again is intentional, and that's another reason to provide the internal containers to hold referenced objects, because those can't just be added to the spec arbitrarily.

webron:
@jasonh-n-austin - let me know when the PR is ready to be merged.

guidelines/REUSE.md
joelwurtz: If json reference must be pointer in the swagger spec, maybe adding different jsonReference with a pattern would help like the following : 

```
"jsonReferenceSchema" : {
    "type": "object",
    "properties": {
        "$ref": {
            "type": "string",
            "required": true,
            "pattern": "^#/definitions/[^/]+?$"
        }
    }
}
```

webron: Doesn't have to be a pointer, can also be a URL. I don't think the schema should deal with that level of validation.

jasonh-n-austin:
@webron @whitlockjc @BigstickCarpet refactored based on PR/issue discussion. 
* Changed language to 'local' and 'remote'
* Removed some excess samples for brevity
* Removed references to JSON Pointer, made it clear that it's JSON Reference all the way down.
* Other minor edits to capture feedback

guidelines/REUSE.md
```diff
@@ -1,85 +1,100 @@
...
 We encourage reuse and patterns through references.
 
 ## What is reusable
+
 The following types are reusable, as defined by the spec:
 
-* Operations
 * Parameters
-* Responses
 * Models (or Schema Objects in general)
```
whitlockjc: Minor nit but for consistency, either italicize this or remove it from the line below.

guidelines/REUSE.md
```diff
@@ -1,85 +1,100 @@
...
+* Responses
+* Operations (_Operations can only be remote references_)
 
 ## Reuse strategy
-When reusing components in an API design, a pointer is created from the definition to target design.  The references are maintained in the structure, and can be updated by modifying the source definitions.  This is different from a "copy on design" approach where references are injected into the design itself.
 
-The reuse technique is transparent between JSON or YAML and is lossless when converting between the two.
+When reusing components in an API design, a reference is created from the definition to target entity. The references are maintained in the structure, and can be updated by modifying the source definitions. This is different from a "copy on design" approach where references are injected into the design itself.
```
whitlockjc: I personally find this paragraph to be painful to read.  Since most of the verbiage predates this PR, this is not necessarily a complaint about the PR but instead a suggestion in general.  Here are my complaints:

* The terms _"component"_ and _"entity"_ are not used anywhere else in specification documentation so they are confusing here.  I know what is being said because I know Swagger very well but someone potentially new to Swagger might not get it.
* _"The references are maintained in the structure"_ seems _duh_ to me because where else would they be maintained?
* _"API design"_ in the first sentence should probably be _"API design document"_
* The last sentence really isn't helpful because that's an implementation detail, and some implementations may very well inject the dereferenced value into the document
* Since OpenAPI references work identically regardless of JSON or YAML, I don't think we need to mention it
* YAML anchors injecting the dereferenced value into its resulting object is an implementation detail.  To the end user, they would get the same result using YAML anchors as they would with JSON References.  While I do realize that the resulting in-memory representation can be different when using JSON References vs. YAML anchors, I don't know if this is worth mentioning since it's a tooling thing that most API design document authors will never see.

I wouldn't be very useful if I just critiqued so here is a version that addresses these concerns.  I'm not saying we need to use this version but I am hoping that it will be useful in coming to some sort of agreement:

As you author your API design documents, there may come a time when you have common object definitions that you would like to reuse to avoid duplication.  *(For example, imagine you have multiple path definitions that each share a common path parameter, or you have a common response structure.)*  The OpenAPI specification allows you to reuse common object definitions through the use of _"references"_.

A reference is technically just a placeholder, a construct in your API design document that says _"the real content for this portion of the document is defined elsewhere"_.  To create a reference, at the location in your document where you want to reuse some other definition, you create an object that has a `$ref` property whose value is a `URI` pointing to where the definition is *(more on this in later sections)*.  Here are two examples:

**JSON** *(JavaScript syntax used to allow comments for demonstration purposes)*
```js
{
  // ...
  definitions: {
    Person: {
      type: 'object',
      properties: {
        friends: {
          type: 'array',
          items: {
            $ref: '#/definitions/Person'
          }
        }
      }
    }
  }
}
```

**YAML**
```yaml
# ...
definitions:
  Person:
    type: object
    properties:
      friends:
        type: array
        items:
          $ref: '#/definitions/Person'
```

In reality, the OpenAPI _"reference"_ concept is not something we came up with.  In fact, it is known as a [JSON Reference](http://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03).  The major difference between JSON References and OpenAPI references is that JSON References can point to anything, anywhere but OpenAPI references are intended only to point to definitions as defined in the section above.

**Note:** Those of you familiar with YAML know that YAML already has a feature very similar to what is described above.  YAML _"anchors"_ also allow you to reuse common definitions/values but since that is a YAML feature, we will not go into further detail on the matter.

guidelines/REUSE.md
```diff
@@ -1,85 +1,100 @@
...
+
 
-Whether you reference definitions internally or externally, you can never override or change their definitions from the referring location. The definitions can only be used as-is.
+##### URL
 
-### Definitions
-Reuse schema definitions by creating a repository of definitions.  This is done by simply hosting a file or set of files for commonly used definitions across a company or organization.  In the case of multiple files, the models can be referenced directly as such:
+Remote files can be hosted on an HTTP server (rather than the local file system). 
+
+One risk of this approach is that environment specific issues could arise if DNS is not taken into account (as the reference can only contain one hostname).
```
whitlockjc: I'm not sure this is worth mentioning.  I think it's obvious that if you cannot access the referenced document for any reason that things will fail.

guidelines/REUSE.md
```diff
@@ -205,8 +232,11 @@ To include these parameters, you would need to add them individually as such:
 ```
 
 ### Operations
+
 Again, Operations can be shared across files.  Although the reusability of operations will be less than with Parameters and models. For this example, we will share a common `health` resource so that all APIs can reference it:
```
whitlockjc: Consistency in casing for `Parameters` and `models`.

whitlockjc:
Please don't hate me.  :)  The biggest issue I had wasn't your fault and could likely be ignored for now and addressed later.  The other pieces of feedback are all small.

guidelines/REUSE.md
jasonh-n-austin: I really hated this paragraph, and was out of energy for it. Thanks for the addition, I'll reword a bit for consistency of voice and get it in there.

guidelines/REUSE.md
jasonh-n-austin: It seems like a common request, "how do I handle multiple environments with <insert spec format here>". This was my attempt at stating, "you don't". 
In other words, it's clarifying that hostnames are not parameterizable, at least in the word of the spec.

guidelines/REUSE.md
jasonh-n-austin: Good point on consistency...I'm also trying to move to 'Definitions' to match the convention (we had model/schema/definition pretty inconsistently all over).

jasonh-n-austin:
@whitlockjc love it...keep it coming. Doc updates with no feedback are meaningless; updates with lots of feedback are always higher quality.

guidelines/REUSE.md
whitlockjc: Glad to help.

guidelines/REUSE.md
whitlockjc: But why would someone expect it to be if there is no mention of this anywhere except for path parameters?  I'm not against keeping it, I just thought by adding the details that it could be confusing.  Your call.

guidelines/REUSE.md
whitlockjc: Good idea.

guidelines/REUSE.md
jasonh-n-austin: @whitlockjc I don't get this sentence:
> The major difference between JSON References and OpenAPI references is that JSON References can point to anything, anywhere but OpenAPI references are intended only to point to definitions as defined in the section above.

With remote URL references, you could potentially point anywhere. For that matter, with remote relative references, you can point to files anywhere in the file system, in theory. The convention aspect of local references is already covered in #local-references.

whitlockjc: Agreed but if you do it the _"OpenAPI Way"_, you should be held to the same _"local references should only point to certain locations"_ even for remote references.  That's the only reason why I lumped this into one.  But as for your point that it is mentioned elsewhere, works for me.

jasonh-n-austin:
@whitlockjc commit'd fixes, with some rewording for brevity and consistency.

Gimmee some :shipit: ! :smiley_cat:

guidelines/REUSE.md
webron: Sorry for the late reply. Just read the comments on this section and there are two issues brought up that I'm not sure I agree with. I've had a quick chat with @whitlockjc directly (sorry @jasonh-n-austin), and think there may be room for discussion. Note I'm adding this comment after reading the comments above but before reading the modified doc. The two issues are:
- There's no difference between JSON References and YAML Anchors, and there's no difference between 'injecting' and 'referencing' (I believe there's a difference in both cases).
- Local JSON References act differently in the OpenAPI spec (they don't).

Since it's late for today, I hope we can continue the discussion tomorrow. I'd rather not dive into the details now to not pull in people to a discussion at this hour.

jasonh-n-austin: @webron I agree: if you read the most recent update, I think I toned down all of that language quite a bit from what @whitlockjc had proposed.

guidelines/REUSE.md
```diff
@@ -1,85 +1,140 @@
...
 
 ## Reuse strategy
-When reusing components in an API design, a pointer is created from the definition to target design.  The references are maintained in the structure, and can be updated by modifying the source definitions.  This is different from a "copy on design" approach where references are injected into the design itself.
 
-The reuse technique is transparent between JSON or YAML and is lossless when converting between the two.
+When authoring API design documents, common object definitions can be utilized to avoid duplication. For example, imagine multiple path definitions that each share a common path parameter, or a common response structure. The OpenAPI specification allows reuse of common object definitions through the use of "references".
+
+A reference a construct in your API design document that indicates "the content for this portion of the document is defined elsewhere". To create a reference, at the location in your document where you want to reuse some other definition, create an object that has a `$ref` property whose value is a URI pointing to where the definition is (more on this in later sections).
```
whitlockjc: _"A reference a"_ should be _"A reference is a"_.

guidelines/REUSE.md
```diff
@@ -1,85 +1,140 @@
...
-The reuse technique is transparent between JSON or YAML and is lossless when converting between the two.
+When authoring API design documents, common object definitions can be utilized to avoid duplication. For example, imagine multiple path definitions that each share a common path parameter, or a common response structure. The OpenAPI specification allows reuse of common object definitions through the use of "references".
+
+A reference a construct in your API design document that indicates "the content for this portion of the document is defined elsewhere". To create a reference, at the location in your document where you want to reuse some other definition, create an object that has a `$ref` property whose value is a URI pointing to where the definition is (more on this in later sections). 
+
+OpenAPI's provides reference capabilities using the [JSON Reference](https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03) specification. 
+
+### JSON Example
+
+``` json
```
whitlockjc: If you want to use `json` as the language, remove the commented out ellipsis.  If you want to keep the comment, change it back to `js`.

guidelines/REUSE.md
```diff
@@ -1,85 +1,140 @@
...
+    type: object
+    properties:
+      friends:
+        type: array
+        items:
+          $ref: '#/definitions/Person'
+```
 
-Referenes can be made either inside the Swagger definition file or to external files. References are done using [JSON Reference](http://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03).
+Note: YAML has a very similar feature, [YAML anchors](http://yaml.org/spec/1.2/spec.html#id2765878). Examples from this point will only be in JSON, using JSON References.
```
whitlockjc: Well played.  :)

whitlockjc:
Two minor nits, one of which is a formatting issue.  After that, I don't see why this can't be merged.  Good job.

jasonh-n-austin:
OK fixed up content tweaks. @webron should be good for merge. Squirrel me!

guidelines/REUSE.md
```diff
@@ -92,14 +147,14 @@ _Assuming file https://my.company.com/definitions/Model.json_
       "format": "int64"
     },
     "tag": {
-      "description": "a complex, shared property.  Note the absolute reference",
+      "description": "A complex, shared property.  Note the absolute reference",
```
webron: I know this was in the original example as well, but the `description` is obviously ignored and does not override the value inside the reference (as previously mentioned in the doc). I get this is used here for documentation to the readers, but it may also send the wrong message. I'll merge as-is, but we may consider removing it after all.

webron:
Thanks all for the great work on this, and apologies for the delay in merging. I think we're good to go.

PR #547, opened 2016-01-26
Splitting guidelines

webron:
With the additional information about JSON References, perhaps it makes to have a dedicated guideline page for JSON References. It would help users who seek that information specifically.

----
Original authors credit - @olensmar and @jasonh-n-austin

### DIFF
--- a/guidelines/JSON_References.md
+++ b/guidelines/JSON_References.md
@@ -1,0 +1,130 @@
+# Guidelines for Referencing
+
+The OpenAPI Specification relies on JSON Refereneces as the machinsm for reusability (DRY). This guide gives you a brief introduction to JSON References and its capability. It does not provide the full details which are covered in its own specification, but aims to provide you with information about the reference types.
+
+All references should follow the [JSON Reference](https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03) specification.
+
+JSON Reference provides guidance on the resolution of references, notably:
+
+> If the URI contained in the JSON Reference value is a relative URI,
+then the base URI resolution MUST be calculated according to
+[RFC3986], section 5.2. Resolution is performed relative to the
+referring document.
+
+Whether you reference definitions locally or remote, you can never override or change their definitions from the referring location. The definitions can only be used as-is.
+
+### Local references
+
+When referencing locally (within the current document), the target references should follow the conventions, as defined by the spec:
+
+* Parameters -> `#/parameters`
+* Responses -> `#/responses`
+* Definitions (Models/Schema) -> `#/definitions`
+
+An example of a local definition reference:
+
+_Example from https://github.com/OAI/OpenAPI-Specification/blob/master/examples/v2.0/json/petstore.json_
+``` json
+          // ... 
+          "200": {
+            "description": "pet response",
+            "schema": {
+              "type": "array",
+              "items": {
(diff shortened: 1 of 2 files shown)


guidelines/REUSE.md
```diff
@@ -275,7 +203,7 @@ To include these parameters, you would need to add them individually as such:
 
 Again, Operations can be shared across files.  Although the reusability of operations will be less than with Parameters and Definitions. For this example, we will share a common `health` resource so that all APIs can reference it:
```
jasonh-n-austin: > For this example

The example is missing...might be best to eliminate this intro, or move it to `JSON_References.md`

guidelines/REUSE.md
```diff
@@ -187,19 +123,11 @@ _Assuming file https://my.company.com/definitions/models.json_
...
-
-
 ### Parameters
 
 Similar to model schemas, you can create a repository of `parameters` to describe the common entities that appear throughout a set of systems.  
 
-Refer to [Guidelines for Referencing](#guidelines-for-referencing) for referencing strategies.
+Refer to [Guidelines for Referencing](#JSON_References.md) for referencing strategies.
 
 Using the same technique as above, you can host on either a single or multiple files.  For simplicity, the example below assumes a single file.
```
jasonh-n-austin: > the example below

Again, no example here, so it doesn't read right.

guidelines/JSON_References.md
```diff
@@ -0,0 +1,130 @@
+# Guidelines for Referencing
+
+The OpenAPI Specification relies on JSON Refereneces as the machinsm for reusability (DRY). This guide gives you a brief introduction to JSON References and its capability. It does not provide the full details which are covered in its own specification, but aims to provide you with information about the reference types.
```
jasonh-n-austin: > machinsm

mechanism

guidelines/REUSE.md
```diff
@@ -314,59 +242,3 @@ Which points to the reference in the `operations.json` file:
 ```
 
 Remember, you cannot override the definitions, but in this case, you can add additional operations on the same path level.
```
jasonh-n-austin: This line is duplicated in both docs...intentional?

whitlockjc:
I wonder if we should do a GitBook specification.  It's very easy on the eyes, hierarchical by default and it should align pretty well with our plans to take this large monolithic document and break it up.

This is more of a general idea than a review of this PR.

webron:
Well, shame on me for not paying attention to the details. Will get those fixed. Any thoughts on the split itself and whether it should be done at all?

@whitlockjc never heard of GitBook before you mentioned it now, so no opinion there.

jasonh-n-austin:
After reading through this, I'm not convinced it's a good split. There's a lot of duplication of content, and the guidelines don't read well without inline examples.
I don't have any experience with GitBook either (just signed up, looks cool), I'd be curious to see an example of what @whitlockjc has in mind.

webron:
Fair enough, @jasonh-n-austin - not overly concerned with pushing it. I'll try giving it another go and see if I can make more sense of a split, but if it feels artificial, then there's no point to it.

If we don't do the split, I'll probably submit a different PR to slightly organize the doc (links, topic rearrangements...). Will ping for additional review in that case.

webron:
Closing for now. May revisit in the future.

PR #570, opened 2016-02-17
Yaml sample files

saitest:


webron:
@olensmar - do you recall what you had in mind here?

PR #594, opened 2016-03-08
Convert response codes in YAML specs to be strings

sjaensch:
According to the spec, response codes should be of type string. This is the case in the JSON examples, but not in the YAML ones, causing issues with consumers that are implemented according to spec. Let's fix the examples so they don't set a bad... example.


webron:
It's not necessarily a bad example, but for the sake of cleanliness, I'll merge it.

ePaul:
I always assumed that any processor would convert the integers to strings here. Do we really need to use quotation marks here?

webron:
@ePaul in YAML keys are not necessarily string. Having something like
```yaml
200: hello
"200": goodbye
```
is legal and those are entirely separate properties. However, when only one exists, converters will normally convert integers to strings, but afaik it is not explicitly defined. So for the sake of purity, the change does make sense. We've applied a similar change to the examples in the spec itself a few months ago.

sjaensch:
Thanks for merging this in! @ePaul to expand on @webron's answer, 200 and "200" are different things in YAML, just like in JSON. Check out what I had to do for [bravado, the Python client library](https://github.com/Yelp/bravado/pull/198/files#diff-18a7d25729cc82f5d28614de17daee37R107).

webron:
Well, not 'just like in JSON' because in JSON keys are always strings.

PR #609, opened 2016-03-23
Items Object should not be an array of object

EricHenry:
As per the Swagger / OpenAPI specification, a property with an "array" type requires an items property as well that is an object (specifically a subset of a JSON Schema object). However in the Uber Examples, the items property is an array of objects.

### DIFF
--- a/examples/v2.0/json/uber.json
+++ b/examples/v2.0/json/uber.json
@@ -346,11 +346,9 @@
         },
         "history": {
           "type": "array",
-          "items": [
-            {
+          "items": {
               "$ref": "#/definitions/Activity"
-            }
-          ]
+           }
         }
       }
     },
--- a/examples/v2.0/yaml/uber.yaml
+++ b/examples/v2.0/yaml/uber.yaml
@@ -261,7 +261,7 @@ definitions:
       history:
         type: array
         items:
-          - $ref: '#/definitions/Activity'
+          $ref: '#/definitions/Activity'
   Error:
     properties:
       code:


whitlockjc:
An array of objects is completely valid.  The `items` value can be an object itself which applies a single JSON Schema to all items in the array or `items` can be an array of objects that will apply the schema to specific array items via its index.

I defer to @webron on this one but this is how I interpreted it and how the tooling I've written supports things.

EricHenry:
@whitlockjc thanks for the quick response! You are absolutely right, the JSON Schema specification says that the items property of an array type "...MUST be either an object or an array. ..." [[5.3.1.1]](http://json-schema.org/latest/json-schema-validation.html#anchor37). 

However, the Swagger/OpenAPI specification is a bit more limiting. It defines that the "items" property is an [Items Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#itemsObject). Which is specifically identified as being a subset of a JSON-Schema's item. 

If it is the case that you can have an array of Items Objects, then I propose that the Swagger / OpenAPI spec should explicit say so since it seems that is not clearly expressed. Or should be included in future versions of the spec. 

What is peculiar is that the swagger editor (with built validation) does not throw an error if given an array, however it doesn't know how to process the given array of objects. Not knowing how their validation system works, I would speculate that the intent of the spec was not to handle arrays.
screenshot from 2016-03-23 13-39-58

I am looking forward to this discussion and getting @webron input.

fehguy:
@EricHenry you are correct.  The Swagger editor is an implementation which supports the spec, but that doesn't make it prone to errors.  Support for arrays under items actually is a bug--per the subset of JSON schema that the spec supports, it should not allow it.

Maintaining a human readable documentation (the 2.0 readme) is the goal, and unfortunately sometimes there are errors in it.  The schema that is built manually from that readme's intent is not the official documentation of the spec, so tooling that use it (like the editor) may not reflect the spec itself.

EricHenry:
@fehguy thanks for your response! I was in no way trying to attack the Swagger editor, I think it is a wonderful tool, and I use it quite often. I just wanted to make sure that the examples reflect the spec, to lessen potential errors or confusion. 

I agree that maintaining human readable documentation is difficult, especially with such a large spec. Do you feel that the readme is specific enough in describing the Items Object?

EricHenry:
Hey guys any update or other thoughts on this PR?

webron:
Yeah, the PR is right. Array values for `items` are not supported by the spec (2.0 at least).

adjenks:
Sorry to necro, but... I really need array values for items... How do I represent this? Is there some other funky way I can specify a specifically structured array?
E.g. Item 1 must be a string, Item 2 must be a number, etc.

MikeRalphson:
Noting the array form of items is now supported in OAS 3.1 using the [prefixItems](http://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.10.3.1.1) keyword.

PR #628, opened 2016-04-07
Wrap '$ref's to prevent 'definitions' override

IvanGoncharov:
From [JSON Reference](https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03):
> Any members other than "$ref" in a JSON Reference object SHALL be  ignored.

> Implementations MAY choose to replace the reference with the referenced value.

### DIFF
--- a/schemas/v2.0/schema.json
+++ b/schemas/v2.0/schema.json
@@ -40,11 +40,19 @@
     },
     "consumes": {
       "description": "A list of MIME types accepted by the API.",
-      "$ref": "#/definitions/mediaTypeList"
+      "allOf": [
+        {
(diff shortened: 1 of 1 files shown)


whitlockjc:
LGTM

webron:
I don't mind merging it, just not sure what it's meant to serve. The description is just a documentation tool to anyone who bothers reading the schema, what difference does it make if it's being ignored when processing?

IvanGoncharov:
@webron description used during validation to provide context for errors.

whitlockjc:
@IvanGoncharov is right, most JSON Schema validators will use the `description` field when a rule fails validation to help relay extra information to the user.

Issue #496, opened 2015-10-27
Proposal: add '$schema' property at root of the spec 2.x


dolmen:
JSON schema specifies a standard way for a document to tell what JSON Schema document can be used to validate it. This is the [`/$schema` property](http://json-schema.org/latest/json-schema-core.html#anchor22).

Unfortunately this property is strictly disallowed in the Swagger 2.0 JSON schema (`/additionalProperties` is set to false): Swagger authors decided to use instead the `/swagger` property (following previous versions). This is enough for humans, but the presence of `/$schema` could be useful for machines to validate Swagger specifications that use vendor extensions and for which a JSON schema specifying the schema of those extensions is available. That would allow to write a Swagger spec that explicitely points to the JSON schema that validates those extensions.

So I propose to add `/$schema`, with a default value of http://swagger.io/schema/v2.0/schema.json (see [swagger-api/swagger.io#48](https://github.com/swagger-api/swagger.io/issues/#48) which is a request to have canonical URLs for Swagger JSON schemas).

As this feature is incompatible with the 2.0 schema (`/$schema` is forbidden in 2.0 because `/additionalProperties` is [set to false](../blob/master/schemas/v2.0/schema.json#L11)), setting this value in a document will only be possible if a version 2.0.1/2.1 of the specification is published with that extension.

An implementation in the JSON Schema is proposed in #485.

MrTrick:
I agree, it would be very useful. 

Several validation libraries will use the "$schema" property to automatically validate the document against that schema. I can include `"$schema": "http://swagger.io/schema/v2.0/schema.json"` in my swagger spec, but all it will do is tell me that it's invalid **because of** that attribute. ^_^

dolmen:
@webron Proposal for OpenAPI.next

jpstrikesback:
#561 provides a nice reason to endorse this, some of the benefits past that are listed above.

dolmen:
@webron about [your comment in #561](561#issuecomment-190402975) 

> The OpenAPI spec is not JSON Schema. It uses it to describe input/output, it provides it as a limited form of validation, but it is not the same.

[...]

PR #631, opened 2016-04-08
updates for identification of spec, schema

### LINKED ISSUE #561: Remove swagger: 2.0 from spec
The current OpenAPI spec has `"swagger": "2.0"` in it.  Since it's no longer called the swagger spec, this should go away.

I do believe we should continue to declare the version in the specification though, to make tooling more tolerant.  Some points about that.

* One major issue with the Swagger 2.0 tooling was that it was looking for the "2.0" version in the spec itself.  That meant, a "2.0.1" spec--even if structurally compatible--would not work.  We should advise all tooling vendors as to versioning process and how they should detect versions inside the spec.  Reference the concerns in #107.  In a couple cases we probably should have bumped the spec version in 2.0 to 2.0.1, etc.  Let's not make that mistake again

* Declaring a semver approach for the spec makes sense.  We don't need to go too far into the details in this ticket as to what constitutes a major, minor, patch version, but in general, if we put a version in the next spec version (which I advocate), it should be in the `x.y.z` format.

* Making it "2.0" as a string vs. number was dumb.  The above point solves that.

fehguy:
Fixes #561

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -71,7 +71,7 @@ The schema exposes two types of fields. Fixed fields, which have a declared name
 
 The OAS representation of the API is made of a single file. However, parts of the definitions can be split into separate files, at the discretion of the user. This is applicable for `$ref` fields in the specification as follows from the [JSON Schema](http://json-schema.org) definitions.
 
-By convention, the OpenAPI Specification (OAS) file is named `oas.json`.
+By convention, the OpenAPI Specification (OAS) file is named `openapi.json` or `openapi.yaml`.
 
 ### Data Types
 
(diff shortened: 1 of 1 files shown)


fehguy:
Also fixes #496

versions/3.0.md
```diff
@@ -71,7 +71,7 @@ The schema exposes two types of fields. Fixed fields, which have a declared name
 
 The OAS representation of the API is made of a single file. However, parts of the definitions can be split into separate files, at the discretion of the user. This is applicable for `$ref` fields in the specification as follows from the [JSON Schema](http://json-schema.org) definitions.
 
-By convention, the OpenAPI Specification (OAS) file is named `swagger.json`.
+By convention, the OpenAPI Specification (OAS) file is named `oas.json`.
```
fehguy: or oas.yaml???

earth2marsh: or `openapi.yaml` and `openapi.json`

jasonh-n-austin: +1

darrelmiller: I prefer openapi.* if we must suggest something, although I prefer if we didn't express a preference for a filename.

versions/3.0.md
```diff
@@ -106,7 +106,8 @@ This is the root document object for the API specification. It combines what pre
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="oasSwagger"></a>swagger | `string` | **Required.** Specifies the OpenAPI Specification version being used. It can be used by the Swagger UI and other clients to interpret the API listing. The value MUST be `"2.0"`.
+<a name="oasSwagger"></a>swagger | `string` | **Required.** Specifies the OpenAPI Specification version being used. It can be used by tooling vendors and clients to interpret the version.  The structure shall be `major`.`minor`.`patch`, where `patch` versions _must_ be compatible with the existing `major`.`minor` tooling. Typically patch versions will be introduced to address errors in the documentation, and tooling should typically be compatible with the corresponding `major`.`minor` (3.0.*). Patch versions will correspond to patches of this document.
```
earth2marsh: It looks like this is still called `swagger`, no? 

In the past, one challenge with versions is distinguishing the version of the spec vs the version of the API described by the spec.

fehguy: my mistake, will fix this.

earth2marsh: Another minor question, but changing if the anchor is being changed, is there any benefit in retaining the word `swagger`?

fehguy:
Schema version + spec version will _not_ necessarily be the same.

@OAI/tdc to review by Weds and provide comments?

versions/3.0.md
```diff
@@ -106,7 +106,8 @@ This is the root document object for the API specification. It combines what pre
...
+<a name="oasSchema"></a>schema | `string` | An optional URL to a formal OAS schema. If present, tooling _may_ choose to use this for validating the document, and the location _must_ be a formally approved OAS schema produced by the OpenAPI working group.
```
fehguy: Kill this, will cause more harm than good @OAI/tdc 

Will publish a formal location of the schema (https://openapis.org/schemas/3.0-latest.json)

versions/3.0.md
webron: Another vote for `openapi.*`.

webron:
Regarding the version - reiterating what we discussed (for documentation sake) - the benefits of a patch-level version granularity are clear, but we're also not going to be update the spec (normally) for every minor change. We'll do our best to collect changes and not have a too-frequent release cycle.

webron:
Other than that, :+1: from me for the PR (assuming suggested file name and removal of `schema`).

fehguy:
I have removed the schema reference and `swagger` from the key name.

fehguy:

darrelmiller:
👍

jasonh-n-austin:
:+1:

earth2marsh:
👍

PR #636, opened 2016-04-08
Add examples object

### LINKED ISSUE #544: $ref in response example
I would like to use `$ref` in the response example in order to break up my specification into more readily manageable chunks.  Examples also benefit from being externalized because they can be easily revisioned, and included verbatim in reference implementations.

desired structure:

``` yaml
      responses:
        200:
          description: Success
          schema:
            $ref: 'profile-models.yaml#/ProfileSearchResponse'
          examples:
            application/json:
              $ref: 'ProfileExample.yaml#/ProfileSearchResponseExample'
```

Today, the children of examples are included directly, meaning the example itself using this example is:

``` json
{
  "$ref": "ProfileExample.yaml#/ProfileSearchResponseExample"
}
```

fehguy:
Fixes #544

DO NOT MERGE--needs documentation updates.

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -1154,7 +1154,11 @@ description: Pets operations
 
 #### <a name="examplesObject"></a>Examples Object
 
-Anywhere an `example` may be given, allow a $ref object.  This does mean that `example` can be either a string primitive or an object, like `additionalProperties`:
+Anywhere an `example` may be given, allow a $ref object.  This does mean that `example`, structurally, can be either a string primitive or an object, like `additionalProperties`.
+
+In locations where the field being provided an `example` is a scalar value _or_ has it's content-type definition determined by a higher-level construct (a response payload, for example, uses the `produces` attribute to select the correct message format), the plural `examples` shall be used, and the payload format be specified as a key to the example.
+
+In all cases, the payload is expected to be compatible with the type schema for the value that it is accompanying. Tooling vendors may choose to valide compatibility automatically, and reject the example value(s) if they are not compatible.
 
 ```yaml
 # in a model
@@ -1163,9 +1167,9 @@ definitions:
     name:
       type: string
       example:
-        $ref: http://foo.bar#/examples/name-example.json
+        $ref: http://foo.bar#/examples/name-example
 
-# in a parameter:
+# in a parameter, note the plural `examples` as the content-type is set by `consumes`:
   parameters:
     - name: address
       in: body
@@ -1173,25 +1177,25 @@ definitions:
         $ref: '#/definitions/Address'
       examples:
         'application/json':
-          $ref: http://foo.bar#/examples/address-example.json
+          $ref: 'http://foo.bar#/examples/address-example.json'
         'application/xml':
-          $ref: http://foo.bar#/examples/address-example.xml
+          $ref: 'http://foo.bar#/examples/address-example.xml'
         'text/plain':
-          $ref: http://foo.bar#/examples/address-example.txt
+          $ref: 'http://foo.bar#/examples/address-example.txt'
         default:
-          $ref: http://foo.bar#/examples/address-example.whatever
-    - name: zip
-      in: query
-      type: string
-      format: zip
+          $ref: 'http://foo.bar#/examples/address-example.whatever'
+    - name: 'zip'
+      in: 'query'
+      type: 'string'
+      format: 'zip'
       example: 
-        $ref: http://foo.bar#/examples/zip-example
-# in a response:
+        $ref: 'http://foo.bar#/examples/zip-example'
+# in a response, note the plural `examples`:
   responses:
     200:
       description: your car appointment has been booked
-        schema:
-          $ref: '#/definitions/SuccessResponse'
+      schema:
+        $ref: '#/definitions/SuccessResponse'
       examples:
         'application/json':
           $ref: http://foo.bar#/examples/address-example.json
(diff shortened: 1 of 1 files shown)


versions/3.0.md
```diff
@@ -1152,6 +1152,51 @@ name: pet
 description: Pets operations
 ```
 
+#### <a name="examplesObject"></a>Examples Object
+
+Anywhere an `example` may be given, allow a $ref object.  This does mean that `example` can be either a string primitive or an object, like `additionalProperties`:
```
ePaul: This is a change to the 2.0 behavior, where the example property of a model is of the same type as what the model described. (I.e. `type: object` means it would be an object, `type: string` means it is a string, `type: number` means it is a number.) Now it seems to be a string containing formatted JSON (or a reference to such)?

If this is intentional, that needs to be updated too.

webron: @ePaul actually, in 2.0 it is the `default` field value has to be the same type as it's 'container' (be it a parameter or a model property). The examples have on such constraint.

webron:
I'm a bit confused by this. Is this meant to add `example` or `examples` to parameters as well (which they don't right now)?

Also, adding `$ref` support to a primitive-valued field is a conceptual change which we haven't had until now. I'm not against it, just making sure we understand the impact of it.

versions/3.0.md
ePaul: @webron Maybe I'm misunderstanding this. [Example Object (for responses)](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#exampleObject):

> The name of the property MUST be one of the Operation produces values (either
> implicit or inherited). The value SHOULD be an example of what such a response
> would look like.

The example for this (with a JSON object) seem to indicate that this is an instance matching the schema.

[In the schema object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#schemaObject) we have:

--------------|--------|-----------
example | Any | A free-form property to include a an example of an instance for this schema.

Also here I would understand (and the examples agree) that this usually include the JSON value, not a string representation thereof. (While it might not be a "constraint", given the type "Any", the text + examples strongly suggest this.)

fehguy:

darrelmiller:
a) this is awesome and has a big 👍 from me
b) Am I correct in understanding that this change means that you can't blindly run an OpenAPI doc through a $ref processor because you could end up trying to merge XML into a JSON doc?

versions/3.0.md
webron: @ePaul - when writing the spec, while making mistakes here and there, I've taken great care to word things as they are. Type `Any` means just that - any valid JSON value would fit there, regardless of what it relates to. The word `SHOULD` is used on purpose and not the word `MUST`. The spec really does not enforce any sort of constraint here, and I don't think it should (for a few reason).

Going back to the `default` field value, you'll see that it uses a stronger restriction saying it `MUST` follow the type assigned to its container.

webron:
@darrelmiller it's a bit ugly now, but technically the approach in 2.0 was that non-json complex values would be pretty much strings. The use of a reference doesn't really change that (you won't be able to magically reference an XML document because it would result in an invalid JSON).

As for the multiple schema - not necessarily. 2.0 has a single schema for JSON and XML for example, although it's obviously limited. With that, it would still be possible to provide multiple examples. This may change when we tackle the content negotiation (and thus supporting multiple schemas).

:+1: from me for the proposal, but the PR does not cover the changes that need to be made in the spec itself. I can expand the PR if the proposal itself is accepted by the @OAI/tdc.

versions/3.0.md
ePaul: @webron Okay, maybe the spec writers meant not to put a restriction here (I guess you were there – I was not), but then I don't think this is a good idea to allow just "anything at all".
It should clearly define how the examples are meant to be used.

Allowing just strings (or refs to such) now is also okay, as long as it is made explicit.

versions/3.0.md
```diff
@@ -1152,6 +1152,55 @@ name: pet
 description: Pets operations
 ```
 
+#### <a name="examplesObject"></a>Examples Object
+
+Anywhere an `example` may be given, allow a $ref object.  This does mean that `example`, structurally, can be either a string primitive or an object, like `additionalProperties`.
+
+In locations where the field being provided an `example` is a scalar value _or_ has it's content-type definition determined by a higher-level construct (a response payload, for example, uses the `produces` attribute to select the correct message format), the plural `examples` shall be used, and the payload format be specified as a key to the example.
+
+In all cases, the payload is expected to be compatible with the type schema for the value that it is accompanying. Tooling vendors may choose to valide compatibility automatically, and reject the example value(s) if they are not compatible.
```
jasonh-n-austin: @fehguy  "choose to valide" => "choose to validate"

jasonh-n-austin:
@fehguy a point of confusion: you updated the `Examples Object` section (https://github.com/OAI/OpenAPI-Specification/blob/issue-544/versions/3.0.md#examples-object), but reference `example` throughout the copy (which would be https://github.com/OAI/OpenAPI-Specification/blob/issue-544/versions/3.0.md#exampleObject).
`example` is left in it's current state (type=any), and `examples` is now `$ref`-able. With that in mind, it might reduce confusion to change the copy to explicitly state `examples` (vs `example`), as they are quite different in their usage.
...or change how `example` works too, but I think that has XML implications.

fehguy:
thanks @jasonh-n-austin.  I'm going to make a note to clarify the documentation on this and we can close this out.

earth2marsh:
👍

jasonh-n-austin:
:+1:

Issue #589, opened 2016-03-01
[Meta Issue] Specification documentation structure and supporting docs


webron:
This does not cover what the spec actually support, but rather how the specification documentation itself is provided. This issue will also cover suggestions for supporting docs.

Please don't add comments directly to this issue.

fehguy:
@OAI/tdc please send some feedback for documentation structure

RobDolinMS:
I'll look to @webron for some advice / mentoring on this.

RobDolinMS:
Will look at adding an index / table of contents (ToC): #673

fehguy:
@OAI/tdc here's an inventory of the documentation items for review / examples:

## Metadata
- [ ] OpenAPI Object
- [ ] Info Object
- [ ] Host Object
- [ ] Contact Object
- [ ] License Object
- [ ] Tag Object

- [ ] ## Paths + operations + body parameters
- [ ] Paths Object
- [ ] Path Item Object
- [ ] Operation Object
- [ ] External Documentation Object
- [ ] Parameter Object
- [ ] Request Body Object
- [ ] Items Object

- [ ] ## Responses + payloads
- [ ] Responses Object
- [ ] Response Object
- [ ] Headers Object
- [ ] Example Object
- [ ] Links Object
- [ ] Response Payload Values
- [ ] Header Object
- [ ] Examples Object
- [ ] Reference Object
- [ ] Schema Object
- [ ] XML Object

- [ ] ## Reuse
- [ ] Components Object
- [ ] Definitions Object
- [ ] Parameters Definitions Object
- [ ] Responses Definitions Object

- [ ] ## Security
- [ ] Security Definitions Object
- [ ] Security Scheme Object
- [ ] Scopes Object
- [ ] Security Requirement Object

handrews:
This meta-issue has served its purpose, and any remaining items are being tracked under new labeling now.

Issue #540, opened 2016-01-15
Allow additional field deprecatedMessage for operation


lieldulev:
field name `deprecatedMessage`
type `string`. 
**optional**.
default `""`. 
Meaningful only if `depreacted` is **true**.

Usually deprecating a command/function/endpoint comes with a message, helping user to migrate to alternative endpoint/api/service.

https://en.wikipedia.org/wiki/Deprecation#Software_deprecation

whitlockjc:
For simplicity, why not allow the `deprecated` property take a string?  Then we don't have to be so verbose.

lieldulev:
Are you suggesting `deprecated` to take either boolean or string?
Or just change it to string and have empty string ( `""` ) as the default value and mean **not deprecated**?

whitlockjc:
Yes, it takes a boolean or a string.  The idea is for simple cases, turning it on *(`deprecated: true`)* would be enough.  But for the case you want to allow, it could be done with the same property *(`deprecated: 'This field has been replaced by that field'`)*.  Heck, you might even allow GFM in the value to allow for styled rendering of the deprecation message.

Processing rules:

* Property not set: Not deprecated
* Property is false: Not deprecated
* Propery is true: Deprecated
* Property is a string: Deprecated

Just an idea...

IvanGoncharov:
@whitlockjc @lieldulev I think we already have two nice fields for that `title` could be `[Deprecated] ...` and `description` could be `**Deprecated**: ...`.
So it's just matter of including both fields in warning output.

whitlockjc:
Well, I figure this proposal is more for a general property and not just for the overall API itself.  Imagine being able to mark operations, paths, parameters, ... as deprecated as your API evolves.

IvanGoncharov:
IMHO, I think Swagger should have minimal amount features which allow to describe API.
So if you want deprecate path, just mark every operation inside it as deprecated(usual it 2-3 operations).

Deprecation attribute for parameter can be useful, but parameters also have `description` attribute.
So solution is the same as for operations, just include `description` into warning message.

whitlockjc:
The issue with relying on description is now you have to parse it for special tokens and that is just more work.  *(It's work all around, I know.)*  We could easily get away with this using a vendor extension but to me, deprecating various parts of your API design is something that is globally useful which makes it a good candidate for core specification support.

IvanGoncharov:
Why you need to parse anything?
[...]

Issue #146, opened 2014-10-09
Support for multiple request/response models based on headers


jahlborn:
I guess this is a 2.0+ request as 2.0 is already "released".

Basically, there is currently no way to express a single path which may have different request/response formats based on additional criteria, namely headers.  two possible scenarios:
1. An API which supports both xml and json, but the models for each format are different enough that they cannot be expressed via a single json schema w/ xml annotations (even if the xml format _is_ expressible using the xml annotations).
2. An API which supports versioning the representation via the media-type.  e.g. application/my+v1 returns a json model which is sufficiently different from the application/my+v2 model such that i cannot describe them both using one unified model.

If supporting different responses for arbitrary headers is too big of a change, just supporting multiple media-types (for request and response) would be sufficient for the cases described here.

nickretallack:
Here's what I want:

```yaml
consumes:
  json: application/json
  protobuf: application/vnd.google.protobuf
produces:
  json: application/json
  protobuf: application/vnd.google.protobuf
paths:
  /things:
    post:
      operationId: post-things
      parameters:
        - name: json
          in: body
          description: JSON payload
          mime: json
          schema:
            type: array
            items:
              type: string
        - name: protobuf
          in: body
          description: Protobuf payload
          mime: protobuf
          schema:
            title: ThingPostMessageProtobuf
      tags:
        - Products
      responses:
        200:
          description: An array of products
          schemas:
            json:
              schema:
                type: array
                items:
                  type: boolean
            protobuf:
              title: ThingResponseMessageProtobuf
```

* Give friendly names to the mime types used in consumes/produces
* Allow multiple different "body" parameters with some property ("mime"?) to differentiate them.
 * This property references the friendly names defined in "consumes".
 * The ContentType header determines which one to use.
* Allow multiple response schemas, keyed by the friendly names in "produces"
 * The Accept header determines which one to use.
* The changes to "consumes" and "produces" also apply to their appearances inside of the Path Objects.

nickretallack:
Or alternatively, to make the "schemas" thing consistent, and show off the other location of consumes/produces:

```yaml
paths:
  /things:
    post:
      operationId: post-things
      consumes:
        json: application/json
        protobuf: application/vnd.google.protobuf
      produces:
        json: application/json
        protobuf: application/vnd.google.protobuf
      parameters:
        - name: data
          in: body
          description: Data
          schemas:
            json:
              type: array
              items:
                type: string
            protobuf:
[...]

PR #637, opened 2016-04-08
added deprecated fields

### LINKED ISSUE #524: [Operation Object] - contain Deprecated field why isn't a [Schema Object] contain such
Hi,

I was able to deprecate the operations using the deprecated field name from the Operation Object.

When I tried to deprecated the model/some fields/properties of a model. The Swagger Spec's Schema Object doesn't seem to have a way to accommodate it.

Any inputs will be helpful. I bumped into this issue on the Springfox project - https://github.com/springfox/springfox/issues/1098

fehguy:
Fixes #524

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -586,6 +586,7 @@ Field Name | Type | Description
 <a name="parameterIn"></a>in | `string` | **Required.** The location of the parameter. Possible values are "query", "header", "path", "formData" or "body".
 <a name="parameterDescription"></a>description | `string` | A brief description of the parameter. This could contain examples of use.  [GFM syntax](https://help.github.com/articles/github-flavored-markdown) can be used for rich text representation.
 <a name="parameterRequired"></a>required | `boolean` | Determines whether this parameter is mandatory. If the parameter is [`in`](#parameterIn) "path", this property is **required** and its value MUST be `true`. Otherwise, the property MAY be included and its default value is `false`. 
+<a name="parameterDeprecated"></a> deprecated | `boolean` | Specifies that a parameter is deprecated and should be transitioned out of usage.
 
 If [`in`](#parameterIn) is `"body"`:
 
@@ -1242,6 +1243,7 @@ Field Name | Type | Description
 <a name="schemaXml"></a>xml | [XML Object](#xmlObject) | This MAY be used only on properties schemas. It has no effect on root schemas. Adds Additional metadata to describe the XML representation format of this property.
 <a name="schemaExternalDocs"></a>externalDocs | [External Documentation Object](#externalDocumentationObject) | Additional external documentation for this schema. 
 <a name="schemaExample"></a>example | Any | A free-form property to include a an example of an instance for this schema.
+<a name="schemaDeprecated"></a> deprecated | `boolean` | Specifies that a schema is deprecated and should be transitioned out of usage.
 
 ##### Patterned Objects 
 


IvanGoncharov:
@fehguy I think `schemaDeprecated` need additional clarification.
Because my first reaction was simular to @webron from [this](https://github.com/OAI/OpenAPI-Specification/issues/524#issuecomment-164625607) comment.

versions/3.0.md
```diff
@@ -1242,6 +1243,7 @@ Field Name | Type | Description
 <a name="schemaXml"></a>xml | [XML Object](#xmlObject) | This MAY be used only on properties schemas. It has no effect on root schemas. Adds Additional metadata to describe the XML representation format of this property.
 <a name="schemaExternalDocs"></a>externalDocs | [External Documentation Object](#externalDocumentationObject) | Additional external documentation for this schema. 
 <a name="schemaExample"></a>example | Any | A free-form property to include a an example of an instance for this schema.
+<a name="schemaDeprecated"></a> deprecated | `boolean` | Specifies that a schema is deprecated and should be transitioned out of usage.
```
ePaul: As mentioned by @IvanGoncharov, this needs more clarification.
You don't actually deprecate a schema, you deprecate a property in a object model (or you deprecate a parameter, or a whole path/operation, but this is a different issue).
The problem (inherited from JSON Schema) is that an object's property consists of just a name and a schema of its value, and thus we can only add stuff to the value schema, as done here.
Maybe add something similar to the "schemaXml" cited above:

> This MAY be used only on properties schemas. It has no effect on root schemas. If used with a value of
> `true`, it specifies that this property is deprecated and should be transitioned out of usage.


An alternative would be to do this similar to `required` and add a property to the containing schema, with a list of deprecated properties.


In addition, when something is deprecated I often feel the need to mention why it is deprecated, and what to use instead. Of course this can be put into the description, but maybe it is better to make this a string property, with the value being the deprecation documentation? (When using the alternative mentioned above, it could be made an object mapping property name → deprecation info, instead of an array of property names.)

IvanGoncharov: > but maybe it is better to make this a string property, with the value being the deprecation documentation?

@ePaul It was discussed here: https://github.com/OAI/OpenAPI-Specification/issues/540
And it's a slippery slope since next thing people ask to implement date of deprecation.
IMHO it's very simple rule of thumb if you can programmatically act based on field value it's one thing.
But if it text for a user to read than description can be used for that. 

One thing that could be added to Swagger spec, is recommendation to add full content of description field into the text of deprecation warning.

ePaul: Thanks for the link. You are right, I guess putting it in the description is enough. (And the recommendation to do so should be here.)

darrelmiller:
Deprecating a model that used for POSTing a request body might make sense if an alternate format is introduced.

ePaul:
> Deprecating a model that used for POSTing a request body might make sense if an alternate format is introduced.

So this is deprecating one of the `oneof`/`anyof` branches? Or how else would an alternate format be introduced? (Maybe with support of #146.)

webron:
@darrelmiller - can you elaborate? Wouldn't the method signature change as well? This may very well be related to what defines a unique operation (including potentially content negotiation as @ePaul mentioned).

:+1: for parameter deprecation.

I understand @ePaul's comment regarding it only being effective in a property definition and not a top-level schema. However, the downside of this is that if there's a model being referenced from multiple places, each referencing place will have to declare it as deprecated. When it comes to reusable component management, this doesn't allow the model to be declared as deprecated so its potential users know not to use it.

So :+1: for model-level deprecation as well, but we should consider the implications on various use cases, and perhaps this would introduce additional limitations to it.

versions/3.0.md
webron: I think such recommendations should be covered outside the spec and within some best practice docs - #589.

ePaul:
@webron Maybe my wording was suboptimal ... in my imagination, top-level schemas become property schemas as soon as they are referenced by some property of another schema. Thus adding the `deprecated` marker in any top-level schema means all properties of this type are deprecated.

fehguy:
I don't see harm in deprecating a model.  I'd like to propose the intent of this PR stays, it applies to models and model properties.

fehguy:

darrelmiller:
👍

jasonh-n-austin:
My only suggestion is to provide a `alternative`, which points to the parameter/field/schema which replaces the deprecated value. With that in mind, it might be wise to make `deprecated` an object (with additional properties beyond `alternative`, like `deprecationDate` etc). 

All that said, I'm a :+1: for 3.0, it seems like a big onion to peel to make it work for every scenario, and the description does the job.

jasonh-n-austin:
One more thing to add: this also leads us into a documentation-only concern. I worry that this isn't useful in any codegen scenarios. IMO I think we need a bigger discussion about where the line is drawn between contract vs documentation format.

ePaul:
@jasonh-n-austin Codegen can generate deprecated-annotations (if the target language supports them), or could have an option to omit all deprecated elements.

fehguy:
closing as approved

earth2marsh:
One advantage to using a date rather than a boolean is that once you know of a planned deprecation in the future, you can indicate when it is being retired.

I don't have a strong opinion on this issue and will abstain from voting.

Issue #588, opened 2016-03-01
Resource types / patterns for Path Item Objects


ralfhandl:
OData services mainly consist of "entity sets" which are collections of "entities" that are uniquely identified by their key property values within their entity set.

So a "read-only entity set" will support
- GET on the set, with a predefined set of query options for paging ($top, $skip), filtering, sorting, free-text search, ...
- GET on the individual entity by specifying the entity set plus the key property values

And "updatable entity sets" will in addition support
- POST on the set to add new entities
- PATCH and DELETE on individual entities

Basically an OData-aware client only needs the entity set name, the structure definition for each entity, and the "readonly/updatable" information to know the path templates and expected request and response bodies, allowed headers, possible response codes etc.

Spelling this out for OData-unaware clients takes a lot of highly repetitive content in the /paths section.

It would be great to have resource types or parameterized templates/patterns for Path Item objects that would allow to specify

/Products:
   ReadOnlyEntitySetOfType
      { $ref: #/definitions/Product }
/People:
  UpdatableEntitySetOfType
     { $ref: #/definitions/Person }

Please note that we foresee (well, actually have :-) services with dozens or hundreds of entity sets, so this would significantly reduce the size of the OpenAPI documents.

webron:
Parent issue: #560

fehguy:
@OAI/tdc removing from OpenAPI.Next proposal and making a future consideration.

webron:
Agreed.

ralfhandl:
See https://github.com/OAI/moonwalk/discussions/30

handrews:
As noted in the previous comment, the best place to continue this is in the Moonwalk repository, in the linked discussion.  Closing out of this repo - @ralfhandl if you think there's something that we might consider here for 3.x please comment and I'll be happy to re-open.

ralfhandl:
Let's tackle it in Moonwalk first and then check whether we want to downport it to 3.x

Issue #581, opened 2016-02-28
Clarify when consumes/produces take effect


ePaul:
The 2.0 specification has the top-level `consumes` and `produces` properties, as well as for each operation.

When the operation is one whose HTTP method's definition doesn't include a payload semantic (most prominently GET, but also HEAD and DELETE, maybe OPTIONS), there is no point in saying the operation "consumes" anything (and there is no point in sending a Content-Type header when there is no body.)

On the operation level thus "consumes" should be forbidden for such operations, and the top-level "consumes" property should be declared in the description to not apply for such operations.

_This was triggered by Issue https://github.com/swagger-api/swagger-codegen/issues/2254._

QianJin2013:
+1

webron:
Not getting into the argument of whether it makes sense to send a payload to a GET but at the end of the day, the HTTP spec does not explicitly prohibits sending a payload with GET operations, and the fact is that some APIs do that. Saying that GET requests don't consume anything is not always true (and also DELETE, HEAD and OPTIONS).

As for the codegen, as long as there's no body parameter, the `consumes` value is meaningless, regardless of which HTTP verb is used (including POST, PUT), and that should be the guiding line.

ePaul:
@webron The guideline "consumes only applies where there is a body parameter" (and maybe in extension also "produces only applies for responses with a schema") sounds good. My proposal is to include an explicit mention of this in the next version of the specification.

webron:
Sure thing, that makes perfect sense, and the part about produces is accurate as well. I've changed the title to reflect the intent here. Note that this is just a clarification and not a change in behavior really - I'd expect existing tools to behave that way with the current version as well, including the codegen.

webron:
Parent issues: #589, #560

webron:
Well, both `consumes` and `produces` have been removed and replaced by better mechanisms for their respective representations.

Issue #577, opened 2016-02-22
Allow documenting HATEOAS APIs (pathless operation / interface / URI class / operation type)


ePaul:
## TLDR

_(added after some discussion)_
- "interfaces" (like path items, just without path parameters, and not linked to a path template)
- a path item can refer to an interface (which is implemented), possibly adding path parameters to it
- a string property (containing an URI) in a schema can refer to an interface (which means the consumer can know which operations can be used on this URI)
## Background / Current Situation

I work at a company where our [Restful API Guidelines](http://zalando.github.io/restful-api-guidelines/index.html) prescribe both ["document your API using OpenAPI"](http://zalando.github.io/restful-api-guidelines/general-guidelines/GeneralGuidelines.html) and ["Use HATEOAS"](http://zalando.github.io/restful-api-guidelines/hyper-media/Hypermedia.html) ("Hypertext as the Engine of Application State", one of the core principles of REST).

Unfortunately both do not work together nicely:
- Swagger/OpenAPI 2.0's model is "we have paths which can be distinguished in some way, and describe the operations on those paths" (including their data format).
- The HATEOAS model is "The client gets a result which might contain links to other resources, and can decide to follow them, no matter which URI format they have".

Currently the intersection between both is "have links in your result, but just to URIs which are also described as paths in your Swagger definition, and describe in descriptions what kind of URIs you can expect".
## Ideas

A core point of HATEOAS are media types and link relations. An API definition should mainly be a description of the media types and what operations they imply for links included in them. (Paraphrased from a [blogpost of Roy Fielding](http://roy.gbiv.com/untangled/2008/rest-apis-must-be-hypertext-driven).)

Media type definitions are analogous to what OpenAPI's `definitions` are currently, might even be an extension of that. For models with a property of type `string`, format `uri` we can then additionally refer to the _URI class_ of that URI.

A **URI class** is defined analogously to the current path definitions, but with some name to be used in model definitions. The URI class defines what operations are possible on an URI, which parameters are allowed/needed (no path parameters, I guess), and what can be expected in return. (The URI class is not a property of the URI itself, but of its usage in a specific place. The same URI could appear with multiple URI classes.)

_I'm not yet sure how link relations can come into this – I guess we might need a way to define the URI class depending on link relation or similar._

[...]

Issue #555, opened 2016-02-05
Message format 'wrappers'


jharmn:
Some existing hypermedia formats (such as JSON API, HAL, JSON Collection) define a standard 'wrapper' for responses (not unlike the old SOAP envelope, in some ways). 

Current OpenAPI conventions are inconvenient and very redundant in describing these formats. An OpenAPI method of describing APIs with these formats would produce specs that are much DRYer.

https://github.com/OAI/OpenAPI-Specification/issues/136
https://github.com/OAI/OpenAPI-Specification/issues/519
https://github.com/OAI/OpenAPI-Specification/issues/522
https://github.com/OAI/OpenAPI-Specification/issues/476

Solutions are not described here yet, but this should serve as a point to consolidate concepts from prior feedback when evaluating potential solutions.

@darrelmiller input here would be great.

dilipkrish:
@darrelmiller is probably more qualified to comment on this, but I'd like to suggest addressing what I believe are gaps in the spec:
1. being able to describe link relations
2. being able to describe additional response headers.
3. being able to describe models in other dialects xml, bson, protobuf etc.

Just throwing an ideas out here for problem # 1 based on the pet store example, please excuse if its a bit contrived

```yaml
  /pets:
    get:
      description:  description....
      operationId: findPets
      parameters:
        - name: tags
          in: query
          description: tags to filter by
          required: false
          type: array
          collectionFormat: csv
          items:
            type: string
        - name: limit
          in: query
          description: maximum number of results to return
          required: false
          type: integer
          format: int32
      responses:
        200:
          description: pet response
          headers: 
             # Location in the case of create resource  
             - name: "Location"
               description: Location of created resource
               type: string
             # Pagination headers
             - name: "totalPages"
               description: total number of pages
               type: integer
               format: int32
          body:
            type: array
            items:
              $ref: '#/definitions/Pet'
             # each link is a urlRef to another Path object, with a specific method
            links: 
             - rel: "create"               
               $urlRef: "#paths/pet"
               method: "post" 
             - rel: "delete"               
               $urlRef: "#paths/pet" 
               method: "delete" 
          ...
```

Also I find it odd, that we're trying to describe other _rich media types_ within swagger. inception? :smile: If an api is using json-api or HAL, CJ, Apiary's Api Blue Print etc. for that matter; is there a need to describe it via swagger as well? I suspect there will be [transformers](https://apitransformer.com/#formats) for those cases in the future.

[...]

PR #640, opened 2016-04-08
added notes for additional issues

fehguy:
FOR DISCUSSION ONLY, DON'T MERGE

Addresses:
#214, #469 
#169 (move to different parent)
#182 (move to different parent)
#213 
#239 (move to different parent)
#208 
#445 
#577 
#555 
#541 
#508 
#581 
#588 
#132 
#146 (move to different parent)
#341

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -1,5 +1,133 @@
 # OpenAPI Specification
 
+### DO NOT MERGE
+
+These are comments for discussion and should be turned into separate PRs
+
+#### NOTES
+
+## Both HTTP and HTTPS ports cannot be specified #214
+
+Proposal: We have schemes, host (+ port), basePath, all which are typically linked to gether.  Suggest we produce a new object:
+
+```yaml
+hosts:
+  - scheme: http
+    host: foo.com
+    port: 8080
+    basePath: /bar
+  - scheme: https
+    host: foo.com
+    port: 8443
+    basePath: /baz
+```
+
+And remove the separate `host`, `basePath`, `schemes`.  Because the port will never be the same between http, https, it makes no sense to support multiple ones
+
+Fixes #489, #214
+
+
+## Why don't host and basePath support path templating? #169
+
+- [ ] Proposal: this belongs under parent #574
+
+## Support an operation to have multiple specifications per path (e.g. multiple POST operation per path) #182
+
+- [ ] Proposal: this is a payload issue, move to parent #586
+
+## Allow for overloading/aliasing of route path #213
+
+- [ ] Proposal: change in #70 would address this
+
+## Allow references between operations inside descriptions. #239
+
+- [ ] Proposal: this is really about _links_ and belongs in the #586 issue
+
+## Swagger.next: Let's not use hashes anymore #208
+
+- [ ] Proposal: while true, moving to arrays causes more problems than it solves.  Suggest we close this with no change.
+
+## Group multiple parameter definitions for better maintainability #445
+
+- [ ] Proposal: close.  Using global parameters will help with this, and references to arrays of parameters is just wrong.a
+
+
+## Allow documenting HATEOAS APIs (pathless operations) #577
+
+- [ ] Proposal: do not support pathless operations.  It's against the spirit of the specification as being declarative.  The support for links _inside_ a path though, will be covered by #574 and #586.  Remove reference to #560 and link to others.
+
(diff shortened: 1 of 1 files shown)


versions/3.0.md
```diff
@@ -1,5 +1,133 @@
...
+
+- [ ] Proposal: this is a payload issue, move to parent #586
+
+## Allow for overloading/aliasing of route path #213
+
+- [ ] Proposal: change in #70 would address this
+
+## Allow references between operations inside descriptions. #239
+
+- [ ] Proposal: this is really about _links_ and belongs in the #586 issue
```
ePaul: This is about links in the documentation, not about links in the payload. I don't see how this relates.

versions/3.0.md
```diff
@@ -1,5 +1,133 @@
...
+- [ ] Proposal: while true, moving to arrays causes more problems than it solves.  Suggest we close this with no change.
+
+## Group multiple parameter definitions for better maintainability #445
+
+- [ ] Proposal: close.  Using global parameters will help with this, and references to arrays of parameters is just wrong.a
+
+
+## Allow documenting HATEOAS APIs (pathless operations) #577
+
+- [ ] Proposal: do not support pathless operations.  It's against the spirit of the specification as being declarative.  The support for links _inside_ a path though, will be covered by #574 and #586.  Remove reference to #560 and link to others.
```
ePaul: As this is my proposal, I'd like to get a bit more explanation for your "against the spirit" judgement.

How is an operation declaration (without declaring at which path such an operation might be available) not declarative?

(I can see that generating a server implementation from such a definition is difficult, but that is a different reasoning.)

fehguy: If the server is self-documenting and has no predefined paths, you don't need OAS IMHO.  "/" is a root path.

ePaul: (I guess we should discuss this at #577 instead of here?)

versions/3.0.md
```diff
@@ -1,5 +1,133 @@
...
+    basePath: /bar
+  - scheme: https
+    host: foo.com
+    port: 8443
+    basePath: /baz
+```
+
+And remove the separate `host`, `basePath`, `schemes`.  Because the port will never be the same between http, https, it makes no sense to support multiple ones
+
+Fixes #489, #214
```
IvanGoncharov: @fehguy Please also consider #641 since it's tightly connected to this change.

versions/3.0.md
```diff
@@ -1,5 +1,133 @@
...
+- [ ] Proposal: this is really about _links_ and belongs in the #586 issue
+
+## Swagger.next: Let's not use hashes anymore #208
+
+- [ ] Proposal: while true, moving to arrays causes more problems than it solves.  Suggest we close this with no change.
+
+## Group multiple parameter definitions for better maintainability #445
+
+- [ ] Proposal: close.  Using global parameters will help with this, and references to arrays of parameters is just wrong.a
+
```
webron: I think we can easily support it with the new construct proposed at #663, and it would make people happy without complicating things. Have parameterGroup as an array of parameters or refs to parameters. Only thing to decide is what happens with conflicts. Granted, it may be a bit more challenging for tools.
Prominent example, paging parameters that are not needed globally but can be reused for quite a few operations.

versions/3.0.md
```diff
@@ -1,5 +1,133 @@
...
+  /health:
+    $ref: '#/x-paths/health'
+  /store:
+    get:
+      ...
+  /users:
+    get:
+      ...
+```
+
```
webron: This requires further discussion.

We might as well just allow $ref’s, and multiple of them. We already break the $ref today with the Operation Object (and there’s currently no issue to change that). I’m generally in favor of breaking from $ref’s restrictions, and it makes for a much cleaner solution. If the concern is about $refs disappearing for being multiple keys, we can also have a single $ref with an array of strings. We can even call it something else like `references` to not be confused with `$ref`, and that would be rendered by the tools as multiple `$ref` occurrences. 

Using JSON Pointers as keys is odd, and it’s actually not a JSON Pointer if I understand correctly the usage, it’s the value of a JSON Reference which is not the same. Seems hack-ish and would be a pain to document in the spec.

fehguy: Please propose an alternate solution then

fehguy:
Closing as issues, PRs have been created, taking into account feedback on this.

Issue #641, opened 2016-04-09
Add support for different execution environment, e.g. sandbox


IvanGoncharov:
Many APIs have multiple environments for execution of the same calls.
Examples are sandbox, geographical locations(to minimise latency), different instances of the same service, etc.

Currently, it impossible to describe multiple environments in the same document but @fehguy proposal from #640 to add array of hosts to spec:
```yaml
 hosts:
   - scheme: http
     host: foo.com
     port: 8080
     basePath: /bar
   - scheme: https
     host: foo.com
     port: 8443
     basePath: /baz
 ```
It seems like good solution for this problem, but it missing few essential details:
 * `title` field
 * `description` field
 * some mechanism to mark default host

webron:
I assume both `title` and `description` would be optional.

IvanGoncharov:
@webron Yes, definitely. It would be really strange to produce different titles for `HTTP` and `HTTPS` hosts.

IvanGoncharov:
@webron I also think if you introduce support for multiple hosts without marking default.
Previously it was choose between HTTP and HTTPS so it was done by tools automatically.
Simplest possible solution would be to add `default` property to elements of `hosts`.
I see only one problem with it: JSON Schema can't validate that it set to `true` for a single host.

webron:
JSON Schema constraints shouldn't dictate the spec. JSON Schema cannot and will not be able to fully validate the spec anyways, so it's a non-issue.

PR #646, opened 2016-04-11
added host section

fehguy:
Fixes issue #214

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -108,9 +108,7 @@ Field Name | Type | Description
 ---|:---:|---
 <a name="oasSwagger"></a>swagger | `string` | **Required.** Specifies the OpenAPI Specification version being used. It can be used by the Swagger UI and other clients to interpret the API listing. The value MUST be `"2.0"`.
 <a name="oasInfo"></a>info | [Info Object](#infoObject) | **Required.** Provides metadata about the API. The metadata can be used by the clients if needed.
-<a name="oasHost"></a>host | `string` | The host (name or ip) serving the API. This MUST be the host only and does not include the scheme nor sub-paths. It MAY include a port. If the `host` is not included, the host serving the documentation is to be used (including the port). The `host` does not support [path templating](#pathTemplating).
-<a name="oasBasePath"></a>basePath | `string` | The base path on which the API is served, which is relative to the [`host`](#oasHost). If it is not included, the API is served directly under the `host`. The value MUST start with a leading slash (`/`). The `basePath` does not support [path templating](#pathTemplating). 
-<a name="oasSchemes"></a>schemes | [`string`] | The transfer protocol of the API. Values MUST be from the list: `"http"`, `"https"`, `"ws"`, `"wss"`. If the `schemes` is not included, the default scheme to be used is the one used to access the OpenAPI definition itself.
+<a name="oasHosts"></a>hosts | [Hosts Object](#hostsObject) | An array of Host objects which provide `scheme`, `host`, `port`, and `basePath` in an associative manner.
 <a name="oasConsumes"></a>consumes | [`string`] | A list of MIME types the APIs can consume. This is global to all APIs but can be overridden on specific API calls. Value MUST be as described under [Mime Types](#mimeTypes).
 <a name="oasProduces"></a>produces | [`string`] | A list of MIME types the APIs can produce. This is global to all APIs but can be overridden on specific API calls. Value MUST be as described under [Mime Types](#mimeTypes).
 <a name="oasPaths"></a>paths | [Paths Object](#pathsObject) | **Required.** The available paths and operations for the API.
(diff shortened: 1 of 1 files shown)


fehguy:
Moved from https://github.com/OAI/OpenAPI-Specification/pull/640.

versions/3.0.md
```diff
@@ -108,9 +108,7 @@ Field Name | Type | Description
...
-<a name="oasHost"></a>host | `string` | The host (name or ip) serving the API. This MUST be the host only and does not include the scheme nor sub-paths. It MAY include a port. If the `host` is not included, the host serving the documentation is to be used (including the port). The `host` does not support [path templating](#pathTemplating).
-<a name="oasBasePath"></a>basePath | `string` | The base path on which the API is served, which is relative to the [`host`](#oasHost). If it is not included, the API is served directly under the `host`. The value MUST start with a leading slash (`/`). The `basePath` does not support [path templating](#pathTemplating). 
-<a name="oasSchemes"></a>schemes | [`string`] | The transfer protocol of the API. Values MUST be from the list: `"http"`, `"https"`, `"ws"`, `"wss"`. If the `schemes` is not included, the default scheme to be used is the one used to access the OpenAPI definition itself.
+<a name="oasHosts"></a>hosts | [Hosts Object](#hostsObject) | An array of Host objects which provide `scheme`, `host`, `port`, and `basePath` in an associative manner.
```
webron: Should be `Host Object`

versions/3.0.md
```diff
@@ -183,6 +181,18 @@ license:
 version: 1.0.1
 ```
 
+#### <a name="hostObject"></a>Host Object
```
webron: Anchor should be `hostObject`

webron:
:+1:

This should also also affect the same fields as described at https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#operationObject.

versions/3.0.md
fehguy: I figure there is a big documentation update once we agree on the merges, will make sure this is taken into account.

darrelmiller:
👍

whitlockjc:
LGTM

jasonh-n-austin:
The biggest issue I see with this:
- Codegen: how does a client know which host to call? It seems like there are two options to make this useable:
  - `default`: which host is used if not specified
  - `name`/`key`: an arbitrary name for the environment, so the client can specify which environment they intend to call.

IMO in Swagger 2.0, `host` is useless in most situations. The client has to use some other system (dns, consul, etcd etc) to discover the correct host for the given environment. Adding an ambiguous set of hosts for any given environment makes it impossible to know which one to call programmatically.
Additionally, spec publishers really don't want to expose the internal hostnames of their dev/qa/staging environments in their public specs (granted, this could potentially be stripped with preprocessing, but would require `name`/`key`).

:-1: as it stands

IvanGoncharov:
@jasonh-n-austin Please take a look at #641 it should addresses all your concerns.
> name/key: an arbitrary name for the environment, so the client can specify which environment they intend to call.

I think best ID is URL generated from this element, so user can choose based on it.
At the same time it would be cool to support optional `title` and `description` to describe non-standard hosts like Sandbox.

> IMO in Swagger 2.0, host is useless in most situations. 

I strongly disagree, it's very useful for public APIs like Instagram, Facebook, etc.

> Additionally, spec publishers really don't want to expose the internal hostnames of their dev/qa/staging environments in their public specs (granted, this could potentially be stripped with preprocessing, but would require name/key).

In that case you can strip based on some kind of vendor extension, e.g. `x-internal: true`.

darrelmiller:
I didn't realize the intent of the hosts array was to enable the identification of different environments.  I thought it was for enumerating identical service instances.  Therefore it wouldn't matter which host you chose, the effect would be the same.  This would allow the client to pick a closest server, or round robin requests as well as pick the scheme they want to use.

Making OpenAPI responsible for identifying multiple different environments seems like a fairly significant increase in responsibility and certainly opens the questions about default environment and which hosts are equivalent and which are not.  Are we sure we want to go there?  The original issue was not trying to solve that problem.

webron:
@darrelmiller - that's a fair point. However, that's not really the *intent*. I think that's the difference between use and abuse. The *spec* shouldn't say it should be used for that or intended to be used for that, but users may end up doing so - we can't stop it.

I agree the goal here is just to be able to fine-tune the host definition is which is a bit limited today.

fehguy:
@jasonh-n-austin my view is that the hosts object is an ordered list, we can make it clear that there is an implicit order of precedence.

Given that and the other votes, I'm merging this and we can always revisit.

sidhantdas007:
can someone just give an example of how they have implemented the multiple swagger host? I am a bit new to swagger

MikeRalphson:
@sidhantdas007 this can only be done natively in OpenAPI 3.0 (the successor to the Swagger specification). A simple example: https://github.com/APIs-guru/openapi-directory/blob/openapi3.0.0/APIs/bbci.co.uk/1.0/openapi.yaml#L1-L5

You can mention additional `host`s in OpenAPI 2.0 only by using specification (vendor) extensions, but this will have no support in tooling other than your own.

Issue #629, opened 2016-04-07
Swagger.Next: Forbid properties other than $ref inside JSON Pointer


IvanGoncharov:
From [JSON Reference](https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03):

> Any members other than "$ref" in a JSON Reference object SHALL be  ignored.
> 
> Implementations MAY choose to replace the reference with the referenced value.

So technically JSON Reference doesn't forbid this behaviour, but it only makes the situation worse.
Since people expect different behaviour and it split lib/tools into two camps.
Ones who replace the reference and one who iterates through `$ref`s.

@whitlockjc I believe you have some background on this issue.

IvanGoncharov:
+ it really easy to implement such check in JSON Schema just add `additionalProperties: false`

whitlockjc:
:+1:

webron:
This is pretty much the opposite of #556 and honestly I'd rather see #556 implemented than this, for the value it provides. I mean, both this and #556 don't follow the JSON Reference spec as it is (which is fine by me), but if we break away, why not make it more useful?

webron:
Parent: #579

IvanGoncharov:
@webron This issue is really different from #556 
This issue is compatible with JSON Pointer, OpenAPI is still could be resolved by tools intended for JSON Pointer. It's intended just a way to prevent user from unexpected behavior, so it just about validation.

In any case I think JSON Pointer compatibility is more important than this two issues put together.

webron:
This issue is *incompatible* with JSON Pointer because it adds a restriction that JSON Pointer doesn't have. If the JSON Pointer's spec writer wanted to disallow extra properties, they would have used different wording in the spec. I believe the wording there was intentional and we should not ignore it. And considering this feature wants to restrict JSON Pointer more than it is, and #556 calls for the opposite...

fehguy:
I'm removing the meta issue from this--it's really part of our larger discussion about JSON schema support.

webron:
Yeah, the `Meta Issue` tag was by mistake, meant to add the openapi.next.

webron:
Tackling PR: #741

webron:
We've decided to follow the JSON Reference spec for our own Reference Object, and as such, allow additional properties, but saying they will be ignore per the spec.

handrews:
[...]

Issue #556, opened 2016-02-08
Extra JSON Reference properties ignored - reduces reusability of references


EvilJimJafar:
Hi everyone.

I feel that the ignoring of extra properties when using JSON References goes against the goal of reusability.

Take this example from a fictional Swagger contract:

```
{
    addresses: [
        "homeAddress": {
            "description": "The premises at which electricity is provided.",
            "$ref": "#/definitions/Address"
        },
        "invoiceAddress": {
            "description": "The billing address - must be where the customer's credit card is registered.",
            "$ref": "#/definitions/Address"
        }
    ]
}
```

The "description" properties help to distinguish between the two uses of the "Address" Reference. Of course you could argue that better naming of the "homeAddress" and "invoiceAddress" properties could achieve this, but we are not always in control of property names in our data models and sometimes you need a more verbose description.

The Swagger editor (http://editor.swagger.io/#/) is an example of an application that could be improved by removing this limitation.

Is this something that has been raised previously? Please see here for a discussion on the Swagger Google Group: https://groups.google.com/forum/#!topic/swagger-swaggersocket/ewgimdO2cOI

webron:
Currently, we use JSON References as-is. While there's a huge advantage to that tooling-wise, it does hinder reusability in some cases, and the issue has been raised (unofficially) a few times. Overriding the `description` is one case, but there are also cases of wanting to override whether fields are required and so on. Not sure if we should change it, given the cost, but we should consider it.

qcho:
I think this is a very good Proposal.
In my example I want to set:
```
      parameters:
        - $ref: '#/parameters/fields'
          default: 'local_default'
```

whitlockjc:
I would be willing to update [json-refs](https://github.com/whitlockjc/json-refs) to have a flag that allowed you to handle this situation.  Based on OpenAPI feature requests, I can see a few different things we'd need:

* Merging keys, like the example above
* Overriding/Replacing keys, possibly like the example below

The JSON Reference spec clearly says these should be ignored.  But...there's nothing stopping json-refs from allowing extra functionality on top in an opt-in manner.

whitlockjc:
Realize, this has nothing specific to do with OpenAPI other than a number of the Node.js tools around OpenAPI use json-refs.  A decision on either side does not have to reflect or impact the ability/decision of the other.  I hope that is clear.

fehguy:
[...]

PR #650, opened 2016-04-12
Issue #577: interface for paths

ePaul:
This draft implementation of #577 simply reuses the $ref property of a path item – not sure if that is the right thing to do, or if we should use a new `interface` property instead (like for the schemas).

This is also still missing the inclusion of the "interfaces" object in either the main object, or the schemas/definitions object (depending on #633).

I opened this pull request, so the discussion in #577 has something concrete to talk about.


ePaul:
After reading the [comment by @webron in #536](https://github.com/OAI/OpenAPI-Specification/pull/536#issuecomment-170695322) about $ref in path items, I guess I should avoid doing the same here. Though #556 and #629 have some more discussion on that.

fehguy:
This is covered in another ticket and recommends invalid an invalid structure for JSON references.  Closing...

Issue #222, opened 2014-12-02
Allow different Content Types in File type parameters


thorntech:
For multipart uploads that contain a file and some other part, like a JSON object, it would be very useful to be able to specify a content type on the File parts other than the actual binary file. This is critical for Java applications that need the content type of the part in order to know how to deserialize the json or xml into a Java object. 

For what its worth, I was able to get this working by modifying swagger-ui,js to use a Blob as the FormData. I had to add a "contentType" field to the Parameter object. I know this violates the spec, but this worked for my project. 

swagger-ui.js line 1797:

```
if (param.paramType === 'form') {
      if (param.type.toLowerCase() !== 'file' && map[param.name] !== void 0) {
        if(typeof param.contentType !== 'undefined') {
            bodyParam.append(param.name, new Blob([map[param.name]], { type: param.contentType}));
        }
        else {
            bodyParam.append(param.name, map[param.name]);
        }
      }
    }
```

In the swagger doc, I can define the file and the json object parameters like so: 

```
        {
         "name": "story",
         "description": "The story object in json format",
         "required": true,
         "paramType": "form",
         "type": "Story",
         "contentType": "application/json"
        },
        {
         "name": "file",
         "description": "The base64 encoded multipart file attachment. ",            
         "required": true,
         "paramType": "form",
         "type": "File"
        }
```

Refer to this discussion / issue in swagger-core for more detail - https://github.com/swagger-api/swagger-core/issues/636

christopheblin:
> I'll just add and say that for now vendor extensions can be used to extend this functionality

Could you elaborate on this please.
I'm having the same problem but I have no clue what you mean with this sentence...

webron:
@christopheblin - of course. If you take the file definition from above as an example:

```json
        {
         "name": "file",
         "description": "The base64 encoded multipart file attachment. ",            
         "required": true,
         "paramType": "form",
         "type": "file",
         "x-mimetype": "image/jpeg"
        }
```
This could indicate that the specific file parameter needs to be of an `image/jpeg` mime type.

webron:
Parent: #565

jmdacruz:
This is a very common pattern on REST APIs when you have to upload some binary content and its related metadata in JSON. In those cases, it is desirable to be able to define:
[...]

Issue #595, opened 2016-03-10
Query parameters objects (like params[variable]=toto, ....)


Nox-404:
Hi,
I'm gonna be a bit lazy and just post it back here, so here is a question/suggestion/request

Is there a way to describe object parameters like this :
- filter[name]=a
- filter[ip]=10.
- ...

What i got :
![Before](http://krondor.fr/img/actual.png)

What i need :
![Before](http://krondor.fr/img/new.png)

> The field between the brackets should be able to be a select (enum) or an input field
> 
> And when one is filled, it should add another one beneath it (and with an enum it should remove or not the selected value (configurable))

I originally posted it [there](https://github.com/swagger-api/swagger-ui/issues/2027)

whitlockjc:
I think this is more about tooling support.  The OpenAPI specification allows you to describe parameters the way you requested already so nothing needs to change.  Since you've already created an issue in swagger-ui *(https://github.com/swagger-api/swagger-ui/issues/2027)*, that's the best course of action.

Since you were already talking to @webron in the swagger-ui issue, I'll let him decide how to handle this issue.

webron:
If I understand what @Nox-404 is asking, that's not really supported by the spec now. Query parameter *names* are not parameterized and can't be controlled that way. You can define an enum for the values, but not for the name.

What concerns me here is the control on the type of the parameter. So if you have `name` it'd be a string, but if you have `age` it'll be an integer? Not sure how we'd represent that elegantly.

Nox-404:
@webron yes that's what i mean, and i don't have any idea on how to describe it yet too

webron:
Parent: #565

whitlockjc:
I don't understand how this can't be done by OpenAPI already.  Take a look at https://github.com/apigee-127/swagger-tools/issues/294 and you'll see that people are already doing this, although it's working due to the fact that the query string parsing library supports it.  But in the referenced issue, you'll see how some are already doing this and it was a tooling issue to support it.

webron:
That's a fair point. It could be a display issue only. Tools could have a way to collect `filter[*]` entries under one roof, though again, there's a problem with the description and the type. 

It could also be a convenience issue - instead of having to repeat the same definition for each filter with a fixed name, you can have one with an enum of 'types' for the filter. Not sure this is something we need to support, but it's fair to consider.

raymondfeng:
We have to workaround the issue today by passing stringified json to complex query parameters, for example:
```
[...]

Issue #437, opened 2015-08-10
json array support for collectionFormat


dtorok:
The spec lists several possible formats in the `collectionFormat` field for an array parameter.
My suggestion is to include `json` too because
- we use json everywhere, no need to introduce a new serialization format here
- json serializer is there already on both the client and server side
- it's as simple as csv (2 more bytes)

The swagger fragment would look like this:

``` json
        - name: id__in
          in: query
          required: false
          type: array
          collectionFormat: json
          items:
            type: integer
            format: int64
```

The URL would look like this:

``` url
https://example.com/api/users/?id__in=[1,2,3]
```

mohsen1:
Are you proposing that `items` can be any JSON Schema also?

webron:
We actually considered adding support for it around 2.0, but a substantial set of libraries that actually support this format.

jasonh-n-austin:
Link to parent #565

fehguy:
This has been added by https://github.com/OAI/OpenAPI-Specification/pull/804

Issue #412, opened 2015-07-15
Multidimensional arrays as parameters.


IvanGoncharov:
Currently Swagger 2.0 support specs like that:

``` json
{
    "swagger": "2.0",
    "info": {
        "version": "0.0.0",
        "title": "Simple API"
    },
    "paths": {
        "/{foo}": {
            "parameters": [
                {
                    "name": "foo",
                    "in": "path",
                    "type": "array",
                    "items": {
                      "type": "array",
                      "items": {
                        "type": "string"
                      }
                    }
                }
            ],
            "get": {
                "responses": {
                    "200": {
                        "description": "OK"
                    }
                }
            }
        }
    }
}
```

By default `collectionFormat` is `csv` and this create problems with interpreting strings like 'a,b,c,d'.
Should it be interpreted like:

``` json
[
  ["a","b","c","d"]
]
```

Or maybe like this:

``` json
[
  ["a", "b"],
  ["c", "d"]
]
```

It should be forbidden to create nested arrays with same `collectionFormat`.
This change will prevent undefined behaviour situation and could be done in `2.0` as bugfix.
JSON Schema couldn't enforce such constraints, but it could be specified in spec.

jasonh-n-austin:
Link to parent #565

fehguy:
Done in https://github.com/OAI/OpenAPI-Specification/pull/804

ePaul:
@fehguy I don't actually see this in the specification ... where do we forbid this (which would be the corresponding structure to what @IvanGoncharov posted):

```yaml
parameters:
  - in: path
    name: foo
    schema:
      type: array
      items:
        type: array
        items:
          type: string
    example: 'a,b,c,d'
    required: true
```

fehguy:
we don't forbid this, we support it by allowing a schema object as the type for the parameter.

ePaul:
The original issue said

> It should be forbidden to create nested arrays with same `collectionFormat`.

Now there isn't a `collectionFormat` anymore, but the `style` property, which only applies on top level.

So for `style: simple` (which is the default for path parameters) there is still no way of knowing whether `/a,b,c,d` would be parsed as `[["a","b"],["c","d"]]`, as `[["a","b","c","d"]]` or as any other structure.

There doesn't even seem to be a way of using different styles for both levels of array (which could solve this potentially).

So I think the issue is still open.

PR #654, opened 2016-04-15
added schema for all parameters

fehguy:
This is NOT a complete change, but conceptually it loosens the requirement that all parameters except `body` parameters be of a simple type.  Instead, a schema can be provided for all parameter types.

For parameter types of type `body`, the schema will be serialized by the `consumes` attribute.  For all other parameter types, a serialization strategy needs to be defined (and is still TBD for definition).

This serialization strategy would be responsible for turning a complex object into JSON, for example.

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -591,6 +591,8 @@ There are five possible parameter types.
   * `application/x-www-form-urlencoded` - Similar to the format of Query parameters but as a payload. For example, `foo=1&bar=swagger` - both `foo` and `bar` are form parameters. This is normally used for simple parameters that are being transferred.
   * `multipart/form-data` - each parameter takes a section in the payload with an internal header. For example, for the header `Content-Disposition: form-data; name="submit-name"` the name of the parameter is `submit-name`. This type of form parameters is more commonly used for file transfers.
 
+For complex parameter schemas, a serialization strategy is required. For parameter of type `in: body`, the serialization will be handled by the `consumes` attribute.  For all other types, a serialization strategy must be declared.
+
 ##### Fixed Fields
 Field Name | Type | Description
 ---|:---:|---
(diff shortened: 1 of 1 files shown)


fehguy:
This addresses #412, #437, #69, #401, #179 (by inclusion of the `schema`), #301 

For issue #222, we can also add a format for `file` to indicate the encoding type.

For issue #300, since schemas can be reused, this should address that as well.

This solves #595 if we can define the serialization strategy

versions/3.0.md
```diff
@@ -599,35 +601,7 @@ Field Name | Type | Description
...
-<a name="parameterMaxLength"></a>maxLength | `integer` | See http://json-schema.org/latest/json-schema-validation.html#anchor26.
-<a name="parameterMinLength"></a>minLength | `integer` | See http://json-schema.org/latest/json-schema-validation.html#anchor29.
-<a name="parameterPattern"></a>pattern | `string` | See http://json-schema.org/latest/json-schema-validation.html#anchor33.
-<a name="parameterMaxItems"></a>maxItems | `integer` | See http://json-schema.org/latest/json-schema-validation.html#anchor42.
-<a name="parameterMinItems"></a>minItems | `integer` | See http://json-schema.org/latest/json-schema-validation.html#anchor45.
-<a name="parameterUniqueItems"></a>uniqueItems | `boolean` | See http://json-schema.org/latest/json-schema-validation.html#anchor49.
-<a name="parameterEnum"></a>enum | [*] | See http://json-schema.org/latest/json-schema-validation.html#anchor76.
-<a name="parameterMultipleOf"></a>multipleOf | `number` | See http://json-schema.org/latest/json-schema-validation.html#anchor14.
+<a name="parameterSchema"></a>schema | [Schema Object](#schemaObject) | The schema defining the type used for the parameter.
```
ePaul: Allow a reference object (i.e. a reference to a schema object) here too?

versions/3.0.md
```diff
@@ -713,11 +689,12 @@ name: token
...
-items:
-  type: integer
-  format: int64
-collectionFormat: csv
+schema:
+  type: array
+  items:
+    type: integer
+    format: int64
+  collectionFormat: csv
```
fehguy: with this change, `collectionFormat` may not make sense.  It should be added to the serialization strategy

fehguy:
need to reconsider this when looking at 6570 support

fehguy:
@OAI/tdc please post other comments here--i will not merge until a serialization strategy is provided and there is more time for feedback

versions/3.0.md
webron: @ePaul - The `Schema Object` allows a `$ref` as its value, so there's no need to be explicit about it. It's the same as how `body` parameters behave today.

webron:
This does _not_ solve #69 nor #401.

To me the serialization strategy can be controlled by a `consumes` property at each parameter, which will override any upper-level `consumes` value (either at the operation or OpenAPI Object level). This will also help resolve #222. This gets tricky since primitives cannot really be serialized as `application/json` for example, so we may explicitly mention that primitives are passed as-is.

To me, the main issue is arrays and the existing `collectionFormat`. No mime-type will actually cover the existing cases in it. Brain is a bit fried right now, the only solution I can come up with right now is fairly ugly - will try to think about it more later.

I do _not_ advocate using complex models for query/path/header parameters (body and form are okay) - but in favor of simplicity I :+1: the concept. Just need to work through the open questions.

fehguy:
* Move type definitions out of serializable parameters
* Add a `schema` to all
* come up with a serialziation strategy for all non-body parameters

Vanuan:
Is there any rough implementation?

I'm looking for a simple query parameter of type object serialized into a JSON string.

ePaul:
@Vanuan I'm not aware of any. We are currently working on the next version of the specification (3.0), and only if that it out, the tools can reliably be updated to support it. (I guess some tools will start implementation earlier, though.)

Vanuan:
Is there some criteria or time frame for 3.0 to be released?
I mean, are you working towards supporting describing specific APIs, features or do you plan to release new spec each year?
For example you could say that in 3.0 you'd want to support describing such APIs:

- Facebook https://developers.facebook.com/tools/explorer/
- Twitter
- Sailsjs blueprint API http://sailsjs.org/documentation/reference/blueprint-api

Or you could say that you target supporting specific features, like an ability to specify fields like this:
?fields=id,name
or specifying objects in queries like this:
?filter={"name":{"includes":"john"}}

Is GitHub issues is the only place or is there some highlevel roadmap with milestone?
In Readme it's mentioned that OpenAPI doesn't plan to support every REST API use case. Does it mean that applications would need to adapt to OpenAPI features to fully employ its benefits?

webron:
@Vanuan we're aiming for a summer release for 3.0. The changes are done via specific feature requests and not aiming to support specific APIs.

Github issues is the place.

It means that not all APIs could be described with OpenAPI. Whether applications change their APIs or don't use OpenAPI is their call.

Issue #301, opened 2015-03-09
Standardize Parameter Models


dilipkrish:
It seems odd that ```Serializable``` parameters be treated differently that Body Parameters. In the interest of _consistency_, I would think that regardless of the type of parameter it should always have a **schema** property.

```yaml
"/api/pet/findByStatus": {
            "get": {
                ...
                "summary": "Finds Pets by status",
                "description": "Multiple status values can be provided with comma seperated strings",
                ...
                "parameters": [
                    {
                        "name": "status",
                        "in": "query",
                        "description": "Status values that need to be considered for filter",
                        "required": true,
                        "type": "string", //<-- NOTE: Seems like this should be standardized
                        "default": "available"
                    }
                ],
```
I would expect it to be rendered as shown below
```yaml
"/api/pet/findByStatus": {
            "get": {
                ...
                "summary": "Finds Pets by status",
                "description": "Multiple status values can be provided with comma seperated strings",
                ...
                "parameters": [
                    {
                        "name": "status",
                        "in": "query",
                        "description": "Status values that need to be considered for filter",
                        "required": true,
                        "schema": {
                            "type": "string"
                        }, 
                        "default": "available"
                    }
                ],
```

webron:
This was done intentionally to avoid allowing to define objects as non-body parameters. It would probably change for some types of form parameters but is unlikely to change for path/query/header parameters.

dilipkrish:
I see, yet still it seems like an un-needed branching logic we have to deal with. Its seems pretty obvious that one cannot have objects in query/path/header and cookies. From a consumption-tooling perspective I'd prefer to not have different logic based on the type of parameter and rely on a schema property that will give information about the model.

webron:
It's not really that obvious. In 1.2, you technically *could* have models as types for query and path parameters, which pretty much lead to this structure. 

I don't disagree the divergence is inconvenient for tooling (and iirc, I wasn't too happy about it when we decided to do that in 2.0), but sometimes it's about finding the balance between helping the tools and helping people create proper specs. 

I tend to think it's more important to protect from human-errors than make tooling-development easier. It's all in the numbers.

MrTrick:
There is a valid concern about ensuring people don't define impossible things. However, within a technically valid swagger spec someone could declare, e.g.:

 * "produces":["image/jpg"] and a response that is JSON.
 * A parameter "foo" that is "type":"integer", "maximum":3, "minimum":78

[...]

Issue #161, opened 2014-10-16
Proposal: Parameters in cookie


BigstickCarpet:
Currently, the only way we can define a cookie parameter is like this:
````yaml
parameters:
  - name: Cookie
    in: header
    type: string
    required: true
````
But that just requires a cookie - any cookie - to be present.   There's no way to distinguish between third-party cookies (e.g. google analytics, advertising, etc.) and our API's cookies.   It would be nice to be able to do this:
````yaml
parameters:
  - name: myApiCookie
    in: cookie
    type: string
    required: true
````

cemo:
definetely +1

frozenspider:
+1, we just ran into that issue and was quite surprised by the lack of cookies support

webron:
Parent: #565.

jasonh-n-austin:
Example in the wild of cookie-based auth in Wordpress API http://v2.wp-api.org/guide/authentication/#cookie-authentication
JIRA does cookies as well
https://developer.atlassian.com/jiradev/jira-apis/jira-rest-apis/jira-rest-api-tutorials/jira-rest-api-example-cookie-based-authentication

ePaul:
Cookies do not fit into the REST model of resources and their manipulation with HTTP operations, as they introduce some kind of session state hold by the client and sent back to the server – instead all needed state should be in the resource representations retrieved and the links in them to other resources (i.e. the hypermedia).

I guess cookies used as authentication are a border case which might be needed sometimes – but there is #15 for that.

BigstickCarpet:
@ePaul - I agree that it's bad practice to use cookies for state or resource representations.  But using them for authentication is common practice and should be supported by Swagger, IMHO.

There are two scenarios in which Cookie parameters are needed, both having to do with cookie-based authentication:

1. **Documenting auth endpoints**
I should be able to use Swagger to document my authentication endpoints, just like I use it to document the rest of my API endpoints.  And since my auth endpoints rely on a specific cookie being set, I need to be able to specify a cookie parameter in Swagger.

2. **Indicating that an endpoint requires an auth cookie**
Certain parts of my API may require authentication, and other parts may not.  I need to be able to indicate this in Swagger.  This could be done via the `securityDefinitions` object, but that doesn't support cookie-based auth yet either.  Cookie parameters would accomplish the same goal though, since the it would allow me to indicate that an endpoint requires a specific cookie to be set.

GMTA:
[...]

Issue #15, opened 2014-03-26
Allow session cookie auth


ajtowns:
Session cookies are one method of authentication available in Jira. See: https://docs.atlassian.com/jira/REST/6.2/#d2e4887

The process is that you visit a URL authenticating via some other method, and the response includes cookies that serve as your authentication for ongoing access.

A new Authorization Object type in 5.1.5 of "cookie" or "session-key" could work here, with "keyname" being required and specifying the cookie id used, and "login-url" giving a pointer to the API endpoint that needs to be accessed to set the cookie?

retrospectacus:
Putting your cookie in the header can be done by giving your securityDefinition name "Cookie" and its actual name goes in the data. It does not yet fully work (see issue 305) and of course you must obtain the cookie elsewhere...so   +1

BigstickCarpet:
@retrospectacus - The problem with `{type: "apiKey", in: "header", name: "cookie"}` is that there's no way to specify _which_ cookie is the API key.   The `Cookie` HTTP header can contain _many_ cookies, and a typical HTTP request will include cookies for all sorts of things - ad trackers, Google analytics, session cookies, etc.   So it's not sufficient to say that my API requires _any_ cookie.  I need to be able to say that I require a _specific_ cookie.  Like this: `{type: "apiKey", in: "cookie", name: "myAuthCookie"}`

retrospectacus:
@BigstickCarpet I completely agree - my described method is (will be?) a very poor workaround until in:cookie is implemented.

BigstickCarpet:
Ah, ok.  Then we're in agreement.  I'm using the same workaround right now in all my Swagger APIs.

cemo:
+1

earth2marsh:
While I think this is an excellent candidate for Swagger.Next consideration, I am going to argue the other side here…

Cookies may be widely used on web applications, but they aren't a good fit for modern APIs (see references below). Swagger has taken a position on API design vis-a-vis its constraints around operation signatures, which means it already has a precedent of not aspiring to describe all possible APIs. This simplifies the tooling, and helps keep Swagger nimble. Perhaps excluding cookies should be part of Swagger's general opinion on API design?

On the pros' side, it's generally a good thing for Swagger to be a more descriptive format. 

On the cons' side, supporting cookies in the spec also implies supporting cookies in the tooling, including SDK generation and Swagger-UI, etc, not to mention the documentation and example overhead. That baggage slows down future development.

References:
[...]

PR #655, opened 2016-04-15
added cookie parameter

fehguy:
This simply adds the `cookie` parameter type, which addresses the following:

#15, #161, #301

There are a number of issues in other projects which are addressed with this fix as well.

In a follow-up issue, a response of type `cookie` will be added

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -582,7 +582,7 @@ Describes a single operation parameter.
 
 A unique parameter is defined by a combination of a [name](#parameterName) and [location](#parameterIn).
 
-There are five possible parameter types.
+There are six possible parameter types.
 * Path - Used together with [Path Templating](#pathTemplating), where the parameter value is actually part of the operation's URL. This does not include the host or base path of the API. For example, in `/items/{itemId}`, the path parameter is `itemId`.
 * Query - Parameters that are appended to the URL. For example, in `/items?id=###`, the query parameter is `id`.
 * Header - Custom headers that are expected as part of the request.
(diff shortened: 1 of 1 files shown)


versions/3.0.md
```diff
@@ -582,20 +582,20 @@ Describes a single operation parameter.
...
   * `application/x-www-form-urlencoded` - Similar to the format of Query parameters but as a payload. For example, `foo=1&bar=swagger` - both `foo` and `bar` are form parameters. This is normally used for simple parameters that are being transferred.
   * `multipart/form-data` - each parameter takes a section in the payload with an internal header. For example, for the header `Content-Disposition: form-data; name="submit-name"` the name of the parameter is `submit-name`. This type of form parameters is more commonly used for file transfers.
-
+* Cookie - Used to pass a specific cookie value to the API.  Cookies are passed specifically through the `Set-Cookie: {name}={value}` syntax.
```
ePaul: Hmm, no. `Set-Cookie:` is used by the server, the client sends `Cookie:`.

>  Cookie - Used to pass a specific cookie value to the API.  Cookies are passed in an HTTP `Cookie` header, using the `{name}={value}` syntax as defined by [RFC 6265](https://tools.ietf.org/html/rfc6265).

fehguy: my bad, i will fix this

fehguy:
There is architecturally no difference between headers and cookies, except for the name and a more rigid serialization scheme

fehguy:
@OAI/tdc please accept or reject, if we choose to accept, I will add response cookie definitions and global cookie definitions to the PR.

earth2marsh:
[Documenting some of the @OAI/tdc debate on this topic…] At face value, this change seems to make complete sense. However, it cuts to the heart of an important issue — the spec exists so that tooling can leverage it. The wider the spectrum of APIs that OAS can describe, the heavier the burden on tooling authors.

Without going into the pros/cons of using cookies in API design, cookies are easy to work with in web browser clients, but they're less easy to deal with in just about everything else. By supporting cookies, we increase the surface area of OSS projects and commercial offerings that may wish to be OpenAPI certified (assuming that such a program exists in the future). 

The issue of cookies will set a precedence. Cookies are great for the web but arguably less appropriate to modern APIs. Should the OpenAPI Specification aspire to be more descriptive, or should OpenAPI aspire to make difficult choices to keep OAS more actionable?

jasonh-n-austin:
Having done some research on the subject, there are two sides to it. For most APIs, this is not a recommended approach. However, there are a number of use cases where browser-integrated apps provide a much better DX, at little risk, from utilizing cookies. We have documented a number of significant commercial offerings (Atlassian, Wordpress, etc) where cookies have been used successfully.

IMO OpenAPI should not be overly opinionated from an API design perspective. The freedom to make design choices that are appropriate to your use cases is the defining freedom which has made APIs great. Adding too many constraints simply excludes APIs which fall somewhat out of the norm...the result is poorly documented/automated APIs, or other spec formats taking advantage by adding the functionality.
I'm in favor of adding cookies support.

👍   :shipit:

whitlockjc:
I agree with @jasonh-n-austin on this one.  We shouldn't force design or best practices on anyone.  And there are many APIs out there using cookies as some form of the contract and that's OpenAPI's purpose is to describe said contract.

webron:
The spec is and will probably be opinionated on API design to some extent. The reason it's opinionated, to me, is normally not because of good/bad API practices but rather what it means for tooling support. Another reason the spec may *seem* opinionated (but it's not) is the case of simplicity over functionality.

For the latter, I don't see added complexity so it's fine. For the former, I'm not sure what kind complexity it will introduce to tooling, if at all. As such, I'm going to abstain here.

What I'm concerned about is the serialization of cookies. How exactly are those going to be described? Just type string? Or are people going to expect a way to enumerate 'fields'? For completeness, I think it's important to include an example here of what it's going to look like.

Issue #643, opened 2016-04-11
Remove references to `vendor extensions` in favor of `extensions` or `x-extensions`


earth2marsh:
During the working group 2.0 spec process, extensions were proposed to suit vendor usecases, and therefore were initially referred to as `vendor extensions.` 

However, there are plenty of reasons to extend the spec without being a vendor, hence the name is too narrow. This proposal recommends deprecating any reference to `vendor extensions` in favor of `extensions` or `x-extensions`.

IvanGoncharov:
@earth2marsh Totally agree with you.
Moreover, it motivates vendors to create their own dialects like `x-ms-*`(see [this](https://github.com/Azure/azure-rest-api-specs/blob/master/documentation/creating-swagger.md#Enum-x-ms-enum)).
And the same true for other vendors, see #616 

Please also consider adding following **recommendations**:
 * to reuse existing extension
 * create more generic extensions, and provide a public specification for them.
 * add vendor namespace only if an extension is specific to this particular vendor.

DavidBiesack:
agree dropping 'vendor' is good. (I suggest the term 'specification extensions' over x-extensions, but it's not hugely important.)

*Namespaces* for extension names perhaps should be addressed as a separate issue, though. Vendors or other non-vendor organizations will still likely use some x-mynamespacehere- prefix to avoid conflict. That will happen regardless of the term OAS uses (extension or vendor extension).

fehguy:
Fixed in #492

PR #656, opened 2016-04-15
Specification Extension instead of Vendor Extension

ralfhandl:
See #643

### DIFF
--- a/versions/2.0.md
+++ b/versions/2.0.md
@@ -126,7 +126,7 @@ Field Name | Type | Description
 
 Field Pattern | Type | Description
 ---|:---:|---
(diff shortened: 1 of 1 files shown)


ePaul:
Hmm, not sure if it is necessary to update the 2.0 spec with this.

ralfhandl:
Not sure either. Finding other pull requests for the 2.0 spec and noticing changes in the public version at http://swagger.io/specification/ I updated both, so @fehguy et al. can decide which PRs to merge.

fehguy:
we are not changing the 2.0 spec.

ralfhandl:
Fine with me. What about the corresponding change #657 to the 3.0 spec?

PR #657, opened 2016-04-15
Specification Extension instead of Vendor Extension

ralfhandl:
#643 for V3.0

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -106,7 +106,7 @@ This is the root document object for the API specification. It combines what pre
 
 Field Name | Type | Description
 ---|:---:|---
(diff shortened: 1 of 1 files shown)


DavidBiesack:
:+1:

webron:
I think we should potentially change the structure of the extensions, so I'd wait with the name change proposal alone.

jasonh-n-austin:
@webron I'd say the name change doesn't hurt, we've all agreed it should change anyways. If there are structural changes, we can address that separately (ready: easier to review PR).

👍

webron:
:+1:

fehguy:
👍

PR #661, opened 2016-04-19
Make clear that info/termsOfService should be an URL.

ePaul:
As discussed in previous pull request #255, current tooling already
interprets it like that. This pull request just redoes the same change
to the 3.0 version instead of the 2.0 one (where they were not applied).

Original change by Matti Schneider <hi@mattischneider.fr>,  2015-01-29 10:51:26.

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -136,7 +136,7 @@ Field Name | Type | Description
 ---|:---:|---
 <a name="infoTitle"></a>title | `string` | **Required.** The title of the application.
 <a name="infoDescription"></a>description | `string` | A short description of the application. [GFM syntax](https://help.github.com/articles/github-flavored-markdown) can be used for rich text representation.
-<a name="infoTermsOfService"></a>termsOfService | `string` | The Terms of Service for the API.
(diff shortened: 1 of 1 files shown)


IvanGoncharov:
@ePaul Just to make it simetrical to [license's url](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#licenseObject) can you please add `MUST be in the format of a URL`.

ePaul:
Hmm, instead of this we could make termsOfService an object analogously to the license object, with an URL field and a way of inlining the actual terms.

IvanGoncharov:
@ePaul I don't think it's a good idea.

> analogously to the license object

It's only `name`, so there is no mechanism for inlining.
A license is a special case since there are many standard ones, so you can just read it's `name` and know what it about without reading it.

> a way of inlining the actual terms.

I think inlining of licenses or ToS is really bad practice, it's huge chunks of text only indirectly related to APIs. Moreover, if you generate docs based on spec with something like that inlined you force a reader to scroll through a huge amount of text.

webron:
:+1:

Trivial change. @tdc please vote.

fehguy:
no brainer.  I'll just merge this

Issue #278, opened 2015-02-19
Why does #/bodyParameter require a "name"?


themihai:
I'm wondering why ``bodyParameter`` schema requires a "name".  Makes sense  only if the parameter is sent/set ``in`` ``query`` or perhaps ``path`` but doesn't make sense if the parameter  ``in`` value is  ``body``  isn't it? Unlike the query string parameters json/xml objects don't have a name.

``
  "bodyParameter": {
      "type": "object",
      "required": [
        "name",
        "in",
        "schema"
      ],
``

webron:
It's for documentation purposes. In previous versions, it had to be 'body'. In 2.0, it can be pretty much anything that you may find descriptive to the body itself.

themihai:
Wouldn't it make sense to use a different term(e.g. label) although there is a "description" field which pretty much solves the documentation issue? Isn't ``name`` used to name the parameters of ``simpleType`` How do you document a parameter which is of simpleType? (i.e a string)

webron:
I don't follow your last question.

fehguy:
here's a practical usage of why you need to have names on parameters:

https://github.com/swagger-api/swagger-js#need-to-send-an-object-to-your-api-via-post-or-put

```js
var pet = {
  id: 100,
  name: "dog"};

swagger.apis.pet.addPet({body: pet});
```

This is doable because the input parameter has a name attached to it.  And of course, the operation can have query params, header params, path params, etc.

themihai:
See the example below. It seems to me that in this case "name" is actually the name of the parameter and is no longer documentation. If the "name" would serve as documentation only it shouldn't be required (just like description) isn't it? My concern is that serving a double purpose adds confusion to the schema even if it doesn't need (there is a "description" field, right?) 


``"parameters":[{"name":"username","in":"path","description":"The name that needs to be fetched. Use user1 for testing. ","required":true,"type":"string"}],  ``

webron:
@fehguy just gave you an actual use case example as to why it is needed. There's no real double purpose here.

Also, the schema is not the issue, it reflects the spec itself, which requires it.

themihai:
@webron the issue is about the specifications. Let me know if there is any other place where it should be discussed.  
When ``in`` is ``query``  the ``name`` field of the schema represents the query field when ``in`` is ``body``  it serves as documentation so this sounds like a double purpose. 
[...]

Issue #565, opened 2016-02-12
[Meta issue] Restructure request parameters


jharmn:
This is a meta issue for tracking improvements for the structure of request parameters.

Please don't add comments to this issue--rather open new tickets, if they apply, and reference this as the parent.

jasonh-n-austin:
Link to parent #560

fehguy:
Unlinking this from #560, it's worthy of it's own meta issue.

webron:
Oh, they're restructured.

Issue #266, opened 2015-02-05
require `consumes` or `produces`?


fehguy:
Without requiring to state what is consumed or produced by a server, it's really up to the client to guess what format to send/receive, and up to the server to handle all the types it doesn't support with proper, known error responses.  A simple work around would be to require that `consumes` is required whenever a parameter of type `body` or `formData` are supplied, and `produces` is required whenever a `schema` is attached to a response code or the default response.

webron:
`schema` doesn't mean model anymore, it could also be a primitive. We can say it's mandatory of the type is `object`. 

We could definitely not validate it using basic json schema as the values propagate.

ePaul:
If the response has a `schema` property, this means the content is in the body of the HTTP response, right? So if there is a response, there should be a `produces` property.

webron:
@ePaul - really? we should force people to define `text/plain` if they return a numeric value?

ePaul:
If the numeric value is in the body, it needs some media type (i.e. `Content-Type` header of the HTTP response). This could be `text/plain`, of course, or `application/x.single-number-in-ascii`, or even some JSON type (because a single number is also a valid JSON value in [RFC 7159](https://tools.ietf.org/html/rfc7159)).

webron:
But is RFC 7159 accepted? It looks like it's a proposal. http://json.org/ suggest that only objects and arrays are valid JSON, while RFC 7159 suggests the change that JSON values are also valid JSON text. And indeed, many JSON parsers die if you pass a simple string value to them.

It looks like IANA refers `application/json` to be  [RFC 4627](https://www.ietf.org/rfc/rfc4627.txt) which suggests that simple types are not officially supported for the mime type.

And let's face it, when was the last time you saw an API use `application/x.single-number-in-ascii`? :)

ePaul:
I don't really want to discuss which media-type is appropriate here, but just mention that one should be used. (I also think an API returning a single number/string/boolean is not a good idea.)

[RFC 7231](https://tools.ietf.org/html/rfc7231#section-3.1.1.5) says that each response with body should have a Content-Type header, except if the server has no idea what it is. I would guess in the case of an API described by an OpenAPI description you can't really say "I have no idea what this is".

webron:
[...]

PR #670, opened 2016-04-25
Request bodies moved out of parameters

jasonh-n-austin:
## Problem description

Request bodies are currently specified in `parameters` as `in: body`. This has been addressed as an incongruence with other parameters throughout a variety of issues in meta #565, as well as discussions with tooling vendors, as well as many exceptions made for request bodies throughout the spec. 

A number of issues arise from treating request body like other parameter locations (path, query, header) :
* `name` is meaningless in body, and is required. To underscore this, from the current spec (I would also argue this serves no documentation purpose):

> The name of the body parameter has no effect on the parameter itself and is used for documentation purposes only

* Body parameters must be unique in the list of parameters. This currently has to be enforced/validated by tooling, as the array structure does not currently create any natural constraint. From [Path Item Object](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#pathItemObject) 

> There can be one "body" parameter at most.

* In trying to add serialization schemes for complex non-body parameters (in #654), some trouble arises.
  * Request body is usually a media type like `application/json`, `application/xml`
  * Other parameters are typically either some sort of collection, and infrequently json, xml, etc.

* In adding `examples` to the body parameter (#636), the `examples` aren’t really applicable to other parameter types.
* Throughout the spec, there are exceptions made for the inconsistency in the body parameter vs the other parameter types.

## Proposed solution

* Move `body` parameter into it’s own key, peer-level to `parameters`.
* Simplify the request body object (vs parameter object), to remove `name`, `in`, `required`, and `deprecated`.
* Eliminate the use of `examples` in parameters (as media types are largely irrelevant). This was already unclear in the spec, and probably needs revision in another PR (as mentioned in #636).

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -346,7 +346,8 @@ Field Name | Type | Description
 <a name="pathItemHost"></a>host | `string` | The host (name or ip) serving the path. This optional value will override the top-level [host](#oasHost) if present. This MUST be the host only and does not include the scheme nor sub-paths. It MAY include a port. If the `host` is not included, the host serving the documentation is to be used (including the port). The `host` does not support [path templating](#pathTemplating).
 <a name="pathItemBasePath"></a>basePath | `string` | The base path on which the API is served, which is relative to the [`host`](#pathItemHost). This optional value will override the top-level [basePath](#oasBasePath) if present. If it is not included, the API is served directly under the `host`. The value MUST start with a leading slash (`/`). The `basePath` does not support [path templating](#pathTemplating). 
 <a name="pathItemSchemes"></a>schemes | [`string`] | The transfer protocol of the API. Values MUST be from the list: `"http"`, `"https"`, `"ws"`, `"wss"`. This optional value will override the top-level [schemes](#oasSchemes) if present. If the `schemes` is not included, the default scheme to be used is the one used to access the OpenAPI definition itself.
-<a name="pathItemParameters"></a>parameters | [[Parameter Object](#parameterObject) <span>&#124;</span> [Reference Object](#referenceObject)] | A list of parameters that are applicable for all the operations described under this path. These parameters can be overridden at the operation level, but cannot be removed there. The list MUST NOT include duplicated parameters. A unique parameter is defined by a combination of a [name](#parameterName) and [location](#parameterIn). The list can use the [Reference Object](#referenceObject) to link to parameters that are defined at the [OpenAPI Object's parameters](#oasParameters). There can be one "body" parameter at most.
+<a name="pathItemParameters"></a>parameters | [[Parameter Object](#parameterObject) <span>&#124;</span> [Reference Object](#referenceObject)] | A list of parameters that are applicable for all the operations described under this path. These parameters can be overridden at the operation level, but cannot be removed there. The list MUST NOT include duplicated parameters. A unique parameter is defined by a combination of a [name](#parameterName) and [location](#parameterIn). The list can use the [Reference Object](#referenceObject) to link to parameters that are defined at the [OpenAPI Object's parameters](#oasParameters). 
+<a name="pathItemRequestBody"></a>requestBody | [[Request Body Object](#requestBodyObject) <span>&#124;</span> [Reference Object](#referenceObject)] | The request body applicable for all the operations described under this path. This request body can be overriden at the operation level, but cannot be removed there.
 
 ##### Patterned Fields
 
@@ -445,7 +446,8 @@ Field Name | Type | Description
 <a name="operationId"></a>operationId | `string` | Unique string used to identify the operation. The id MUST be unique among all operations described in the API. Tools and libraries MAY use the operationId to uniquely identify an operation, therefore, it is recommended to follow common programming naming conventions.
 <a name="operationConsumes"></a>consumes | [`string`] | A list of MIME types the operation can consume. This overrides the [`consumes`](#oasConsumes) definition at the OpenAPI Object. An empty value MAY be used to clear the global definition. Value MUST be as described under [Mime Types](#mimeTypes).
 <a name="operationProduces"></a>produces | [`string`] | A list of MIME types the operation can produce. This overrides the [`produces`](#oasProduces) definition at the OpenAPI Object. An empty value MAY be used to clear the global definition. Value MUST be as described under [Mime Types](#mimeTypes).
-<a name="operationParameters"></a>parameters | [[Parameter Object](#parameterObject) <span>&#124;</span> [Reference Object](#referenceObject)] | A list of parameters that are applicable for this operation. If a parameter is already defined at the [Path Item](#pathItemParameters), the new definition will override it, but can never remove it. The list MUST NOT include duplicated parameters. A unique parameter is defined by a combination of a [name](#parameterName) and [location](#parameterIn). The list can use the [Reference Object](#referenceObject) to link to parameters that are defined at the [OpenAPI Object's parameters](#oasParameters). There can be one "body" parameter at most.
+<a name="operationParameters"></a>parameters | [[Parameter Object](#parameterObject) <span>&#124;</span> [Reference Object](#referenceObject)] | A list of parameters that are applicable for this operation. If a parameter is already defined at the [Path Item](#pathItemParameters), the new definition will override it, but can never remove it. The list MUST NOT include duplicated parameters. A unique parameter is defined by a combination of a [name](#parameterName) and [location](#parameterIn). The list can use the [Reference Object](#referenceObject) to link to parameters that are defined at the [OpenAPI Object's parameters](#oasParameters).
+<a name="operationRequestBody"></a>body | [[Request Body Object](#requestBodyObject) <span>&#124;</span> [Reference Object](#referenceObject)] | The request body applicable for this operation. If a request body is already defined at the [Path Item](#pathItemParameters), the new definition will override it, but can never remove it.
 <a name="operationResponses"></a>responses | [Responses Object](#responsesObject) | **Required.** The list of possible responses as they are returned from executing this operation.
 <a name="operationSchemes"></a>schemes | [`string`] | The transfer protocol for the operation. Values MUST be from the list: `"http"`, `"https"`, `"ws"`, `"wss"`. The value overrides the OpenAPI Object [`schemes`](#oasSchemes) definition. 
 <a name="operationDeprecated"></a>deprecated | `boolean` | Declares this operation to be deprecated. Usage of the declared operation should be refrained. Default value is `false`.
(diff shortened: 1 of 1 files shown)


ePaul:
> * Eliminate the use of examples in parameters (as media types are largely irrelevant). This was already unclear in the spec, and probably needs revision in another PR (as mentioned in #636).

While I think media types need separate considerations (see #146), I think the plural `examples` is still useful, i.e. the ability to give multiple examples. (Though this could live inside the schema object, not at the parameter level.)

IvanGoncharov:
@jasonh-n-austin Cool PR :+1: definitely make spec less confusing.
But I disagree with removing `required` and `deprecated`.

You can make POST requests without body parameter, so you should have the ability to write `request: false`. For details, see [this disscussion](http://stackoverflow.com/questions/4191593/is-it-considered-bad-practice-to-perform-http-post-without-entity-body).

If you use the same schema to describe multiple body parameters, how you will deprecate only one of them? So body need it's own `deprecate` property.

versions/3.0.md
```diff
@@ -346,7 +346,8 @@ Field Name | Type | Description
...
+<a name="pathItemParameters"></a>parameters | [[Parameter Object](#parameterObject) <span>&#124;</span> [Reference Object](#referenceObject)] | A list of parameters that are applicable for all the operations described under this path. These parameters can be overridden at the operation level, but cannot be removed there. The list MUST NOT include duplicated parameters. A unique parameter is defined by a combination of a [name](#parameterName) and [location](#parameterIn). The list can use the [Reference Object](#referenceObject) to link to parameters that are defined at the [OpenAPI Object's parameters](#oasParameters). 
+<a name="pathItemRequestBody"></a>requestBody | [[Request Body Object](#requestBodyObject) <span>&#124;</span> [Reference Object](#referenceObject)] | The request body applicable for all the operations described under this path. This request body can be overriden at the operation level, but cannot be removed there.
```
ePaul: Actually I would omit it at the path item level, and have it only at the operation level.

I can't think of a single request body definition which would be plausibly the same for multiple, or even all, operations on the same path.

ePaul:
Deprecating a request body is only sensible when the operation previously had a body parameter, but in the future should be called without one. (Otherwise you would deprecate the whole operation.)

I guess this happens, but not that often, so I'm ambivalent if it is necessary.

Which might be happening more often would be passing a different body parameter (maybe with a different Content-Type), and then the old one will be deprecated. Enabling this at all belongs to #146, though. I guess a solution to that will also cover the case of "no body parameter".

versions/3.0.md
```diff
@@ -786,9 +734,89 @@ schema:
   type: file
 ```
 
+#### <a name="requestBodyObject"></a>Request Body Object
+
+Describes a single request body.
+
+If the `requestBody` attribute is not present, or `schema` is not defined, the request body is not required.
```
jasonh-n-austin: @IvanGoncharov this concept is meant to preclude the need for `required`. If no request body is needed, then one should not be specified. You could also specify a schema with nothing specified and get close the same net effect.

jasonh-n-austin:
@IvanGoncharov I think @ePaul got it right here. In the case of an operation with a request body, the entire operation would be deprecated, but a request body would be very unlikely to be deprecated.

I intentionally didn't tackle multiple request bodies in this PR, to try and take a bite sized chunk of the complexity. Perhaps when we consider content-type-specific requests the `deprecated` issue comes back into focus. 

On multiple request bodies (to be addressed in a future PR), to emphasize why it's not in here:
The additional complexity here is that the current structure uses one schema (the OpenAPI subset of JSON Schema) to describe all media types (which is, in itself, an issue), and examples are delineated by content-type. 
If multiple request bodies (by content-type) are specified, then examples must be modified to unlink them from content-type.

versions/3.0.md
jasonh-n-austin: I don't think it's super common, but I could see a case where the same type is used on multiple paths, with some exceptions. I'm not super tied to it, but since v2.0 has the concept for `parameters`, it seemed fair to keep it in splitting out `requestBody`.

versions/3.0.md
```diff
@@ -445,7 +446,8 @@ Field Name | Type | Description
...
+<a name="operationParameters"></a>parameters | [[Parameter Object](#parameterObject) <span>&#124;</span> [Reference Object](#referenceObject)] | A list of parameters that are applicable for this operation. If a parameter is already defined at the [Path Item](#pathItemParameters), the new definition will override it, but can never remove it. The list MUST NOT include duplicated parameters. A unique parameter is defined by a combination of a [name](#parameterName) and [location](#parameterIn). The list can use the [Reference Object](#referenceObject) to link to parameters that are defined at the [OpenAPI Object's parameters](#oasParameters).
+<a name="operationRequestBody"></a>body | [[Request Body Object](#requestBodyObject) <span>&#124;</span> [Reference Object](#referenceObject)] | The request body applicable for this operation. If a request body is already defined at the [Path Item](#pathItemParameters), the new definition will override it, but can never remove it.
```
whitlockjc: This should be `requestBody` instead of `body`.

jasonh-n-austin: `body` should be `requestBody` here

whitlockjc:
My personal opinion is that having a path-level `requestBody` is not all that useful.  Sure, there are some cases where the request body will have the same schema across all operations with a body but it could easily be confusing for operations where that isn't the case.  I would much rather _reuse_ things using references much like we do elsewhere using a `$ref` and so specifying a request body would be opt-in instead of opt-out in this case.  I realize the thinking behind this was in the same vein around the ability now to set `body` parameters at the path-level but I'm not sure how useful even that is.

Would this require a global `requests` or `requestBodies` object to store the reusable schemas?

versions/3.0.md
```diff
@@ -595,25 +597,24 @@ Describes a single operation parameter.
...
 * Form - Used to describe the payload of an HTTP request when either `application/x-www-form-urlencoded`, `multipart/form-data` or both are used as the content type of the request (in the OpenAPI Specification's definition, the [`consumes`](#operationConsumes) property of an operation). This is the only parameter type that can be used to send files, thus supporting the `file` type. Since form parameters are sent in the payload, they cannot be declared together with a body parameter for the same operation. Form parameters have a different format based on the content-type used (for further details, consult http://www.w3.org/TR/html401/interact/forms.html#h-17.13.4):
   * `application/x-www-form-urlencoded` - Similar to the format of Query parameters but as a payload. For example, `foo=1&bar=swagger` - both `foo` and `bar` are form parameters. This is normally used for simple parameters that are being transferred.
```
jasonh-n-austin: @webron to look at how to deal with form urlencoded parameters.

webron:
This is a good proposal. I'll add my comments to specific sections tackling the proposal as-is, then go over the formData parameters.

versions/3.0.md
webron: Adding to the people above and @whitlockjc's comment on the main thread - this just doesn't seem like a real case we need to handle. The idea behind parameters being at the POI was that people won't need to repeat common parameters - this makes sense for path parameters, query parameters and obviously header parameters. This doesn't make sense as much for payloads (form and body). You could argue there's a case where a single body can be used for both POST and PUT, and that a specific path doesn't have a GET/DELETE/... calls as well. So yes, but it feels like the edge of the edge. 

Does adding support to it cost much? Probably not. Will it be used by enough APIs to justify it? Probably even less. So in favor of simplifying the tooling and considering there's a very easy alternative, I'd suggest we drop this from the POI.

versions/3.0.md
```diff
@@ -595,25 +597,24 @@ Describes a single operation parameter.
 
 A unique parameter is defined by a combination of a [name](#parameterName) and [location](#parameterIn).
 
-There are six possible parameter types.
+There are five possible parameter types.
 * Path - Used together with [Path Templating](#pathTemplating), where the parameter value is actually part of the operation's URL. This does not include the host or base path of the API. For example, in `/items/{itemId}`, the path parameter is `itemId`.
 * Query - Parameters that are appended to the URL. For example, in `/items?id=###`, the query parameter is `id`.
 * Header - Custom headers that are expected as part of the request.
-* Body - The payload that's appended to the HTTP request. Since there can only be one payload, there can only be *one* body parameter. The name of the body parameter has no effect on the parameter itself and is used for documentation purposes only. Since Form parameters are also in the payload, body and form parameters cannot exist together for the same operation.
```
webron: Removing this description removes the restriction of using body parameters alongside formData parameters (which just can't happen). We either need to add it (probably in both places) or discuss moving formData parameters to the `requestBody` as well.

versions/3.0.md
webron: If we move `formData` parameters to the `requestBody` as well (and maybe even if not), perhaps a more suitable name would be `payload` (not a big fan of mixed-case names, `body` may be too obscure to some).

versions/3.0.md
webron: I don't get it. If a `requestBody` is defined, how can it not have `schema`? Seems like `schema` needs to be required in this case.

Not sure we can avoid `required` here. To me, not stating a `requestBody` doesn't make it optional, it makes it 'ignored', 'unaccepted'.

We can argue whether there are cases that a payload can be optional, question is do we really want to not allow specifying it. We can make it implicitly required allowing people to explicitly set it as not-required if needed, covering most cases and reducing bloat.

versions/3.0.md
```diff
@@ -786,9 +734,89 @@ schema:
...
+##### Request Body Examples
+
+A request body with a referenced model definition.
+```js
+{
+  "description": "user to add to the system",
+  "schema": {
+    "$ref": "#/definitions/User"
+  },
+  "examples": {
```
webron: Looks like we can't avoid discussing how `examples` are written. I don't know what it means to reference an external non-json/yaml file. Also, `examples` values were intended to be taken as-is, so the `$ref` *is* the example in this case. Furthermore (and related to the two points above) - the only real way to represent non-json examples is to quote them as strings, but for JSON many people opt to just using JSON as-is without quoting. This makes processing a bit more difficult.

This is probably out of scope for this PR though and should be handled separately.

versions/3.0.md
```diff
@@ -786,9 +734,89 @@ schema:
   type: file
 ```
 
+#### <a name="requestBodyObject"></a>Request Body Object
+
+Describes a single request body.
+
+If the `requestBody` attribute is not present, or `schema` is not defined, the request body is not required.
+
+The serialization of the request body should be specified by the `consumes` attribute.
```
webron: `should`? Perhaps `MUST` - see #266. If we stick with `should`, it should be written as `SHOULD`.

webron:
Re `examples` - we may need to revisit it when we handle the content negotiation ticket.
Not sure how/why this changes the `example` for regular parameters though.

Re `required` - added my comments. Not sure we can avoid it.

Re `deprecated` - I tend to agree that if the body is deprecated, then the operation is deprecated. An alternative (and followup to above) to change it as marked to not being `required` - that should cover it.

@whitlockjc raises an important issue regarding reusable components - since requestBody is no longer a parameter, do we need to allow it under the new components construct? Not sure we can avoid it.

versions/3.0.md
ePaul: I'm not sure if/how the `consumes` attribute actually specifies how the request body is being serialized. Do you mean something like

>  If we have an `application/json` or `application/...+json` media type, it is serialized as JSON, if we have an `application/xml` or `application/...+xml` media type, it is serialized as XML?

If so, it should actually be written out.

versions/3.0.md
jasonh-n-austin: Agreed, I'll drop it.

versions/3.0.md
jasonh-n-austin: Adding `required` back in. 

I'm specifying that it defaults to `true`, to avoid having to specify `required: true` for every request body (when that is probably the 90+% case).

versions/3.0.md
jasonh-n-austin: I'll try to rephrase a bit.

I'd like to spend more time tackling this concept in another PR, where I'm planning to move the media type up one level, something like:
```
request:
  application/json:
    schema: 
      $ref: "#/definitions/User"
```

I think once it's in this format, it's more self-descriptive.

darrelmiller: @jasonh-n-austin  Really glad you are heading in that direction.  That was definitely the next logical step for me.

jasonh-n-austin:
* Fixed a typo and refined some language.
* Added `required`
* Removed path-level body

Now to look at `formData`.

webron:
I'm trying to collect my thoughts around `formData` - my take is that we should try eliminating it and converging it with the `requestBody` - there are a few gotchas there but combining this PR with #654 allows us to solve *many* issues regarding `formData` (validation in conjunction with `requestBody`, allowing multiple file uploads, setting types for files and so on).

Hate bringing it up but regarding the name, are we set on `requestBody`? Was thinking of just `payload`, unless we're set on having the word `request` as part of it.

webron:
Ugh, forgot to say, I think this can be merged, and I'll tackle the `formData` in a separate PR.

wparad:
2 cents, please help rid the whole of `formdata`, either `body` or `payload`.

webron:
This also resolved #278.

versions/3.0.md
```diff
@@ -786,9 +733,88 @@ schema:
...
+
+Describes a single request body.
+
+The `Content-Type` of the request body must be specified by the `consumes` attribute, either at the [top-level](#oasConsumes) or [operation level](#operationConsumes).
+
+##### Fixed Fields
+Field Name | Type | Description
+---|:---:|---
+<a name="requestBodyDescription"></a>description | `string` | A brief description of the request body. This could contain examples of use.  [GFM syntax](https://help.github.com/articles/github-flavored-markdown) can be used for rich text representation.
+<a name="requestBodySchema"></a>schema | [Schema Object](#schemaObject) | The schema defining the type used for the request body.
```
webron: Just noticed this - shouldn't the schema field be required? What does a request body mean without a schema?

ralfhandl: Some of our APIs use AtomPub-style "media resources", see https://tools.ietf.org/html/rfc5023#section-9.6. In this case the request body of a POST can have any media type, e.g. a picture or PDF document, so there's no schema.

darrelmiller: Many media types full describe the semantics of a body.

Issue #688, opened 2016-05-10
Proposal: Hypermedia and linking between resources


erikvanzijst:
[Bitbucket's API](https://api.bitbucket.org/swagger.json) uses [HAL-style](https://tools.ietf.org/html/draft-kelly-json-hal-07) links to make things more discoverable. An example of this is how a repository object embeds a link to its list of pull request and a pull request object links to its comments and reviewers.

Embedding links has a number of disadvantages though.
## Limitations of Custom Links in Schema Objects

As our API grows, so does the number of associated resources and by extension the number of links embedded in every response, which has a compute and bandwidth cost. The extensive [list embedded in our repository object](https://api.bitbucket.org/2.0/repositories/evzijst/interruptingcow) reflects this:

``` json
  "links": {
    "watchers": {
      "href": "https://api.bitbucket.org/2.0/repositories/evzijst/interruptingcow/watchers"
    },
    "branches": {
      "href": "https://api.bitbucket.org/2.0/repositories/evzijst/interruptingcow/refs/branches"
    },
    "tags": {
      "href": "https://api.bitbucket.org/2.0/repositories/evzijst/interruptingcow/refs/tags"
    },
    "commits": {
      "href": "https://api.bitbucket.org/2.0/repositories/evzijst/interruptingcow/commits"
    },
    "clone": [
      {
        "href": "https://bitbucket.org/evzijst/interruptingcow.git",
        "name": "https"
      },
      {
        "href": "ssh://git@bitbucket.org/evzijst/interruptingcow.git",
        "name": "ssh"
      }
    ],
    "self": {
      "href": "https://api.bitbucket.org/2.0/repositories/evzijst/interruptingcow"
    },
    "html": {
      "href": "https://bitbucket.org/evzijst/interruptingcow"
    },
    "avatar": {
      "href": "https://bitbucket.org/evzijst/interruptingcow/avatar/32/"
    },
    "hooks": {
      "href": "https://api.bitbucket.org/2.0/repositories/evzijst/interruptingcow/hooks"
    },
    "forks": {
      "href": "https://api.bitbucket.org/2.0/repositories/evzijst/interruptingcow/forks"
    },
    "downloads": {
      "href": "https://api.bitbucket.org/2.0/repositories/evzijst/interruptingcow/downloads"
    },
    "pullrequests": {
      "href": "https://api.bitbucket.org/2.0/repositories/evzijst/interruptingcow/pullrequests"
    }
  }
```

[...]

PR #710, opened 2016-06-09
added links proposal

fehguy:
To support HAL-style linking, as suggested in #688, I propose the following:

* A mechanism for design-time, _static_ links between responses and operations to make operations on responses visible to consumers
* A reusable linking structure to allow link definitions to be DRY
* A mechasim for parameterized link traversal to avoid server-side computation of links through reuse of existing, OAS-defined operations

Additional considerations:
* This is not intended to be a _run time_ linking solution.  This solution does not require users to redesign their runtime API.  It _may_ be able to represent the intent of runtime HAL-style links in responses.  It does _not_ intend to represent dynamically generated, arbitrary links.
* A clear separation of concerns, between payloads and operations.  Therefore no _operation data_ shall be tied to payload schemas.

Spec changes
* Introduction of a `Link Object` to represent a single `link`
* Addition of a  `links` object in the `Response Object` to define what links are available from a specific operation
* Addition of a `links` section in the `Components Object` to allow reuse across operations
* Introduction of a `Link Parameters Object` to represent parameters to build a link from
* A mechanism for reading values from response payloads and using them for client-side link computation

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -938,7 +938,7 @@ default:
 ```
 
 #### <a name="responseObject"></a>Response Object
-Describes a single response from an API Operation.
+Describes a single response from an API Operation, including design-time, static `links` to operations based on the response.
 
 ##### Fixed Fields
 Field Name | Type | Description
@@ -947,6 +947,8 @@ Field Name | Type | Description
 <a name="responseSchema"></a>schema | [Schema Object](#schemaObject) | A definition of the response structure. It can be a primitive, an array or an object. If this field does not exist, it means no content is returned as part of the response. As an extension to the [Schema Object](#schemaObject), its root `type` value may also be `"file"`. This SHOULD be accompanied by a relevant `produces` mime-type.
 <a name="responseHeaders"></a>headers | [Headers Object](#headersObject) | A list of headers that are sent with the response.
 <a name="responseExamples"></a>examples | [Example Object](#exampleObject) | An example of the response message.
+<a name="responseLinks"></a>links | [Links Object](#linksObject) | An object representing operations related to the response payload.
+
 
 ##### Patterned Objects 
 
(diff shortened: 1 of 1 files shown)


versions/3.0.md
```diff
@@ -1123,6 +1125,344 @@ application/json:
...
+href: 'http://colors.my-server.com/colors/green'
+href: 'http://colors.my-server.com/colors/blue'
+```
+
+As with all links, it at the the clients' discretion to follow them, and permissions and the existence of a value is not guaranteed soley by the existence of a link relationship. [**I assume you mean that the existance of the link relationship does not guarantee a successful response from the call**]
+
+
+### Example
+
+The example below shows how relationships in the BitBucket API can be represented with the proposed OAI link scheme.  This example uses `operationId` values to link responses to possible operations.
```
fehguy: Note, this is a very long example and should probably be shortened for the documentation.

versions/3.0.md
```diff
@@ -1123,6 +1125,344 @@ application/json:
...
+The `Link Object` is responsible for defining a possible operation based on a single response.
+
+Field Name  |  Type  | Description
+  ---       | :---:  |     ---
+href      |  url   | a relative or absolute URL to a linked resource. This field is mutually exclusive with the `operationId` field.
+ operationId| string | the name of an _existing_, resolvable OAS operation, as defined with a unique `operationId`.  This field is mutually exclusive with the `href` field.  Relative `href` values _may_ be used to locate an existing Operation Object in the OAS.
+parameters  | Link Parameters Object | an Object representing parameters to pass to an operation as specified with `operationId` or identified via `href`.
+description | string | a description of the link, supports GFM.
+
+#### <a name="responsePayload"></a>Response Payload Values
```
fehguy: It may be desirable to allow `request` values to be used in the links

versions/3.0.md
```diff
@@ -1123,6 +1125,344 @@ application/json:
...
+color: ["red", "green", "blue"]
+```
+
+The variables generated can be used in locations prescribed by the definition.
+
+
+### Variable substitution
+In all cases, _variables_ identified in the response payload may be substitued **only** in locations prescribed in the link definitions.  All such locations are prefixed by a `$response` keyword and surrounded by curly brackets `{}`. 
+
+### Example
```
fehguy: Need to add an example using query parameters

versions/3.0.md
```diff
@@ -1123,6 +1125,344 @@ application/json:
...
+```
+
+And the array example:
+
+```yaml
+ColorSelection:
+  href: 'http://colors.my-server.com/colors/{$response.color}'
+```
+
+Would produce the following links:
```
fehguy: This needs more explanation.  There are potentially cartesian products of links generated with nested arrays.  Since this is a client-generated link, and _optional_, it should be acceptable

versions/3.0.md
```diff
@@ -1123,6 +1125,344 @@ application/json:
...
+---|:---:|---
+<a name="linkName"></a>link name | [Link Object](#linkObject) | A short name for the link, following the naming constraints of the <a href="#definitionsName">definitions name</a>.  The link shall reference a single Link Object, or a JSON Reference to a single link object
+
+
+### Link Object
+The `Link Object` is responsible for defining a possible operation based on a single response.
+
+Field Name  |  Type  | Description
+  ---       | :---:  |     ---
+href      |  url   | a relative or absolute URL to a linked resource. This field is mutually exclusive with the `operationId` field.
```
fehguy: We need to provide more information when following _non-matched_ operations (i.e. absolute references outside of the spec).  For example, headers, HTTP method, etc. should be provided.

versions/3.0.md
```diff
@@ -1123,6 +1125,344 @@ application/json:
...
+ operationId| string | the name of an _existing_, resolvable OAS operation, as defined with a unique `operationId`.  This field is mutually exclusive with the `href` field.  Relative `href` values _may_ be used to locate an existing Operation Object in the OAS.
```
fehguy: How will remote `operationId` values work?  Can they be referenced with a JSON reference?

versions/3.0.md
```diff
@@ -1123,6 +1125,344 @@ application/json:
...
+
+Would produce the following links:
+
+```yaml
+href: 'http://colors.my-server.com/colors/red'
+href: 'http://colors.my-server.com/colors/green'
+href: 'http://colors.my-server.com/colors/blue'
+```
+
+As with all links, it at the the clients' discretion to follow them, and permissions and the existence of a value is not guaranteed soley by the existence of a link relationship. [**I assume you mean that the existance of the link relationship does not guarantee a successful response from the call**]
```
fehguy: Should add documentation as to _what_ happens if a required parameter is not provided or is `null`.

fehguy:
@OAI/tdc this is per the last TDC call.  Please add comments, note I have added several open issues that need your help with responses or documentation.

versions/3.0.md
```diff
@@ -1123,6 +1125,344 @@ application/json:
...
+    { "color": "green" },
+    { "color": "blue" }
+]
+```
+
+will be extracted as such:
+
+```json
+color: ["red", "green", "blue"]
+```
```
ePaul: I'm not sure that this special-case handling for arrays is really useful and generalizable for more complex objects.
What happens if you have an array of arrays? Or an object containing an array, which in turn contains an object with a property we want to access?

As I understand, the point is that if some information appears multiple times in the response, we get multiple links generated. So we need a way to extract that information in an array.
Why don't we use something like [JSONPath](http://goessner.net/articles/JsonPath/) to extract what we want from the response, instead of defining our own ad-hoc syntax?

With that, in your example the link definitions would look like this (the rest of the example remains same):

```yaml
  links:
    UserRepositories:
      # returns array of '#/components/definitions/repository'
      operationId: getRepositoriesByOwner
      parameters:
        username: $.username
    UserRepository:
      # returns '#/components/definitions/repository'
      operationId: getRepository
      parameters:
        username: $[*].owner.username
        slug: $[*].slug
    RepositoryPullRequests:
      # returns '#/components/definitions/pullrequest'
      operationId: getPullRequestsByRepository
        params: 
          username: $.owner.username
          slug: $.slug
    PullRequestMerge:
      # executes /2.0/repositories/{username}/{slug}/pullrequests/{pid}/merge
      operationId: mergePullRequest
      parameters:
        username: $.user.username # Should be $response.author.username?
        slug: $.repository.slug
        pid: $.id
```

Apart from `$response` → `$`, the main difference is in the array access in the UserRepository link.
That also shows a possible problem: Does this create one link for each combination of username and slug (not wanted), or aligns the links from the user names and slugs with each other? We want the latter, but how to specify this if the array is not at top-level, and we even might be accessing different arrays in both parameters?

fehguy: @ePaul recall that link generation is a client-side activity, so the consumer is responsible for generating the links themselves.

As far as nested arrays, yes, the goal is to produce _multiple_ links.  Again, it is up to the consumer of the response to decide if it should follow them or not.  And with nested arrays, yes there can be combinatorics.  That is actually desirable.

As far as syntax, I'm not really partial to any syntax other than:

1) Making sure it's not trying to do _too_ much.  This applies for the hesitation for general JSON Schema support or path templating
2) It's supported and understood well enough

If we can use something else, then great.  If the other syntaxes are far more than we need, we can choose to support a subset or not use them.

ePaul: > @ePaul recall that link generation is a client-side activity, so the consumer is responsible for generating the links themselves.

Yes, the goal is to make it possible, given an OpenAPI definition and a response from one operation, to know exactly which link URIs are to be generated. (Whether an user of a potential client library actually retrieves this list, and/or follows them, is of course not decided by the API.)

My main point was that the proposed spec change was a bit nebulous enough about the array effect, letting an implementor guess what to do. This certainly needs rewording. (I can try to find a better wording, though not right now ... a bit too tired to think clearly.)


The second point is, that the proposed syntax doesn't indicate where there is an array expected, like how JSONPath's `[*]` does it.
(While this might be intended (i.e. to match HAL's links, which can either come as an object or an array of objects), this contradicts the whole point of being "deterministic" which Swagger/OpenAPI had until now.)

Inventing a new sub-syntax seemed odd, thus I proposed JSONPath as I found that recently and it allows doing this retrieval (I've done exactly this).
Of course, it also allows more things, which we might not actually need, so defining the actually used subset (like just `.` and `[*]`) might be a way to go. (Then implementors can choose to use an existing JSONPath implementation, or implement the used subset themselves.)

I understand that the `$response` keyword was selected to be able to later extend this to request properties/parameters.

ePaul: Another thing I noted while trying to write down my proposal:

* In the `href` field we have an URL template in which parts within `{}` will be interpreted as some kind of expression and then replaced.

* In the link parameters object, we have "static or dynamic values". The specification text doesn't really say what this is, but from the examples it looks like "dynamic values" are strings starting with `$response` (which are interpreted as the same kind of expressions as before), while static values are everything else (i.e. booleans, numbers, objects, arrays, or strings not starting with `$response`). I guess only primitives are sensible for simple parameters (header/query/path), but for body parameters having an object or array would make sense. When using the dynamic values, we could also extract objects/arrays from the response, but then it is not clear how this would interact with the array expansion logic.

versions/3.0.md
```diff
@@ -1123,6 +1125,344 @@ application/json:
...
+    UserRepositories:
+      # returns array of '#/components/definitions/repository'
+      href: 'https://na2.gigantic-server.com/2.0/repositories/{$response.username}'
+```
+
+
+### Link Parameters
+Using the `operationId` to reference an operation in the definition has many benefits, including the ability to define media-type options, security requirements, response and error payloads.  Many operations require parameters to be passed, and these may be dynamic depending on the response itself.
+
+To specify parameters required by the operation, we can use a **Link Parameters Object**.  This object contains parameter names along with static or dynamic valus:
```
ePaul: An operation can have multiple parameters with the same name (e.g. one path parameter, one header parameter, one query parameter and one form parameter – while the path parameter's name can be changed in the definition without affecting any existing implementations, this doesn't work for the other parameter locations).

The current specification would pass the same value to all of them, or choose randomly one of them?

fehguy: I don't think you can have multiple parameters with the same name.  @webron ?

versions/3.0.md
ePaul: Here is a proposal trying to make it a bit clearer.

This stays with the `$response` prefix for the expressions, but adds a `[*]` whenever we are accessing an array element. It also shows how a property expression can be validated against a schema.

-----

### Link Object
The `Link Object` is responsible for defining a possible operation based on a single response.

  ---       | :---:  |     ---
href      |  url   | a relative or absolute URL template (see below) to a linked resource. This field is mutually exclusive with the `operationId` field.
 operationId| string | the name of an _existing_, resolvable OAS operation, as defined with a unique `operationId`.  This field is mutually exclusive with the `href` field.  Relative `href` values _may_ be used to locate an existing Operation Object in the OAS.
parameters  | Link Parameters Object | an Object representing parameters to pass to an operation as specified with `operationId` or identified via `href`.
description | string | a description of the link, supports GFM.

#### Link URI templates 

The `href` field is a string which can contain variable placeholders which will be filled from the response. Those placeholders are surrounded by braces (`{}`) and contain each an expression which follows the property expression grammar explained below. For use in templates, only expressions with a primitive result schema are allowed.

When evaluating a template against a response, each expression is evaluated and produces a list of values (possibly just one). The cross product of those lists is created, and for each element of the cross product we get one link by replacing the values into the URL template.

*(TODO: this needs an example. Also, it doesn't really solve when we have properties in objects in an array, where we don't really want a cross product, but zipping their result lists.)*

#### Property Expressions

A property expression (used in URL templates for links, as well as in links parameters inside a response) can have any of the following forms. Its validity depends on the response's schema definition, and with an actual response object (i.e. an object matching a schema) it can be evaluated to a result list (a list of values, each matching the schema).

Form | Validity condition | Schema | Evaluation (result list)
-----|---------|----|---
`$response` | The response has a schema. | The response's schema. | A list with the whole response value as the single element, or an empty list if there is no response value.
*expression* `.` *property-name* | The *expression*'s schema has type `object`, and includes a property with name *property-name* | The schema of that property | For each element `x` of the result of *expression*, if `x` has a field named *property-name*, include that field's value in the result list, else nothing.
*expression* `[*]` | The *expression*'s schema has type `array`, and has a defined `items` schema. | The `expression`'s schema's `items` schema. | For each element `x` of the result of *expression* and for each element `y` of `x`, include `y` in the result list.

*(TODO: add examples)*

### Link Parameters

Using the `operationId` to reference an operation in the definition has many benefits, including the ability to define media-type options, security requirements, response and error payloads.  Many operations require parameters to be passed, and these may be dynamic depending on the response itself.

To specify parameters required by the operation, we can use a **Link Parameters Object**.  This object contains parameter names along with static or dynamic values.

####  Fields

---|:---:|---
{name} | Any | {name} must be the name of a parameter defined for the operation identified by operationId. The value must be either a literal value matching the type of that parameter's schema, or a string with a property expression with has (in the context of the current response) a schema which is compatible to the named parameter's schema.

*(TODO: What happens if the operation has multiple parameters of the same name, like one header parameter and one query parameter of same name?)

---------

This omits all the examples. If wanted, I can do a pull request with my changes relative to this branch, including the given examples (adapting them to the new syntax).

Here is a still open problem (with both my wording as the existing one):

Given the Bitbucket API in your example, with a link definition like this:
```yaml
    UserRepository:
      # returns '#/components/definitions/repository'
      operationId: getRepository
      parameters:
        username: $response[*].owner.username
        slug: $response[*].slug
```
Assume an example response body like this:
```json
[ { "slug": "OpenAPI-Specification", "owner": { "username" : "OAI" } },
  { "slug": "swagger-codegen", "owner": { "username": "swagger" } } ]
```

We want that this produces two links (`/2.0/repositories/OAI/OpenAPI-Specification` and `/2.0/repositories/swagger/swagger-codegen`), not four of them (i.e. neither `/2.0/repositories/swagger/OpenAPI-Specification` nor `/2.0/repositories/OAI/swagger-codegen`).

versions/3.0.md
ePaul: The [2.0 specification says](http://swagger.io/specification/#parameterObject):

> A unique parameter is defined by a combination of a name and location.

The same is in the [current state of the 3.0 version](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#parameterObject).

While when designing an API such things should be avoided, they can happen when describing an existing API, and the wording sounds like that was intentional. (I was not there when this was decided, so I don't know.)

versions/3.0.md
```diff
@@ -1123,6 +1125,344 @@ application/json:
...
+---|:---:|---
+<a name="referenceRef"></a>$ref | `string` | If present, a reference to another Links Object.  Note, the presence of `$ref` in the Links Object will cause all _Patterned Objects_ to be ignored. 
+
+Field Pattern | Type | Description
+---|:---:|---
+<a name="linkName"></a>link name | [Link Object](#linkObject) | A short name for the link, following the naming constraints of the <a href="#definitionsName">definitions name</a>.  The link shall reference a single Link Object, or a JSON Reference to a single link object
+
+
+### Link Object
+The `Link Object` is responsible for defining a possible operation based on a single response.
```
dilipkrish: Why not just defer to the web linking spec [rfc5988](https://tools.ietf.org/html/rfc5988) to define what a link looks like? Will be especially useful if we want to link to an external specification for e.g. [links in HAL](https://tools.ietf.org/html/draft-kelly-json-hal-08#section-5). The only addition to that would be an `operationId` property to dereference operations within a swagger specification.

fehguy: That would dramatically increase the scope of this feature, including adding templating, etc.  I'm not sure we want to go that far with links in the OAI specs

dilipkrish: By "increase scope" do you mean the scope of this in the spec or the tooling? Wouldn't that be a concern of the tool implementers if it's the latter? Granted one of them is swagger-core 😀 

Putting it another way, is the spec the 🐔 Or the 🐣 ? I think it's similar to the json schema support and supporting oneOf, allOf etc. the spec obviously will have higher fidelity. IMHO There is bound to be a impedance mismatch between tooling and spec, but  I'd like the spec to not be limited in scope as a result of that. 

I think this goes back to the open-world/ closed-world view discussion that @darrelmiller brought up

fehguy: Increasing scope of the feature.  The tooling has to follow the feature definition, and if the scope of it is such that no tooling will work, well, we'll be back at square one.  So I'm less worried about swagger-core having to support it, but at the feature definition level, I don't have much interest in making something that no tooling will support, just because "someone else has described something similar elsewhere on the internet".  That will get us in trouble.

dilipkrish: 👍🏼

I'm asking cos if I were to implement a super set of the link object in [springfox](http://springfox.io) then I'd like that to be an acceptable/ compliant spec implementation i.e. Open world spec.  

Also could we add media type as an optional. That will be useful to disambiguate a path+representation, when not using `operationId`.

versions/3.0.md
dilipkrish: Can I suggest making examples an embedded gist?

versions/3.0.md
```diff
@@ -1123,6 +1125,344 @@ application/json:
...
+For computing links, and providing instructions to execute them, a mechanism is defined for accessing values in a response and using them as variables while invoking the linked operation.
+
+Field Name | Type | Description
+---|:---:|---
+<a name="referenceRef"></a>$ref | `string` | If present, a reference to another Links Object.  Note, the presence of `$ref` in the Links Object will cause all _Patterned Objects_ to be ignored. 
+
+Field Pattern | Type | Description
+---|:---:|---
+<a name="linkName"></a>link name | [Link Object](#linkObject) | A short name for the link, following the naming constraints of the <a href="#definitionsName">definitions name</a>.  The link shall reference a single Link Object, or a JSON Reference to a single link object
+
```
ralfhandl: Why again the limitation to `a-zA-Z0-9.-_\`? Our APIs use link names as well as definitions names that are valid variable names in Java, JavaScript, Swift, and C#, i.e. a Unicode letter followed by Unicode letters and Unicode numbers, not just a-zA-Z0-9.

versions/3.0.md
```diff
@@ -1123,6 +1125,344 @@ application/json:
...
+### Link Object
+The `Link Object` is responsible for defining a possible operation based on a single response.
+
+Field Name  |  Type  | Description
+  ---       | :---:  |     ---
+href      |  url   | a relative or absolute URL to a linked resource. This field is mutually exclusive with the `operationId` field.
+ operationId| string | the name of an _existing_, resolvable OAS operation, as defined with a unique `operationId`.  This field is mutually exclusive with the `href` field.  Relative `href` values _may_ be used to locate an existing Operation Object in the OAS.
+parameters  | Link Parameters Object | an Object representing parameters to pass to an operation as specified with `operationId` or identified via `href`.
+description | string | a description of the link, supports GFM.
+
```
ralfhandl: In addition to the `description` a `title` and `^x-` specification extensions would really be helpful.

DavidBiesack:
@fehguy can you elaborate on what you mean by "I don't think that mime-type traversals should be supported when following links (perhaps explicitly forbidden)"? I don't understand what this implies.

For example, in our APIs we use content type negotiation and wish to use different links to expose different flavors of the operation. For example, for a report resource, we may want to describe multiple links to the same resource/operation such as 
```
method: GET
uri: /reports/{reportId}
  link 1:  rel: self with response media type application/vnd.sas.report+json
  link 2:  rel: pdf with response media type application/pdf
  link 3:  rel: msword with a response media type application/vnd.openxmlformats-officedocument.wordprocessingml.document
```
(This ties back to our old friend, #146 ) 

I would not want this type of API specification to be forbidden.

versions/3.0.md
darrelmiller: @fehguy Being able to echo request parameters into link parameters without the server having to add them to the response would be convenient.  The example that springs to mind would be where someone makes some API multi-tenant using a path segment. e.g.

     /{tenantid}/my/resource/here

Allowing links to be created as 

     /{tenantid}/my/linked/resource

without having to add the tenant id to the body would be nice.

versions/3.0.md
webron: It is true for 2.0, unfortunately. I suggest we change that in 3.0. This will be a breaking change, in term of migration, but it makes the spec and tools easier.

versions/3.0.md
darrelmiller: I think there is value in being able to access the request parameters.  Allowing links to use inbound parameters without the server having to echo those parameters in the response body.

versions/3.0.md
darrelmiller: @fehguy  This might be a useful point to consider using URI Template syntax for specifying query parameters.  It would allow defining optional query parameters and all the logic of how to resolve those.  It also would help with resolving your red, green, blue scenario.

versions/3.0.md
fehguy: you bet!

versions/3.0.md
fehguy: only described parameters would be considered as part of the request linking

```yaml
paths:
  /foo:
    get:
      parameters: []
    # ...
```
`/foo?bar=true`

bar, if not a modeled parameter, will _not_ be accessible.

```yaml
paths:
  /foo:
    post:
      body:
        schema:
           type: object
           properties:
             id:
               type: string
    # ...
```

reaching inside the body parameter if complex is supported, but may need a special, distinct locator other than `body`
````
$request.body.id
```

versions/3.0.md
fehguy: What else is needed other than the following?
* URL
* http method
* request headers
* accept headers (perhaps included in headers)
* post body (parameter)

versions/3.0.md
fehguy:
```
href: '/foo'
method: 'get'
```

```
paths:
  /foo:
    $ref: 'http://gigantic-server.com/#paths/fooPath'
```

on `gigantic-server.com`:

```
paths:
  fooPath:
    get:
      operationId: conflict # !!!
```

fehguy:
Need to consider further the content-type negotiation between links

versions/3.0.md
shanemhansen: Typo: existence

versions/3.0.md
darrelmiller: @fehguy Having considered this more, the only time URI Templates would be useful is for external links where there are optional query parameters.  If I understand correctly, an external link that used query parameters would look something like,

components:
  
     links:
       UserRepositories:
          # returns array of '#/components/definitions/repository'
          href: 'https://na2.gigantic-server.com/2.0/repositories?username={$response.username}'&active={$response.activeOnly}

With this format, if response.activeOnly was null then the resulting url would contain `&active=` which may not be acceptable to the target server.   Unfortunately, defining a URI Template for this scenario wouldn't work either because the syntax that allows optional parameters uses a technique that would prevent us from naming the query parameters correctly.

versions/3.0.md
fehguy: @dilipkrish looks like we can't embed gists in the README directly.  I'll move this away to an external gist instead.

versions/3.0.md
fehguy: @ePaul i suggest that we bring wording suggestions into the PR instead of comments.

fehguy:
I've updated this PR.  Some comments are now out-of-place and resolved, so I'm removing them.

versions/3.0.md
fehguy: these have been added

PR #711, opened 2016-06-10
fixing typo

dret:


fehguy:
I think this is harmless, but be sure any changes for the next version of the spec goes to the OpenAPI.next branch...  @webron ?

webron:
Yeah, thanks @dret. Let's focus on changing the next version and not the current one, unless critical. I believe some other of @dret's PRs have been merged, including possibly the equivalent to this in 3.0.

Issue #712, opened 2016-06-10
Mime type -> Media type


dret:
while many people still say "mime type" (and everybody knows what they're talking about), the correct term now is "media type" (https://tools.ietf.org/html/rfc6838). if nobody objects, i'll create a PR to make this terminology change.

darrelmiller:
I certainly have no objections :-)

dret:
closed by https://github.com/OAI/OpenAPI-Specification/commit/d9dc7fd4e0a7f61d49e787c29a705487fa5a319b

PR #714, opened 2016-06-12
Terminology fix: "Mime type" -> "Media type"

dret:
this PR addresses #712. currently, it changes the anchor #mimeType to #mediaType, so if there are external references to this anchor, then those should be updated as well. alternatively, we could add a second anchor, effectively aliasing the section to be addressable by both identifiers.


dret:
thanks!

PR #720, opened 2016-06-27
Change markdown standard from GFM to CommonMark

darrelmiller:
Github Flavored markdown is not well defined and the page referred to by this specification that used to explain the differences between GFM and regular markdown has been removed by Github.  [CommonMark](http://commonmark.org/) is an attempt to create a well specified version of Markdown.  There is a detailed specification and many examples.  Using a style of markdown that is well documented and rigorously defined should make it easier for tooling writers to produce more consistent and compatible results.

As can be seen from the quick [reference for CommonMark](http://commonmark.org/help/) the vast majority of existing markdown in OpenAPI documents is likely to render identically.  One feature that has yet to be standardized is syntax for tables.  This is being worked on but reaching consensus for a complex feature is challenging.

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -1150,7 +1150,7 @@ href        | url    | a relative or absolute URL to a linked resource. This fie
 operationId | string | the name of an _existing_, resolvable OAS operation, as defined with a unique `operationId`.  This field is mutually exclusive with the `href` field.  Relative `href` values _may_ be used to locate an existing Operation Object in the OAS.
 parameters  | Link Parameters Object | an Object representing parameters to pass to an operation as specified with `operationId` or identified via `href`.
 headers     | Link Headers Object    | an Object representing headers to pass to the linked resource.
-description | string | a description of the link, supports GFM.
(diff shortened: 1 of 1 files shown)


webron:
👍

darrelmiller:
@ralfhandl  Well <strong>as far</strong> as I <em>can tell</em> <code>GitHub Flavored Markdown</code> already supports raw HTML tags.  That's what I used to format this line.  So there is nothing new there.

ralfhandl:
Well, in the meantime I tried my two favorite OpenAPI tools:
* Swagger UI has no problems with raw HTML tables in descriptions
* Swagger Editor completely ignores raw HTML tables in descriptions

Both accept GFM table syntax with the `|` separator character, with different degree of completeness.

Probably the spec should explicitly state that tools are expected to support CommonMark combined with raw HTML in descriptions.

webron:
Markdown, by definition, supports inline HTML, regardless of the flavor used. It seems CommonMark doesn't deviate from that support, so there's no need to explicitly mention it.

fehguy:
@darrelmiller can u rebase and merge!?

PR #729, opened 2016-07-09
Revised typo and language to MAY

jameswatts:


darrelmiller:
👍

webron:
I'm ok with fixing the typo, a bit less inclined to change the wording (as minor as it may be). Would much rather concentrate on fixing relevant wording issues in 3.0.

jameswatts:
@webron should I revise the PR to remove the language change?

webron:
If you don't mind. And feel free to submit something similar (with the revised language if applicable) to 3.0 so we won't forget there.

jameswatts:
@webron done.

jameswatts:
Proposed language update for **3.0** here: #760

Issue #724, opened 2016-07-04
Linking API to OpenAPI and documentation


dret:
would it make sense to recommend best practices for how an API should make its OpenAPI description (and potentially additional documentation) discoverable? we don't know how exactly an API is designed, but if it uses/supports hypermedia, we could recommend to use the link relations proposed here: https://tools.ietf.org/html/draft-wilde-service-link-rel
we might add text such as:
- If an API wants to make its OAI description discoverable from the API itself, it is recommended to use the "service-desc" link relation.
- If an API wants to make its external documentation discoverable from the API itself (the documentation that may be linked from the OpenAPI object via the "externalDocs" field), it is recommended to use the "service-doc" link relation.

i'd be more than happy to make a PR to that effect, but first wanted to raise an issue and discuss it.

eskoviak:
I think this would be an excellent feature. I have an use case in mind wherein having this sort of resource would be useful.

darrelmiller:
@dret We've discussed the use of described-by before to link an API to an OpenAPI description.  Do you see service-desc as being more suited for this scenario?

dret:
On 2016-07-04 15:58, Darrel wrote:
> @dret <https://github.com/dret> We've discussed the use of described-by
> before to link an API to an OpenAPI description. Do you see service-desc
> as being more suited for this scenario?

i am certainly biased because i have cooked up the service-desc and 
service-doc link relation types myself for exactly this kind of scenario.

described-by seems a bit odd to me in this case because it seems to be 
used mostly for cases where the "described-by" resource is metadata 
about the linking resource, such as who authored it, when, and these 
kinds of things. but the specification itself is so broad that with 
enough willingness, described-by is almost the only link relation you'll 
ever need on the web:

"The relationship A describedby B asserts that resource B provides a 
description of resource A. There are no constraints on the format or 
representation of either A or B, neither are there any further 
constraints on either resource."

in the end, all the link relation types have fuzzy definitions that 
invite a variety of interpretations. but i don't think described-by 
would be a great fit, because that might better be used to link to a 
resource with specific metadata about the linking resource.

dret:
On 2016-07-04 13:03, Ed Skoviak wrote:
> I think this would be an excellent feature. I have an use case in mind
> wherein having this sort of resource would be useful.

strictly speaking this would just be a best practice and not an actual 
feature that the spec could or should define. but it could be something 
to make this aspect of connectedness of resources on the web a bit 
better than it is today.

[...]

Issue #110, opened 2014-09-15
Define formal media type (MIME type) for Swagger Object 


philipashlock:
A media type of `application/swagger+json` has been [suggested on the mailing list](https://groups.google.com/forum/#!searchin/swagger-swaggersocket/application$2Fswagger$2Bjson/swagger-swaggersocket/Xd5cWuTusVs/QyRTUlS8Xi4J) before. I would recommend formally declaring this as part of the spec. 

RAML [defines its media type](http://raml.org/spec.html#overview) as `application/raml+yaml`

API Blueprint [defines its base media type](https://github.com/apiaryio/api-blueprint-ast#media-types) as `application/vnd.apiblueprint.ast` plus similar ones for [other serializations](https://github.com/apiaryio/api-blueprint-ast#media-types)

philipashlock:
For what it's worth, the White House open data policy is currently documenting `application/swagger+json` for metadata provided by federal agencies: https://project-open-data.cio.gov/v1.1/api/

If this should be changed, please flag it is an issue on https://github.com/project-open-data/project-open-data.github.io/issues

webron:
:+1:

olensmar:
I'm all for it and would like to go so far and suggest different mime-types for different encodings and Swagger versions

application/swagger-v20+json
application/swagger-v20+yaml
application/swagger-v12+json
application/swagger-v12+xml

(not really sure we need to add xml for 1.2 but it was in the spec as far as I can remember)

since Swagger 1.2 has separate files for resource listings and api declarations I suggest we make the api declaration implicit and have a separate mime-type for the resource listing

application/swagger-v12-resource-listing+json
application/swagger-v12-resource-listing+xml

or alternatively we make both the api declaration and resource listing explicit;

application/swagger-v12-resource-listing+json
application/swagger-v12-api-declaration+json
application/swagger-v12-resource-listing+xml
application/swagger-v12-api-declaration+xml

If we're aligned I can submit this to IANA... let me know your thoughts

/Ole

dilipkrish:
@olensmar Not really sure we need to separate resource listing and api declaration. Its implicit by definition that if its application/swagger-v12+json we're going to get a document that conforms to the swagger12 spec. likewise for swagger2 also. In the case of swagger 1.2 the protocol to access the resource listing to get to the api declarations is a shared understanding described by the spec. 

As an aside, related to #708 would you be able to assist/ hand-hold so I could also register

application/swagger-v20+hal+json

[...]

PR #734, opened 2016-07-12
first stab at discoverability (#724)

dret:
not sure where this would go, feel free to move around. it's a bit hard to get a feel for the spec's structure with no available toc/outline view.

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -2411,3 +2411,13 @@ Two examples for this:
 
 1. The [Paths Object](#pathsObject) may be empty. It may be counterintuitive, but this may tell the viewer that they got to the right place, but can't access any documentation. They'd still have access to the [Info Object](#infoObject) which may contain additional information regarding authentication.
 2. The [Path Item Object](#pathItemObject) may be empty. In this case, the viewer will be aware that the path exists, but will not be able to see any of its operations or parameters. This is different than hiding the path itself from the [Paths Object](#pathsObject) so the user will not be aware of its existence. This allows the documentation provider a finer control over what the viewer can see.
+
+
+## <a href="discovery"></a>OpenAPI and Documentation Discovery
+
+Service providers may want to make an OpenAPI description discoverable from (some) resources provided by the API. One typical example could be an API's "home" resource (the intended starting point of the API), which for example might use [a standardized home document format](https://tools.ietf.org/html/draft-nottingham-json-home-04). (This home document in turn may be made discoverable [by "home" links](https://tools.ietf.org/html/draft-wilde-home-xml-04#section-4.3.1) from other resources provided by the API.)
+
+For improving the discoverability of description and documentation available for an API, the following recommendations are made:
+
+* It is recommended that links from API resources to the OpenAPI description of the API are using [the "service-desc" link relation type](https://tools.ietf.org/html/draft-wilde-service-link-rel-01#section-4.2).
+* It is recommended that additional (primarily human-readable) documentation, which often may be the documentation that is linked from the OpenAPI object via the "externalDocs" field, is linked from API resources by using (the "service-doc" link relation type](https://tools.ietf.org/html/draft-wilde-service-link-rel-01#section-4.1).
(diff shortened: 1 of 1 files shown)


DavidBiesack:
I recommend making the wording more like a guideline, less like a very loose suggestion. For example, rather than "Service providers may want to make an OpenAPI description discoverable", consider "Service providers wishing make an OpenAPI description discoverable may ...."  You can also be more concise by dropping the sentence "In such a scenario, it can help if there is a predictable way in which API description and documentation is made discoverbale from the API itself. " which is not necessary. Change "the following recommendations are made:" and "It is recommended that links" to just "Provide a link to the Open API Specification with the service-desc link relation", and so on.

DavidBiesack:
The proposal should specify the JSON Home "hints"/"formats" should contains the supported media type,format, and versions (such as application/openapi+json;version=2.0 vs application/openapi+yaml;version=2.0 as described in #110 )

dret:
On 2016-07-12 17:59, David Biesack wrote:
> The proposal should specify the JSON Home "hints"/"formats" should
> contains the supported media type,format, and versions (such as
> application/openapi+json;version=2.0 vs
> application/openapi+yaml;version=2.0 as described in #110
> <https://github.com/OAI/OpenAPI-Specification/issues/110> )

i am not sure that this needs to be recommended. media type hints are 
optional and if people feel like using them, that's ok. if they don't 
feel like using them, that's ok, too. why would you move away from that 
optional model and tell them that they should use hints?

dret:
On 2016-07-12 17:56, David Biesack wrote:
> I recommend making the wording more like a guideline, less like a very
> loose suggestion. For example, rather than "Service providers may want
> to make an OpenAPI description discoverable", consider "Service
> providers wishing make an OpenAPI description discoverable may ...." You
> can also be more concise by dropping the sentence "In such a scenario,
> it can help if there is a predictable way in which API description and
> documentation is made discoverbale from the API itself. " which is not
> necessary. Change "the following recommendations are made:" and "It is
> recommended that links" to just "Provide a link to the Open API
> Specification with the service-desc link relation", and so on.

thanks for the feedback. i did a bit of rewriting at 
https://github.com/OAI/OpenAPI-Specification/pull/734/commits/0932f82b35b1196ca3d1cbf2e2f5c13c39d3d6a6 
and hope that this addresses your issues.

DavidBiesack:
If we are recommending service-desc then I think it makes sense to recommend a consistent and complete representation of the link so that people have some guidance to follow . I did use the word "should". I'd rather not leave people guessing on what media type to use (and perhaps guessing wrong) or making them search somewhere to find out, or simply omit the type because the guidance does not mention it.

The rewrite looks much improved, thanks. Minor tweak: change "are using [the "service-desc" link relation type]" to "may use [the "service-desc" link relation type]"

dret:
thanks for the feedback!

On 2016-07-14 15:16, David Biesack wrote:
> If we are recommending service-desc then I think it makes sense to
> recommend a consistent and complete representation of the link so that
> people have some guidance to follow . I did use the word "should". I'd
> rather not leave people guessing on what media type to use (and perhaps
> guessing wrong) or making them search somewhere to find out, or simply
> omit the type because the guidance does not mention it.

personally, i think it would be fine to omit the type, unless you have a 
specific reason to add such a hint to the link. i am curious if others 
have any opinion about mentioning/recommending/requiring a type hint.

> The rewrite looks much improved, thanks. Minor tweak: change "are using
> [the "service-desc" link relation type]" to "may use [the "service-desc"
> link relation type]"

https://github.com/OAI/OpenAPI-Specification/pull/734/commits/542f10d3b3566e8486aef2d83bb715bae009b188 
fixes this and uses the stronger "should" language you're suggesting above.

dret:
anything else that's required here? thanks for the feedback so far!

dret:
anything i can help to move this forward? it has been in the pipeline for a while. thanks!

dret:
anything i can help to move this forward? it has been in the pipeline for a while. thanks!

webron:
@dret - thanks for the submission. we're got a lot to cover and didn't get around to it this time, but it's on the queue for our next discussion.

RobDolinMS:
@webron Would you please give this a look and comment on what should happen next?

webron:
This is more of @darrelmiller's domain. If I understand it properly, I don't think it'll go into 3.0.0.. It won't make it into RC1 for sure.

PR #740, opened 2016-07-21
Grammar - Removes duplicate articles a and an

baileylo:
Fixes grammar mistake


ePaul:
@baileylo could you apply the same change to [the 3.0 version in the OpenAPI.next branch](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md)?

PR #741, opened 2016-07-21
JSON Schema Support Proposal

webron:
## **WIP**

This PR will build up to a proposal for the changes in handling JSON Schema in OpenAPI.next.
This comment will be modified as the proposal evolves.

At this time, please avoid adding comments to this proposal. Comments _will_ be deleted. When the status of the PR changes, comments and discussions would be welcome.

This PR does not cover representation of XML and alternative data structures.

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -1294,7 +1294,7 @@ $ref: 'Pet.yaml'
 $ref: 'definitions.yaml#/Pet'
 ```
 
-#### <a name="schemaObject"></a>Schema Object
+#### <a name="schemaObject"></a>Schema Object 
 
 The Schema Object allows the definition of input and output data types. These types can be objects, but also primitives and arrays. This object is based on the [JSON Schema Specification Draft 4](http://json-schema.org/) and uses a predefined subset of it. On top of this subset, there are extensions provided by this specification to allow for more complete documentation.
 


fehguy:
# JSON Schema

## Combining Schemas

  - **allOf**: An instance is valid against this keyword if and only if it is valid against all the schemas in this keyword's value.  To validate against allOf, the given data must be valid against all of the given subschemas.
  ```
  {
    "allOf": [
      { "type": "string" },
      { "maxLength": 5 }
    ]
  }
  ```

  - **oneOf**: An instance is valid against this keyword if and only if it is valid against exactly one schema in this keyword's value.
  ```
  {
    "oneOf": [
      { "type": "number", "multipleOf": 5 },
      { "type": "number", "multipleOf": 3 }
    ]
  }
  ```

  - **anyOf**: An instance is valid against this keyword if and only if it is valid against at least one schema in this keyword's value.  To validate against anyOf, the given data must be valid against any (one or more) of the given subschemas.
  ```
  {
    "anyOf": [
      { "type": "string", "maxLength": 5 },
      { "type": "number", "minimum": 0 }
    ]
  }
  ```

* Analysis
  - allOf.  It's already supported.  It results in a single schema, and the result is a single model definition for code generation.

  Example:
  ```
  post:
    operationId: doSomething
    schema:
      allOf:
      - type: object
        properties:
          a:
            type: string
      - type: object
        properties:
          b:
            type: string
    responses:
      200:
        description: a successful response
        schema:
          allOf:
          - type: object
            properties:
              c:
                type: string
          - type: object
            properties:
              d:
                type: string
  ```

  For this example, code generators will do the following:
  1. Create an input model.  It would be an "inline model", with string properties `a`, `b`.
  2. Create an output model.  It would be an "inline model", with string properties `c`, `d`.

  ```
  OutputModel doSomething(InputModel input) { ... }
  ```

  Pitfalls:
  * inline models are ugly and potentially duplicative

  - oneOf.  We create discrete models for each input array item as well as discrete models for each output array item.
  ```
  post:
    operationId: doSomething
    schema:
      oneOf:
      - type: object
        properties:
          a:
            type: string
      - type: object
        properties:
          b:
            type: string
    responses:
      200:
        description: a successful response
        schema:
          oneOf:
          - type: object
            properties:
              c:
                type: string
          - type: object
            properties:
              d:
                type: string
  ```
  For this example, a code generator would need to do the following:
  1.  Create input models.  Say, `A`, `B`
  2.  Create output models, `C`, `D`
  3.  Input models implement an interface, for example `AB`
  4.  Create an output model interface, `CD`, `C` and `D` implement it
  ```
  class A implements AB;
  class B implements AB;
  class C implements CD;
  class D implements CD;

  pubilc AB doSomething(CD input) {
    Object response = invoke(serialize(cd));
    C c_type = deserialize(response, C.class);
    D d_type = deserialize(response, D.class);
    // only one is allowed!
    if(c_type != null && d_type == null)
      return c_type;
    if(c_type == null && d_type != null)
      return d_type;
    throw new InvalidTypeException(...)
  }
  ```
  
  Pitfalls:
  * Refences models in array must implement appropriate interface
      - That interface is the _union_ of all properties in the arrays, which isn't always possible, since same property name may have different type
      - May create many interfaces that don't make much sense.  Reduces the value of interfaces in general
      - Referenced models need to implement many interfaces if used in multiple `oneOf` scenarios, multiplying the combinatorics
  * More straight forward with validation than modeling, but validation ignored for generated clients

  - **anyOf**.  This means that we would have separate, discrete models for each element in the array.  For an operation response, the code generator will need to try to deserialize the payload into each model definition.  The first one that matches would be the accepted type.  For requests, the post body would need to implement an interface, and each of the schemas implement that interface.
  ```
  post:
    operationId: doSomething
    schema:
      anyOf:
      - type: object
        properties:
          a:
            type: string
      - type: object
        properties:
          b:
            type: string
    responses:
      200:
        description: a successful response
        schema:
          anyOf:
          - type: object
            properties:
              c:
                type: string
          - type: object
            properties:
              d:
                type: string
  ```
  For this example, a code generator would need to do the following:
  1.  Create input models.  Say, `A`, `B`
  2.  Create output models, `C`, `D`
  3.  Input models implement an interface, for example `AB`
  4.  Create an output model interface, `CD`, `C` and `D` implement it
  ```
  class A implements AB;
  class B implements AB;
  class C implements CD;
  class D implements CD;

  pubilc AB doSomething(CD input) {
    Object response = invoke(serialize(cd));
    C c_type = deserialize(response, C.class);
    if(c_type != null)
      return c_type;
    D d_type = deserialize(response, D.class);
    if(d_type != null)
      return d_type;
    throw new InvalidTypeException(...)
  }
  ```
Pitfalls:
* Same as `oneOf`

Recommendation
* Add support for `oneOf`, `allOf`.  Both are about the same to implement
* Add support for `not` but make it clear that it's only for validation


## Type Primitives
Previously, for primitive types, the `format` _must_ be part of a strict enumeration.  This makes it difficult to extend, for example:

```
type: string
format: uuid
```

May not be specific enough for UUID type 1 vs. type 4.

Recommendation
* `type` _must_ be of a supported set of values.  You may not invent new `type`s period.
* `format` _is recommended_ to be from a bounded set of values.  You _may_ invent new formats as you see fit, however tooling which does not support the format **must** fall back to the default `type` without a format.  This behavior must be made clear in the specification.  For example:

```
type: integer
```

will fall back to:
```
type: integer
format: int64
```
for example (this is just for illustration).

We should also include `time`, `duration` formats. Also consider how to handle dates with different formats

Add extension mechanism, via "look here first" repository

## required, readOnly, writeOnly
Requested extensions to support these value types.

Recommendation:
* I see no reason why to not allow these attributes.  In many (most?) cases they don't make much sense but they will do little harm.

## Null
Requested to support the null value.

Recomendation:
* Add a `nullable` attribute to indicate that a property may be null.

## Discriminator
Because of the lack of `oneOf` support, users must use a discriminator and a base type to handle polymorphic models.  It requires full knowledge of all subtypes in advance, which is pretty much impossible.

Recommendation:
* `oneOf` support reduces the need for this, however the discriminator is important.
* Discriminator currently has only a single value--the attribute which would indicate the `type` of the object to deserialize to.  It is missing a mapping of values to type, which should be added.

## Unsupported fields
`id`, `$schema`, `additionalItems`, `definitions`, `patternProperties`, `dependencies`
Make it explicitly clear what is supported and what is not.

## file primitive
Suggest handling with binary data definition

## description on enums
Enums currently only allow for a value.  Many requests to have a description and other metadata to follow

Recommendation:
* Support `enum` objects with a proposed schema:
  ```
  value
  name
  description
  ```

## binary data definition

## Clarify additionalProperties default setting

fehguy:
@OAI/tdc please add comments here before we make a formal PR

DavidBiesack:
This pretty closely matches what I would look for.

(Separately, it would be nice to have an Open API spec validation tool that detects impossible schema, for example, if using oneOf, there may be a common input that is compatible with both C and D. (This could arise if C and D both share some members, but their differences are not `required`. A tool that can detect that would be useful... but I admit that this may be undecidable in general.)

Looking forward to the PR. Thanks

jasonh-n-austin:
👍 ❤️ :shipit: 😍

RomanGotsiy:
How about documentation? How to represent `oneOf` or `anyOf` in documentation?
How to render HTML-form?
By bringing `oneOf` and `anyOf` into the OpenAPI you're restricting it only to validation... Tooling other than validation either won't support it at all or support will be limited/buggy.
It's not only my guesses. Let's take an example of `discriminator` which is much simpler than `oneOf` and `anyOf`. However, there are lots of tools that doesn't support it nicely (even SwaggerUI).

> Add support for `not` but make it clear that it's only for validation

How to understand this? Does it mean that code-generation or documentation should ignore schema if it contains `not`?

fehguy:
@RomanGotsiy one step at a time please.  Read the comments, we are not restricting `oneOf` or `anyOf` to just validation, that's what the majority of the discussion was about--how to represent it in tooling.

`not` means that if the payload matches the `not` schema, it should be rejected.  There is nothing to generate, code wise, but the schema should be used to reject the request.

RomanGotsiy:
@fehguy 
> one step at a time please
yes, sorry. This time I will try to be more structural :)

> we are not restricting oneOf or anyOf to just validation
you misunderstood me. I mean `oneOf` and `anyOf` are difficult to support in anything beyond validation, so, most likely, it won't be supported by the majority of tooling (discriminator is a vivid example).

So yes, de jure you are not restricting `oneOf` or `anyOf` to just validation, but de facto... it is a question.

I just want to get attention to `visualization`. Your [approving changes guideline](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/DEVELOPMENT.md#approving-changes) says the following:

> Visualization. Can the specification change be graphically visualized somehow in a UI or other?

fehguy:
The answer is yes, and details will be coming.

ralfhandl:
Question on `nullable`: how will this work for structured properties that are `$ref`erencing a reusable schema? Can the `nullable` attribute be placed within the JSON Reference?

````JSON
"Address": {
    "$ref": "#/components/definitions/PeopleAPI.Address",
    "nullable": true
}
````

**Comment by fehguy** [comment]
@ralfhandl we haven't tackled the details yet but my guess is an allOf would be required

**Comment by ralfhandl** [comment]
@fehguy Using `allOf` might make it more difficult for tools. 

Following the discussion in #556 I experimented with `description` nested within JSON References versus `allOf`. In the `allOf` variant both Swagger UI and Swagger Editor were more likely to "give up" on type expansion and just mark a property's type as "undefined", whereas additional properties within JSON References allowed correct expansion of types.

**Comment by tedepstein** [comment]
There's a conflict between the {{allOf}} semantics and the default value of {{false}} for {{additionalProperties}}.  

Consistent with JSON schema, an {{allOf}} schema is valid "if and only if it is valid against all the schemas in this keyword's value."

In JSON Schema, this example:
```
{
  "a" : "Jack",
  "b" : "Jill"
}
```
...is valid against this schema:
```
    allOf:
    - type: object
      properties:
        a:
          type: string
    - type: object
      properties:
        b:
          type: string
```
This holds true because:
* The example is valid w.r.t. the first schema (defining only property 'a'); and
* The example is valid w.r.t. the second schema (defining only property 'b'); and

But both of those conditions depend on the default of {{additionalProperties: true}}.  If additionalProperties were false, then the example wouldn't be valid w.r.t. either of those subschemas:
* The example is _invalid_ w.r.t. the first schema because the additional property 'b' is disallowed.
* The example is _invalid_ w.r.t. the second schema because the additional property 'a' is disallowed.

IIUC, based on @webron's earlier comments, the intention in the current 2.0 and planned 3.0 versions of OpenAPI is to redefine the default value of additionalProperties to false, and disallow setting additionalProperties to true.

We can talk about how to resolve this.  But first:
* Is the problem clear enough?
* @webron, do I understand your earlier comments correctly?

ePaul:
@fehguy I think in your format example of `type:integer` the fallback description is backward (i.e. it should be »`type: integer, format: int64` falls back to `type:integer`«.)

fehguy:
@ePaul no, exactly the opposite.  If you don't specify the format, the underperscribed type, `integer` would have a fallback to `int64` in my example.  `integer` on its own, is simply ambiguous.

ePaul:
Hmm, then the text needs to be clearer, I've read it differently. (And the spec would need to say which format is the default for each type.)

My approach would be to say:

* `type: string` without format = any JSON string, without any restrictions or special interpretation.
* `type: number` without format = any JSON number (= a decimal number with any number of digits before and after the decimal point, possibly represented on wire with e-notation). In Java this would correspond to BigDecimal.
* `type: integer` without format = any JSON integer (= a decimal number without a decimal point, with any number of digits, possible represented on wire with e-notation using a positive exponent). In Java this would correspond to BigInteger.

(For types `boolean`, `object` and `array` there are no formats anyways, and `null` is not a real type.)


And all formats unsupported by an implementation would fall back to those meanings – this allows any application to work around the non-support. This is how I interpreted the current spec anyways.

We could introduce a `format: abitrary` or similar for this meaning.

webron:
Closed in favor of (the now merged) #894.

PR #742, opened 2016-07-22
Links proposal (updated)

fehguy:
The previous PR was updated, and the comments got all jacked up.  So here's a new PR, the [old one](https://github.com/OAI/OpenAPI-Specification/pull/710) has been closed but you can see comments there in all their glory.

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -938,7 +938,7 @@ default:
 ```
 
 #### <a name="responseObject"></a>Response Object
-Describes a single response from an API Operation.
+Describes a single response from an API Operation, including design-time, static `links` to operations based on the response.
 
 ##### Fixed Fields
 Field Name | Type | Description
@@ -947,6 +947,8 @@ Field Name | Type | Description
 <a name="responseSchema"></a>schema | [Schema Object](#schemaObject) | A definition of the response structure. It can be a primitive, an array or an object. If this field does not exist, it means no content is returned as part of the response. As an extension to the [Schema Object](#schemaObject), its root `type` value may also be `"file"`. This SHOULD be accompanied by a relevant `produces` mime-type.
 <a name="responseHeaders"></a>headers | [Headers Object](#headersObject) | A list of headers that are sent with the response.
 <a name="responseExamples"></a>examples | [Example Object](#exampleObject) | An example of the response message.
+<a name="responseLinks"></a>links | [Links Object](#linksObject) | An object representing operations related to the response payload.
+
 
 ##### Patterned Objects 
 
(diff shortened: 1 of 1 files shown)


fehguy:
To support HAL-style linking, as suggested in #688, I propose the following:

* A mechanism for design-time, _static_ links between responses and operations to make operations on responses visible to consumers
* A reusable linking structure to allow link definitions to be DRY
* A mechasim for parameterized link traversal to avoid server-side computation of links through reuse of existing, OAS-defined operations

Additional considerations:
* This is not intended to be a _run time_ linking solution.  This solution does not require users to redesign their runtime API.  It _may_ be able to represent the intent of runtime HAL-style links in responses.  It does _not_ intend to represent dynamically generated, arbitrary links.
* A clear separation of concerns, between payloads and operations.  Therefore no _operation data_ shall be tied to payload schemas.

Spec changes
* Introduction of a `Link Object` to represent a single `link`
* Addition of a  `links` object in the `Response Object` to define what links are available from a specific operation
* Addition of a `links` section in the `Components Object` to allow reuse across operations
* Introduction of a `Link Parameters Object` to represent parameters to build a link from
* A mechanism for reading values from response payloads and using them for client-side link computation

versions/3.0.md
```diff
@@ -1123,6 +1125,395 @@ application/json:
...
+
+Payload Variables:
+```yaml
+id: df71a505-07bc-458a-a5c0-73c0340d1ec7
+firstname: Ash
+lastname: Williams
+missingValue: null
+```
+
+In situations where variables appear in an array, an array of variables will be extracted.  For example:
```
fehguy: This needs more documentation

versions/3.0.md
```diff
@@ -1123,6 +1125,395 @@ application/json:
...
+
+Source Location | reference identifier | example value | example reference | notes
+---|:---|:---|:---
+HTTP Method            | `$method`         | `/users/{$method}`                | The allowable values for the `$method` will be those for the HTTP operation
+Requested content type | `$accepts`        | `/users/3?format={$accepts}`      |  
+Request parameter      | `$request`        | `/users/{$request.id}`            | Request parameters must be declared in the `parameters` section for the operation or they cannot be used in substitution.  This includes request headers
+Request body           | `$requestBody`    | `/users/{$requestBody.user.uuid}` | For operations which accept payloads, references may be made to portions of the `requestBody` or the entire body itself
+Request URL            | `$url`            | `/track?url={$url}`               | 
+Response value         | `$response`       | `{$response.uuid}`                | Only the payload in the response can be accessed with the `$response` syntax
+Response header        | `$responseHeader` | `{$responseHeader.Server}`        | Single header values only are available.
```
fehguy: Added fields per https://github.com/OAI/OpenAPI-Specification/pull/710#discussion_r67555409

versions/3.0.md
```diff
@@ -1123,6 +1125,395 @@ application/json:
...
+The `Link Object` is responsible for defining a possible operation based on a single response.
+
+Field Name  |  Type  | Description
+  ---       | :---:  |     ---
+href      |  url   | a relative or absolute URL to a linked resource. This field is mutually exclusive with the `operationId` field.
+ operationId| string | the name of an _existing_, resolvable OAS operation, as defined with a unique `operationId`.  This field is mutually exclusive with the `href` field.  Relative `href` values _may_ be used to locate an existing Operation Object in the OAS.
+parameters  | Link Parameters Object | an Object representing parameters to pass to an operation as specified with `operationId` or identified via `href`.
+description | string | a description of the link, supports GFM.
+
+Locating a linked resource may be performed by either a `href` or `operationId`. In the case of an `operationId`, it must be unique and resolved in the scope of the OAS document.  Because of the potential for name clashes, consider the `href` syntax as the preferred method for specifications with external references.
```
fehguy: Clarified `href` vs. `operationId`

versions/3.0.md
```diff
@@ -1123,6 +1125,395 @@ application/json:
...
+      properties: 
+        id: 
+          type: integer
+        title: 
+          type: string
+        repository: 
+          $ref: '#/components/definitions/repository'
+        author: 
+          $ref: '#/components/definitions/user'
+```
```
fehguy: Will move this into a gist link eventually

versions/3.0.md
```diff
@@ -1123,6 +1125,395 @@ application/json:
...
+
+For computing links, and providing instructions to execute them, a mechanism is defined for accessing values in a response and using them as variables while invoking the linked operation.
+
+Field Name | Type | Description
+---|:---:|---
+<a name="referenceRef"></a>$ref | `string` | If present, a reference to another Links Object.  Note, the presence of `$ref` in the Links Object will cause all _Patterned Objects_ to be ignored. 
+
+Field Pattern | Type | Description
+---|:---:|---
+<a name="linkName"></a>link name | [Link Object](#linkObject) | A short name for the link, following the naming constraints of the <a href="#definitionsName">definitions name</a>.  The link shall reference a single Link Object, or a JSON Reference to a single link object
```
fehguy: @ralfhandl the naming constraints are for consistency.  No reason to have different naming structures here from the model definitions.

versions/3.0.md
```diff
@@ -1123,6 +1125,396 @@ application/json:
...
+### Link Object
+The `Link Object` is responsible for defining a possible operation based on a single response.
+
+Field Name  |  Type  | Description
+  ---       | :---:  |     ---
+href      |  url   | a relative or absolute URL to a linked resource. This field is mutually exclusive with the `operationId` field.
+ operationId| string | the name of an _existing_, resolvable OAS operation, as defined with a unique `operationId`.  This field is mutually exclusive with the `href` field.  Relative `href` values _may_ be used to locate an existing Operation Object in the OAS.
+parameters  | Link Parameters Object | an Object representing parameters to pass to an operation as specified with `operationId` or identified via `href`.
+description | string | a description of the link, supports GFM.
+^x-         | Any    | Allows extensions to the OpenAPI Schema. The field name MUST begin with `x-`, for example, `x-internal-id`. The value can be `null`, a primitive, an array or an object. See [Specification Extensions](#specificationExtensions) for further details.
```
fehguy: @ralfhandl I added `description` and `x-*` extensions

versions/3.0.md
```diff
@@ -1123,6 +1125,397 @@ application/json:
...
+
+### Link Object
+The `Link Object` is responsible for defining a possible operation based on a single response.
+
+Field Name  |  Type  | Description
+---|:---:|---
+href        | url    | a relative or absolute URL to a linked resource. This field is mutually exclusive with the `operationId` field.
+operationId | string | the name of an _existing_, resolvable OAS operation, as defined with a unique `operationId`.  This field is mutually exclusive with the `href` field.  Relative `href` values _may_ be used to locate an existing Operation Object in the OAS.
+parameters  | Link Parameters Object | an Object representing parameters to pass to an operation as specified with `operationId` or identified via `href`.
+headers     | Link Headers Object    | an Object representing headers to pass to the linked resource.
```
fehguy: @DavidBiesack With the addition of headers for the linked resource, we can traverse media types now.

versions/3.0.md
fehguy: returns header value as a string, up to the caller to parse it if an array, etc.

fehguy: need to explain how to reach into non-json bodies

fehguy: If we continue with `consumes` and `produces` we should stay consistent with `accepts` and `$responseHeader.Content-Type`.

Clarify case sensitivity as well

fehguy:
approved by @OAI/tdc

kevinswiber:
I was approaching this PR with some hope that it could be used to support hypermedia types being represented in OAS, but I see now that there's a requirement for the links to be "static."  This requirement alone breaks the coupling semantics for [Web Linking](https://tools.ietf.org/html/rfc5988), on which (nearly?) all the hypermedia types depend.

And that's okay if that's not the supported use case for this feature.  But it leaves me a bit bewildered.  Can someone help by defining use cases or existing APIs that could benefit from this feature in its current form?  It seems like this introduces a lot of hopping around between references to grok what's going on.  What are the intended benefits?  Maybe it's obvious to most people.  Hopefully, it is.  But right now, I'm in the dark.

Thanks.

fehguy:
@kevinswiber consider reading the comments in the previous (mangled) PR and the corresponding issues for the rationale

PR #749, opened 2016-08-09
Fix error in json example

robertpanzer:
The XML property example in JSON nested the name property inside the id property which is wrong I guess and the two properties should be next to each other.


ePaul:
Thanks for the fix.

Could you have a look whether the same problem also affects [the 3.0 version in the OpenAPI.next branch](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md) and submit a PR to fix this too (into the OpenAPI.next branch, not into master)?

Issue #721, opened 2016-06-28
Making representations a first class concept to address a number of structural concerns


darrelmiller:
This is a proposal to address a number of the issues aggregated under #560.
Specifically:
- Having multiple schemas for requests and responses #270 
- It removes the produces/consumes ambiguity described here #581 
- It extends on @jasonh-n-austin 's addition of the RequestBody object #670 
- It avoids some of the issues around formData by making it obsolete #222 

The basic premise is that OpenAPI models `operations` that interact with `resources` using a specific HTTP method.  The interaction involves sending `representations` and retrieving those `representations`.  A single `operation` may support different `representations` identified by a `media-type`.  

Consider the following `GET` operation that returns one of two `representations`.

``` yaml
get:
  description: Returns pets based on ID
  summary: Find pets by ID
  operationId: getPetsById
  responses:
    '200':
      description: pet response
      representations: 
         application/json:
            schema:
              type: array
              items:
                $ref: '#/definitions/Pet'
         text/html:
    default:
      description: error payload
      representations: 
        application/json:
          schema:
            $ref: '#/definitions/ErrorModel'
        text/html:
parameters:
- name: id
  in: path
  description: ID of pet to use
  required: true
  type: array
  items:
    type: string
  collectionFormat: csv
```

The `description` and `headers` properties of a [response object](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#response-object) stay, but the `schema` and `examples` move under the `representation object` that is defined as a property of the `representations object`.  The `examples` property will need to change to either an array of objects, or a single object because all examples would be for the same media type.  

By defining the supported media types on the `representations object`, there is no longer a need for a `produces` array.

It is important to note that different `representations` should not be semantically different when accompanied with the same class of status code.  A request to a resource should always the same **thing** (for some unfortunately nebulous definition of thing).  However, the syntax of that representation may be different and the amount of information contained may be different, but from a consumer's perspective. it is same concept, regardless of the representation.  This is why the `description` property is the same for all representations.

The following is an example of a POST request that may send a HTML form as a request body.

``` yaml
tags:
- pet
summary: Updates a pet in the store with form data
description: ""
operationId: updatePetWithForm
parameters:
- name: petId
  in: path
  description: ID of pet that needs to be updated
  required: true
  type: string
requestbody:
  description: Updated status of the pet
  required: false
  representations:
    application/x-www-form-urlencoded:
      schema: 
        properties:
[...]

Issue #665, opened 2016-04-20
Create serialization strategy for non-body params


fehguy:
Follow-up for #564, meta issue #565

ePaul:
@fehguy I think the #564 should be #654 instead. (Feel free to delete this comment after fixing it in the initial comment.)

ePaul:
I just found that there currently is an [Internet draft discussed in the HTTP working group](https://tools.ietf.org/html/draft-ietf-httpbis-jfv-00). This basically defines a format to transport JSON arrays (while omitting the brackets) in HTTP headers.

I guess this could work for custom headers with OpenAPI schema definitions for them, as well as for future official HTTP headers following an RFC coming out of this internet draft.

Note that this defines a serialization scheme just for arrays, not for objects, because that way it works nicely with multiple occurences of a Header, which is equivalent to putting the values together in one, separated by commas.

It will not work for most existing headers using different syntax, for that we would need some different serialization scheme.

*(originally posted at #652, but @fehguy pointed out it belongs here.)*

dcolens:
I bet this was discussed before, but what about using something like http://api.jquery.com/jquery.param/ ?

fehguy:
Covered in #878

PR #761, opened 2016-08-18
Issue 721 - Content objects for describing request/response payloads

darrelmiller:


zeeshanejaz:
@darrelmiller is this solely for the purpose of enabling profiles? Sorry I have been late to the party. Can you reference issues that are solved by this. Thanks.

ePaul:
I guess this could also solve at least parts of #146, beside #721?

ePaul:
As already mentioned in [a comment by @fehguy in #721](https://github.com/OAI/OpenAPI-Specification/issues/721#issuecomment-234666646), there is still missing some explanation on how the schema of an `application/x-www-form-urlencoded` (or `multipart/form-data`) representation is mapped to the actual value (or the other way around), i.e. a serialization format.

As this is also missing for other kinds of parameters, I'm okay with addressing it in a separate PR, I just wanted to mention it here again. I guess #665 will have to address that too.

darrelmiller:
@zeeshanejaz The issues addressed by the PR mentioned in the original issue #721. Profiles are simply a way I have found to address the request that some people want to have multiple schemas for the same status code and media type.

darrelmiller:
@ePaul Being able to specify multiple response representations opens the door to supporting #146 but I think it is outside of the spec to try and tie a particular input parameter to the selection of the output representation.  There is still a server implementation behind the API and it is up to the server to decide which representation will be returned.  If the server wants to expose a parameter that directly selects the representation, then it is free to do so.  I don't think the OpenAPI spec needs to get involved at this level.  
As far as generating client code is concerned, the response content type header should provide everything that is needed to deserialize the payload into the appropriate structure.  It is true that most languages don't support the notion of a method that can return multiple different types, so some compromise will need to be made on the signature of calling method.  If having a precise signature is important, then perhaps using multiple paths is a better option.

darrelmiller:
@ePaul I agree that some discussion of how form-data can be described by schemas is important.  But the discussion of how non-JSON data can be described is also bigger than just forms.  Hopefully we can address that subject very soon.

fehguy:
@OAI/tdc let's get your input on this

versions/3.0.md
```diff
@@ -295,16 +292,17 @@ Field Pattern | Type | Description
...
+        "200": {          
           "description": "A list of pets.",
-          "schema": {
-            "type": "array",
-            "items": {
-              "$ref": "#/definitions/pet"
+          "representations" : {
+            "application/json" : {
+                "schema": {
+                            "type": "array",
```
whitlockjc: Indentation looks weird when rendered via Markdown.

versions/3.0.md
```diff
@@ -367,24 +365,28 @@ Field Pattern | Type | Description
...
         }
       },
       "default": {
         "description": "error payload",
-        "schema": {
-          "$ref": "#/definitions/ErrorModel"
+        "representations": {
+          "text/html" : {
+            "schema": {
+                      "$ref": "#/definitions/ErrorModel"
```
whitlockjc: Indentation again.

whitlockjc:
How do representations support the `void` response, where there is no body?

versions/3.0.md
```diff
@@ -476,42 +477,46 @@ Field Pattern | Type | Description
...
+      "application/x-www-form-urlencoded":{
+        "schema" : {
+           "properties": {
+              "name" 
+                "description" : "Updated name of the pet",
+                "type": "string"
+              "status":
+                "description" : "Updated status of the pet",
+                "type": "string"
+           },
```
dilipkrish: It seems like we're going from a flat array of parameters to an object with those fields. Is the intention that we treat the object as an array of form variables?

versions/3.0.md
```diff
@@ -476,42 +477,46 @@ Field Pattern | Type | Description
...
-      "in": "formData",
-      "description": "Updated status of the pet",
-      "required": false,
-      "type": "string"
     }
   ],
+  "requestBody" : {
+    "representations" : {
+      "application/x-www-form-urlencoded":{
+        "schema" : {
```
dilipkrish: missing "type" : "object"

versions/3.0.md
darrelmiller: This is my current proposal for mapping form-style key/value pairs into JSON schema.

darrelmiller:
@whitlockjc I would represent `void` responses by not including a `representations` object.

Issue #737, opened 2016-07-12
Asyncronous API callbacks


RobPhippen:
> This issue description is **work in progress** This is one of three items agreed to be split out from #716 (see [this note](https://github.com/OAI/OpenAPI-Specification/issues/716#issuecomment-232086961)).
## Summary

This issue covers a syntax to enable asynchronous callbacks to be specified.

fehguy:
Some feedback on this.

Conceptually:
I don't know that this makes sense.  To me, there are differences between `async` responses and `callbacks`.  An async response will be sent, in-band, after a connection is negotiated with the server.  An example would be a websocket or an HTTP-long poll.  The "trigger" for the response is typically very specific to HTTP connection types, such as interrupting the connection stream, negotiating a 101, etc.

On the other hand, a `callback` is more of what a webhook would perform--that is, posting directly to a URL, as specified by the caller.  While it is async, the connection is broken between negotiation and processing.

This proposal doesn't really cover either of those two.  If we're looking to define an async response, we'll have to make it clear how the connection is negotiated and what the trigger would be, if any (it could simply be a socket, but that's outside raw HTTP 1.1).  If we're looking to define a callback, we need to make the callback URL part of the request.  I guess that's what #735 is for, but it doesn't seem like they're related correctly

Structurally:
* On the `x-async-responses`:  A formal support means no extension (`x-`).  I suggest `asyncResponses`
* Use a hash instead of array for items in the `asyncResponses`
* The value for the items in the `asyncResponses` is essentially an operation, it should simply adhere to that schema
* `discriminator` doesn't make sense to me.  If the payload is what varies, then we should use the standard payload mechanism for determining what type of object is being sent back

darrelmiller:
There is also other precedents for the use of the term async in HTTP.  The `prefer` header uses the [respond-async](https://tools.ietf.org/html/rfc7240#section-4.1) to describe an interaction that returns a 202 and the client must make a second request to get the results.
It would probably be wise not to introduce more ambiguity in what async means for HTTP.

fehguy:
Closing as https://github.com/OAI/OpenAPI-Specification/pull/763 has been merged.

Issue #736, opened 2016-07-12
WebHook Event Discrimination


RobPhippen:
## Summary

This issue proposes a mechanism to allow different logical events that occur on a single WebHooks callback API to be discriminated.
## Relationship to other issues

This item is one of three split out from #716 as agreed at the OAI TDC on June 17th 2016.

The three items are;
- #735 , which covers webhook event subscription, but which is restricted to the case of modelling a single logical event stream for each physical callback API registered
- #736 (this item), which can be considered to extend #735 to allow multiple logical event streams to be 
- #737 , which covers asynchronous callbacks
## Objective of the proposal

This Issue extends #735 to allow discrimination between different logical event streams on the same physical callback URL. The rest of this document assumed that #735 has been read.
### `eventDiscriminator`

This element extends the definition of `x-event-streams` to allow multiple logical event streams on the same physical `callback URL`. It extends the definition of 'parameter`within`x-event-streams` to add this element. It is only required where the 'discriminator' is contained within a header element. If the callback response body can be modelled via the existing swagger schema discriminators, then this can be used instead.
- `eventDiscriminator:`
  - `Parameter:` `parameterName`
  - This introduced to cope with cases where the webhooks callback URL can receive more than one type of event and the event type is conveyed in the header or query perameters. In these cases, the polymorphic Swagger `discriminator` cannot be used. The `Parameter` field specifies which parameter is used to discriminate between different events that may occur.
  - If `eventDiscriminator` is present, then the parameter named has a value equal to the `event-id` when that event occurs
  - If `eventDiscriminator` is absent, then event types are discriminated using the Swagger discriminator, as specified for the schema of the body parameter
### Discriminator in a header: Swagger YAML snippet

We have chosen to illustrate example using github because it has a subcription API.

Please refer to the Github documentation on ['create a hook'](https://developer.github.com/v3/repos/hooks/#create-a-hook)
for further information on the API definition provided by Github for webhooks.

**Please note** that all elements beginning with `x-github` or `x-hub` are part of the Github API
specification, and _do not form part of this proposal_.
Finally, the Swagger below is not comprehensive - for example, not all objects are fully modeled.

``` yaml
paths:
  /repos/owner/repo/hooks:
    post:
      operationId: 'CreateHook'
      summary: Create a webhook in GITHUB
      description: |
        Call this API to create a new WebHook in Github
      parameters:
        - name: subscription
          in: body
          description: Details to subscribe for a callback on a given set of events
          required: true
[...]

Issue #735, opened 2016-07-12
WebHook Event Subscription


RobPhippen:
## Summary

This issue proposes a mechanism to allow an event subscription API to be associated with the stream of events that may occur as a 'side effect' of the subscription. 

This is one of three items agreed to be split out from #716 (see [this note](https://github.com/OAI/OpenAPI-Specification/issues/716#issuecomment-232086961)).
## Objective of the proposal
- **Subcription/Publication association**: allow a programmer to discover the events that can occur once they have registered
## Background

Many SaaS systems provide some kind of event interface as core to their behaviour.
There are many approaches to implementing this. No single approach is as prevalent as JSON/HTTP APIs have become for service implementation however webhooks is significtly more common than any other single approach.
## WebHooks

Webhooks are a convention of use of normal Web interfaces, typically using JSON/HTTP,
in which an `event subscriber` can register a `callback URL` with an `event publisher`.
When an event occurs, the `event publisher` invokes an HTTP method on that URL (Typically `POST`) to send the information related to the event.
### WebHooks in real life

Webhooks are not standardized nor formally specified, but are in common use in a substiantial
number of systems.

The lack of standardisation means that there is variation in each of the following
areas;
- Whether or not a `subscription API` is made available - many systems instead require
  registration via an administrative UI
- If a `subscription API` is available, its precise form varies between different systems
- Some systems effectively require a different `callback URL` to be registered for each event stream,
  while others support subscription to multiple event streams with a single `callback URL`

The following conventions are commonly observed;
- When an event occurs, where the publisher supports subscription to multiple different event streams on the same URL
  it is common for them to provide a parameter, or a data field within a parameter, as part of the
  POSTed message that specifies which event has occurred.
- When an event occurs, almost all systems support the use of HTTP POST to convey the event information.
  Some _mandate_ the use of HTTP POST, while others allow other HTTP methods to be selected as an option
### Scope of this proposal

This proposal covers the particular case where 
- a webhooks `subscription API` is offered that allows a `callback URL` to be registered, and which allows the event of interest to be specified
- a single subscription results in a single logical event stream being delivered to the registered callback URL

The proposal does _not_ specify a standardised form for the subscription API. Instead, it enables a subscription API to be identified via the presence of the `x-event-streams` element, which models the event stream that occurs as a result of the subscription operation.
### Related Proposals

An additional proposal, #736, builds on this proposal to add a mechanism to discriminate between multiple logical event streams on the same physical callback address.
[...]

Issue #716, opened 2016-06-16
Proposal for Swagger Async Responses and WebHooks


RobPhippen:
# Swagger Asynchronous Reponses and WebHooks
## Background

This document provides an outline of the proposed swagger format for documenting a webhook subscription and callback interface.
### Background

Many SaaS systems provide some kind of event interface as core to their behaviour.
There are many approaches to implementing this. No single approach is as prevalent as JSON/HTTP APIs have become for service implementation however webhooks is significtly more common than any other single approach.

This document concentrates on extending Swagger to cope with the WebHooks pattern.
## WebHooks

Webhooks are a convention of use of normal Web interfaces, typically using JSON/HTTP,
in which an `event subscriber` can register a `callback URL` with an `event publisher`.
When an event occurs,
the `event publisher` invokes an HTTP method on that URL (Typically `POST`) to send the information related to the event.
### WebHooks in real life

Webhooks are not standardized nor formally specified, but are in common use in a substiantial
number of systems.

The lack of standardisation means that there is variation in each of the following
areas;
- Whether or not a `subscription API` is made available - many systems instead require
  registration via an administrative UI
- If a `subscription API` is available, its precise form varies between different systems
- Some systems effectively require a different `callback URL` to be registered for each event stream,
  while others support subscription to multiple event streams with a single `callback URL`

The following conventions are commonly observed;
- When an event occurs, where the publisher supports subscription to multiple different event streams on the same URL
  it is common for them to provide a parameter, or a data field within a parameter, as part of the
  POSTed message that specifies which event has occurred.
- When an event occurs, almost all systems support the use of HTTP POST to convey the event information.
  Some _mandate_ the use of HTTP POST, while others allow other HTTP methods to be selected as an option
## Objective of the proposal

This objective of this proposal is to allow the `subscription` and `event publication` interactions
involved in the WebHooks convention to be minimally modeled via Swagger, such that a Swagger-enabled SDK could;
- **Understand event structures**: Provide an abstraction of each event that could occur once subcription has occurred, reducing the amount of code that a programmer has to author to process the events.
- **Discriminate between events**: discriminate between events of different types that may occur so that the appropriate program logic can be invoked
- **Subcription/Publication association**: (by allowing the event set+discriminator to be associated with a subscription API)
  allow a programmer to discover the events that can occur once they have registered
- **Documentation**: extend the documentation benefits of swagger to events.
## Not covered by the proposal

[...]

PR #763, opened 2016-08-19
Callback feature

fehguy:
Addresses #716, #735, #736, #737.

To support webhooks, a proposed `callback` mechanism is added.  Essentially, callbacks are out-of-band `operations` against some remote server.  The callback subscription mechanism is not prescribed, as it is actually implementation dependent.  The mechanism for responses, however, is created by added a `callbacks` section to any operation.

The `callbacks` section has named callbacks, which are essentially operations.  The operations can be applied to a URL as extracted from the `request` options using a syntax similar to with the `links` technique.

An example below describes the GitHub webhook mechanism, with comments, for discussion.

```
paths:
  /repos/{owner}/{repo}/hooks/{id}:
    post:
      body: |
        description: Repository service hooks (like email or Campfire) can have at most one configured at a time. Creating hooks for a service that already has one configured will update the existing hook.
        Repositories can have multiple webhooks installed. Each webhook should have a unique config. Multiple webhooks can share the same config as long as those webhooks do not have any events that overlap.
        schema:
          $ref: '#/components/definitions/WebhookRequest'
      responses:
        201:
          description: Webhook created
          schema:
            $ref: '#/components/definitions/WebhookResponse'
      callbacks:
        createWebhook:
          $ref: '#/components/callbacks/createWebhook'
components:
  callbacks:
    createWebhook:
      # extract the url parameter from the request
      '$request.body.url':
        post:
          parameters:
          - in: header
            name: X-GitHub-Event
            type: string
            description: Name of the [event](https://developer.github.com/webhooks/#events) that triggered this delivery.
            required: true
          - in: header
            name: X-Hub-Signature
            type: string
            description: HMAC hex digest of the payload, using the [the hook's `secret`](https://developer.github.com/v3/repos/hooks/#create-a-hook) as the key (if configured).
          - in: header
            name: X-GitHub-Delivery
            type: string
            description: Unique ID for this delivery.
          body:
            name: payload
            schema:
              $ref: '#/components/definitions/Event'
          responses:
            200:
              description: webhook successfully processed an no retries will be performed
  definitions:
    WebhookRequest:
      required:
      - name
      - config
      properties:
        name:
          type: string
          description: Use web for a webhook or use the name of a valid service. (See /hooks for the list of valid service names.)
        config:
          $ref: '#/components/definitions/WebhookConfig'
        events:
          type: array
          items:
            type: string
            enum:
            - push
          default: [push]
          description: Determines what events the hook is triggered for.
        active:
          type: boolean
          description: Determines whether the hook is actually triggered on pushes.
    WebhookResponse:
      properties:
        id:
          type: string
        url:
          type: string
          format: url
        test_url:
          type: string
          format: url
        ping_url:
          type: string
          format: url
        name:
          type: name
        events:
          type: array
          items:
            type: string
            enum:
            - push
            - pull_request
        active:
          type: boolean
        config:
          $ref: '#/components/definitions/WebhookConfig'
        updated_at:
          type: string
          format: date-time
        created_at:
          type: string
          format: date-time
    WebhookConfig:
      type: object
        properties:
          url:
            type: string
            format: url
          content_type:
            type: string
            enum:
            - json
        additionalProperties:
          type: string
          description: Key/value pairs to provide settings for this hook. These settings vary between hooks and some are defined in the [github-services](https://github.com/github/github-services) repository. Booleans are stored internally as "1" for true, and "0" for false. Any JSON true/false values will be converted automatically.
    Event:
      anyOf:
      - $ref: '#/components/definitions/commit_comment'
      - $ref: '#/components/definitions/create'
      - $ref: '#/components/definitions/delete'
      - $ref: '#/components/definitions/deployment'
      - $ref: '#/components/definitions/deployment_status'
      - $ref: '#/components/definitions/fork'
      - $ref: '#/components/definitions/gollum'
      - $ref: '#/components/definitions/issue_comment'
      - $ref: '#/components/definitions/issues'
      - $ref: '#/components/definitions/member'
      - $ref: '#/components/definitions/membership'
      - $ref: '#/components/definitions/page_build'
      - $ref: '#/components/definitions/public'
      - $ref: '#/components/definitions/pull_request_review_comment'
      - $ref: '#/components/definitions/pull_request'
      - $ref: '#/components/definitions/push'
      - $ref: '#/components/definitions/repository'
      - $ref: '#/components/definitions/release'
      - $ref: '#/components/definitions/status'
      - $ref: '#/components/definitions/team_add'
      - $ref: '#/components/definitions/watch'
```

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -453,6 +453,7 @@ Field Name | Type | Description
 <a name="operationParameters"></a>parameters | [[Parameter Object](#parameterObject) <span>&#124;</span> [Reference Object](#referenceObject)] | A list of parameters that are applicable for this operation. If a parameter is already defined at the [Path Item](#pathItemParameters), the new definition will override it, but can never remove it. The list MUST NOT include duplicated parameters. A unique parameter is defined by a combination of a [name](#parameterName) and [location](#parameterIn). The list can use the [Reference Object](#referenceObject) to link to parameters that are defined at the [OpenAPI Object's parameters](#oasParameters).
 <a name="operationRequestBody"></a>requestBody | [[Request Body Object](#requestBodyObject) <span>&#124;</span> [Reference Object](#referenceObject)] | The request body applicable for this operation. 
 <a name="operationResponses"></a>responses | [Responses Object](#responsesObject) | **Required.** The list of possible responses as they are returned from executing this operation.
+<a name="operationCallbacks"></a>callback responses | [Callback Responses Object](#callbackObject) | The list of possible callback responses as they are returned from executing this operation.
 <a name="operationSchemes"></a>schemes | [`string`] | The transfer protocol for the operation. Values MUST be from the list: `"http"`, `"https"`, `"ws"`, `"wss"`. The value overrides the OpenAPI Object [`schemes`](#oasSchemes) definition. 
 <a name="operationDeprecated"></a>deprecated | `boolean` | Declares this operation to be deprecated. Usage of the declared operation should be refrained. Default value is `false`.
 <a name="operationSecurity"></a>security | [[Security Requirement Object](#securityRequirementObject)] | A declaration of which security schemes are applied for this operation. The list of values describes alternative security schemes that can be used (that is, there is a logical OR between the security requirements). This definition overrides any declared top-level [`security`](#oasSecurity). To remove a top-level security declaration, an empty array can be used.
@@ -1049,6 +1050,36 @@ Response with no return value:
 ```yaml
 description: object created
 ```
+
+#### <a name="callbackObject"></a>Callback Object
+
+A container for possible out-of band callbacks from an operation. A callback may be returned from an operation, calling back to the path specified in the operation object.
+
+##### Patterned Fields
+Field Pattern | Type | Description
+---|:---:|---
+<a name="responseName"></a>Callback name | [Callback Operation Object](#operationObject) <span>&#124;</span> [Operation Object](#operationObject) | An operation object used to define a callback payload structure
+<a name="responsesExtensions"></a>^x- | Any | Allows extensions to the OpenAPI Schema. The field name MUST begin with `x-`, for example, `x-internal-id`. The value can be `null`, a primitive, an array or an object. See [Specification Extensions](#specificationExtensions) for further details.
+
+
+##### Callback Object Example
+
+A callback to the URL specified by the `url` parameter in the request
+
+
+```yaml
+myWebhook:
+  '$request.url':
+    post:
+      body:
+        name: postResponse
+        schema:
+          $ref: '#/definitions/SomePayload'
+      responses:
+        200:
+          description: webhook successfully processed an no retries will be performed
+```
+
 
 #### <a name="headersObject"></a>Headers Object
 Lists the headers that can be sent as part of a response.


fehguy:
@OAI/tdc for discussion today ^^

versions/3.0.md
```diff
@@ -1050,6 +1051,36 @@ Response with no return value:
 description: object created
 ```
 
+#### <a name="callbackObject"></a>Callback Object
+
+A container for possible out-of band callbacks from an operation. A callback may be returned from an operation, calling back to the path specified in the operation object.
```
ePaul: »A callback may be returned from an operation« sounds wrong. Maybe better »A callback may be caused/triggered by an operation«?

ePaul:
This `$request.body.url` syntax doesn't seem to be defined. Something similar is there for the variable substitution in the link object, but there it is in `{...}`. I suggest unifying this (and defining it a bit clearer).

versions/3.0.md
```diff
@@ -1050,6 +1051,36 @@ Response with no return value:
 description: object created
 ```
 
+#### <a name="callbackObject"></a>Callback Object
```
ePaul: This is called here »Callback Object«, but »Callback Responses Object« in the Operation object.

versions/3.0.md
```diff
@@ -1050,6 +1051,36 @@ Response with no return value:
...
+
+A container for possible out-of band callbacks from an operation. A callback may be returned from an operation, calling back to the path specified in the operation object.
+
+##### Patterned Fields
+Field Pattern | Type | Description
+---|:---:|---
+<a name="responseName"></a>Callback name | [Callback Operation Object](#operationObject) <span>&#124;</span> [Operation Object](#operationObject) | An operation object used to define a callback payload structure
```
ePaul: Both links here go to the same target (the operation object), which doesn't match the example (which looks more like the top level Paths object, with some variable substitution mixed in in the path).

DavidBiesack:
in the example, I think `X-Hub-Signature` should be `X-GitHub-Signature`

DavidBiesack:
The proposal should address security considerations -- can a man in the middle or other attack alter the callback URL in the request in order to redirect sensitive data?

fehguy:
@DavidBiesack I'm not sure how you think this proposal introduces any security issues?  Please explain.

fehguy:
@OAI/tdc please review and add your feedback.  Would be great to get any other examples or feedback on this structure.

darrelmiller:
I'm all 👍 on this.

RobPhippen:
A few comments;
* Although it's 'just a name' (and so could be any valid string, I assume) the name `createWebHook` as the identifier/name for the callback seemed odd to me. I wanted to log this in case there is a deeper reason for the name to be what it is
  * I might have used the term `eventCallback` or `gitHubEventHook`
* Setting aside the syntax used in `'$request.body.url':`, its *placement* in the structure of the document leads me to think that the entire callback is part of a structure that hangs off this item. 
  * If this interpretation is correct, I think this means that this update does not model 'standalone' webhooks declarations. 
  * In the last OAI TDC discussion (that I was on) we spent some time discussing that it is useful and necessary to enable this. As I recall, the proposal was to allow callback declarations to exist within the components section without needing to be attached to a subscription API
  * In your [comment on #716] you had syntax like this

```
callbacks:
      # callbacks are sent to the url provided
      # this could also be a query param, etc.
      url: '$request.url'
      schema:
        type: object
        properties:
          a:
            type: string
```
... which could allow WebHooks to be specified standalone by allowing `url:` to be optional.

* Discriminating between events: the `anyOf:` does a nice job of enumerating all of the different event structures that could occur. Can we also provide a way of specifying that `X-GitHub-Event` provides a discriminator?
* On the syntax of `'$request.body.url':`, if I've understood the semantics, and checking the schemas, I think this should be `'$request.body.config.url:'`

RobPhippen:
Hi @DavidBiesack github does in fact use `x-hub-signature` for this specific item;
https://developer.github.com/webhooks/securing/

RobPhippen:
One more thought; could we add an element something like;
``` yaml
   subscriptionList: '$request.body.config.events'
```
Which uses your syntax to make it clear which element of the webhooks subscription interaction specifies the list of events required.

whitlockjc:
This implementation doesn't seem anything like what I'm see elsewhere.  I think calling things what they are in the spec, `webhooks` or `events`, make more sense than `callbacks`.  To me, I'd expect to define a set of `events` and then dictate which operations emitted those `events`.  Each `event` would describe the payload sent via the webhook and the expected response to return.

Here is an example of GitHub's Webhook documentation: https://developer.github.com/webhooks/  I realize this feedback might be a little late.

mhailstone:
Here is a portion of our Event Hub definition updated with this callback proposal:

PR #796, opened 2016-09-29
Hyphenate adjective?

it33:
Sorry, total nit, if "Swagger-specific" is a term describing "annotations", maybe it's easier to read with a hyphen? 

I don't think what's there is wrong, I was just scanning this document and got tripped up over "The Swagger specific annotations...", so if I tripped, maybe others will too. 

Just a thought. Overall, love the project!


it33:
@fehguy thanks! Wow, you are fast! :)

fehguy:
except when i'm not :)

PR #798, opened 2016-09-29
updated #761 for naming consistency

### LINKED ISSUE #721: Making representations a first class concept to address a number of structural concerns

fehguy:
This is a follow-up PR for #761, which fixes #721.  The `representations` element has been added, along with support for multiple examples throughout the `schema` object.  In addition, the request body examples from #670 has been added to the PR.

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -58,7 +58,7 @@ YAML, being a superset of JSON, can be used as well to represent a OAS (OpenAPI 
 
 For example, if a field is said to have an array value, the JSON array representation will be used:
 
-```js
+```json
 {
    "field" : [...]
 }
@@ -128,8 +128,6 @@ Field Name | Type | Description
 <a name="oasVersion"></a>openapi | `string` | **Required.** Specifies the OpenAPI Specification version being used. It can be used by tooling Specifications and clients to interpret the version. The structure shall be `major`.`minor`.`patch`, where `patch` versions _must_ be compatible with the existing `major`.`minor` tooling. Typically patch versions will be introduced to address errors in the documentation, and tooling should typically be compatible with the corresponding `major`.`minor` (3.0.*). Patch versions will correspond to patches of this document.
 <a name="oasInfo"></a>info | [Info Object](#infoObject) | **Required.** Provides metadata about the API. The metadata can be used by the clients if needed.
 <a name="oasHosts"></a>hosts | [Hosts Object](#hostsObject) | An array of Host objects which provide `scheme`, `host`, `port`, and `basePath` in an associative manner.
-<a name="oasConsumes"></a>consumes | [`string`] | A list of media types the APIs can consume. This is global to all APIs but can be overridden on specific API calls. Value MUST be as described under [Media Types](#mediaTypes).
-<a name="oasProduces"></a>produces | [`string`] | A list of media types the APIs can produce. This is global to all APIs but can be overridden on specific API calls. Value MUST be as described under [Media Types](#mediaTypes).
 <a name="oasResponses"></a>responses | [Responses]
 <a name="oasPaths"></a>paths | [Paths Object](#pathsObject) | **Required.** The available paths and operations for the API.
 <a name="oasSchemas"></a>components | [Components Object](#componentsObject) | An element to hold various schemas for the specification.
@@ -167,7 +165,7 @@ Field Pattern | Type | Description
 
 ##### Info Object Example:
 
-```js
+```json
 {
   "title": "Swagger Sample App",
   "description": "This is a sample server Petstore server.",
@@ -231,7 +229,7 @@ Field Pattern | Type | Description
 
 ##### Contact Object Example:
 
-```js
+```json
 {
   "name": "API Support",
   "url": "http://www.swagger.io/support",
@@ -264,7 +262,7 @@ Field Pattern | Type | Description
 
 ##### License Object Example:
 
-```js
+```json
 {
   "name": "Apache 2.0",
   "url": "http://www.apache.org/licenses/LICENSE-2.0.html"
@@ -307,21 +305,22 @@ Field Pattern | Type | Description
 
 ##### Paths Object Example
 
-```js
+```json
 {
   "/pets": {
     "get": {
       "description": "Returns all pets from the system that the user has access to",
-      "produces": [
-        "application/json"
-      ],
       "responses": {
         "200": {
           "description": "A list of pets.",
-          "schema": {
-            "type": "array",
-            "items": {
-              "$ref": "#/definitions/pet"
+          "representations": {
+            "application/json": {
+              "schema": {
+                "type": "array",
+                "items": {
+                  "$ref": "#/definitions/pet"
+                }
+              }
             }
           }
         }
@@ -335,15 +334,15 @@ Field Pattern | Type | Description
 /pets:
   get:
     description: Returns all pets from the system that the user has access to
-    produces:
-    - application/json
     responses:
       '200':
         description: A list of pets.
-        schema:
-          type: array
-          items:
-            $ref: '#/definitions/pet'
+        representations:
+          application/json:
+            schema:
+              type: array
+              items:
+                $ref: '#/definitions/pet'
 ```
 
 #### <a name="pathItemObject"></a>Path Item Object
@@ -379,30 +378,34 @@ Field Pattern | Type | Description
 
 ##### Path Item Object Example
 
-```js
+```json
 {
   "get": {
     "description": "Returns pets based on ID",
     "summary": "Find pets by ID",
     "operationId": "getPetsById",
-    "produces": [
-      "application/json",
-      "text/html"
-    ],
     "responses": {
       "200": {
         "description": "pet response",
-        "schema": {
-          "type": "array",
-          "items": {
-            "$ref": "#/definitions/Pet"
+        "representations": {
+          "*": {
+            "schema": {
+              "type": "array",
+              "items": {
+                "$ref": "#/definitions/Pet"
+              }
+            }
           }
         }
       },
       "default": {
         "description": "error payload",
-        "schema": {
-          "$ref": "#/definitions/ErrorModel"
+        "representations": {
+          "*": {
+            "schema": {
+              "$ref": "#/definitions/ErrorModel"
+            }
+          }
         }
       }
     }
@@ -428,20 +431,21 @@ get:
   description: Returns pets based on ID
   summary: Find pets by ID
   operationId: getPetsById
-  produces:
-  - application/json
-  - text/html
   responses:
     '200':
       description: pet response
-      schema:
-        type: array
-        items:
-          $ref: '#/definitions/Pet'
+      representations:
+        application/json:
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/Pet'
     default:
       description: error payload
-      schema:
-        $ref: '#/definitions/ErrorModel'
+      representations:
+        "*": 
+          schema:
+            $ref: '#/definitions/ErrorModel'
 parameters:
 - name: id
   in: path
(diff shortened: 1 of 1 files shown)


fehguy:
@OAI/tdc please review

versions/3.0.md
```diff
@@ -128,8 +128,6 @@ Field Name | Type | Description
 <a name="oasVersion"></a>openapi | `string` | **Required.** Specifies the OpenAPI Specification version being used. It can be used by tooling Specifications and clients to interpret the version. The structure shall be `major`.`minor`.`patch`, where `patch` versions _must_ be compatible with the existing `major`.`minor` tooling. Typically patch versions will be introduced to address errors in the documentation, and tooling should typically be compatible with the corresponding `major`.`minor` (3.0.*). Patch versions will correspond to patches of this document.
 <a name="oasInfo"></a>info | [Info Object](#infoObject) | **Required.** Provides metadata about the API. The metadata can be used by the clients if needed.
 <a name="oasHosts"></a>hosts | [Hosts Object](#hostsObject) | An array of Host objects which provide `scheme`, `host`, `port`, and `basePath` in an associative manner.
-<a name="oasConsumes"></a>consumes | [`string`] | A list of media types the APIs can consume. This is global to all APIs but can be overridden on specific API calls. Value MUST be as described under [Media Types](#mediaTypes).
```
DavidBiesack: For incompatible changes to OAS from 2.0 to 3.0, we need a guide (separate from the spec) on how to convert (or interpret). I.e. if `consumes`/`produces` are removed here, how does one transform a 2.0 document to 3.0 and preserve the same API definition? This also applies to removing  `consumes`/`produces` on an operation. It's only slightly clearer there, since the `representations`  exist, but there are still some ambiguities: if produces lists more than one media type, which one is mapped to the 2xx response bodies, and what happens to the others?  Example 2: how are the top level  `consumes`/`produces` merged with the operation  `consumes`/`produces` to yield the same effective `representations` in request and response bodies?

versions/3.0.md
```diff
@@ -761,67 +751,83 @@ schema:
 
 Describes a single request body.
 
-The `Content-Type` of the request body must be specified by the `consumes` attribute, either at the [top-level](#oasConsumes) or [operation level](#operationConsumes).
```
DavidBiesack: Rather than delete this, instead indicate that the set of allowed Content-Type values is derived from the media types in the `representations` object in the Request Body Object.

versions/3.0.md
```diff
@@ -466,8 +470,6 @@ Field Name | Type | Description
 <a name="operationDescription"></a>description | `string` | A verbose explanation of the operation behavior. [CommonMark syntax](http://spec.commonmark.org/) can be used for rich text representation.
 <a name="operationExternalDocs"></a>externalDocs | [External Documentation Object](#externalDocumentationObject) | Additional external documentation for this operation.
 <a name="operationId"></a>operationId | `string` | Unique string used to identify the operation. The id MUST be unique among all operations described in the API. Tools and libraries MAY use the operationId to uniquely identify an operation, therefore, it is recommended to follow common programming naming conventions.
-<a name="operationConsumes"></a>consumes | [`string`] | A list of media types the operation can consume. This overrides the [`consumes`](#oasConsumes) definition at the OpenAPI Object. An empty value MAY be used to clear the global definition. Value MUST be as described under [Media Types](#mediaTypes).
```
DavidBiesack: I thinks the spec needs a **Representations Object** section to define `representations`.

versions/3.0.md
```diff
@@ -969,7 +989,6 @@ Describes a single response from an API Operation, including design-time, static
 Field Name | Type | Description
 ---|:---:|---
 <a name="responseDescription"></a>description | `string` | **Required.** A short description of the response. [CommonMark syntax](http://spec.commonmark.org/) can be used for rich text representation.
-<a name="responseSchema"></a>schema | [Schema Object](#schemaObject) | A definition of the response structure. It can be a primitive, an array or an object. If this field does not exist, it means no content is returned as part of the response. As an extension to the [Schema Object](#schemaObject), its root `type` value may also be `"file"`. This SHOULD be accompanied by a relevant `produces` media type.
 <a name="responseHeaders"></a>headers | [Headers Object](#headersObject) | A list of headers that are sent with the response.
 <a name="responseExamples"></a>examples | [Example Object](#exampleObject) | An example of the response message.
```
DavidBiesack: This PR does not fix the mention of `produces` (which no longer exists) in [Example Object](https://github.com/OAI/OpenAPI-Specification/blob/650000a84424b3a3c7ae8f375e0774210299daa3/versions/3.0.md#example-object) and [Examples Object](https://github.com/OAI/OpenAPI-Specification/blob/650000a84424b3a3c7ae8f375e0774210299daa3/versions/3.0.md#examples-object)

darrelmiller:
@fehguy I'm confused.  I had made these changes in #761 but in my last update I had renamed `representations` to `content`.  Did my changes get merged into yours, or did you have to redo what I had done?

versions/3.0.md
```diff
@@ -979,56 +998,75 @@ Field Name | Type | Description
 
 Field Pattern | Type | Description
 ---|:---:|---
+<a name="representations"></a>`*` | [Schema Object](#schemaObject)<span>&#124;</span> [Reference Object](#referenceObject) | A schema describing the response value for a specific `Content-Type`
 <a name="responseExtensions"></a>^x- | Any | Allows extensions to the OpenAPI Schema. The field name MUST begin with `x-`, for example, `x-internal-id`. The value can be `null`, a primitive, an array or an object. See [Specification Extensions](#specificationExtensions) for further details.
 
+Representations may take the form of a wildcard (`*`) to designate any `Content-Type`, or a regular expression for matching a specific type
```
DavidBiesack: The statement is unclear, especially since "representations" is not defined in this spec (in [Response Object](https://github.com/OAI/OpenAPI-Specification/blob/650000a84424b3a3c7ae8f375e0774210299daa3/versions/3.0.md#response-object) ).  To be more precise, use a Representations Object definition in the spec:

   representations  |   [ Representations Object ]   | specify a mapping between Content-Media types and their representation definitions

Please include an example with a regular expression

versions/3.0.md
```diff
@@ -761,67 +751,83 @@ schema:
...
 A body parameter that is an array of string values:
-```js
+```json
 {
   "description": "user to add to the system",
-  "schema": {
-    "type": "array",
-    "items": {
-      "type": "string"
+  "representations": {
```
DavidBiesack: The use of a JSON array schema with `text/plain` is confusing. If it is an array of strings, it should be application/json. if it is a text (i.e. one user per line) then it is not an array. (Repeated in the YAML example.)

versions/3.0.md
```diff
@@ -761,67 +751,83 @@ schema:
...
+    schema:
+      $ref: '#/definitions/User'
+    example:
+      $ref: 'http://foo.bar/examples/user-example.xml'
+  '*':
+    schema:
+      $ref: 'http://foo.bar/examples/user-example.whatever'
 ```
 
 A body parameter that is an array of string values:
```
DavidBiesack: This is no longer a "body parameter" but a "request body"

fehguy:
Oh weird, I looked at your commit but I see the word `content` instead of `representations` throughout it.  Maybe you didn't push your update?  Happy to use yours :)

versions/3.0.md
```diff
@@ -926,26 +934,34 @@ It can be used to cover undeclared responses.
...
+<a name="responsesCode"></a>[HTTP Status Code](#httpCodes) | [Response Object](#responseObject) <span>&#124;</span> [Reference Object](#referenceObject) | Any [HTTP status code](#httpCodes) can be used as the property name (one property per HTTP status code). Describes the expected response for that HTTP status code.  [Reference Object](#referenceObject) can be used to link to a response that is defined at the [OpenAPI Object's responses](#oasResponses) section. This field must quoted for compatibility between JSON and YAML (i.e. "200"), and may contain the uppercase character, `X` to designate a wildcard, such as `2XX` to represent all response codes between `[200-299]`.
```
DavidBiesack: I know the `XX` wildcard was added previously, but are there (or should there be) constraints on the use of `X`? I.e. certainly `2XX` and `4XX` make sense; but do `2X0` or `XX0` make sense? Should we limit this to just `1XX, 2XX. 3XX, 4XX, 5XX`?

versions/3.0.md
```diff
@@ -1649,19 +1702,15 @@ definitions:
       example:
         $ref: http://foo.bar#/examples/name-example
 
-# in a request body, note the plural `examples` as the content-type is set by `consumes`:
+# in a request body, note the plural `examples` as the content-type is set to `*`:
```
DavidBiesack: Use `Content-Type` not `content-type`

DavidBiesack: I'm unclear on "is set to `*`" as the example uses `application/json`. Also, this does not _set_ the Content-Type, it only _matches_ it.

versions/3.0.md
darrelmiller: Yes, this is going to be a bit of a challenge.  I think the only safe option is to project the mediatypes defined in produces and consumes down into the requestBody and responseObjects and create representation/content objects for each media-type.  This could cause some unfortunate bloat in some specs.  It is my understanding that operation level 'produces/consumes' overrides higher level declarations.

versions/3.0.md
darrelmiller: There was some confusion regarding us calling this `representations` vs `content` object and I have many similar updates in PR #721 .  I'm going to merge the two PRs as the "idea" has been agreed upon and then we can raise issues against the wording in the spec.  /cc @fehguy

versions/3.0.md
darrelmiller: I'm fixing that.

darrelmiller: Being fixed.

versions/3.0.md
fehguy: @DavidBiesack the docs + migration information will come later.  It doesn't make sense to do it right now until we're done with all structural changes.

DavidBiesack:
already merged -- is there another PR for some of the changes being worked on? (sorry if I'm not allowing time to actually create the PR!)

Can `representations` (or the `content` objects if that is the correct name?) be defined in `components` and reused with `$ref`? I don't see any examples that show this such as `$ref: '#/components/representations/pet'` `$ref: '#/components/content/pet'`

FWIW, I prefer `representations` over `content`; that is literally half of **REST** after all :smirk: . But I admit it is unfortunatly 114% longer

darrelmiller:
@DavidBiesack I merged both Tony's PR and my original PR in.  We have the Callbacks PR to merge in and I need to create a PR for the changes to support URI Template stuff and because everything is in one file we keep causing ourselves merge conflicts.  So, at this point it's probably best to raise issues with links to the main document and we'll tag them with the '3.0 needs documentation' tag.

We can reuse responses, so I'm not sure there is a need for reusing content objects.  But I'm open to being convinced otherwise.

Initially I preferred representation, but once I realized that I could describe it as a content object that contains "Content  Type" descriptions, and that ties directly to the HTTP header, it felt more natural.  Most people are not familiar with the "representations" terminology.

PR #807, opened 2016-10-03
Added bearer/JWT support and OpenIdConnect

darrelmiller:


darrelmiller:
@OAI/tdc Thoughts?

versions/3.0.md
```diff
@@ -2974,10 +2974,13 @@ Supported schemes are basic authentication, an API key (either as a header or as
 ##### Fixed Fields
 Field Name | Type | Validity | Description
 ---|:---:|---|---
-<a name="securitySchemeType"></a>type | `string` | Any | **Required.** The type of the security scheme. Valid values are `"basic"`, `"apiKey"` or `"oauth2"`.
+<a name="securitySchemeType"></a>type | `string` | Any | **Required.** The type of the security scheme. Valid values are `"apiKey"`, `"http"`, `"oauth2"`, `"openIdConnect"`.
```
fehguy: I'm going to assume `openIdConnect` you mean http://openid.net/connect/ ?  Is that a generic protocol?  Is it ok to use the name in the spec? 

I'm wondering if there is a more generic term to describe it.

darrelmiller: The license to use/reference OpenID related works seems pretty unrestricted http://openid.net/specs/openid-connect-discovery-1_0.html#Notices 
To be more generic we could have a `IdentityProviderURL`.  OpenID provides both a WebFinger protocol and a .well-known URL for getting at the OpenID configuration.  If other Identity Providers come along they would probably have their own "discovery" mechanism.

versions/3.0.md
```diff
@@ -3019,6 +3024,22 @@ name: api_key
 in: header
 ```
 
+###### JWT Bearer Sample
+
+```json
+{
+  "type": "scheme",
```
fehguy: `type` should be `http`

versions/3.0.md
```diff
@@ -3019,6 +3024,22 @@ name: api_key
...
+  "scheme" : "bearer",
+  "bearerFormat" : "JWT",
```
fehguy: The `bearerFormat` may not be necessary, except to indicate that the token will be prefixed by the term `Bearer` when being passed to the client.  More for documentation than anything.

fehguy:
"Shall we merge?  Cha... Cha... Cha..."

curtisgibby:
Now that this has been merged into the `OpenApi.next` branch, when will that be released as the stable "production" branch? (OpenAPI version 3?) I'm relying on a downstream project that I'm sure won't update until this is released on your side. (Eagerly awaiting JWT support in https://github.com/swagger-api/swagger-ui and https://github.com/zircote/swagger-php .)

webron:
https://www.openapis.org/blog/2017/01/24/a-new-year-a-new-specification

Issue #607, opened 2016-03-21
version 3.0: additional formats


ralfhandl:
As there are several issues proposing new formats, here a list of the possible future full picture

| Common Name | Type | Format | Comments |
| --- | --- | --- | --- |
| octet/(unsigned) byte | `integer` | `uint8` | **new: unsigned 8 bits** |
| signed byte | `integer` | `int8` | **new: signed 8 bits** |
| short | `integer` | `int16` | **new: signed 16 bits** |
| integer | `integer` | `int32` | signed 32 bits |
| long | `integer` | `int64` | signed 64 bits |
| big integer | `integer` |  |  |
| float/single | `number` | `float` |  |
| double | `number` | `double` |  |
| decimal | `number` | `decimal` | **new: decimal floating-point number**, recipient-side internal representation as a binary floating-point number may lead to rounding errors |
| big decimal | `number` |  |  |
| string | `string` |  |  |
| byte | `string` | `byte` | base64 encoded characters |
| url-safe binary | `string` | `base64url` | **new: base64url encoded characters** - #606 |
| binary | `string` | `binary` | any sequence of octets |
| boolean | `boolean` |  |  |
| date | `string` | `date` | As defined by `full-date` - [RFC3339](http://xml2rfc.ietf.org/public/rfc/html/rfc3339.html#anchor14) |
| dateTime | `string` | `date-time` | As defined by `date-time` - [RFC3339](http://xml2rfc.ietf.org/public/rfc/html/rfc3339.html#anchor14) |
| time (of day) | `string` | `time` | **new: As defined by `partial-time` - [RFC3339](http://xml2rfc.ietf.org/public/rfc/html/rfc3339.html#anchor14)** - #358 |
| duration | `string` | `duration` | **new: As defined by `xs:dayTimeDuration` - [XML Schema 1.1](http://www.w3.org/TR/xmlschema11-2/#dayTimeDuration)** - #359 |
| uuid | `string` | `uuid` | **new: Universally Unique Identifier (UUID) [RFC4122](https://www.ietf.org/rfc/rfc4122.txt)** |
| password | `string` | `password` | Used to hint UIs the input needs to be obscured. |

amarzavery:
:+1:

DavidBiesack:
[...]

PR #811, opened 2016-10-06
Additional formats as discussed in #607

ralfhandl:
Additional formats as discussed in #607

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -95,23 +95,33 @@ An additional primitive data type `"file"` is used by the [Parameter Object](#pa
 <a name="dataTypeFormat"></a>Primitives have an optional modifier property `format`.
 OAS uses several known formats to more finely define the data type being used.
 However, the `format` property is an open `string`-valued property, and can have any value to support documentation needs.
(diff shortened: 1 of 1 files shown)


darrelmiller:
I think these additional formats would be valuable additions to a "format registry" that indicates commonly used formats.  I'm not convinced that all of these formats should be defined in the specification.  This would require all tooling to support all of the formats and I suspect many would not be.

ralfhandl:
> I'm not convinced that all of these formats should be defined in the specification. 

Which are the new ones that CAN be defined in the specification?

Given that there's no "format registry" and the Swagger/OpenAPI specification seems to be the only official document that defines formats beyond the rather rudimentary list in http://json-schema.org/latest/json-schema-validation.html#rfc.section.7.3, I'll happily settle for any non-empty subset of the new formats.

webron:
@ralfhandl - thanks for the PR. The @OAI/tdc has discussed the issued and decided that we don't feel the extended format descriptions should be in the spec.

However, there's an understandable need to define such additional formats and we've considered two alternative approaches. These will be handled in #845.

Issue #199, opened 2014-11-11
Support JSON Schema "not"


dschenkelman:
http://json-schema.org/latest/json-schema-validation.html#anchor91

webron:
Parent: #579.

webron:
Tackling PR: #741

webron:
`not` is in 3.0!

Issue #377, opened 2015-05-25
Allow anyOf in specification


paulhill:
We don't necessarily need to support anyOf for tooling. 
I know the code generation authors like the definition of type hierarchy that comes with allOf inheritance. 
But it would be so nice to be able to specify an anyOf array.
Even if swagger spec validation just ignored it that would be an improvement.
It seems so restrictive to disallow certain types of schema validation just because tool authors aren't supporting it.

:-/

paulhill:
A big part of what I liked about swagger 1 was that I could re-used my json-schema validation as my model definitions. That went away with swagger 2.0

webron:
Actually, we necessarily need to support it. 

While strictly speaking, Swagger is the specification, the specification itself means nothing without proper tooling support.

While you may not be looking it, once we put it in, users would start asking for it (and rightfully so). This is true for other features that are not yet fully implemented in the tools (though are easier to implement).

We're trying to think of ways to add support for anyOf in the future though.

coding2012:
Just as a use-case here. We use a Swagger file to define our API so that we get pretty documentation, etc.. This works great. We have the need to use an anyOf - so right now we we provide a "dummied up" schema that is "pretty close" but not correct. Then we document that in the Swagger doc parts. At runtime, we point to a different place inside the Swagger.json file that is using anyOf (because our any of is making references we keep all of the JSON Schema in the same file). Anyway while I realize it will be a long time if ever a code-generation tool could support that, we do our server in JS (and clients too). So just throwing it out there that we use Swagger, and anyOf support would be great, as we have a slightly awkward work-around.

webron:
Tackling PR: #741

fehguy:
This has been merged.

PR #812, opened 2016-10-07
Addressed metadata documentation cleanup from #589

fehguy:
See https://github.com/OAI/OpenAPI-Specification/issues/589#issuecomment-250816162

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -127,7 +127,7 @@ Field Name | Type | Description
 ---|:---:|---
 <a name="oasVersion"></a>openapi | [OpenAPI Version String](#oasVersion) | **Required.** Specifies the OpenAPI Specification version being used. It can be used by tooling Specifications and clients to interpret the version. The structure shall be `major`.`minor`.`patch`, where `patch` versions _must_ be compatible with the existing `major`.`minor` tooling. Typically patch versions will be introduced to address errors in the documentation, and tooling should typically be compatible with the corresponding `major`.`minor` (3.0.*). Patch versions will correspond to patches of this document.
 <a name="oasInfo"></a>info | [Info Object](#infoObject) | **Required.** Provides metadata about the API. The metadata can be used by the clients if needed.
-<a name="oasHosts"></a>hosts | [Host Object](#hostObject) | An array of Host objects which provide connectivity information to a target host.
+<a name="oasHosts"></a>hosts | [Server Object](#serverObject) | An array of Server Objects which provide connectivity information to a target host.
 <a name="oasPaths"></a>paths | [Paths Object](#pathsObject) | **Required.** The available paths and operations for the API.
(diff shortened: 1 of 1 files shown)


fehguy:
Fixes #199, #377

darrelmiller:
In the OpenAPI object the `hosts` key is still present but in the example it is shown as `servers`.  Do we have a preference?

Also the description of the Host Object seems to have got stuck in the middle of the description of the Info Object.

DavidBiesack:
I prefer `servers` over `hosts`. The example shows 
```
servers:
-   url: https://development.gigantic-server.com/v1
```
and I think "server" (as in "API server" / "service") is a better term here than host, as `host` should refer to just the  host name `development.gigantic-server.com`. I don't think we should create a new definition of `host` that is not compatible with the [definition of host in a URL](https://tools.ietf.org/html/rfc3986#section-3.2.2)

fehguy:
@darrelmiller i will fix that.

darrelmiller:
Need to add a new doc that discusses migration guidelines and expectations from tooling vendors in regards to being compatible between minor versions.

versions/3.0.md
```diff
@@ -196,9 +205,73 @@ An object representing a Host.
...
+<a name="hostTemplates"></a>templates | [Templates Object](#hostTemplatesObject) | An object holding templates for substitution in the URL template
+
+This object can be extended with [Specification Extensions](#specificationExtensions). 
+
+##### Host Object Example
+
+The following shows how multiple hosts can be described in the host object array
+
+```yaml
+servers:
```
earth2marsh: alternatives for `servers`: `instances` … ?

fehguy:
the terms `hosts` and `servers` need to be updated accordingly in the operations / paths objects

versions/3.0.md
```diff
@@ -196,9 +205,73 @@ An object representing a Host.
...
+- url: {scheme}://{host}:{port}/{basePath}
+  description: The production API server
+  templates:
+    scheme:
+      enum:
+        - https
+        - http
+      default: https
```
earth2marsh: `default` currently applies to optional parameters and represents the server-assumed default, like the number of results in a page or an offset. By applying it to a required element, we confuse an already tricky aspect of a spec. An alternative could be `example`.

versions/3.0.md
```diff
@@ -196,9 +205,73 @@ An object representing a Host.
...
+  description: Production server
+```
+
+The following shows how templates an be used for a host configuration
+
+```yaml
+servers:
+- url: {scheme}://{host}:{port}/{basePath}
+  description: The production API server
+  templates:
```
earth2marsh: As discussed, introducing `templates` when other path templates use `parameters` seems less elegant, but I don't have a better alternative to supply at this time.

fehguy:
@webron to add documentation about what happens when the `servers` attribute is empty or not present

PR #813, opened 2016-10-07
Fixed typos in Operation Object section

ralfhandl:
"schemes" was listed twice in table, kept nicer line
JSON example lacked a few curly braces


webron:
@OAI/tdc - merged as it really just fixes mistakes.

PR #814, opened 2016-10-10
Minor corrections for OpenAPI 3.0 document

brandur:
Adds a few minor corrections for the 3.0 definition document (or I hope these are correct anyway):
- Moves a parameter object example over to use the new 3.0 `schema` type definition instead of using the 2.0 `type` and `items` fields.
- Moves a stray `collectionFormat` out of `schema` definition and back into the root of a parameter object.

Totally unrelated, but does anyone know if there's any thinking around a release date for OpenAPI 3.0? I've built out a 2.0 API definition, but unfortunately a number of quirks in that spec prevent me from accurately describing my API, and I'm stuck building in some very hacky workarounds until something changes. I'm very much looking forward to the 3.0 release :)

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -466,9 +466,11 @@ This object can be extended with [Specification Extensions](#specificationExtens
       "in": "path",
       "description": "ID of pet to use",
       "required": true,
-      "type": "array",
-      "items": {
-        "type": "string"
+      "schema": {
+        "type": "array",
+        "items": {
+          "type": "string",
+        }
       },
       "collectionFormat": "csv"
     }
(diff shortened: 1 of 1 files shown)


webron:
Thanks for this. There are plenty of places and examples that need to be fixed and we started the process of clearing those out - every little bit helps.
As for the `collectionFormat`, we're working on finalizing what that would look like in the spec (it may change entirely).

brandur:
> Thanks for this. There are plenty of places and examples that need to be fixed and we started the process of clearing those out - every little bit helps.

No problem! Let me know if there's anything else I can fix and/or adjust.

webron:
@brandur - with additional changes (see #842 for example), I'm afraid the updates are no longer relevant. Thank you for the PR though.

Issue #673, opened 2016-04-26
Please add an Index to the specification


germanp:
This will make easier searching and reading.

fehguy:
Parent issue #589

RobDolinMS:
Do we want to number the spec sections as well?  /cc @fehguy @webron @darrelmiller

RobDolinMS:
I've started on this in a branch: https://github.com/RobDolinMS/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md  

Would welcome any feedback (or a thumbs-up) from more experienced spec contributors /cc @fehguy @webron @darrelmiller

webron:
Thanks @RobDolinMS, looking good. I'd suggest waiting with this for a bit longer though as content may easily shift around or even be renamed, giving you extra work at the end.

RobDolinMS:
I figured it would be good to get this added now so folks can use it (and comment on level of detail, if numbering would be useful, etc.)

I'm glad to pick-up an issue to do an edit/update pass as part of finalizing the spec.

webron:
Done!

PR #815, opened 2016-10-11
[3.0.md] Add table of contents

RobDolinMS:
Adds a table of contents to the Open API v3.0 spec


RobDolinMS:
This resolves #673

fehguy:
Certainly this will need to be updated after some time, but let's merge for general structure now!?

PR #816, opened 2016-10-15
Update README.md

cyberhicham:
typo

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# The OpenAPI Specification (fka The Swagger Specification)
+# The OpenAPI Specification (aka The Swagger Specification)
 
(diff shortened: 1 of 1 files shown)


darrelmiller:
fka stands for "formerly known as" whereas "aka" stands for "also known as", which is not the case.  What used to be the Swagger specification is no longer known as that.  It is now known as "OpenAPI Specification.

cyberhicham:
Oops, my bad !

ePaul:
Though maybe we should write this phrase out?

cyberhicham:
Maybe, as this is the first time I saw fka (feels wrong for me ^^)

PR #821, opened 2016-10-28
Incorrect definition name: pet

dzuluaga:
The definition for on the working example is "body" instead of "pet". Check out here http://petstore.swagger.io/v2/swagger.json.

### DIFF
--- a/examples/v2.0/yaml/petstore-expanded.yaml
+++ b/examples/v2.0/yaml/petstore-expanded.yaml
@@ -58,7 +58,7 @@ paths:
       description: Creates a new pet in the store.  Duplicates are allowed
       operationId: addPet
       parameters:
-        - name: pet
+        - name: body
           in: body
           description: Pet to add to the store
           required: true


examples/v2.0/yaml/petstore-expanded.yaml
```diff
@@ -58,7 +58,7 @@ paths:
       description: Creates a new pet in the store.  Duplicates are allowed
       operationId: addPet
       parameters:
-        - name: pet
+        - name: body
```
fehguy: You can actually name the body parameter whatever you like. So 'pet' is perfectly legal in this case

dzuluaga:
Yes. Thanks @fehguy. I noticed that. I was just suggesting to make it consistent with the working example in the URL above, which has name: body.

webron:
Thanks for this. A few thoughts:

1. The current example is not a mistake. In fact, it's a great example that the body name can be anything.
2. We're trying to keep 2.0-related changes a minimum only where required.

As such, I'm going to close this for now. We can reopen if people strongly about it.

PR #824, opened 2016-11-03
Fix server field names

jobol:
it is disturbing to not have the real field names


DavidBiesack:
1+

fehguy:
Merging as the previous value was in error.

PR #825, opened 2016-11-04
Docs/examples cleanup for Request/Response/Content/Types

jharmn:
* "media type" vs "media-type" consistency
* Broken or inconsistent formatting in samples
* A few tweaks here and there

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -85,10 +85,10 @@ Version | Date | Notes
 Path templating refers to the usage of curly braces ({}) to mark a section of a URL path as replaceable using path parameters.
 
 ##### <a name="mediaTypes"></a>Media Types
-Media type definitions are spread across several resources.
-The media type definitions should be in compliance with [RFC 6838](http://tools.ietf.org/html/rfc6838).
+Media-type definitions are spread across several resources.
+The media-type definitions should be in compliance with [RFC 6838](http://tools.ietf.org/html/rfc6838).
 
-Some examples of possible media type definitions:
+Some examples of possible media-type definitions:
 ```
   text/plain; charset=utf-8
   application/json
@@ -869,7 +869,8 @@ Field Name | Type | Description
 Field Pattern | Type | Description
 ---|:---:|---
 <a name="requestBodyRepresentation"></a>`*` | [Schema Object](#schemaObject) | The schema defining the request body.
-<a name="parameterExtensions"></a>^x- | Any | Allows extensions to the OpenAPI Schema. The field name MUST begin with `x-`, for example, `x-internal-id`. The value can be `null`, a primitive, an array or an object. See [Vendor Extensions](#vendorExtensions) for further details.
+
+This object can be extended with [Specification Extensions](#specificationExtensions). 
 
 
 ##### Request Body Examples
@@ -884,7 +885,7 @@ A request body with a referenced model definition.
           "$ref": "#/definitions/User"
         },
       "examples": [
-        { "$ref": 'http://foo.bar#/examples/address-example.json'}
+        { "$ref": "http://foo.bar#/examples/address-example.json"}
       ]
     },
     "application/xml" : {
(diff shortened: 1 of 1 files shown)


versions/3.0.md
```diff
@@ -85,10 +85,10 @@ Version | Date | Notes
 Path templating refers to the usage of curly braces ({}) to mark a section of a URL path as replaceable using path parameters.
 
 ##### <a name="mediaTypes"></a>Media Types
-Media type definitions are spread across several resources.
-The media type definitions should be in compliance with [RFC 6838](http://tools.ietf.org/html/rfc6838).
+Media-type definitions are spread across several resources.
+The media-type definitions should be in compliance with [RFC 6838](http://tools.ietf.org/html/rfc6838).
```
DavidBiesack: leave as "Media type" with a space, not a hyphen - this is how it is used on http://tools.ietf.org/html/rfc6838 !!

versions/3.0.md
```diff
@@ -964,61 +966,71 @@ Describes a set of supported content types. A content object can be used in [req
...
+    - "<Users/>"
+  text/plain:
     examples:
-      - "Bob,Diane,Mary,Bill"
+    - Bob,Diane,Mary,Bill
+    - ''
 ```
 
 #### <a name="contentTypeObject"></a>Content Type Object
-Each content type object provides schema and examples for a the media type identified by its key.  Content Type objects can be used in a [content object](#contentObject).
```
DavidBiesack: Revert this one also; should be "media type"

versions/3.0.md
```diff
@@ -1416,11 +1426,11 @@ Allows sharing examples for operation requests and responses.
 ##### Patterned Fields
 Field Pattern | Type | Description
 ---|:---:|---
-<a name="exampleMediaType"></a>{[media type](#mediaTypes)} | Any | The name of the property MUST be one of the Operation `produces` values (either implicit or inherited). The value SHOULD be an example of what such a response would look like.
```
DavidBiesack: Revert this one also; should be "media type"

versions/3.0.md
```diff
@@ -1416,11 +1426,11 @@ Allows sharing examples for operation requests and responses.
 ##### Patterned Fields
 Field Pattern | Type | Description
 ---|:---:|---
-<a name="exampleMediaType"></a>{[media type](#mediaTypes)} | Any | The name of the property MUST be one of the Operation `produces` values (either implicit or inherited). The value SHOULD be an example of what such a response would look like. 
+<a name="exampleMediaType"></a>{[media-type](#mediaTypes)} | Any | The name of the property MUST be one of the Operation `produces` values (either implicit or inherited). The value SHOULD be an example of what such a response would look like.
```
DavidBiesack: Revert this one also; should be "media type"

versions/3.0.md
```diff
@@ -85,10 +85,10 @@ Version | Date | Notes
 Path templating refers to the usage of curly braces ({}) to mark a section of a URL path as replaceable using path parameters.
 
 ##### <a name="mediaTypes"></a>Media Types
-Media type definitions are spread across several resources.
-The media type definitions should be in compliance with [RFC 6838](http://tools.ietf.org/html/rfc6838).
+Media-type definitions are spread across several resources.
+The media-type definitions should be in compliance with [RFC 6838](http://tools.ietf.org/html/rfc6838).
 
-Some examples of possible media type definitions:
```
DavidBiesack: Revert this one also; should be "media type"

versions/3.0.md
```diff
@@ -964,61 +966,71 @@ Describes a set of supported content types. A content object can be used in [req
 
 Each key in the content object is the media-type of the [Content Type Object](#contentTypeObject).
```
DavidBiesack: this should be "media type" ; `media-type` is not a member of the content type object; rather the key is a string which contains a media type.

versions/3.0.md
```diff
@@ -964,61 +966,71 @@ Describes a set of supported content types. A content object can be used in [req
...
 
+The media-type may take the form of a wildcard (`*`) to designate any `Content-Type`, or a regular expression for matching a specific type
```
DavidBiesack: Should be just "media type" as in the RFC

versions/3.0.md
```diff
@@ -964,61 +966,71 @@ Describes a set of supported content types. A content object can be used in [req
...
+  text/plain:
     examples:
-      - "Bob,Diane,Mary,Bill"
+    - Bob,Diane,Mary,Bill
+    - ''
 ```
 
 #### <a name="contentTypeObject"></a>Content Type Object
-Each content type object provides schema and examples for a the media type identified by its key.  Content Type objects can be used in a [content object](#contentObject).   
+Each content type object provides schema and examples for a the media-type identified by its key.  Content Type objects can be used in a [content object](#contentObject).
```
DavidBiesack: use "media type"

fehguy:
Looks good to me.  We can sanitize the wording later as I'm certain there will be more cleanup to do.

versions/3.0.md
```diff
@@ -884,49 +885,50 @@ A request body with a referenced model definition.
...
 
 ```yaml
 description: user to add to the system
-content: 
-  'application/json':
+content:
+  application/json:
     schema:
-      $ref: '#/definitions/User'
+      "$ref": "#/definitions/User"
```
webron: I'd avoid quoting keys unless necessary (for example, in case of numbers).
For consistency, we should use single quotes where possible in the YAML values (following a recommendation from YAML's author).

PR #828, opened 2016-11-04
Components Review

darrelmiller:


versions/3.0.md
```diff
@@ -395,7 +396,12 @@ Field Pattern | Type | Description
 <a name="parametersDefinitionsObject"></a> | [Parameters Definitions Object](#parametersDefinitionsObject) | An object to hold parameters to be reused across operations. Parameter definitions can be referenced to the ones defined here.
 <a name="responseHeadersDefinitionsObject"></a> | [Response Headers Definitions Object](#responseHeadersDefinitionsObject) | Response headers to reuse across the specification.
 <a name="securityDefinitionsObject"></a> | [Security Definitions Object](#securityDefinitionsObject) | Security definitions to reuse across the specification.
+<a name="linksDefinitionObject"></a> | [Link Definitions Object](#linksDefinitionsObject) | Link definitions to reuse across the specification.
+<a name="callbacksDefinitionObject"></a> | [Callback Definitions Object](#callbacksDefinitionsObject) | Callback definitions to reuse across the specification.
 
+All the fixed fields declared above are objects that MUST contain keys that only use the following characters: `a-zA-Z0-9.-_\`.
```
DavidBiesack: Start the regex with - otherwise this implies all characters between. and _

`-a-zA-Z0-9._\`

Also, is \ intentional or were you trying to escape some other character? This actually renders as if *backslash-space* is in the set of characters, which some my interpret as allowing a space???

versions/3.0.md
```diff
@@ -2831,15 +2837,14 @@ animals:
...
-An object to hold data types that can be consumed and produced by operations.
+An object to hold schemas for data types that can be consumed and produced by operations.
 These data types can be primitives, arrays or models.
 
 ##### Patterned Fields
 
 Field Pattern | Type | Description
 ---|:---:|---
 <a name="definitionsName"></a>{name} | [Schema Object](#schemaObject) | A single definition, mapping a "name" to the schema it defines.
-The following characters are legal in the `definitionsName`: `a-zA-Z0-9.-_\`.  All other characters, including extended ASCII characters, are not legal in the `definitionsName`.  The name may be used to namespace the definitions--it is the tooling Specification's job to ensure that the namespace is honored.
```
DavidBiesack: Same comment about - and \

versions/3.0.md
darrelmiller: I hadn't actually considered it regex expression and had merely thought of it as an informal list of valid characters.  However, obviously having some more formal definition would be good and as we don't tend to do ABNF here, I guess regex is a good option.  How does something like this look for a regex?

[a-zA-Z0-9.\-_\\]+

Does that help or confuse more?

darrelmiller:
@OAI/tdc This PR was reviewed by at the last TDC.  Comments on the regex syntax would be welcomed and let me know if this is good to merge.

versions/3.0.md
DavidBiesack: That regex is fine (I prefer putting `-` at the beginning so that it does not need escaping, but that is merely style.) 

However, I still don't know why `\` is in there.  That will also lead to confusing identifiers such as `doesThisContain\nNewlines\tAndTabs...OrNot`. Please add some prose indicating what `\` is used for.

darrelmiller: I agree.  I'll find out why the \ is there.  The constraint was previously just for definitions, I moved it to be applicable to all reusable components.

fehguy: Yes, honestly I think the `\` will only cause confusion, and as said on the @OAI/tdc call, it's easier to add that support than take it away.

PR #831, opened 2016-11-07
Fix typo in 2.0.md

estliberitas:


DavidBiesack:
I think "fka" was intentional, as "formerly known as" not "also known as". Might be better to spell it out thoogh.

estliberitas:
@DavidBiesack ouch, sorry, did not know the origins :) Then, it's up to you guys.

webron:
Yup, `fka` is the right term. Thanks for taking the time to submit it though.

estliberitas:
@webron np ;)

PR #839, opened 2016-11-17
Tweak grammar and diction

gibson042:


webron:
Thank you, but we're really not accepting changes to 2.0 unless insanely critical. If the same changes apply to 3.0, please send a PR against it and we'll review it.

PR #841, opened 2016-11-18
Tweak grammar and diction

gibson042:


versions/3.0.md
```diff
@@ -119,12 +119,11 @@ For example, if a field is said to have an array value, the JSON array represent
 }
 ```
 
-While the API is described using JSON it does not impose a JSON input/output to the API itself.
+Even though the API is described with JSON, it is not required to use JSON for input or output.
```
DavidBiesack: The 3.0 spec uses both JSON and YAML, so I think the entire sentence can be removed. JSON/YAML representation is discussed in the beginning of this section.

gibson042: I personally like the explicit separation, and have updated the sentence. But if you still want it removed completely, just say the word.

PR #842, opened 2016-11-18
added parameter object updates for #840

fehguy:


fehguy:
@OAI/tdc This PR addresses the `parameter` object docs of #589.  It introduces two concepts which have been discussed but not agreed upon:

1) Parameters have previously had a simple definition type. For example:

```yaml
in: query
type: string
name: search
```

To preserve this simplicity, I have added back this section from the 2.0 specification.

2) To support complex parameters, and address the serialization needs, we introduce, instead of the `schema` property, the `content` property as used elsewhere:

```yaml
in: query
name: search
content:
  application/json:
    schema:
      type: object
```

This allows the following:

* defining serialization strategies for complex types
* keeping a simple structure for the 99% of use cases where primitives and arrays of primitives are passed

webron:
Regarding the simple case - restrictions related to `formData` parameters should be removed as it no longer exists as a type.

versions/3.0.md
```diff
@@ -743,7 +743,36 @@ Field Name | Type | Description
...
+<a name="parameterAllowEmptyValue"/>allowEmptyValue | `boolean` | Sets the ability to pass empty-valued parameters. This is valid only for either `query` or `formData` parameters and allows you to send a parameter with a name only or  an empty value. Default value is `false`.
+<a name="parameterItems"></a>items | [Items Object](#itemsObject) | **Required if [`type`](#parameterType) is "array".** Describes the type of items in the array.
+<a name="parameterCollectionFormat"></a>collectionFormat | `string` | Determines the format of the array if type array is used. Possible values are: <ul><li>`csv` - comma separated values `foo,bar`. <li>`ssv` - space separated values `foo bar`. <li>`tsv` - tab separated values `foo\tbar`. <li>`pipes` - pipe separated values <code>foo&#124;bar</code>. <li>`multi` - corresponds to multiple parameter instances instead of multiple values for a single instance `foo=bar&foo=baz`. This is valid only for parameters [`in`](#parameterIn) "query" or "formData". </ul> Default value is `csv`.
```
fehguy: We should probably take this out

webron:
@fehguy to clean out the PR and be merged. @whitlockjc and and @darrelmiller to discuss additional options for simplification.

PR #847, opened 2016-11-19
Replace links

fmvilas:
Hi all,

Just saw the links in the Component Object table were broken and here is a PR to fix it. Hope it helps.

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -389,16 +389,16 @@ All objects defined within the components object will have no effect on the API 
 
 ##### Fixed Fields
 
-Field Pattern | Type | Description
+Field Name | Type | Description
 ---|:---:|---
-<a name="definitionsObject"></a> | [Definitions Object](#definitionsObject) | A hash containing payload definitions for the specification.
-<a name="responsesDefinitionsObject"></a> | [Responses Definitions Object](#responsesDefinitionsObject) | Reusable responses objects.
-<a name="parametersDefinitionsObject"></a> | [Parameters Definitions Object](#parametersDefinitionsObject) | An object to hold parameters to be reused across operations. Parameter definitions can be referenced to the ones defined here.
-<a name="requestBodyDefinitionsObject"></a> | [Request Body Definitions Object](#requestBodyDefinitionsObject) | An object to hold request body definitions to be reused across operations. Request Body definitions can be referenced to the ones defined here.
-<a name="responseHeadersDefinitionsObject"></a> | [Response Headers Definitions Object](#responseHeadersDefinitionsObject) | Response headers to reuse across the specification.
-<a name="securityDefinitionsObject"></a> | [Security Definitions Object](#securityDefinitionsObject) | Security definitions to reuse across the specification.
-<a name="linksDefinitionObject"></a> | [Link Definitions Object](#linksDefinitionsObject) | Link definitions to reuse across the specification.
-<a name="callbacksDefinitionObject"></a> | [Callback Definitions Object](#callbacksDefinitionsObject) | Callback definitions to reuse across the specification.
+<a name="componentsDefinitionsObject"></a>definitions | [Definitions Object](#definitionsObject) | A hash containing payload definitions for the specification.
+<a name="componentsResponsesDefinitionsObject"></a>responsesDefinitions | [Responses Definitions Object](#responsesDefinitionsObject) | Reusable responses objects.
+<a name="componentsParametersDefinitionsObject"></a>parametersDefinitions | [Parameters Definitions Object](#parametersDefinitionsObject) | An object to hold parameters to be reused across operations. Parameter definitions can be referenced to the ones defined here.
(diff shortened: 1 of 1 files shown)


webron:
Thanks for this. I'm afraid this doesn't follow the standard naming convention we use so I can't merge it yet.

Object field anchors contain the object name only (without `Object`) and do not have a hyphen. 

However, looking at the current documentation it looks like that table is broken more than that, because it doesn't even contain the field names and the header is wrong.

versions/3.0.md
```diff
@@ -391,14 +391,14 @@ All objects defined within the components object will have no effect on the API
...
-<a name="requestBodyDefinitionsObject"></a> | [Request Body Definitions Object](#requestBodyDefinitionsObject) | An object to hold request body definitions to be reused across operations. Request Body definitions can be referenced to the ones defined here.
-<a name="responseHeadersDefinitionsObject"></a> | [Response Headers Definitions Object](#responseHeadersDefinitionsObject) | Response headers to reuse across the specification.
-<a name="securityDefinitionsObject"></a> | [Security Definitions Object](#securityDefinitionsObject) | Security definitions to reuse across the specification.
-<a name="linksDefinitionObject"></a> | [Link Definitions Object](#linksDefinitionsObject) | Link definitions to reuse across the specification.
-<a name="callbacksDefinitionObject"></a> | [Callback Definitions Object](#callbacksDefinitionsObject) | Callback definitions to reuse across the specification.
+<a name="componentsObject-definitionsObject"></a> | [Definitions Object](#definitionsObject) | A hash containing payload definitions for the specification.
```
webron:
```
<a name="componentsDefinitionsObject"></a>definitions | [Definitions Object](#definitionsObject) | A hash containing payload definitions for the specification.
```
in this case (notice the anchor and the added field name).

The first header of the table, `Field Pattern` should be `Field Name`.

fmvilas: Got it. Will fix today. Thanks!

fmvilas:
@webron Please review the field names as it's something I don't know.

versions/3.0.md
```diff
@@ -389,16 +389,16 @@ All objects defined within the components object will have no effect on the API
...
-<a name="linksDefinitionObject"></a> | [Link Definitions Object](#linksDefinitionsObject) | Link definitions to reuse across the specification.
-<a name="callbacksDefinitionObject"></a> | [Callback Definitions Object](#callbacksDefinitionsObject) | Callback definitions to reuse across the specification.
+<a name="componentsDefinitionsObject"></a>definitions | [Definitions Object](#definitionsObject) | A hash containing payload definitions for the specification.
+<a name="componentsResponsesDefinitionsObject"></a>responsesDefinitions | [Responses Definitions Object](#responsesDefinitionsObject) | Reusable responses objects.
+<a name="componentsParametersDefinitionsObject"></a>parametersDefinitions | [Parameters Definitions Object](#parametersDefinitionsObject) | An object to hold parameters to be reused across operations. Parameter definitions can be referenced to the ones defined here.
+<a name="componentsRequestBodyDefinitionsObject"></a>requestBodyDefinitions | [Request Body Definitions Object](#requestBodyObject) | An object to hold request body definitions to be reused across operations. Request Body definitions can be referenced to the ones defined here.
```
ePaul: `#requestBodyObject` → `#requestBodyDefinitionsObject`?
Ah, that one is still missing.

versions/3.0.md
```diff
@@ -389,16 +389,16 @@ All objects defined within the components object will have no effect on the API
...
+<a name="componentsResponseHeadersDefinitionsObject"></a>responseHeadersDefinitions | [Response Headers Definitions Object](#responseObject) | Response headers to reuse across the specification.
```
ePaul: The correct link target here, `#responseHeadersDefinitionsObject`, is also still missing.

versions/3.0.md
```diff
@@ -389,16 +389,16 @@ All objects defined within the components object will have no effect on the API
...
+<a name="componentsSecurityDefinitionsObject"></a>securityDefinitions | [Security Definitions Object](#securityDefinitionsObject) | Security definitions to reuse across the specification.
+<a name="componentsLinksDefinitionObject"></a>linkDefinitions | [Link Definitions Object](#linkObject) | Link definitions to reuse across the specification.
+<a name="componentsCallbacksDefinitionObject"></a>callbackDefinitions | [Callback Definitions Object](#callbackObject) | Callback definitions to reuse across the specification.
```
ePaul: Both `#callbackDefinitionsObject` and `#linkDefinitionsObject` are still missing. Or are they the same as the callback Object and links object linked here? Then the name of the links here should be changed, too.

ePaul:
As just mentioned in my review, some of the "* definitions objects" included in the components object are not yet described in the specification. They should certainly be added near the existing similar ones. I don't have the overview about the currently open issues – if there is not yet an issue to track this, please create one.

webron:
Thanks for the PR. That section was overhauled eventually, so no longer needed.

fmvilas:
Cool, thanks for the good job you're doing here guys. I'm sorry I don't have enough time to contribute more often to OpenAPI.

Issue #843, opened 2016-11-18
Relative vs Absolute URLs


darrelmiller:
There are a number of properties that are defined in the OpenAPI specification.  Most of them do not specify if the URLs can be relative or absolute, and if they are allowed to be relative, where they are relative to.

Here is the list of URLs in the specification:
 - TermsOfService
 - ContactUrl
 - LicenseUrl
 - Server.Url
 - ExternalDocUrl
 - Link href
 - openIdConnectUrl
 - authorizationUrl
 - tokenUrl
 - refreshUrl

My suggestion is that wording should be added to indicate that all URLs should be allowed to be relative unless specifically prohibited and relative URLs are relative to the serverURL.  Only the Server.Url should be indicated as an URL that MUST be absolute.

Example:

```YAML
openapi: 3.0.0
info:
  title: My Awesome API
  termsOfService: ./terms/
servers:
- url: https://sandbox.gigantic-server.com/v1
  description: User testing server
- url: https://api.gigantic-server.com/v1
  description: Production server
```

The terms of service for the testing server would be at `https://sandbox.gigantic-server.com/terms` and for the production server `https://api.gigantic-server.com/terms`

Note that the terms for the production server is found at `https://api.gigantic-server.com/terms` and not `https://api.gigantic-server.com/v1/terms` due to the lack of a trailing slash on the server URL.

DavidBiesack:
The example does not match up. The server `url` in the example is `https://sandbox.gigantic-server.com/v1` so appending the termsOfService relative URL `./terms/` to it yields ``https://sandbox.gigantic-server.com/v1./terms/` which is probably not the intent. 

I don't think we can/should strip anything (`v1`) off the server url. For example:

```YAML
openapi: 3.0.0
info:
  title: My Awesome API
  termsOfService: /terms
servers:
- url: https://sandbox.gigantic-server.com/v1
  description: User testing server
- url: https://api.gigantic-server.com/v1
  description: Production server
```

with the resulting url `https://sandbox.gigantic-server.com/v1/terms` for the test server, and  `https://api.gigantic-server.com/v1/terms` for prod

darrelmiller:
@DavidBiesack We could choose to say that relative resolution is done via simple string concatenation, but that is not how it is defined in [RFC 3986](https://tools.ietf.org/html/rfc3986#section-5.4.1) and it is not how browsers do it, nor how libraries like .net do it.  Can't say for Java.

[...]

PR #855, opened 2016-12-12
make path → URL resolution clearer.

ePaul:
You can't append a string to an object (at least not with a canonical result), so we should make clear that we want to append to the `url` field of the server object (possibly after filling in the templates there).

The highlighting of "appended" makes it clearer that there is no relative URL resolution as mentioned in https://github.com/OAI/OpenAPI-Specification/issues/843#issuecomment-261575257.

The task list in https://github.com/OAI/OpenAPI-Specification/issues/589#issuecomment-250816162 says the cleanup of the Paths object is already finished, so I felt free to add this edit now.

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -408,15 +408,15 @@ All the fixed fields declared above are objects that MUST use keys that match th
 
 #### <a name="pathsObject"></a>Paths Object
 
-Holds the relative paths to the individual endpoints.
-The path is appended to the [`Server Object`](#serverObject) in order to construct
+Holds the relative paths to the individual endpoints and their operations.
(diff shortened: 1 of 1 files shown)


ePaul:
I just rebased this branch to solve the conflicts.

darrelmiller:
The changes look good to me.  I wish we could avoid using the word "endpoint" but unfortunately there is no good word for thing that the `path object` represents.

ePaul:
@darrelmiller maybe "path family"?

webron:
Let's not introduce new terminology now. Suggest opening a ticket for future clarifications. This can still make it in 3.0.0-FINAL (which will just be called 3.0.0).

PR #871, opened 2017-01-27
Issue870

curtisr7:
Fix for OAI/OpenAPI-Specification/issues/870.

Note that I have two additional commits to clean up random trailing whitespaces and replaced the few tab indents with spaces. 

Please let me know if you don't want the space/tab commits and I can remove.


webron:
Thanks for the PR. We've changed so many things and cleaned those up before releasing RC0.

Issue #874, opened 2017-01-30
3.0.md Parameter Object typo


curtisr7:
The parameter object definition states that there are four possible parameter types[1], but the `in` field states that there are five types[2].

I *believe* that the bulleted list should include formData.

Another nit. Should the bulleted types[1] be lower case rather than upper cased? ie: `path`, `query`, etc, etc?

[1] https://github.com/OAI/OpenAPI-Specification/blame/OpenAPI.next/versions/3.0.md#L734
[2] https://github.com/OAI/OpenAPI-Specification/blame/OpenAPI.next/versions/3.0.md#L745

darrelmiller:
The issues with the parameter types have been fixed.  FormData is now handled with a content object, schema object and encoding object.

Issue #800, opened 2016-09-29
Example(s) in Content Type Objects


darrelmiller:
In the definition of a [Content Type Object](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#content-type-object) we currently allow either an array of 'examples' or a single 'example' object.

There are conflicting opinions on how we should do this:
- Only have an `example` object and allow secondary examples to be defined using x- extensions
- Only have an `examples` array and require a single item array for single examples
- Add another object layer in examples array to provide example metadata, to be able to make a distinction between the examples.
- Leave it as is and support both `example` and `examples`

DavidBiesack:
I vote against using `x-` extensions. that renders them generally unusable by OpenAPI tools. 

I would like to see each example be an Example Object with at least a `description`, as may examples require or benefit from at least some explanation.

I like the freedom of either singular `example` **or** `examples`  but not both (i.e. `oneOf`). It's a simple transformation from singular to list in tools. We should support the `examples` array. Most cases probably have only one example, but we should not make it more cumbersome the majority by requiring `examples: [ ... ]`

jharmn:
This issue is moot with the current spec. `description` is at the Request & Response Object level, not in `content` or `content type` at all.
@darrelmiller closing, reopen if I missed something.

Issue #430, opened 2015-07-31
Improve binary data definitions


webron:
Following the discussion in #50, for the next version of the spec we should provide a better way of defining transfer of binary data. This should also include support for multiple file upload (#254) and supporting file transfer using other mime types (#326).

evigeant:
As mentioned here: https://github.com/swagger-api/swagger-core/issues/1531#issuecomment-161080235

I would like the spec to support a binary format which would indicate that it needs to be streamed. I believe a lot of APIs performing file upload and download would benefit from this as the current `file` type is too restrictive (only for multipart upload, cannot be used for generated content).

I reproduce here my proposal for an additional `string` format:

### Data Types ###
Common Name | type | format | Comments
----------- | ------ | -------- | --------
string | `string` | | |
byte | `string` | `byte` | base64 encoded characters
binary | `string` | `binary` | any sequence of octets
blob | `string` | `blob` | a long sequence of octets (treated as a stream)

The generators would then use an appropriate data type (ex: `java.io.InputStream`) to represent these parameters.

The name `blob` is only a suggestion, it comes from the database world, however it does convey the notion of an object too large to be handled in the normal sense.

webron:
Well, I'll copy my comment too ;)

Not sure the spec should deal with how things are handled but rather what they are. To me, in your case, `binary` and `blob` are the same. If you want to have something that's implementation-specific, that's where the extensions come in.

evigeant:
Sorry for the cross-posting, I looked around a lot in the issues to find problems similar to what I was having. I'll continue the discussion here only.

I understand your point and I agree that the spec should document the data type and not the implementation details in general. However, this specific problem is quite common and one could argue that binary data that cannot be handled in memory is different from binary data that can.

But the real issue is that any API for large file download cannot be represented optimally using swagger. What I mean is that if I specify `binary` as my result type, all (most?) clients will be generated using byte[] which will not work because the data cannot be held entirely in memory. This happens regardless of any vendor extensions I put in the spec. Basically, one of the major benefits of using swagger, namely generating client code, is not applicable for any API doing file download or raw file upload (not multipart).

webron:
No worries about the cross posting and happy to see you moved the discussion to the right place.

Thanks for providing the additional details, they should be taken into account. Question is - can we really tell the clients what's 'too big' for the to handle? I get the need to hint them about it, we just need to be careful about how we end up documenting it.

evigeant:
[...]

Issue #418, opened 2015-07-20
Expand 'example' support


webron:
Later on I'll try to add the history behind the issue, but for now, here's an explanation of what's needed:
- Allow providing examples at the parameter level. Right now it's limited to definitions which affect only body parameters. This also means that the same example would be used for all operations referencing it.
- Allow providing examples based on mime-type. This will become even more critical with the added support for versions and other mime-type based features in the future.

Fival85:
@ApiParam(value = "The Ordernumber", example = {"1234", "5678"}) @PathParam("orderNumber")

That would be nice and easy to use :)

jasonh-n-austin:
Link to parent #565

webron:
Done.

PR #878, opened 2017-02-01
documented `requestBody`

### LINKED ISSUE #665: Create serialization strategy for non-body params

fehguy:
This PR adds documentation for `requestBody` to describe inputs into operations.

This fixes #665, #303, #222, #418, #179, #874, #800, #303,

It clarifies #430

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -1116,7 +1116,7 @@ schema:
   type: string
   format: binary
 
-# mutliple files:
+# multiple files:
 schema:
   type: array
   items:
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -1213,7 +1213,7 @@ requestBody:
               type: '#/definitions/Address'
 ```
 
-In scenarios where more control is needed over the Content-Type for `multipart` request bodies, an `encoding` attribute is introduced.  This attribute is _only_ applicable to `mulitpart/*` request bodies.
+In scenarios where more control is needed over the Content-Type for `multipart` request bodies, an `encoding` attribute is introduced.  This attribute is _only_ applicable to `mulitpart/*` and `x-www-form-urlencoded` request bodies.
 
 #### <a name="encodingObject"></a>Encoding Object
 
@@ -1231,7 +1231,7 @@ A single encoding definition applied to a single schema property
 ##### Fixed Fields
 Field Name | Type | Description
 ---|:---:|---
-<a name="contentType"></a>Content-Type | `string` | **Required.** The content-type to use for encoding a specific property.
+<a name="contentType"></a>contentType | `string` | **Required.** The content-type to use for encoding a specific property.
 <a name="headers"></a>Headers | `object` | A string map allowing additional information to be provided as headers, for example `Content-Disposition`.  Note `Content-Type` is described separately and will be ignored from this section.
 <a name="style"></a>Style | `string` | The content-type to use for encoding a specific property.  See (#parameterContent) for details on the `style` property
 <a name="explode"></a>explode | `boolean` | When this is true, property values of type `array` or `object` generate seperate parameters for each value of the array, or key-value-pair of the map.  For other types of properties this property has no effect. The default value is false.
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -145,13 +145,11 @@ By convention, the OpenAPI Specification (OAS) file is named `openapi.json` or `
 Primitive data types in the OAS are based on the types supported by the [JSON-Schema Draft 4](http://json-schema.org/latest/json-schema-core.html#anchor8).
 Models are described using the [Schema Object](#schemaObject) which is a subset of JSON Schema Draft 4.
 
-An additional primitive data type `"file"` is used by the [Parameter Object](#parameterObject) and the [Response Object](#responseObject) to set the parameter type or the response as being a file.
-
 <a name="dataTypeFormat"></a>Primitives have an optional modifier property `format`.
 OAS uses several known formats to more finely define the data type being used.
 However, the `format` property is an open `string`-valued property, and can have any value to support documentation needs.
 Formats such as `"email"`, `"uuid"`, etc., can be used even though they are not defined by this specification.
-Types that are not accompanied by a `format` property follow their definition from the JSON Schema (except for `file` type which is defined above).
+Types that are not accompanied by a `format` property follow their definition from the JSON Schema.
 The formats defined by the OAS are:
 
 
@@ -742,7 +740,7 @@ There are four possible parameter types.
 Field Name | Type | Description
 ---|:---:|---
 <a name="parameterName"></a>name | `string` | **Required.** The name of the parameter. Parameter names are *case sensitive*. <ul><li>If [`in`](#parameterIn) is `"path"`, the `name` field MUST correspond to the associated path segment from the [path](#pathsPath) field in the [Paths Object](#pathsObject). See [Path Templating](#pathTemplating) for further information.<li>For all other cases, the `name` corresponds to the parameter name used based on the [`in`](#parameterIn) property.</ul>
-<a name="parameterIn"></a>in | `string` | **Required.** The location of the parameter. Possible values are "query", "header", "path", "formData" or "cookie".
+<a name="parameterIn"></a>in | `string` | **Required.** The location of the parameter. Possible values are "query", "header", "path" or "cookie".
 <a name="parameterDescription"></a>description | `string` | A brief description of the parameter. This could contain examples of use.  [CommonMark syntax](http://spec.commonmark.org/) can be used for rich text representation.
 <a name="parameterRequired"></a>required | `boolean` | Determines whether this parameter is mandatory. If the parameter is [`in`](#parameterIn) "path", this property is **required** and its value MUST be `true`. Otherwise, the property MAY be included and its default value is `false`. 
 <a name="parameterDeprecated"></a> deprecated | `boolean` | Specifies that a parameter is deprecated and should be transitioned out of usage.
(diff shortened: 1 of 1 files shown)


versions/3.0.md
```diff
@@ -1124,6 +1100,185 @@ application/json:
...
+schema:
+  type: string
+  format: base64
+
+# content transfered in binary (octet-stream):
+schema:
+  type: string
+  format: binary
+
+# mutliple files:
```
Danielku15: Should be `multiple`

versions/3.0.md
```diff
@@ -580,7 +578,7 @@ Field Name | Type | Description
...
-<a name="operationRequestBody"></a>requestBody | [[Request Body Object](#requestBodyObject) <span>&#124;</span> [Reference Object](#referenceObject)] | The request body applicable for this operation. 
+<a name="operationRequestBody"></a>requestBody | [[Request Body Object](#requestBodyObject) <span>&#124;</span> [Reference Object](#referenceObject)] | The request body applicable for this operation.  The `requestBody` is only supported in HTTP methods where the [HTTP 1.1 specification](https://tools.ietf.org/html/rfc7231#section-4.3.1) has explicitly defined semantics for request bodies.
```
whitlockjc: What does _"is only supported"_ mean?  Does that mean you cannot specify a `requestBody` in the OAI document in these situations or that the consumer of the OAI document is not required to do anything with the `requestBody` in these situations?  If this is the latter, what impact does this have on tooling authors?

fehguy: It means the spec does not allow, for example, `requestBody` on GET operations, and as a tooling vendor, you're breaking the spec by supporting it.

whitlockjc: Thanks, carry on.

whitlockjc:
There are a few places where you use _``requestBody``_ and I wonder what you think about using a link to the _"Request Body Object"_ documentation itself instead?  I see that in the _"Content Object"_ section _(https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#content-object) you're already doing this and that's why it came to mind.  Other than addressing this for consistency, LGTM.

versions/3.0.md
webron: The language needs to be stronger. Should say either that it MUST be one of the specific HTTP verbs (meaning using it with GET, DELETE...) will make the spec invalid - OR - that it SHALL be ignored if used with the wrong verbs (meaning, validation will allow it to exist there, but tools can safely ignore).

versions/3.0.md
```diff
@@ -1124,6 +1100,185 @@ application/json:
...
+##### Considerations for file uploads
+
+In contrast with the 2.0 specification, describing `file` input/output content in OpenAPI is 
+described with the same semantics as any other schema type. Specifically:
+
+```yaml
+# content transferred with base64 encoding
+schema:
+  type: string
+  format: base64
```
webron: This is `byte` today. We can change, but needs to be updated in the other section as well.

versions/3.0.md
```diff
@@ -1124,6 +1100,185 @@ application/json:
...
+  type: string
+  format: binary
+
+# multiple files:
+schema:
+  type: array
+  items:
+    type: string
+    format: binary
+```
```
webron: This is incorrect. You can't have multiple file uploads with octet-stream, as they have no separator. For multiple files, the schema has to be named.

versions/3.0.md
```diff
@@ -1124,6 +1100,185 @@ application/json:
...
+              type: string
+          addresses:
+            # default Content-Type for arrays is based on the `inner` type (object shown, so `application/json` in this example)
+            type: array
+            items:
+              type: '#/definitions/Address'
+```
+
+In scenarios where more control is needed over the Content-Type for `multipart` request bodies, an `encoding` attribute is introduced.  This attribute is _only_ applicable to `mulitpart/*` request bodies.
+
```
webron: Language is okay - this means that a use can choose to define it still, but it will be ignored if not `multipart/*`.

versions/3.0.md
```diff
@@ -1124,6 +1100,185 @@ application/json:
...
+#### <a name="encodingObject"></a>Encoding Object
+
+An object representing multipart region encoding for `requestBody` objects
+
+##### Patterned Fields
+Field Pattern | Type | Description
+---|:---:|---
+<a name="encodingObjectProperty"></a>The property to apply encoding to | [Encoding Property](#encodingProperty) <span>&#124;</span> [Encoding](#encoding) | The field name to apply special encoding attributes to.  This field must exist in the schema as a property.  To avoid collisions with specification extensions, properties may not begin with `x-`.
```
webron: What's `Encoding Property`? Take a look at the `type` column.

webron: Would also suggest to reverse the restriction. I think it's better to allow `x-` in properties and not allow extensions here, as you can add the extensions a level up covering this as well.

versions/3.0.md
```diff
@@ -1124,6 +1100,185 @@ application/json:
...
+This object can be extended with [Specification Extensions](#specificationExtensions). 
+
+#### <a name="encoding"></a>Encoding
+
+A single encoding definition applied to a single schema property
+
+##### Fixed Fields
+Field Name | Type | Description
+---|:---:|---
+<a name="contentType"></a>Content-Type | `string` | **Required.** The content-type to use for encoding a specific property.
```
webron: media type

webron:
Thanks for putting this together, @fehguy - this addresses a lot of important issues.

Following our discussions, one thing is missing and that's addressing custom serialization format with `x-www-form-urlencoded`. The defaults are mentioned. IIRC, these should fall under the `Encoding Object`.

If we choose to skip that, then the `Encoding Object` structure can be simplified.

versions/3.0.md
darrelmiller: I'm not sure this is true.  See http://stackoverflow.com/questions/913626/what-should-a-multipart-http-request-with-multiple-files-look-like

webron: That's multipart. This is octet-stream.

versions/3.0.md
```diff
@@ -1124,6 +1100,185 @@ application/json:
...
+            type: object
+            properties: {}
+          profileImage:
+            # default is application/octet-stream, need to declare an image type only!
+            type: string
+            format: binary
+encoding:
+  historyMetadata:
+    # require XML content-type in utf-8 encoding
+    contentType: application/xml; charset=utf-8
```
fehguy: what about `Content-Disposition`?

versions/3.0.md
darrelmiller: You are right.  My bad.

fehguy: Ron just reminded me that we need to have a property with a `name`, holding the `array` property.  Thus **Ron** is right, we can document that an array request body without schema property is meaningless and unsupported by all implementations of the tooling (thx @webron )

fehguy: +1 @webron

versions/3.0.md
fehguy: @webron is wrong.  Should stay `contentType`.

fehguy: + @webron

versions/3.0.md
fehguy: Waiting for @darrelmiller for feedback on this

versions/3.0.md
whitlockjc: I agree that not allowing vendor extensions here would be best.

versions/3.0.md
fehguy: Let's add `style` here, great idea @webron ?

versions/3.0.md
fehguy: this will be a documentation concern

versions/3.0.md
```diff
@@ -1124,6 +1100,179 @@ application/json:
...
+
+#### <a name="encoding"></a>Encoding
+
+A single encoding definition applied to a single schema property
+
+##### Fixed Fields
+Field Name | Type | Description
+---|:---:|---
+<a name="contentType"></a>Content-Type | `string` | **Required.** The content-type to use for encoding a specific property.
+<a name="headers"></a>Headers | `object` | A string map allowing additional information to be provided as headers, for example `Content-Disposition`.  Note `Content-Type` is described separately and will be ignored from this section.
```
whitlockjc: Is this only applicable for multi-part support?  If so, shouldn't we document it as such?

fehguy: yes it is.  I'll address in the PR

webron:
LGTM (though `Content-Type` should be `contentType`).

versions/3.0.md
darrelmiller: This should be updated to say the encoding attribute are also applies to `application/x-www-form-urlencoded`

fehguy: fine.

versions/3.0.md
```diff
@@ -1124,6 +1100,179 @@ application/json:
...
+A single encoding definition applied to a single schema property
+
+##### Fixed Fields
+Field Name | Type | Description
+---|:---:|---
+<a name="contentType"></a>contentType | `string` | **Required.** The content-type to use for encoding a specific property.
+<a name="headers"></a>Headers | `object` | A string map allowing additional information to be provided as headers, for example `Content-Disposition`.  Note `Content-Type` is described separately and will be ignored from this section.
+<a name="style"></a>Style | `string` | The content-type to use for encoding a specific property.  See (#parameterContent) for details on the `style` property
+<a name="explode"></a>explode | `boolean` | When this is true, property values of type `array` or `object` generate seperate parameters for each value of the array, or key-value-pair of the map.  For other types of properties this property has no effect. The default value is false.
+<a name="contentType"></a>Content-Type | `string` | **Required.** The content-type to use for encoding a specific property.
```
darrelmiller: Content Type is here twice.  Should it even be here or do we move it to headers?

darrelmiller:
LGTM (though see comment about content-type, Ron might be wrong!)

versions/3.0.md
fehguy: oops.  I'll remove one of them.

I left it out of the `headers` section intentionally.  Since it describes the serialization strategy it's (a) required and (b) easier to predict in it's own attribute.

fehguy:
@OAI/tdc shall we merge? (cha cha cha)

whitlockjc:
LGTM

darrelmiller:
:shipit:

PR #884, opened 2017-02-11
Updates to parameter and requestBody

webron:
Summary of changes:
- Renamed `commaDelimited` to `simple`, reference the type in the RFC and made not applicable to `query`. Described behavior for primitive value, added exploded example.
- Set default `style` based on parameter type.
- Explained `allowEmptyValue` is inapplicable if example table value is `n/a`.
- Added example of free-form query parameters, made explicit typed example.
- Added support for style to header and cookie params - only one style allowed for each.
- Fixed anchors in Encoding Object.
- Changed the default `style` of `x-www-form-urlencoded` to `form`, set behavior to be like `query` parameters.
- Made `contentType` not required, described default values.

Note that the behavior of `headeer` and `cookie` parameters is _different_ than what we discussed.

#### `header`

For values `[ red, green, blue ]`:

```yaml
/test:
  get:
    parameters:
    - in: header
      name: colors
      required: true
      type: array
      style: simple
      items:
        type: string
```

will be `colors: red,green,blue` regardless of whether `explode` is true/false.
This is different than the discussed result of `explode` true represented as:
`colors: colors=red&colors=green&colors=blue`.

#### `cookie`
By following `style` being `form`, we get the representation of `Set-Cookie: colors=red,green,blue` by default, so no extra explanation needed (unlike using `simple`).

---
Open issues:
- What would `explode` being `true` for `cookie` parameter will look like?
- Many users requested that the `collectionFormat` by default be `multi` and not `csv`. This means that we should set `explode` to `true` for `query` parameters by default. It's a cumbersome explanation (if this then that, otherwise something else), but doable. Opinions welcome.

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -760,7 +760,7 @@ For simpler scenarios, a [`style`](#parameterStyle) and [`schema`](#parameterSch
 Field Name | Type | Description
 ---|:---:|---
 <a name="parameterStyle"></a>style | `string` | Describes how the parameter value will be serialized depending on type of the parameter value. Default values (based on value of `in`): for `query` - `form`; for `path` - `simple`; for `header` - `simple`; for `cookie` - `form`.
-<a name="parameterExplode"></a>explode | `boolean` | When this is true, parameter values of type `array` or `object` generate seperate parameters for each value of the array, or key-value-pair of the map.  For other types of parameters this property has no effect. The default value is false.
+<a name="parameterExplode"></a>explode | `boolean` | When this is true, parameter values of type `array` or `object` generate seperate parameters for each value of the array, or key-value-pair of the map.  For other types of parameters this property has no effect. When [`style`](#parameterStyle) is `form`, the default value is `true`. For all other styles, the default value is `false`.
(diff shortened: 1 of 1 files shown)


fehguy:
`simple`? Seems like that's a bit harder to figure out without cross-referencing...

webron:
Reusing the name from https://tools.ietf.org/html/rfc6570#section-3.2.2, like the other `style` names from there.

fehguy:
OK sure, but it still sucks

webron:
Alternatives welcome.

fehguy:
`commaSeparated` is a good idea...  but we can follow the standard

webron:
We can discuss it tomorrow. I don't think it's a good idea because unlike in 2.0, it cannot be applied to `query` and `formData` (or `requestBody` now).

`csv` has been effectively replaced by both `simple` and `form`, depending on the situation.

darrelmiller:
Were there still open issues here?  I thought we resolved everything.

webron:
Per our discussion, changed the default value of `explode` to `true` when the `style` is `form` and the rest has a default value of `false`. This makes the default from 2.0 be `multi` instead of `csv`.

PR #891, opened 2017-02-17
Callback review and preliminary message expression syntax

darrelmiller:


versions/3.0.md
```diff
@@ -1555,15 +1555,70 @@ description: object created
 
 #### <a name="callbackObject"></a>Callback Object
 
-A container for possible out-of band callbacks from an operation. A callback may be returned from an operation, calling back to the path specified in the operation object.
+A map of possible out-of band callbacks related to the parent operation. Each value in the map is an [operation object](#operationObject) that describes a request that may be initiated by the API provider and the responses expected. The key value used to identify callback object is an expression, that when evaluated at runtime, identifies a URL to used for the callback operation.
```
DavidBiesack: add article "the" : "used to identify callback the object"

DavidBiesack: "is an expression, evaluated at runtime, that identifies... "

versions/3.0.md
```diff
@@ -1555,15 +1555,70 @@ description: object created
...
 
 ##### Patterned Fields
 Field Pattern | Type | Description
 ---|:---:|---
```
DavidBiesack: I believe CommonMark does not support tables with this notation; one must use HTML <table> elements.

versions/3.0.md
```diff
@@ -1555,15 +1555,70 @@ description: object created
...
-<a name="responseName"></a>Callback name | [Callback Operation Object](#operationObject) <span>&#124;</span> [Operation Object](#operationObject) | An operation object used to define a callback payload structure
+<a name="responseName"></a>Callback name | [Callback Operation Object](#operationObject) <span>&#124;</span> [Operation Object](#operationObject) | An operation object used to define a callback request and expected responses
```
DavidBiesack: name="callbackName" perhaps?

versions/3.0.md
```diff
@@ -1555,15 +1555,70 @@ description: object created
...
 
 This object can be extended with [Specification Extensions](#specificationExtensions). 
 
+##### Key Expression
+
+The key used to identify the callback object is also an expression that can be evaluated on the runtime HTTP request/response to identify the URL to be used for the callback request.  A simple example might be "$request.body.url".  However, the expression syntax enables accessing the complete HTTP message and reaching into any payload that can map to an object/array/property data model.
```
DavidBiesack: "evaluated in the context of a runtime HTTP request/response" (It's not clear what "evaluated on the runtime HTTP request/response object means" since those are not executable, but they do provide a *context* for expression evaluation.

versions/3.0.md
```diff
@@ -1555,15 +1555,70 @@ description: object created
...
-<a name="responseName"></a>Callback name | [Callback Operation Object](#operationObject) <span>&#124;</span> [Operation Object](#operationObject) | An operation object used to define a callback payload structure
+<a name="responseName"></a>Callback name | [Callback Operation Object](#operationObject) <span>&#124;</span> [Operation Object](#operationObject) | An operation object used to define a callback request and expected responses
 
 This object can be extended with [Specification Extensions](#specificationExtensions). 
 
+##### Key Expression
+
+The key used to identify the callback object is also an expression that can be evaluated on the runtime HTTP request/response to identify the URL to be used for the callback request.  A simple example might be "$request.body.url".  However, the expression syntax enables accessing the complete HTTP message and reaching into any payload that can map to an object/array/property data model. 
+
+The expression is defined by the following [ABNF](https://tools.ietf.org/html/rfc5234) syntax
```
DavidBiesack: Did you consider using [JSON Pointer](https://tools.ietf.org/html/rfc6901) or [JSONPath](http://goessner.net/articles/JsonPath/) for the `body-reference` expressions? In particular, when extracting elements from arrays, it is often necessary to pass in predicates/filters to match a specific array elements. (I'm extrapolating based on from experience with [UnRAVL](https://github.com/sassoftware/unravl/blob/master/doc/Bind.md) which also allows extracting data from response bodies.)

versions/3.0.md
```diff
@@ -1555,15 +1555,70 @@ description: object created
...
+Here are the examples of how the various expressions evaluate, assuming a the callback operation has a path parameter named `eventType` and a query parameter named `queryUrl`.
+
+Expression | Value 
+---|:---
+$request.path.eventType | myevent
+$request.query.queryUrl | http://clientdomain.com/stillrunning
+$request.header.content-Type | application/json
+$request.body.failed | http://clientdomain.com/stillrunning
+$request.body.successurls[2] | http://clientdomain.com/medium
+$response.header.Location | http://example.org/subscription/1
```
DavidBiesack: This expression notation may need some mechanism to provide a default value if the item is not found, i.e. $request.body.successurls[2] | $response.header.Location
would return the second successurl if it exists, or the location header.

versions/3.0.md
```diff
@@ -1555,15 +1555,70 @@ description: object created
...
+
+```
+      expression = target source *( "." bind-token [ index ]] )
+      target = ("$request." | "$response.")
+      source = ( header-reference | query-reference | path-reference | body-reference ) 
+      header-reference = "header." token
+      query-reference = "query." name  
+      path-reference = "path." name
+      body-reference = "body" *( index | "." name )
+      index = "[" *(digit) "]"
```
DavidBiesack: document is indexing is zero-based (I assume, recommend) or 1-based.

versions/3.0.md
darrelmiller: Fixed.

versions/3.0.md
darrelmiller: :-) I realize this might sound a little hypocritical but we are suggesting CommonMark in OpenAPI definitions, but we are still using GFM in the spec.

versions/3.0.md
darrelmiller: Fixed

versions/3.0.md
darrelmiller: Fixed.  Thanks!

versions/3.0.md
darrelmiller: I had not considered using those for just the `body-reference`.  I had ruled out using them for the complete expression, but using it for just the body-reference  is quite an cool idea.  I'd like to use JSON Pointer because we already need that in tooling to support $refs.  

I have a fairly significant update to this PR because I found it contradicted some of the docs in Links which I am also reviewing.  I'll post an update soon and I'll see if it is easy to incorporate JSONPointer.  Nice thing about that is approach is in the future we could open the door to using other fragment syntaxes for other media types where JSONPointer is not appropriate.

versions/3.0.md
darrelmiller: If we move to JSONPointer, I can punt on this :-)

versions/3.0.md
darrelmiller: In the Links section, the suggestion is that if the item is not found, then the value is null.  I'm not exactly sure what we do if the reference points to something that is not a primitive.

darrelmiller:
@DavidBiesack Huge thanks for the feedback.  I've made most of the fixes.  Updated PR coming soon.

versions/3.0.md
darrelmiller: I'm not sure about the predicates/filters, I do see the value but it adds quite a bit of work for tooling.  I'm not going to include anything for the moment, but I'll bring it up with the TDC tomorrow.

darrelmiller:
This latest update has a number of changes to the variable expressions used for both callbacks and links.

It was necessary to add syntax to access parameters using '$request.path.{name}' '$request.query.{name}' and '$request.header.{name}' since the parameter names are only unique in a single location.

Also, instead of creating a new syntax for reaching into request and response bodies, this proposal suggests using a JSON Pointer for accessing the body instead:
 e.g.
    $response.body(/id)

Another alternative syntax would be

    $response.body#/id

But I'm tending to prefer the function style over the fragment

versions/3.0.md
```diff
@@ -1559,17 +1559,62 @@ Response with no return value:
...
+For example, given the following HTTP request:
+
+```http
+POST /subscribe/myevent?queryUrl=http://clientdomain.com/stillrunning HTTP/1.1
+Host: example.org
+Content-Type: application/json
+Content-Length: 123
+
+{
+  "failedurl" : "http://clientdomain.com/failed"
```
earth2marsh: Suggest `failedUrl` and `successUrls` (camel)

whitlockjc: I like the `$response.body#/id` variant.

whitlockjc:
:shipit:

fehguy:
Well that's a fun read!  Looks good to me, you have my 👍

versions/3.0.md
```diff
@@ -1559,17 +1559,62 @@ Response with no return value:
 description: object created
 ```
 
-#### <a name="callbackObject"></a>Callback Object
+#### <a name="callbacksObject"></a>Callbacks Object
 
-A container for possible out-of band callbacks from an operation. A callback may be returned from an operation, calling back to the path specified in the operation object.
+A map of possible out-of band callbacks related to the parent operation.
+Each value in the map is an [operation object](#operationObject) that describes a request that may be initiated by the API provider and the expected responses.
```
webron: We can fix it later, but should be Operation Object

versions/3.0.md
```diff
@@ -1559,17 +1559,62 @@ Response with no return value:
...
+<a name="callbackName"></a>Callback name | [Callback Operation Object](#operationObject) <span>&#124;</span> [Operation Object](#operationObject) | An operation object used to define a callback request and expected responses
 
 This object can be extended with [Specification Extensions](#specificationExtensions). 
 
+##### Key Expression
+
+The key used to identify the callback object is a variable expression that can be evaluated in the context of a runtime HTTP request/response to identify the URL to be used for the callback request.
+A simple example might be `$request.body#/url`.
+However, using [variable substition](#variableSubstition) syntax the complete HTTP message can be accessed.
+This includes accessing any part of a body that can be accessed using a JSONPointer [RFC6901](https://tools.ietf.org/html/rfc6901).
```
webron: JSON Pointer

versions/3.0.md
```diff
@@ -1688,11 +1735,11 @@ The `Link Object` is responsible for defining a possible operation based on a si
...
-parameters  | Link Parameters Object | an Object representing parameters to pass to an operation as specified with `operationId` or identified via `href`.
-headers     | Link Headers Object    | an Object representing headers to pass to the linked resource.
-description | string | a description of the link, supports [CommonMark syntax](http://spec.commonmark.org/).
+href        | `string`    | a relative or absolute URL to a linked resource. This field is mutually exclusive with the `operationId` field.
+operationId | `string` | the name of an _existing_, resolvable OAS operation, as defined with a unique `operationId`.  This field is mutually exclusive with the `href` field.  Relative `href` values _may_ be used to locate an existing Operation Object in the OAS.
+parameters  | [Link Parameters Object](#linkParameters) | an Object representing parameters to pass to an operation as specified with `operationId` or identified via `href`.
```
webron: `Parameters Object` instead of `Link Parameters Object`

webron:
:shipit: 👍 

I was going to say something about headers being case insensitive but you got it!

Does it make sense to support cookies too in variable substitution? (#probablynot #pleasedontkillme)

versions/3.0.md
webron: I stand corrected!

PR #893, opened 2017-02-20
Examples documentation

fehguy:
Documentation clarification for example, examples


versions/3.0.md
```diff
@@ -765,7 +765,7 @@ Field Name | Type | Description
 <a name="parameterAllowReserved"></a>allowReserved | `boolean` | Determines whether the parameter value should allow reserved characters, as defined by [RFC3986](https://tools.ietf.org/html/rfc3986#section-2.2) `:/?#[]@!$&'()*+,;=` to be included without percent-encoding. This property only applies to parameters with an `in` value of `query`. The default value is `false`.
 <a name="parameterSchema"></a>schema | [Schema Object](#schemaObject) | The schema defining the type used for the parameter.
 
-For more complex scenarios a `content` object can be used to define the media-type and schema of the parameter.
+For more complex scenarios a `content` object can be used to define the media-type and schema of the parameter.  Note that because the `content` object provides the appropriate serialization between examples and content serialization strategies, the `example` is not allowed in the simple `schema` scenario.
```
fehguy: Note that we are removing support for `example`, `examples` for the simple `schema` case.  @OAI/tdc make sure this makes sense!

The reasoning is that if examples are to be specified, there is no reason to avoid using the `content` object which gives explicit example format information

versions/3.0.md
```diff
@@ -1102,10 +1102,10 @@ application/json:
   schema:
     $ref: "#/definitions/Pet"
   examples:
+    # converted directly from YAML to JSON
     - name: Fluffy
       petType: Cat
-    - name: Rover
-      petType: Frog
+    - {"name": "Rover", "petType": "Frog"}
```
fehguy: These two examples are semantically equivalent

versions/3.0.md
```diff
@@ -2153,13 +2153,15 @@ description: Pets operations
...
 #### <a name="examplesObject"></a>Examples Object
 
-Anywhere an `example` may be given, allow a $ref object.
-This does mean that `example`, structurally, can be either a string primitive or an object, like `additionalProperties`.
+Anywhere an `example` may be given, a JSON Reference may be used, with the 
+explicit restriction that examples having a JSON format with object named 
+`$ref` are not allowed. This does mean that `example`, structurally, can be 
+either a string primitive or an object, similar to `additionalProperties`.
 
-In locations where the field being provided an `example` is a scalar value _or_ has it's content-type definition determined by a higher-level construct (a response payload, for example, uses the `produces` attribute to select the correct message format), the plural `examples` shall be used, and the payload format be specified as a key to the example.
```
fehguy: With the move to `requestBody`, this didn't make sense anymore

versions/3.0.md
handrews: Ironically, we added `"examples"` to JSON Schema for draft 06 (which I know you're not moving to anytime soon if ever, just thought I'd mention it).

darrelmiller: @fehguy I'm not sure I agree with this.  If `format` is `phonenumber`, or `postalCode` you may want to include an example.

versions/3.0.md
DavidBiesack: I find the json encoding inside the YAML (while correct) to be confusing; I don't think it is done elsewhere in the 3.0 spec.

versions/3.0.md
fehguy: just kidding!  Wanted to see if y'all were paying attention.

versions/3.0.md
```diff
@@ -1102,10 +1102,10 @@ application/json:
   schema:
```
fehguy: inside a content object

earth2marsh: Capturing a TDC discussion comment "I would expect that any tooling worth its salt would throw this out if it failed to parse, but that it would try to parse it."

whitlockjc:
Does it make sense to allow a `description` property for the `examples` scenario?  If I have N different examples, I'm likely showing the different valid ways this request/response could be structured and it would be nice to have a description.  Yes, this would mean an object for the example instead of a string/object but it could be really useful.  Thoughts?

darrelmiller:
:shipit:

webron:
@whitlockjc - @darrelmiller brought the issue up during the call. While it makes sense, to avoid the extra layer we won't support it now. Maybe in future versions.

Other than that, LGTM.

whitlockjc:
:shipit:

versions/3.0.md
```diff
@@ -1068,8 +1073,8 @@ Each content type object provides schema and examples for a the media type ident
...
 ---|:---:|---
 <a name="contentTypeSchema"></a>schema | [Schema Object](#schemaObject) | The schema defining the type used for the request body.
-<a name="contentTypeExamples"></a>examples | [Examples Array](#examplesArray) | Examples of the content type.
-<a name="contentTypeExample"></a>example | [Example Object](#exampleObject) | Example of the content type.
+<a name="contentTypeExamples"></a>examples | [Examples Array](#examplesArray) | Examples of the content type.  Each example in the Examples Array must be in the correct format as specified in the _content_ type.  The `examples` object is mutually exclusive to the `example` object.  Furthermore, if referencing a `schema` which contains an example, the `examples` value will _override_ the example provided by the schema.
+<a name="contentTypeExample"></a>example | [Example Object](#exampleObject) | Example of the content type.  The example object _must_ be in the correct format as specified in the _content_ type.  The `example` object is mutually exclusive to the `examples` object.  Furthermore, if referencing a `schema` which contains an example, the `example` value will _override_ the the example provided by the schema.
```
DavidBiesack: I thought on the call Ron suggested softening the MUST to a SHOULD. (in the above para, use MUST not must as per rfc)

versions/3.0.md
```diff
@@ -2153,13 +2158,15 @@ description: Pets operations
...
+explicit restriction that examples having a JSON format with object named 
+`$ref` are not allowed. This does mean that `example`, structurally, can be 
+either a string primitive or an object, similar to `additionalProperties`.
 
-In locations where the field being provided an `example` is a scalar value _or_ has it's content-type definition determined by a higher-level construct (a response payload, for example, uses the `produces` attribute to select the correct message format), the plural `examples` shall be used, and the payload format be specified as a key to the example.
-
-In all cases, the payload is expected to be compatible with the type schema for the value that it is accompanying.
-Tooling Specifications may choose to validate compatibility automatically, and reject the example value(s) if they are not compatible.
+In all cases, the payload is expected to be compatible with the type schema 
+for the value that it is accompanying.  Tooling implementations may choose to
```
DavidBiesack: ... more weight to using SHOULD instead of MUSt above.

PR #894, opened 2017-02-21
Update Schema Object

webron:
(This replaced PR #741)

Covers:
- Clarifications about JSON Schema property adjustments to the OAS
- Support for `nullable`
- Support for `writeOnly`, removed `required` restriction
- `discriminator` clarifications
- examples cleanup

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -2260,18 +2260,14 @@ $ref: 'definitions.yaml#/Pet'
 
 The Schema Object allows the definition of input and output data types.
 These types can be objects, but also primitives and arrays.
-This object is based on the [JSON Schema Specification Draft 4](http://json-schema.org/) and uses a predefined subset of it.
-On top of this subset, there are extensions provided by this specification to allow for more complete documentation.
+This object is an extended subset of the [JSON Schema Specification Draft 4](http://json-schema.org/).
 
 Further information about the properties can be found in [JSON Schema Core](http://json-schema.org/latest/json-schema-core.html) and [JSON Schema Validation](http://json-schema.org/latest/json-schema-validation.html).
 Unless stated otherwise, the property definitions follow the JSON Schema specification as referenced here.
 
 The following properties are taken directly from the JSON Schema definition and follow the same specifications:
-- $ref - As a [JSON Reference](https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03)
-- format (See [Data Type Formats](#dataTypeFormat) for further details)
+
 - title
-- description ([CommonMark syntax](http://spec.commonmark.org/) can be used for rich text representation)
-- default (Unlike JSON Schema, the value MUST conform to the defined type for the Schema Object)
 - multipleOf
 - maximum
 - exclusiveMaximum
@@ -2287,25 +2283,32 @@ The following properties are taken directly from the JSON Schema definition and 
 - minProperties
 - required
 - enum
-- type
 
-The following properties are taken from the JSON Schema definition but their definitions were adjusted to the OpenAPI Specification.
-Their definition is the same as the one from JSON Schema, only where the original definition references the JSON Schema definition, the [Schema Object](#schemaObject) definition is used instead.
-- items
-- allOf
-- oneOf
-- anyOf
-- not
-- properties
-- additionalProperties
+The following properties are taken from the JSON Schema definition but their definitions were adjusted to the OpenAPI Specification. 
+- type - Value MUST be a string. Multiple types via an array are not supported.
+- allOf - Inline or referenced schema MUST be of a [Schema Object](#schemaObject) and not a standard JSON Schema.
+- oneOf - Inline or referenced schema MUST be of a [Schema Object](#schemaObject) and not a standard JSON Schema.
+- anyOf - Inline or referenced schema MUST be of a [Schema Object](#schemaObject) and not a standard JSON Schema.
+- not - Inline or referenced schema MUST be of a [Schema Object](#schemaObject) and not a standard JSON Schema.
+- items - Value MUST be an object and not an array. Inline or referenced schema MUST be of a [Schema Object](#schemaObject) and not a standard JSON Schema. `items` MUST be present of the `type` is `array`.
+- properties - Property definitions MUST be a [Schema Object](#schemaObject) and not a standard JSON Schema (inline or referenced).
+- additionalProperties - Value can be boolean or object. If object, definition MUST be a [Schema Object](#schemaObject) and not a standard JSON Schema (inline or referenced).
+- $ref - As a [Reference Object](#referenceObject).
+- description ([CommonMark syntax](http://spec.commonmark.org/) can be used for rich text representation).
+- format - See [Data Type Formats](#dataTypeFormat) for further details. While relying on JSON Schema's defined formats, the OAS offers a few additional predefined formats.
+- default - Unlike JSON Schema, the value MUST conform to the defined type for the Schema Object.
+
+Additional properties defined by the JSON Schema specification that are not mentioned here are strictly unsupported.
 
 Other than the JSON Schema subset fields, the following fields may be used for further schema documentation.
 
 ##### Fixed Fields
 Field Name | Type | Description
 ---|:---:|---
-<a name="schemaDiscriminator"></a>discriminator | `string` | Adds support for polymorphism. The discriminator is the schema property name that is used to differentiate between other schema that inherit this schema. The property name used MUST be defined at this schema and it MUST be in the `required` property list. When used, the value MUST be the name of this schema or any schema that inherits it.
-<a name="schemaReadOnly"></a>readOnly | `boolean` | Relevant only for Schema `"properties"` definitions. Declares the property as "read only". This means that it MAY be sent as part of a response but MUST NOT be sent as part of the request. Properties marked as `readOnly` being `true` SHOULD NOT be in the `required` list of the defined schema. Default value is `false`.
+<a name="schemaNullable"></a>nullable | `boolean` | Allows sending a `null` value for the defined schema. Default value is `false`.
+<a name="schemaDiscriminator"></a>discriminator | `string` | Adds support for polymorphism. The discriminator is the schema property name that is used to differentiate between other schema that inherit this schema. The property name used MUST be defined at this schema and it MUST be in the `required` property list. When used, the value MUST be the name of this schema or any schema that inherits it. See [Composition and Inheritance](#schemaComposition) for more details.
+<a name="schemaReadOnly"></a>readOnly | `boolean` | Relevant only for Schema `"properties"` definitions. Declares the property as "read only". This means that it MAY be sent as part of a response but MUST NOT be sent as part of the request. If property is marked as `readOnly` being `true` and is in the `required` list, the `required` will take effect on the response only. A property MUST NOT be marked as both `readOnly` and `writeOnly` being `true`. Default value is `false`.
+<a name="schemaWriteOnly"></a>writeOnly | `boolean` | Relevant only for Schema `"properties"` definitions. Declares the property as "write only". This means that it MAY be sent as part of a request but MUST NOT be sent as part of the response. If property is marked as `writeOnly` being `true` and is in the `required` list, the `required` will take effect on the request only. A property MUST NOT be marked as both `readOnly` and `writeOnly` being `true`. Default value is `false`.
 <a name="schemaXml"></a>xml | [XML Object](#xmlObject) | This MAY be used only on properties schemas. It has no effect on root schemas. Adds Additional metadata to describe the XML representation format of this property.
 <a name="schemaExternalDocs"></a>externalDocs | [External Documentation Object](#externalDocumentationObject) | Additional external documentation for this schema. 
 <a name="schemaExample"></a>example | Any | A free-form property to include an example of an instance for this schema.
@@ -2314,7 +2317,7 @@ Field Name | Type | Description
 
 This object can be extended with [Specification Extensions](#specificationExtensions). 
 
-###### Composition and Inheritance (Polymorphism)
+###### <a name="schemaComposition"></a>Composition and Inheritance (Polymorphism)
 
 The OpenAPI Specification allows combining and extending model definitions using the `allOf` property of JSON Schema, in effect offering model composition.
 `allOf` takes in an array of object definitions that are validated *independently* but together compose a single object. 
@@ -2323,7 +2326,9 @@ While composition offers model extensibility, it does not imply a hierarchy betw
 To support polymorphism, OpenAPI Specification adds the support of the `discriminator` field.
 When used, the `discriminator` will be the name of the property used to decide which schema definition is used to validate the structure of the model.
 As such, the `discriminator` field MUST be a required field.
-The value of the chosen property has to be the friendly name given to the model under the `definitions` property.
+There are are two ways to define the value of a discriminator for an inheriting instance.
+- Use the definition's name.
+- Override the definition's name by overriding the property with a new value. If exists, this takes precedence over the definition's name.
 As such, inline schema definitions, which do not have a given id, *cannot* be used in polymorphism.
 
 ###### XML Modeling
@@ -2334,8 +2339,6 @@ The [XML Object](#xmlObject) contains additional information about the available
 ##### Schema Object Examples
 
 ###### Primitive Sample
-
-Unlike previous versions of Swagger, OpenAPI Schema definitions can be used to describe primitive and arrays as well.
 
 ```json
 {
@@ -2362,7 +2365,7 @@ format: email
       "type": "string"
     },
     "address": {
-      "$ref": "#/definitions/Address"
+      "$ref": "#/components/definitions/Address"
     },
     "age": {
       "type": "integer",
@@ -2381,7 +2384,7 @@ properties:
   name:
     type: string
   address:
-    $ref: '#/definitions/Address'
+    $ref: '#/components/definitions/Address'
   age:
     type: integer
     format: int32
@@ -2413,7 +2416,7 @@ For a string to model mapping:
 {
   "type": "object",
   "additionalProperties": {
-    "$ref": "#/definitions/ComplexModel"
+    "$ref": "#/components/definitions/ComplexModel"
   }
 }
 ```
@@ -2421,7 +2424,7 @@ For a string to model mapping:
 ```yaml
 type: object
 additionalProperties:
-  $ref: '#/definitions/ComplexModel'
+  $ref: '#/components/definitions/ComplexModel'
 ```
 
 ###### Model with Example
@@ -2599,7 +2602,7 @@ definitions:
       ]
     },
     "Cat": {
-      "description": "A representation of a cat",
+      "description": "A representation of a cat. Note that `Cat` will be used as the discriminator value.",
       "allOf": [
         {
           "$ref": "#/definitions/Pet"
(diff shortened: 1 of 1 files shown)


versions/3.0.md
```diff
@@ -2287,25 +2283,32 @@ The following properties are taken directly from the JSON Schema definition and
...
-- allOf
-- oneOf
-- anyOf
-- not
-- properties
-- additionalProperties
+
+The following properties are taken from the JSON Schema definition but their definitions were adjusted to the OpenAPI Specification. 
+- type - Value MUST be a string. Multiple types via an array are not supported.
+- allOf - Inline or referenced schema MUST be of a [Schema Object](#schemaObject) and not a standard JSON Schema.
```
DavidBiesack: Should this be:
"MUST be of a [Schema Object](#schemaObject) or [Schema Reference](#reference-object)" ? (Ditto on the other items in the list.

DavidBiesack: Please elaborate on what "not a standard JSON Schema." means. However, if it means "not a full JSON Schema but the OpenAPI Specification schema", I think we should just drop this since you already refer explicitly to `[Schema Object](#schemaObject)`

versions/3.0.md
```diff
@@ -2287,25 +2283,32 @@ The following properties are taken directly from the JSON Schema definition and
...
+- oneOf - Inline or referenced schema MUST be of a [Schema Object](#schemaObject) and not a standard JSON Schema.
+- anyOf - Inline or referenced schema MUST be of a [Schema Object](#schemaObject) and not a standard JSON Schema.
+- not - Inline or referenced schema MUST be of a [Schema Object](#schemaObject) and not a standard JSON Schema.
+- items - Value MUST be an object and not an array. Inline or referenced schema MUST be of a [Schema Object](#schemaObject) and not a standard JSON Schema. `items` MUST be present of the `type` is `array`.
```
DavidBiesack: typo: "*of* the type" s/b "*if* the type"

versions/3.0.md
```diff
@@ -2287,25 +2283,32 @@ The following properties are taken directly from the JSON Schema definition and
...
+- properties - Property definitions MUST be a [Schema Object](#schemaObject) and not a standard JSON Schema (inline or referenced).
+- additionalProperties - Value can be boolean or object. If object, definition MUST be a [Schema Object](#schemaObject) and not a standard JSON Schema (inline or referenced).
+- $ref - As a [Reference Object](#referenceObject).
```
DavidBiesack: suggest change "As a Reference Object" to "A Reference Object"

versions/3.0.md
```diff
@@ -2260,18 +2260,14 @@ $ref: 'definitions.yaml#/Pet'
...
 
 Further information about the properties can be found in [JSON Schema Core](http://json-schema.org/latest/json-schema-core.html) and [JSON Schema Validation](http://json-schema.org/latest/json-schema-validation.html).
 Unless stated otherwise, the property definitions follow the JSON Schema specification as referenced here.
 
 The following properties are taken directly from the JSON Schema definition and follow the same specifications:
-- $ref - As a [JSON Reference](https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03)
-- format (See [Data Type Formats](#dataTypeFormat) for further details)
+
 - title
-- description ([CommonMark syntax](http://spec.commonmark.org/) can be used for rich text representation)
```
DavidBiesack: remove parens here (not used elsewhere in OAS when describing CommonMark)

versions/3.0.md
```diff
@@ -2287,25 +2283,32 @@ The following properties are taken directly from the JSON Schema definition and
...
+- allOf - Inline or referenced schema MUST be of a [Schema Object](#schemaObject) and not a standard JSON Schema.
+- oneOf - Inline or referenced schema MUST be of a [Schema Object](#schemaObject) and not a standard JSON Schema.
+- anyOf - Inline or referenced schema MUST be of a [Schema Object](#schemaObject) and not a standard JSON Schema.
+- not - Inline or referenced schema MUST be of a [Schema Object](#schemaObject) and not a standard JSON Schema.
+- items - Value MUST be an object and not an array. Inline or referenced schema MUST be of a [Schema Object](#schemaObject) and not a standard JSON Schema. `items` MUST be present of the `type` is `array`.
+- properties - Property definitions MUST be a [Schema Object](#schemaObject) and not a standard JSON Schema (inline or referenced).
+- additionalProperties - Value can be boolean or object. If object, definition MUST be a [Schema Object](#schemaObject) and not a standard JSON Schema (inline or referenced).
+- $ref - As a [Reference Object](#referenceObject).
+- description ([CommonMark syntax](http://spec.commonmark.org/) can be used for rich text representation).
```
DavidBiesack: remove parens here (not used elsewhere in OAS when describing CommonMark)

versions/3.0.md
```diff
@@ -2599,7 +2602,7 @@ definitions:
       ]
     },
     "Cat": {
-      "description": "A representation of a cat",
+      "description": "A representation of a cat. Note that `Cat` will be used as the discriminator value.",
```
DavidBiesack: Please add an example:

Below are examples showing an array of Pet objects using `petType` as a discriminator:
```yaml
- name: "Fluffy"
  petType: "Cat"
  huntingSkill: "lazy"
- name: "Rover"
  petType: "Dog"
  packSize: 6
```

```json
[
  {
    "name" : "Fluffy",
    "petType" : "Cat",
    "huntingSkill" : "lazy"
  },
  {
    "name" : "Rover",
    "petType" : "Dog",
    "packSize" : 6
  }
]

```

versions/3.0.md
webron: It should be the Schema Object, not Schema Reference (which I assume you meant JSON Reference). The Schema Object contains the JSON Reference in it.

The 'standard JSON Schema' is in reference to the initial point that the Schema Object is an extended subset of JSON Schema. I'd rather keep this explicit here due to questions raised in the past.

webron:
Updates:
- Updated JSON Schema support from Draft 04 to Draft 05 (or wright-00).
- Following that, updated support for External references instead of Canonical dereferencing (same thing, different name).
- Updated references to JSON Pointer.
- Updated fallback of tools to deal with unknown formats.
- Some wording updates.

Not included:
- Updated example with `discriminator` value overriding. Pending review of the `discriminator` as a whole.

Let me know if I missed anything we discussed on the call today.

versions/3.0.md
```diff
@@ -144,14 +144,15 @@ By convention, the OpenAPI Specification (OAS) file is named `openapi.json` or `
 
 ### Data Types
 
-Primitive data types in the OAS are based on the types supported by the [JSON-Schema Draft 4](http://json-schema.org/latest/json-schema-core.html#anchor8).
-Models are described using the [Schema Object](#schemaObject) which is a subset of JSON Schema Draft 4.
+Primitive data types in the OAS are based on the types supported by the [JSON-Schema Draft Wright 00](https://tools.ietf.org/html/draft-wright-json-schema-00#section-4.2) (also known as, JSON Schema Draft 05). Note that `integer` as a type is also supported and is defined as a JSON number without a fraction or exponent part. `null` is not supported as a value.
+Models are described using the [Schema Object](#schemaObject) which is an extended subset of JSON Schema Draft 05.
```
whitlockjc: Does it read better using _"JSON-Schema Draft Wright 00"_ here since that is what you mentioned prior?  Also, was hyphenating _JSON Schema_ to _JSON-Schema_ intentional in the link text?  If so, what's the rules behind when we hyphenate and when we do not?

versions/3.0.md
```diff
@@ -2214,8 +2215,8 @@ definitions:
 A simple object to allow referencing other definitions in the specification.
 It can be used to reference parameters and responses that are defined at the top level for reuse.
 
-The Reference Object is a [JSON Reference](http://tools.ietf.org/html/draft-pbryan-zyp-json-ref-02) that uses a [JSON Pointer](http://tools.ietf.org/html/rfc6901) as its value.
-For this specification, only [canonical dereferencing](http://json-schema.org/latest/json-schema-core.html#anchor27) is supported.
+The Reference Object is a [JSON Reference](https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03) that uses a [JSON Pointer](http://tools.ietf.org/html/rfc6901) as its value.
```
whitlockjc: I'm not sure I understand what's being said here.  Is the _Reference Object_ **only** for local references, as in references within the same document?  If that's the case, the value is itself a _JSON Pointer_ as a _JSON Reference_ allows both local and remote references.

darrelmiller: Reference Object is for all references whether local or remote.

darrelmiller:
:shipit:

versions/3.0.md
whitlockjc: That's what I was hoping for.  So in that case, the _"that uses a JSON Pointer as its value"_ is not accurate is it?  The _JSON Pointer_ version of a `$ref` value is for _local only_ and since a _Reference Object_ is for local and remote references, isn't the _Reference Object_'s `$ref` just a URI per the [JSON References spec](https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03#section-3).

versions/3.0.md
webron: The hyphen is a mistake. Will fix. 

I can change it to JSON Schema Draft Wright 00 everywhere, but it's not the exact name. The exact name is draft-wright-json-schema-00 which is awkward to write across the doc.

versions/3.0.md
webron: You're right, @whitlockjc. The indication of JSON Pointer here is not needed. It was added in 2.0, back in the days where things were still unclear to us all. We now know better - A JSON References, is a JSON Reference, and in *some* cases the value is a JSON Pointer. The actual explanation here is wrong and unneeded. Will remove.

versions/3.0.md
whitlockjc: Makes sense.  I trust your judgement.

versions/3.0.md
whitlockjc: 👍

versions/3.0.md
webron: Fixed.

versions/3.0.md
webron: Completely reworked JSON Reference, see main comment.

webron:
A few last minute updates to this PR:
- Updated the naming of JSON Schema to be consistent throughout the doc.
- Completely rewrote the Reference Object and updated its examples. I didn't realize the wording was relevant to the 2.0 but not 3.0. The wording should be simpler now, explain exactly what it is or isn't, and avoid the potential confusion regarding its resolution.

@tdc - please review again.

darrelmiller:
Yes, the wording is even better now. :shipit:

versions/3.0.md
whitlockjc: I'm with @DavidBiesack on this one, I would expect `allOf` to take either an inline _Schema Object_ or a _JSON Reference_ to a _Schema Object_ defined elsewhere.

whitlockjc: @DavidBiesack I almost responded with a _"+1"_ to your comment but I think the _"Inline or referenced schema"_ as to being either a _Schema Object_ or a _JSON Reference_ to a _Schema Object_ defined elsewhere.  @webron's working seem fine to me.

versions/3.0.md
```diff
@@ -2287,25 +2284,32 @@ The following properties are taken directly from the JSON Schema definition and
...
+
+The following properties are taken from the JSON Schema definition but their definitions were adjusted to the OpenAPI Specification. 
+- type - Value MUST be a string. Multiple types via an array are not supported.
+- allOf - Inline or referenced schema MUST be of a [Schema Object](#schemaObject) and not a standard JSON Schema.
+- oneOf - Inline or referenced schema MUST be of a [Schema Object](#schemaObject) and not a standard JSON Schema.
+- anyOf - Inline or referenced schema MUST be of a [Schema Object](#schemaObject) and not a standard JSON Schema.
+- not - Inline or referenced schema MUST be of a [Schema Object](#schemaObject) and not a standard JSON Schema.
+- items - Value MUST be an object and not an array. Inline or referenced schema MUST be of a [Schema Object](#schemaObject) and not a standard JSON Schema. `items` MUST be present if the `type` is `array`.
+- properties - Property definitions MUST be a [Schema Object](#schemaObject) and not a standard JSON Schema (inline or referenced).
+- additionalProperties - Value can be boolean or object. If object, definition MUST be a [Schema Object](#schemaObject) and not a standard JSON Schema (inline or referenced).
```
whitlockjc: Can it be a _JSON Reference_ to a _Schema Object_?

versions/3.0.md
```diff
@@ -2287,25 +2284,32 @@ The following properties are taken directly from the JSON Schema definition and
...
+- $ref - A [Reference Object](#referenceObject).
```
whitlockjc: Should we mention this at all?  To me, any place you could have a _Schema Object_ you would allow a _JSON Reference_ to a _Schema Object_ but it is an either/or, as in you cannot mix a _JSON Reference_ and a _Schema Object_ in the same object.  I think we could have some extra wording here making sure this is known to be the case, much like we document the `items` property as being applicable and required only when `type: array`.

versions/3.0.md
```diff
@@ -2287,25 +2284,32 @@ The following properties are taken directly from the JSON Schema definition and
...
+- description - [CommonMark syntax](http://spec.commonmark.org/) can be used for rich text representation.
+- format - See [Data Type Formats](#dataTypeFormat) for further details. While relying on JSON Schema's defined formats, the OAS offers a few additional predefined formats.
+- default - Unlike JSON Schema, the value MUST conform to the defined type for the Schema Object.
```
whitlockjc: You don't specifically tell what the type is, or its purpose.  Would it make sense to do this and then throw the existing content into a _"(Note: ...)"_?

versions/3.0.md
```diff
@@ -2323,7 +2327,9 @@ While composition offers model extensibility, it does not imply a hierarchy betw
 To support polymorphism, OpenAPI Specification adds the support of the `discriminator` field.
 When used, the `discriminator` will be the name of the property used to decide which schema definition is used to validate the structure of the model.
 As such, the `discriminator` field MUST be a required field.
-The value of the chosen property has to be the friendly name given to the model under the `definitions` property.
+There are are two ways to define the value of a discriminator for an inheriting instance.
```
whitlockjc: Without an example, this portion is unclear to me.

whitlockjc: Actually, I _might_ get this.  Are you saying that the value of the `discriminator` field must match the name of a definition?  Does this mean you can **only** use values that match keys of `#/components/definitions`?

versions/3.0.md
```diff
@@ -2323,7 +2327,9 @@ While composition offers model extensibility, it does not imply a hierarchy betw
...
+- Use the definition's name.
+- Override the definition's name by overriding the property with a new value. If exists, this takes precedence over the definition's name.
```
DavidBiesack: I'm not sure what this means. Override which property - the discriminator field (`petType` in this example)? If so, how is it overridden and specified? Would I do:

```YAML
  Cat:  ## "Cat" will be used as the discriminator value
    description: A representation of a cat
    allOf:
    - $ref: '#/definitions/Pet'
    - type: object
      properties:
         petType:
              type: string
              enum:
              - Cat
          huntingSkill:
             ...
```

An example would help.

versions/3.0.md
DavidBiesack: Good point -- we rely heavily on external schema files via `$ref`. For example, if Cat and Dog are defined externally, I should still be able to use them. However for a schema included from an external schema.json there is no name as there is for one defined in `/components/definitions`. If all three schema (Pet, Cat, Doc) were defined externally (so that they can be used in multiple OpenAPI documents) and there is no single `/components/definitions` in which they reside, how would this work?

webron: We'll be handling the `discriminator` in a separate PR.

webron:
Thanks @DavidBiesack - I'd still rather leave the wording regarding the JSON Schema intact, based on my experience with users in the past.

versions/3.0.md
darrelmiller: I think it is useful to remind people that even when embedding schema objects in `allOf` what is being embedded is not a regular JSON Schema.  I suspect many people don't realize that the schemas inside OpenAPI are not regular JSON Schemas.

darrelmiller:
Latest updates LGTM 🎈

versions/3.0.md
whitlockjc: Doesn't linking to a _Schema Object_ do this?

versions/3.0.md
```diff
@@ -2253,25 +2254,21 @@ $ref: 'Pet.yaml'
...
 #### <a name="schemaObject"></a>Schema Object
 
 The Schema Object allows the definition of input and output data types.
 These types can be objects, but also primitives and arrays.
-This object is based on the [JSON Schema Specification Draft 4](http://json-schema.org/) and uses a predefined subset of it.
-On top of this subset, there are extensions provided by this specification to allow for more complete documentation.
+This object is an extended subset of the [JSON Schema Specification Wright Draft 00](http://json-schema.org/).
 
-Further information about the properties can be found in [JSON Schema Core](http://json-schema.org/latest/json-schema-core.html) and [JSON Schema Validation](http://json-schema.org/latest/json-schema-validation.html).
+Further information about the properties can be found in [JSON Schema Core](https://tools.ietf.org/html/draft-wright-json-schema-00) and [JSON Schema Validation](https://tools.ietf.org/html/draft-wright-json-schema-validation-00).
```
whitlockjc: Here you call it _"JSON Schema Core"_ but you reference it earlier as _"JSON Schema Specification Wright Draft 00"_.  Make a decision and use it in both places.

versions/3.0.md
```diff
@@ -2287,25 +2284,33 @@ The following properties are taken directly from the JSON Schema definition and
...
+- additionalProperties - Value can be boolean or object. Inline or referenced schema MUST be of a [Schema Object](#schemaObject) and not a standard JSON Schema.
+- description - [CommonMark syntax](http://spec.commonmark.org/) can be used for rich text representation.
+- format - See [Data Type Formats](#dataTypeFormat) for further details. While relying on JSON Schema's defined formats, the OAS offers a few additional predefined formats.
+- default - The default value represents what would be assumed by the consumer of the input as the value of the schema if one is not provided. Unlike JSON Schema, the value MUST conform to the defined type for the Schema Object defined at the same level. For example, of `type` is `string`, then `default` can be `"foo"`` but cannot be `1`.
+
+Alternatively, any time a Scheme Object can be used, a [Reference Object](#referenceObject) can be used in place. This allows referencing definitions in place of defining them inline. Effectively, it is the `$ref` property in JSON Schema but follows the same guidelines and restrictions as the Reference Object. For example, additional properties alongside a `$ref` SHALL be ignored.
```
whitlockjc: Should _"in place"_ be _"in its place"_?

whitlockjc: The third sentence doesn't make sense to me.  What about mentioning the caveat about mixing other properties with `$ref` into the _Reference Object_ documentation and omitting the last two sentences?

versions/3.0.md
webron: JSON Schema is split into two specifications - Core and Validation. Can't keep it as one, and shouldn't reference both every single time.

whitlockjc: Fair enough.

whitlockjc:
:shipit:

webron:
:tada:

PR #897, opened 2017-02-21
Updated Header Object, removed Items Object

webron:
Updated the Header Object to point to the Parameter Object for an exact structure, but added a few changes. There's no reason not to follow the same new structure as we have for parameters (that was the case before pretty much). Content is not duplicated, but referenced with exceptions.

Modified examples in Header Object and Headers Object.

Completely removed the Items Object as the Header Object was the only one referencing it - not needed anymore.

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -1279,7 +1279,7 @@ encoding:
   profileImage:
     # only accept png/jpeg
     contentType: image/png, image/jpeg
-```fi
+```
 
 #### <a name="responsesObject"></a>Responses Object
 
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -740,7 +740,7 @@ A unique parameter is defined by a combination of a [name](#parameterName) and [
 There are four possible parameter types.
 * Path - Used together with [Path Templating](#pathTemplating), where the parameter value is actually part of the operation's URL. This does not include the host or base path of the API. For example, in `/items/{itemId}`, the path parameter is `itemId`.
 * Query - Parameters that are appended to the URL. For example, in `/items?id=###`, the query parameter is `id`.
-* Header - Custom headers that are expected as part of the request.
+* Header - Custom headers that are expected as part of the request. Note that [RFC 7230](https://tools.ietf.org/html/rfc7230#page-22) states header names are case insensitive.
 * Cookie - Used to pass a specific cookie value to the API.
 
 
@@ -1279,7 +1279,7 @@ encoding:
   profileImage:
     # only accept png/jpeg
     contentType: image/png, image/jpeg
-```
+```fi
 
 #### <a name="responsesObject"></a>Responses Object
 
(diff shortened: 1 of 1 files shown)


versions/3.0.md
```diff
@@ -1606,29 +1540,38 @@ Rate-limit headers:
 {
     "X-Rate-Limit-Limit": {
         "description": "The number of allowed requests in the current period",
-        "type": "integer"
```
DavidBiesack: I guess I missed the change that removed 'type' from Parameter Object and **requires** the use of schema. This makes an OAS document more verbose as seen in this example - one line becomes three when a parameter uses a primitive, which is by far the most common case (boolean, string, number)

I suggest we reconsider this.

darrelmiller: This was a tough call that we went back and forward on just a couple of weeks ago.  Eventually, we sided on using the schema object for consistency.  Yes it is more verbose, however making a clear delineation between what are parameter specific properties and what are schema related properties helps in understanding the behavior of a parameter.  

It also has the side benefit of simplifying the documentation and tooling implementation.

whitlockjc:
Other than removing the 4th _Parameter Object_ different for _Header Objects_, LGTM.

webron:
- Removed headers from being required.
- Added a note about header names being case insensitive.

I did not add a validation restriction saying that names SHOULD or MUST be case insensitive. My take is that tools can choose how to behave as producers and consumers. If you feel differently, please add the wording you'd like to be used.

versions/3.0.md
```diff
@@ -741,7 +740,7 @@ A unique parameter is defined by a combination of a [name](#parameterName) and [
 There are four possible parameter types.
 * Path - Used together with [Path Templating](#pathTemplating), where the parameter value is actually part of the operation's URL. This does not include the host or base path of the API. For example, in `/items/{itemId}`, the path parameter is `itemId`.
 * Query - Parameters that are appended to the URL. For example, in `/items?id=###`, the query parameter is `id`.
-* Header - Custom headers that are expected as part of the request.
+* Header - Custom headers that are expected as part of the request. Note that [RFC 7230](https://tools.ietf.org/html/rfc7230#page-22) states header names are case insensitive.
```
whitlockjc: For consistency, should we wrap the _"Note"_ in parentheses like in other places?

versions/3.0.md
```diff
@@ -1591,7 +1525,7 @@ myWebhook:
 
 
 #### <a name="headersObject"></a>Headers Object
-Lists the headers that can be sent as part of a response.
+Lists the headers that can be sent as part of a response. Note that [RFC 7230](https://tools.ietf.org/html/rfc7230#page-22) states header names are case insensitive.
```
whitlockjc: Same as above.

whitlockjc:
Other than minor nit about wrapping the _"Note that"_ into a _"(Note: ...)"_, LGTM.

whitlockjc:
:shipit:

fehguy:
All good from my POV, merge when others have approved. 👍

darrelmiller:
:shipit:

webron:
Merging! 🎉

PR #901, opened 2017-02-21
added version clarification

fehguy:
Partially addresses concern in #589

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -74,12 +74,13 @@ Additional utilities can also take advantage of the resulting files, such as tes
 ## Revision History
 
 Version | Date | Notes
---- | --- | ---
-2.0 | 2015-12-31 | Donation of Swagger 2.0 to the Open API Initiative
-2.0 | 2014-09-08 | Release of Swagger 2.0
-1.2 | 2014-03-14 | Initial release of the formal document.
-1.1 | 2012-08-22 | Release of Swagger 1.1
-1.0 | 2011-08-10 | First release of the Swagger Specification
+---   | --- | ---
+3.0.0 | 2017-02-28 | Implementor's draft of the 3.0 specification
+2.0   | 2015-12-31 | Donation of Swagger 2.0 to the Open API Initiative
+2.0   | 2014-09-08 | Release of Swagger 2.0
+1.2   | 2014-03-14 | Initial release of the formal document.
+1.1   | 2012-08-22 | Release of Swagger 1.1
+1.0   | 2011-08-10 | First release of the Swagger Specification
 
 ## Definitions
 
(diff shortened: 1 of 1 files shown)


webron:
LGTM.

darrelmiller:
:shipit:

whitlockjc:
:shipit:

versions/3.0.md
```diff
@@ -221,7 +222,7 @@ Field Name | Type | Description
 <a name="infoTermsOfService"></a>termsOfService | `string` | A URL to the Terms of Service for the API.
 <a name="infoContact"></a>contact | [Contact Object](#contactObject) | The contact information for the exposed API.
 <a name="infoLicense"></a>license | [License Object](#licenseObject) | The license information for the exposed API.
-<a name="infoVersion"></a>version | `string` | **Required** Provides the version of the application API (not to be confused with the specification version).
+<a name="infoVersion"></a>version | `string` | **Required** Provides the version of the application API (not to be confused with the specification version).  While not required, is it suggested that the `info.version` value correspond to the version of the API definition.  In practice, the version of the API _implementation_ may evolve at an entirely different rate.
```
DavidBiesack: `version` is marked **Requried**, then the second sentence starts "While not required". Some may be confused and think this starts a contradiction. I suggest not adding this sentence at all, as it only raises additional questions (is "application API === "API definition"?) Instead, I suggest rephrasing, introducing fewer new/undefined/ambiguous terms:

**Required** Provides the version of the  API (not to be confused with the [_OpenAPI specification version_](#openapi-version-string) or the _implementation_ version which may evolve at different rates).

versions/3.0.md
```diff
@@ -74,12 +74,13 @@ Additional utilities can also take advantage of the resulting files, such as tes
...
 
 Version | Date | Notes
---- | --- | ---
-2.0 | 2015-12-31 | Donation of Swagger 2.0 to the Open API Initiative
-2.0 | 2014-09-08 | Release of Swagger 2.0
-1.2 | 2014-03-14 | Initial release of the formal document.
-1.1 | 2012-08-22 | Release of Swagger 1.1
-1.0 | 2011-08-10 | First release of the Swagger Specification
+---   | --- | ---
+3.0.0 | 2017-02-28 | Implementor's draft of the 3.0 specification
```
RobDolinMS: I suggest we follow a model of calling this v3.0.0-rc0 (release candidate zero) and then when we have polished, we're at v3.0.0-rc1 (or rc2 or rc3) and then we have a v3.0.0-final (and can delete the release candidates)

versions/3.0.md
RobDolinMS: @DavidBiesack Would you please considering opening this as a separate issue or PR?

fehguy:
@OAI/tdc I've updated the version to be `3.0.0.-rc0` per call.

versions/3.0.md
DavidBiesack: @RobDolinMS - sure, see #905

PR #902, opened 2017-02-22
Fix invalid JSON to use dpouble quotes; use "name": consistently in JSON

DavidBiesack:
One JSON example was using single quotes around a string item ; changed to double quotes.

Also made spacing consistent in JSON examples to all use `"name": value` rather than a mix of no space or a space before the colon.

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -938,7 +938,7 @@ A request body with a referenced model definition.
       "examples": [ "http://foo.bar/examples/user-example.txt" ]
     },
     "*": {
-      "schema": {
+      "example": {
         "$ref": "http://foo.bar/examples/user-example.whatever"
       }
     }
@@ -962,7 +962,7 @@ content:
     examples:
       - 'http://foo.bar/examples/user-example.txt'
   '*':
-    schema:
+    example:
       $ref: 'http://foo.bar/examples/user-example.whatever'
 ```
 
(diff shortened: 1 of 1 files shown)


versions/3.0.md
```diff
@@ -921,35 +921,29 @@ A request body with a referenced model definition.
...
-      "examples": [
-        { "$ref": 'http://foo.bar#/examples/address-example.txt'}
-      ]
+    "text/plain": {
+      "examples": [ "http://foo.bar/examples/user-example.txt" ]
     },
     "*": {
-      "example": {
-        $ref: "http://foo.bar#/examples/address-example.whatever"
+      "schema": {
```
webron: This is actually supposed to be `example`.

darrelmiller: and the * should be */*

DavidBiesack: I replaced this JSON completely, generating it from the YAML (because the two did not match; one used foo.bar# and the other foo.bar. I'll also correct the YAML
```

  '*':
    schema:
      $ref: 'http://foo.bar/examples/user-example.whatever'
```
to
```

  '*':
    example:
      $ref: 'http://foo.bar/examples/user-example.whatever'
```
with the update

DavidBiesack:
ok, thanks @webron -- PR updated

DavidBiesack:
I just saw @darrelmiller 's comment, so I've fixed `*` to `*/*` as well. Thanks for the catch, @darrelmiller . Now we just need a GFM hook to auto-convert JSON <--> YAML so we don't have to manually code both formats in all our examples. :smiley_cat:

PR #905, opened 2017-02-22
Remove ambiguity about info.version

DavidBiesack:
In [OpenAPI.next versions/3.0.md](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#infoObject) `info.version` is marked **Requried**, then the second sentence starts "While not required, ". Some may be confused and think this starts a contradiction. I suggest not adding this sentence at all, as it only raises additional questions (is "application API === "API definition"?) I propose more concise rephrasing.

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -221,7 +221,7 @@ Field Name | Type | Description
 <a name="infoTermsOfService"></a>termsOfService | `string` | A URL to the Terms of Service for the API.
 <a name="infoContact"></a>contact | [Contact Object](#contactObject) | The contact information for the exposed API.
 <a name="infoLicense"></a>license | [License Object](#licenseObject) | The license information for the exposed API.
-<a name="infoVersion"></a>version | `string` | **Required** The version of the API.  Is it suggested that the `info.version` value correspond to the version of the API definition (not to be confused with the [OpenAPI specification version](#oasVersion).) In practice, the version of the API _implementation_ (not specified here) may evolve at an entirely different rate.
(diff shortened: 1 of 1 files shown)


RobDolinMS:
Thanks for opening this @DavidBiesack.  

I'm tagging @fehguy to take a look.  I'm not sure if the original text was a typo (required ... not required) or was explaining that the versions were not required to be consistent.

darrelmiller:
I feel the revised text is a little too concise.  The point I was hoping to make here was that the version property identifies the API Definition, not the API implementation.

fehguy:
Yes, as @darrelmiller stated, the `version` value _is required_.  It is suggested but not required that the value of the `version` is the version of the API definition as opposed to the API implementation.

DavidBiesack:
I've suggested another revision, which I think addresses all the comments. Note: I understand version is *required*; I am suggesting removing the clause "While not required, " which would add confusion. Proposed:

**Required** The version of the API definition (which is distinct from the OpenAPI specification version or the API implementation version which may evolve at different rates.)

webron:
I find that even more confusing - it suggests that the OAS version and API implementation may evolve at different rates (different than the original intent).

DavidBiesack:
but the OAS version and the API implementation version *can* evolve at different rates. The API definition may be version: "v2", and OAS may be version "2.0" then the document updated to OAS "3.0" in april, then in June the implementation version may change to 2.1.0 to 2.1.1 to 2.1.2 to 2.2.0

DavidBiesack:
I'm happy to simplify:

**Required** The version of the API definition (which is distinct from the OpenAPI specification version or the API implementation version.)

webron:
@DavidBiesack if you can use your last suggested wording, we'll merge. Thanks.

DavidBiesack:
updated, thx

PR #911, opened 2017-02-24
Clients should refrain from usage of the declared operation.

paulhill:
Grammar: Clients should refrain from usage of the declared operation.


webron:
That's slightly tricky. While `deprecated` is a general documentation feature for consumers (which I prefer to clients), I'm wondering if it can also have an impact on producers when it comes to code generation.

PR #914, opened 2017-02-24
ordering typo in example?

paulhill:
Not sure if this is a typo or I'm just a bit lost and it's possible to nest the media type like that


darrelmiller:
It was a type.  Thanks for the PR.

PR #918, opened 2017-02-24
[File Structure] Clarify name convention is optional

RobDolinMS:
Liz Lynch found wording confusing.  

Use "MAY" to indicate this is optional.

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -139,7 +139,7 @@ The OAS representation of the API is made of a single file.
 However, parts of the definitions can be split into separate files, at the discretion of the user.
 This is applicable for `$ref` fields in the specification as follows from the [JSON Schema](http://json-schema.org) definitions.
 
-By convention, the OpenAPI Specification (OAS) file is named `openapi.json` or `openapi.yaml`.
+By convention, the OpenAPI Specification (OAS) file MAY be named `openapi.json` or `openapi.yaml`.
 
 ### Data Types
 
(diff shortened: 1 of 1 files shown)


webron:
I'm leaning towards SHOULD rather than MAY.

RobDolinMS:
I'm hesitant on "SHOULD" since it sometimes is used to suggest an action should be done unless there is a good reason not to.  

Other opinions?

webron:
I see it as our recommendation to do that. We don't really dive into discovery much, and as a standard way I think we need to encourage people to do that.

Using MAY feels like with might as well not add that suggestion at all (especially if you read the full sentence). Saying that by convention something MAY be named something to me sounds odd.

ePaul:
If we recommend something like this, we should have a reason.
What is the reason for having a conventional name of a API specification file?

* Allowing only one such file per directory in a file system? (Why would you want to prescribe something like this?)
* Having it reachable in the root of the API it describes? (Then it is not so much about the file name, but about a well known URL path component.)
* any other reason?

RobDolinMS:
Per TDC discussion, let's use "RECOMMENDED"

ePaul:
I still would like to know the reason for recommending a specific name.

RobDolinMS:
I've updated this PR per TDC discussion to use "RECOMMENDED" /cc @ePaul 

@webron If this looks good, please merge.

webron:
@ePaul without diving into details and complicating things - for better or worse, this is how the discovery of the spec has worked so far, and we're too late in the game to change it for 3.0. We recommend doing so, but it's been the common practice until now and people should be familiar with it. It may not be a _technical_ reason, but a reason still. We can open discussion for changes to that in the next major version (and I believe there's a ticket).

PR #921, opened 2017-02-27
Updated keywords

webron:
Updated the case of the keywords; MUST, SHOULD, MAY, SHALL.


webron:
Merging this as a cosmetic change. We can revert if needed.

PR #926, opened 2017-02-27
[Revision History] Move to end of document

RobDolinMS:
Signed-off-by: Rob Dolin <robdolin@microsoft.com>

PR #928, opened 2017-02-27
[ToC] Update to use internal anchors

RobDolinMS:
This updates the ToC through the [Info Object](#infoObject) and its children to use the internal anchors in the document (instead of GitHub-generated anchors)


RobDolinMS:
This change also includes moving revision history to the end of the document (#926)

PR #932, opened 2017-02-28
Fix multipart encoding example

raymondfeng:
This PR fixes the multipart encoding example based on https://gist.github.com/fehguy/fc04aa2420a89b5c9f625c55181f87f1#file-requestbody-yaml-L106. The encoding object is also listed as an optional child element for `contentTypeObject`.


fehguy:
LGTM

PR #934, opened 2017-02-28
Content-Type consistency

earth2marsh:
Needs a second set of eyes!

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -760,7 +760,7 @@ Field Name | Type | Description
 <a name="parameterExamples"></a>examples | [[Example Object](#exampleObject)] | Examples of the content type.  Each example in the Examples array SHOULD be in the correct format as specified parameter encoding.  The `examples` object is mutually exclusive to the `example` object.  Furthermore, if referencing a `schema` which contains an example, the `examples` value SHALL _override_ the example provided by the schema.
 <a name="parameterExample"></a>example | [Example Object](#exampleObject) | Example of the content type.  The example object SHOULD be in the correct format as specified in the parameter encoding.  The `example` object is mutually exclusive to the `examples` object.  Furthermore, if referencing a `schema` which contains an example, the `example` value SHALL _override_ the example provided by the the schema.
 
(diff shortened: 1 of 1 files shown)


versions/3.0.md
```diff
@@ -1080,21 +1080,21 @@ content:
...
-<a name="contentTypeExample"></a>example | [Example Object](#exampleObject) | Example of the content type.  The example object SHOULD be in the correct format as specified in the _content_ type.  The `example` object is mutually exclusive to the `examples` object.  Furthermore, if referencing a `schema` which contains an example, the `example` value SHALL _override_ the the example provided by the schema.
+<a name="contentTypeExamples"></a>examples | [Examples Array](#examplesArray) | Examples of the Content-Type.  Each example in the Examples array SHOULD be in the correct format as specified in the media type.  The `examples` object is mutually exclusive to the `example` object.  Furthermore, if referencing a `schema` which contains an example, the `examples` value SHALL _override_ the example provided by the schema.
```
darrelmiller: I don't really like the use of `Content-Type` here as it makes me think that it is an example of the value in a `Content-Type` header.

versions/3.0.md
```diff
@@ -1080,21 +1080,21 @@ content:
...
+<a name="contentTypeExample"></a>example | [Example Object](#exampleObject) | Example of the Content-Type.  The example object SHOULD be in the correct format as specified in the media type.  The `example` object is mutually exclusive to the `examples` object.  Furthermore, if referencing a `schema` which contains an example, the `example` value SHALL _override_ the the example provided by the schema.
```
darrelmiller: As above.

earth2marsh: works for me. Updated.

earth2marsh:
Addressed @darrelmiller 's comments.

darrelmiller:
I added #943 to rename Content Type Object, so there is a fair amount of overlap with this PR so we need to reconcile the changes in the two if my PR is accepted.

darrelmiller:
These changes were incorporated into PR #943

PR #938, opened 2017-02-28
[Path Item Object Example] Match field order

RobDolinMS:
In the table describing the fields, the order is summary then description.  

This PR updates the examples to match that order.


webron:
The order is not important, and we shouldn't present it that way. I actually think it's better to have examples here and there that don't follow it.

RobDolinMS:
@webron If we want to show how order varies, I'd suggest a specific illustration of this; potentially near the front of the spec.  This variation read like a minor error to me rather than something intentional.

PR #943, opened 2017-02-28
Rename Content Type Object to Media Type Object

darrelmiller:
Now that Encoding has a Content-Type property there is some confusion when using the term `Content-Type` within the spec.  Using the term `Media Type Object` to describe the child objects of a `Content-Object` is consistent with the use of media type identifiers as the keys.


webron:
:shipit:

PR #944, opened 2017-02-28
Updated server templates, all things relative

webron:
- Update server templates, changed naming convention to variables
- Fixed relevant property definitions
- Updated anchors
- Some clarifications re relative URLs

🦄


darrelmiller:
🍒 🍭

PR #952, opened 2017-03-01
update version

webron:
well…


darrelmiller:
Are you sure?

webron:
We _can_ hijack it and call it version 4.0. I'm sure nobody would notice.

darrelmiller:
:shipit:

Issue #956, opened 2017-03-02
3.0.md links to LinkObject are broken


willdady:
As per the title. I had a quick look but can't figure out why it's not working.

`#linksObject` ->  `#links-object` Works fine.
`#linkObject` ->  `#link-object` Doesn't work.

ePaul:
@willdady Thanks for noting. I opened pull request #966 to fix this.

fehguy:
I think this can be closed

PR #959, opened 2017-03-03
Fix a few missing entries and a broken link in the Table of Contents.

timburks:
I noticed a few entries were missing in the TOC, so I did a quick scan to pull out the link targets: 

```grep "#### <a name" 3.0.md | grep -v "#####" | sed -e "s/^.*\"\(.*\)\".*$/\1/"``` 

That gave me a list of objects which I used to update the TOC. I also noticed that the linkObject anchor was "#linkObject", which seemed to break the TOC link in the current doc.


ePaul:
The `#` fixes #956 (I just some minutes ago opened #966 for this, with the same change, not having seen this PR).

I checked the new links in the TOC, all work.
:+1:

PR #964, opened 2017-03-06
Small grammar nits

MikeRalphson:
'Like such' reads badly, and add missing 'is' to 'it at the'. Fixed 3 double 'the the's (thanks @willdady) and reword a clunky negative clause.


ePaul:
:+1:

versions/3.0.md
```diff
@@ -2020,7 +2020,7 @@ href: 'http://colors.my-server.com/colors/green'
 href: 'http://colors.my-server.com/colors/blue'
 ```
 
-As with all links, it at the the clients' discretion to follow them, and permissions and the ability to make a successful call to that link is not guaranteed solely by the existence of a relationship.
+As with all links, it is at the the clients' discretion to follow them, and permissions and the ability to make a successful call to that link is not guaranteed solely by the existence of a relationship.
```
willdady: There is also a double 'the the'.

MikeRalphson: @willdady good eyes. In fact I think the whole sentence needs rewording. Was looking too microscopically.

PR #968, opened 2017-03-07
Missing capitalisation of 'must'

MikeRalphson:


ePaul:
If you are at it, could you fix the other occurrence of `must` in the document (in the OpenAPI object's field table, tags).

MikeRalphson:
@ePaul if you mean

> Not all tags that are used by the [Operation Object](#operationObject) must be declared

I think that "not all .. must be declared" isn't a case of MUST or MUST NOT. **But** it may be better to reword this as

> Not all tags that are used by the [Operation Object](#operationObject) need to be declared

to avoid the ambiguity.

ePaul:
You are right, I somehow didn't read the full sentence. Your rewording seems fine.

MikeRalphson:
Happy to fold in the other change if required.

PR #991, opened 2017-03-14
[master] Validate (not lint) markdown files

MikeRalphson:
Checks for broken links and missing img alt tags ~~and fenced code-blocks with no defined language~~.

This change also updates the Node.js version used on Travis from 0.10.36 which is now unsupported, to a LTS version (6.x)

Issues identified have been fixed.


webron:
I'd rather keep the changes to 1.2 and 2.0 to a minimum. The broken links don't change any of the meaning of the spec and are not 100% required. We'll discuss that and decide if we want to make these changes.

MikeRalphson:
@webron sure, there is an argument that if no-one's noticed up till now, they're not needed. I was just thinking ahead for when v3 gets merged into master.

webron:
Yeah, I like the idea of adding it to 3.0 to avoid such issues from now on. Thanks for the pulling this together.

RobDolinMS:
@MikeRalphson - It looks like there is now a conflict.  Would you be able to resolve?

MikeRalphson:
@RobDolinMS done.

webron:
Sorry for the trouble, @MikeRalphson. I'd rather not merge this for now given the reasons above. Appreciate the work though. We might revisit it in the future and could just reopen and merge.

PR #992, opened 2017-03-14
[OpenAPI.next] validate (not lint) versions/3.*.md with mdv

MikeRalphson:
Checks for broken links and images with missing alt tags ~~and fenced code-blocks with no defined language~~.

This change also updates the Node.js version used on Travis from 0.10.36 (which is now unsupported) to a LTS version (6.x)

Issues identified have been fixed.


webron:
I don't know how, but it looks like a bunch of things got messed up when you look at the rendered view.
For example:
- https://github.com/MikeRalphson/OpenAPI-Specification/blob/11af30e39d1319d2d55828333cedcaec7f3f6dd4/versions/3.0.md#fixed-fields-13
- https://github.com/MikeRalphson/OpenAPI-Specification/blob/11af30e39d1319d2d55828333cedcaec7f3f6dd4/versions/3.0.md#patterned-fields-4

webron:
Can the check to add ```text be configured and removed? It's not really needed and just adds noise.

MikeRalphson:
@webron thanks for the heads-up. The raw diff looked ok. Will rework,

MikeRalphson:
@webron commits reworked but table-formatting issues still remain.

My theory is that GitHub have recently changed their markdown renderer and the `<span>&#124;</span>` hack in table cells no-longer works.

You can see the same problem in an unrelated PR by @willdady [here](https://github.com/willdady/OpenAPI-Specification/blob/70f0b65289b5b6e6abc00b373aa8cc6ea0f47662/versions/3.0.md#fixed-fields-6).

The following markdown

```markdown
Column A|Column B|Column C|Overflow
---|---|---|---
Cell A|Cell <span>&#124;</span>B|Cell C
```

renders as:

Column A|Column B|Column C|Overflow
---|---|---|---
Cell A|Cell <span>&#124;</span>B|Cell C

One option is to use another Unicode bar character:

Column A|Column B|Column C|Overflow
---|---|---|---
Cell A|Cell <span>&#166;</span>B|Cell C

webron:
Thanks. I'll look into it.

webron:
@MikeRalphson - I should have merged the PRs in a different order, but alas... I'm trying to manually resolve issues for some PRs but would you mind resolving these yourself?

webron:
@MikeRalphson I see you're pushing changes, lmk when done.

MikeRalphson:
@webron those four conflicts (between fixed links and fixed table formatting) are resolved.

Issue #880, opened 2017-02-05
clarify regex format


fehguy:
I know it's from JSON schema, but we should make it clear in the docs that we're supporting [ECMA-262](http://www.ecma-international.org/publications/standards/Ecma-262.htm) as the regex format in the spec.

PR #994, opened 2017-03-15
Fix the component name regular expression

MikeRalphson:
Currently the expression given matches everything

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -411,7 +411,7 @@ Field Name | Type | Description
 
 This object can be extended with [Specification Extensions](#specificationExtensions). 
 
-All the fixed fields declared above are objects that MUST use keys that match the regular expression: `[a-zA-Z0-9.\-_]+`.
+All the fixed fields declared above are objects that MUST use keys that match the regular expression: `^[a-zA-Z0-9\.\-_]+$`.
 
 Field Name Examples:
 


darrelmiller:
Thanks for catching this.  However, I don't think what you have is quite what I intended either.
I think we want `^[a-zA-Z0-9\.\-_]+$`

The tool I was using didn't require the period to be escaped, so I didn't.  The slash prior to the hyphen was  an escape character, not an allowance of slashes!

MikeRalphson:
Excellent, the absence of `\\` makes the regex easy enough to read without needing an explanatory text I think. Commit amended.

webron:
Wait, which regex variant do you use? ;)

darrelmiller:
@webron I use this tool http://ultrapico.com/Expresso.htm which uses .net's regex engine :-)

webron:
I'd just suggest we keep it to http://json-schema.org/latest/json-schema-validation.html#rfc.section.5.8 compliance across the doc (not that we have that many regular expressions in it, thankfully).

MikeRalphson:
@webron @darrelmiller it would be nice to be able to state that regular expressions SHOULD be interpreted as per ECMA 262 (Perl-5 style). But the reality is most tooling will use their native regexp parser, which may be POSIX BRE for example.

webron:
See https://github.com/OAI/OpenAPI-Specification/issues/880.

If tooling will choose to use a different regex parser, they'd be non-compliant. Since we agreed our standard is based on JSON Schema, it is what it is. Can't guarantee cross-platform functionality without it.

FWIW, we've had similar issues in the past in some tooling and but us and the community ended up fixing those.

ePaul:
@MikeRalphson I think for expressions in the specification document (i.e. static ones), tool implementation authors can translate them easily to whatever their library supports. So we are free to choose whatever we want here, as long as we make it clear what it is.

For expressions appearing in an API definition document (i.e. in schemas), it is not so easy ... but we still should make clear in the spec which regexp flavor we want there (the same as JSON schema, I guess). If implementations don't support this, they will be not fully functional then.

PR #1016, opened 2017-03-28
Remove trailing spaces

ePaul:
This removes trailing spaces in the whole document.

~~(This PR includes the changes from #1015, because there are conflicts due to changes on neighboring lines. Merging this one first would automatically merge #1015, merging #1015 first will make the diff here clearer to contain only the whitespace changes. If #1015 is rejected, I can rebase this to not include those changes.)~~

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -97,7 +97,7 @@ Some examples of possible media type definitions:
   application/vnd.github.v3.patch
 ```
 ##### <a name="httpCodes"></a>HTTP Status Codes
-The HTTP Status Codes are used to indicate the status of the executed operation. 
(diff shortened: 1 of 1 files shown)


webron:
@ePaul - are you trying to make me suffer? why do we need this?

ePaul:
*Need* may be a too hard word. I just don't like superfluous spaces at the end of lines. (I started to note this when copying some examples, and then applied a regular expression to the whole document.)

Feel free to reject, I'll not be offended if you decide not to merge this.

webron:
@ePaul - Thanks for the work but we're not going to merge this. In the long run, future changes by different people may end up adding back some trailing spaces and it's pretty much a losing battle.

PR #1027, opened 2017-04-04
[master/2.0] Replace <span>&#124;</span> with \|

MikeRalphson:
Whatever cache was holding onto the previously-rendered html of 2.0.md has expired.


webron:
Thanks @MikeRalphson.

Issue #488, opened 2015-10-20
Allow multiple examples with optional label


elmolicious:
Hi, for some routes we need multiple examples for a "Response Object" as well as multiple body payload examples in the "Parameter Object".

e.g: our API route returns an object with optional properties and we want to show 2 examples how it could look like ...

``` json
...
{
    ...
    "examples" : [
        {
            "application/json" :
                {
                    "property" : "value"
                }
        },
        {
            "application/json" :
                {
                     "property" : "value",
                     "optional_property" : "value"
                }
        }
    ]
    ...
}
```

In addition we would like to name/label these examples .
In the specification "Example Object"s you can specify different representations of an example for a given MIME-type. I would suggest adding either an optional "name" or "label" property to the "Example Object" which is a string (maybe even an object to allow localized string values for different languages) or to allow custom properties (e.g x-name).

e.g: 

``` json
...
{
    ...
    "examples" : [
        {
            "application/json" :
              {
                    "property" : "value"
              },
            "name" : "Default Response"
        },
        {
            "application/json" :
                {
                     "property" : "value",
                     "optional_property" : "value"
                },
            "name" : "Response if state of ressource ..."
        }
    ]
    ...
}
```

If there is any good way to solve it with the current swagger specification, i would love to hear it and it would make this feature request obsolete.

Thanks.

mkrufky:
+1  I also have an endpoint that can be used in different ways.  The `required` and `optional` argument fields can vary depending on the function being carried out.  It would be great to be able to represent the same endpoint multiple times, specifying different arguments (and response body), rather than having to represent it all at once, marking all arguments as `optional`

arno-di-loreto:
+1 having a description in addition to the name would be great too
All this could like like this maybe:
```YAML
examples:
  - name: An first example
    description: use case
    content:
      - application/json : <example for media type application/json>
      - application/xml: <example for media type application/xml>
```
That could be useful for documentation but I think also for mock API generation

arno-di-loreto:
Based on how [Example Object](https://github.com/OAI/OpenAPI-Specification/blob/3.0.0-rc0/versions/3.0.md#exampleObject) are defined and used in version 3.0.0-rc0, I think that the it could be slightly modified to provide documentation data this way:

Current Example Object is  *a freeform object, array or primitive value*.

Modified Example Object is an Object with these properties:
- summary (String) : An optional short description or name
- description: (String): An optional longer description 
[...]

PR #1028, opened 2017-04-04
Update Examples Object

DavidBiesack:
Clarify that Examples Object may be used instead of Example Object.

Show JSON examples as well as YAML

Clarify the $ref values used in examples to remove ambiguity.
These are not references to external JSON/XML/TXT resources, but
rather are references to reusable examples in the current OpenAPI definition
or in an external OpenAPI definition's components/examples

This does not yet address issue #488 but cleans up Examples Object
a bit, which is needed whether we alter the OAS as per #488 .

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -1128,20 +1128,20 @@ A request body with a referenced model definition.
       "schema": {
         "$ref": "#/components/schemas/User"
       },
-      "examples": [ "http://foo.bar/examples/user-example.json" ]
+      "examples": [ "http://foo.bar/other-api.yaml#/components/examples/user-example.json" ]
     },
     "application/xml": {
       "schema": {
         "$ref": "#/components/schemas/User"
       },
-      "examples": [ "http://foo.bar/examples/user-example.xml" ]
+      "examples": [ "http://foo.bar/other-api.yaml#/components/examples/user-example.xml" ]
     },
     "text/plain": {
-      "examples": [ "http://foo.bar/examples/user-example.txt" ]
+      "examples": [ "http://foo.bar/other-api.yaml#/components/examples/user-example.txt" ]
     },
     "*/*": {
       "example": {
-        "$ref": "http://foo.bar/examples/user-example.whatever"
+        "$ref": "http://foo.bar/other-api.yaml#/components/examples/user-example.whatever"
       }
     }
   }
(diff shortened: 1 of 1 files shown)


versions/3.0.md
```diff
@@ -1128,20 +1128,20 @@ A request body with a referenced model definition.
       "schema": {
         "$ref": "#/components/schemas/User"
       },
-      "examples": [ "http://foo.bar/examples/user-example.json" ]
+      "examples": [ "http://foo.bar/other-api.yaml#/components/examples/user-example.json" ]
```
whitlockjc: I assume `user-example.json` is a property name for `examples`?

DavidBiesack: yes. This was carried over from the original, but I can change the names if that will help avoid confusion.

whitlockjc: The naming is definitely confusing, to me at least..  `user-example.json` reads like a file name.

DavidBiesack: ok, good feedback - I've updated the PR, and also added a missing JSON block, and also named the `/components/example/` items more uniformly. 

I also fixed some cases where the spec
had `/components/schemas/pet` in some places, `/components/schemas/Pet` in others.

versions/3.0.md
```diff
@@ -2355,7 +2355,12 @@ description: Pets operations
 
 #### <a name="examplesObject"></a>Examples Object
 
-Anywhere an `example` may be given, a JSON Reference MAY be used, with the 
+Instead of a single `example`, you may provide an array of 
+[Example Object](#exampleObject) values named `examples'.
+`example` and `examples` are mutually exclusive.
+
+Anywhere an `example` may be given (including in the Examples Object array), 
+a JSON Reference MAY be used, with the
```
whitlockjc: For consistency, we should use `Reference Object` _(linkfied)_ instead of `JSON Reference`.

versions/3.0.md
```diff
@@ -2365,42 +2370,147 @@ for the value that it is accompanying.  Tooling implementations MAY choose to
 validate compatibility automatically, and reject the example value(s) if they 
 are not compatible.
 
+Here is a single `example` in a model [Schema Object](#schemaObject):
```
whitlockjc: Minor nit, is _"single"_ necessary here?  Basically, you're saying _"Here is an `example` within a model"_.  Thoughts?

versions/3.0.md
```diff
@@ -2365,42 +2370,147 @@ for the value that it is accompanying.  Tooling implementations MAY choose to
...
+    identity:
+      properties:
+        name:
+          type: string
       example:
-        $ref: http://foo.bar#/examples/name-example
+        $ref: "#/components/examples/identity"
+```
+
+Below is an `examples` array in a [Request Body Object](#requestBodyObject),
```
whitlockjc: Same as above, you've already documented `examples` to be an _array_ so it seems unnecessary.

versions/3.0.md
```diff
@@ -2365,42 +2370,147 @@ for the value that it is accompanying.  Tooling implementations MAY choose to
...
+showing two in-line examples and references
```
whitlockjc: _"in-line"_ should be _"inline"_.

whitlockjc:
Thanks for being patient with the nitpicks.  :)

versions/3.0.md
DavidBiesack: will do

versions/3.0.md
DavidBiesack: I'm going to convert it to `examples` (and drop "single) since this is the **Examples Object** section, not  **Example Object**.

versions/3.0.md
DavidBiesack: I'm also going to move this paragraph to *Example Object* which is a better place for it.

versions/3.0.md
DavidBiesack: will do

DavidBiesack:
Anyone willing to review and merge this (before the text has diverged too much making merge difficult)?

DavidBiesack:
May not be needed if the proposal in #1035 is accepted.

DavidBiesack:
Closing - #1035 addressed the changes in this PR

PR #1033, opened 2017-04-10
Updates to rc1

curtisr7:
While reviewing the spec I came across a number of small issues:

* ~~There was a ton of extra trailing whitespace scattered throughout the spec. My ide cleans that up automatically, figured it might be a good addition?~~
* Cleaned up a few tabs 
* `OAuth Flows Object` was missing a space and made it so the markdown wouldn't render.
* Fixed the `JWT Bearer` json sample so it mirrored the yaml sample
* Made a couple small grammar fixes.
* ~~`ResponseObject.responseHeader` is an array of header objects, so added brackets around the type.~~

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -163,7 +163,7 @@ password | `string` | `password` | Used to hint UIs the input needs to be obscur
 ### <a name="relativeReferences"></a>Relative References in URLs
 
 Unless specified otherwise, all properties that are URLs MAY be relative references as defined by [RFC 3986](https://tools.ietf.org/html/rfc3986#section-4.2).
-Relative references are resolved using the URLs defined in the [`Server Object`](#serverObject) as a Base URI.
+Relative references are resolved using the URLs defined in the [`Server Object`](#serverObject) as a base URI.
 
 Relative references used in `$ref` are processed as per [JSON Reference](https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03), i.e. using the URL of the current document as the base URI. See also the [Reference Object](#referenceObject).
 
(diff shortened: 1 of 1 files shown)


versions/3.0.md
```diff
@@ -20,51 +20,51 @@ Additional utilities can also take advantage of the resulting files, such as tes
...
+    - [Schema Object](#schemaObject)
+    - [XML Object](#xmlObject)
+    - [Security Scheme Object](#securitySchemeObject)
+    - [Scopes Object](#scopesObject)
+    - [Security Requirement Object](#securityRequirementObject)
+  - [Specification Extensions](#specificationExtensions)
+  - [Security Filtering](#securityFiltering)
 - [Appendix A: Revision History](#revisionHistory)
-	
+
```
darrelmiller: I'm not sure what the intent of these changes are.  Were these done automatically by the IDE?

curtisr7: Take a look at the commit message. Removing unnecessary whitespace and replacing tabs with spaces.

darrelmiller: Did the IDE do this?  Two trailing spaces in a markdown line is significant.  I would rather not have tools auto removing trailing whitespace.

curtisr7: Yes, the IDE did this... and I wasn't aware that trailing spaces in markdown mean <br>.

darrelmiller:
The headers object is in responses is not an array, it is a map.  The use of the word "lists" is misleading and that should probably be fixed.

Having semantically significant changes included in a PR that also has a significant number of whitespace removals make reviewing changes much harder than it needs to be.

curtisr7:
Appending `?w=1` to the url will remove whitespace changes... but I'll go ahead and remove that commit.

versions/3.0.md
darrelmiller: According to the rich diff, the changes caused some indentation that was previously not there.   Was that intentional?
image

curtisr7: That was unintentional. The problem is that most of Table of Contents is tab spaced, but server variable / server variables is using spaces. Odds are whomever put those two items in were trying to get the list to line up properly and the were using spaces where the rest of the list is using tabs. I'll fix that up also.

darrelmiller: Getting rid of the tabs would be great.

curtisr7:
@darrelmiller I pushed out a new batch of commits this morning to address your comments from yesterday.

darrelmiller:
Looks good, although the OAuth Flows Object Fixed Fields table still broken.

ePaul:
I have #1016 for just the whitespace changes.

curtisr7:
@darrelmiller sorry about being slow here. Was pulled elsewhere.

> OAuth Flows Object Fixed Fields table still broken.
Fixed

darrelmiller:
Changes were made in other more focused PRs.

PR #1035, opened 2017-04-11
Updated Examples

darrelmiller:
This proposal attempts to address the issues raised in #488 

The "Example Object" has been updated.  This object now supports the following scenarios:
    a) multiple examples are required,
    b) when short `summary` or long `description` is needed, or 
    c) when an example needs to be reused using a reference object.

```yaml
examples:
  cat:
    summary: Big cat example
    value:
      name: Puma
      id: 1
  dog:
    summary: Family Dog
    description: This is the story about Ferguson the dog ...
    value:
      name: Ferguson
      id: 2
  frog:
    $ref: "#/components/examples/frog"
```

For simpler scenarios where only one example is required, you can inline a single example.

```yaml
example:
    name: Ferguson
    id: 2
```

For scenarios where the example is for a media type that is not JSON or YAML the value can be externalized using the `externalValue` property instead of the `value` property.  The value retrieved when dereferencing this URL must match the media type/ encoding associated with described example object.

```yaml
  'application/xml':
    examples:
      user:
        summary: User Example in XML
        externalValue: http://foo.bar/examples/user-example.xml
```

For Content Objects, Parameter Objects and Header Objects, you can use either a inline `example` or a map of described example objects called `examples` , but not both.  Schema objects only support inline `example` as in OpenAPI V2.

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -53,7 +53,6 @@ Additional utilities can also take advantage of the resulting files, such as tes
 		- [Callback Object](#callbackObject)
 		- [Headers Object](#headersObject)
 		- [Example Object](#exampleObject)
-		- [Described Example Object](#describedExampleObject)
 		- [Links Object](#linksObject)
 		- [Link Object](#linkObject)
 		- [Link Parameters Object](#linkParametersObject)
@@ -408,7 +407,7 @@ Field Name | Type | Description
 <a name="componentsSchemas"></a> schemas | Map[`string`, [Schema Object](#schemaObject)] | An object to hold reusable [Schema Objects](#schemaObject).
 <a name="componentsResponses"></a> responses | Map[`string`, [Response Object](#responseObject)] | An object to hold reusable [Response Objects](#responseObject).
 <a name="componentsParameters"></a> parameters | Map[`string`, [Parameter Object](#parameterObject)] | An object to hold reusable [Parameter Objects](#parameterObject).
-<a name="componentsExamples"></a> examples | Map[`string`, [Described Example Object](#describedExampleObject)] | An object to hold reusable [Described Example Objects](#describedExampleObject).
+<a name="componentsExamples"></a> examples | Map[`string`, [Example Object](#exampleObject)] | An object to hold reusable [Example Objects](#exampleObject).
 <a name="componentsRequestBodies"></a> requestBodies | Map[`string`, [Request Body Object](#requestBodyObject)] | An object to hold reusable [Request Body Objects](#requestBodyObject).
 <a name="componentsHeaders"></a> headers | Map[`string`, [Header object](#headerObject)] | An object to hold reusable [Header objects](#headerObject).
 <a name="componentsSecuritySchemes"></a> securitySchemes| Map[`string`, [Security Scheme Object](#securitySchemeObject)] | An object to hold reusable [Security Scheme Objects](#securitySchemeObject).
@@ -938,8 +937,8 @@ Field Name | Type | Description
 <a name="parameterExplode"></a>explode | `boolean` | When this is true, parameter values of type `array` or `object` generate separate parameters for each value of the array, or key-value-pair of the map.  For other types of parameters this property has no effect. When [`style`](#parameterStyle) is `form`, the default value is `true`. For all other styles, the default value is `false`.
 <a name="parameterAllowReserved"></a>allowReserved | `boolean` | Determines whether the parameter value SHOULD allow reserved characters, as defined by [RFC3986](https://tools.ietf.org/html/rfc3986#section-2.2) `:/?#[]@!$&'()*+,;=` to be included without percent-encoding. This property only applies to parameters with an `in` value of `query`. The default value is `false`.
 <a name="parameterSchema"></a>schema | [Schema Object](#schemaObject) \| [Reference Object](#referenceObject)] | The schema defining the type used for the parameter.
-<a name="parameterExample"></a>example | [Example Object](#exampleObject) | Example of the media type.  The example object SHOULD be in the correct format as specified in the parameter encoding.  The `example` object is mutually exclusive to the `examples` object.  Furthermore, if referencing a `schema` which contains an example, the `example` value SHALL _override_ the the example provided by the schema.
-<a name="parameterExamples"></a>examples | Map[ `string`, [Described Example Object](#describedExampleObject) \| [Reference Object](#referenceObject)] | Described Examples of the media type.  Each described example SHOULD contain a value in the correct format as specified in the parameter encoding.  The `examples` object is mutually exclusive to the `example` object.  Furthermore, if referencing a `schema` which contains an example, the `examples` value SHALL _override_ the example provided by the schema.
+<a name="parameterExample"></a>example | Any | Example of the media type.  The example object SHOULD be in the correct format as specified in the parameter encoding.  The `example` object is mutually exclusive to the `examples` object.  Furthermore, if referencing a `schema` which contains an example, the `example` value SHALL _override_ the the example provided by the schema. To represent examples of media types that cannot naturally represented in JSON or YAML, a string value can be used to contain the example with escaping where necessary.
+<a name="parameterExamples"></a>examples | Map[ `string`, [Example Object](#exampleObject) \| [Reference Object](#referenceObject)] | Examples of the media type.  Each example SHOULD contain a value in the correct format as specified in the parameter encoding.  The `examples` object is mutually exclusive to the `example` object.  Furthermore, if referencing a `schema` which contains an example, the `examples` value SHALL _override_ the example provided by the schema.
 
 For more complex scenarios a [Content Object](#contentObject) can be used to define the media type 
 and schema of the parameter.  This option is mutually exclusive with the simple scenario
@@ -1131,7 +1130,7 @@ A request body with a referenced model definition.
       "examples": {
           "user" : {
             "summary": "User Example", 
-            "valueUrl": "http://foo.bar/examples/user-example.json"
+            "externalValue": "http://foo.bar/examples/user-example.json"
           } 
         }
     },
@@ -1142,7 +1141,7 @@ A request body with a referenced model definition.
       "examples": {
           "user" : {
             "summary": "User example in XML",
-            "valueUrl": "http://foo.bar/examples/user-example.xml"
+            "externalValue": "http://foo.bar/examples/user-example.xml"
           }
         }
     },
@@ -1150,7 +1149,7 @@ A request body with a referenced model definition.
       "examples": {
         "user" : {
             "summary": "User example in Plain text",
-            "valueUrl": "http://foo.bar/examples/user-example.txt" 
+            "externalValue": "http://foo.bar/examples/user-example.txt" 
         }
       } 
     },
@@ -1158,7 +1157,7 @@ A request body with a referenced model definition.
       "examples": {
         "user" : {
             "summary": "User example in other format",
-            "valueUrl": "http://foo.bar/examples/user-example.whatever"
+            "externalValue": "http://foo.bar/examples/user-example.whatever"
         }
       }
     }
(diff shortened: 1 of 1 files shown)


whitlockjc:
Does this address a specific issue?

darrelmiller:
@whitlockjc Yes, #488

DavidBiesack:
I recommend simpler terminology -- Example Object for `example' and `Examples Object` rather than `Described Example Object` for this new object. (It's plural since it is a map, i.e. a container of multiple examples). The two are distinct, and I think they should be mutually exclusive (as `example` and `examples` are now.)  This just replaces the examples array with an Examples object.

darrelmiller:
@DavidBiesack But the "example" object that is mapped by the "examples" object is not the same object as is used in "example".  That's why we have the new name.  There isn't much value in creating a distinct object for the map of "described example objects" which is why I just identified it as a Map.  This is a convention we started using in Components.
It might be possible to not use "example object" in the "example" property and just use "any" as is done in the schema object and then we could reclaim "example object" to be what I have called "described example object".  The only downside there is we lose the place to document the fact that the "example object" is a literal value or a string that contains an escaped example.

versions/3.0.md
```diff
@@ -1267,9 +1323,9 @@ Each Media Type Object provides schema and examples for a the media type identif
...
-<a name="mediaTypeExample"></a>example | [Example Object](#exampleObject) \| [Reference Object](#referenceObject) | Example of the media type.  The example object SHOULD be in the correct format as specified in the media type.  The `example` object is mutually exclusive to the `examples` object.  Furthermore, if referencing a `schema` which contains an example, the `example` value SHALL _override_ the example provided by the schema.
+<a name="mediaTypeSchema"></a>schema | [Schema Object](#schemaObject) \| [Reference Object](#referenceObject) | The schema defining the type used for the request body.
+<a name="mediaTypeExample"></a>example | [Example Object](#exampleObject) | Example of the media type.  The example object SHOULD be in the correct format as specified in the media type.  The `example` object is mutually exclusive to the `examples` object.  Furthermore, if referencing a `schema` which contains an example, the `example` value SHALL _override_ the the example provided by the schema.
```
DavidBiesack: use **Example Value** and "example value"

versions/3.0.md
```diff
@@ -1267,9 +1323,9 @@ Each Media Type Object provides schema and examples for a the media type identif
...
+<a name="mediaTypeExamples"></a>examples | Map[ `string`, [Described Example Object](#describedExampleObject) \| [Reference Object](#referenceObject)] | Described Examples of the media type.  Each described example SHOULD contain a value in the correct format as specified in the media type.  The `examples` object is mutually exclusive to the `example` object.  Furthermore, if referencing a `schema` which contains an example, the `examples` value SHALL _override_ the example provided by the schema.
```
DavidBiesack: use **Example Object** instead

versions/3.0.md
```diff
@@ -1822,9 +1894,9 @@ X-Rate-Limit-Reset:
 
 #### <a name="exampleObject"></a>Example Object
```
DavidBiesack: becomes `<a name="exampleValue"></a>Example Value`

versions/3.0.md
```diff
@@ -1822,9 +1894,9 @@ X-Rate-Limit-Reset:
...
 
-Allows sharing examples for operation requests and responses. This object can either be a freeform object, array or primitive value.  To represent examples of media types that cannot naturally represented in the OpenAPI definition, a string value can be used to contain the example with escaping where necessary. 
+Allows sharing examples for operation requests and responses. This literal value can either be any JSON or YAML object, array or primitive value.  To represent examples of media types that cannot naturally represented in the OpenAPI definition, a string value can be used to contain the example with escaping where necessary. 
 
-##### Example Example
+##### Example Object Example
```
DavidBiesack: becomes Example Value Example

versions/3.0.md
```diff
@@ -1848,6 +1920,84 @@ breed: Mixed
 
 ```
 
+#### <a name="describedExampleObject"></a>Described Example Object
```
DavidBiesack: becomes `<a name="exampleObject"></a>Example Object`

versions/3.0.md
```diff
@@ -1848,6 +1920,84 @@ breed: Mixed
...
+
+##### Fixed Fields
+Field Name | Type | Description
+---|:---:|---
+<a name="describedExampleSummary"></a>summary | `string` | Short description for the example.
```
DavidBiesack: drop "described" throughout

versions/3.0.md
```diff
@@ -1848,6 +1920,84 @@ breed: Mixed
...
+<a name="describedExampleDescription"></a>description | `string` | Long description for the example.
+<a name="describedExampleValue"></a>value | [Example Object](#exampleObject) | Embedded literal example. The `value` field and `valueUrl` field are mutually exclusive.
+<a name="describedExampleValueUrl"></a>valueUrl | `string` | A URL that points to the literal example. This provides the ability to reference examples that cannot easily be included in JSON or YAML documents.  The `value` field and `valueUrl` field are mutually exclusive.
```
DavidBiesack: I suggest using `$ref : uri` instead of `valueUri`.
This way, it can also be a pointer to a value elsewhere in the OpenAPI definition, or it can be an external resource

versions/3.0.md
```diff
@@ -53,12 +53,12 @@ Additional utilities can also take advantage of the resulting files, such as tes
 		- [Callback Object](#callbackObject)
 		- [Headers Object](#headersObject)
 		- [Example Object](#exampleObject)
```
darrelmiller: That seems like a reasonable approach.  I must admit, I did like the "Described Example" moniker but `Example Value` is more accurate than `Example Object` for the current usage.

DavidBiesack: (Funny.. I felt the Described Example moniker was slightly awkward.)

darrelmiller: It's ok, I'm not too emotionally attached to it :-)

versions/3.0.md
darrelmiller: It can't be done like that as then we would lose the `summary` and `description` properties.  The alternative would be this:
```yaml
examples:
  foo: 
    summary: this is a referenced example value
    value:
      $ref: http://example.com/reuse.json#/example/one
```
The downside to this, is that you can no longer reference example values that are not JSON or YAML.  Based on the conversation with @fehguy the other day, I am recommending that we add wording to the reference object definition that $ref can only point to JSON or YAML content.
By using `valueUrl` we can have something special for examples that may be other formats.  (and later schemas, but shhhhhh, don't say that out loud).

versions/3.0.md
earth2marsh: If the `valueUrl` concept survives, consider whether having a name that is more similar to `externalDocs` which has some similarities here (non-$ref style references), like `externalValue` or `externalExample`

darrelmiller:
Changed made based on TDC review.  

- `Described Example Object` becomes `Example Object`
- The old example goes away and we now just use `Any` for the type of `example`.
- `valueUrl` has been renamed to `externalValue`.

fehguy:
Looks good to me!

versions/3.0.md
```diff
@@ -1822,30 +1893,80 @@ X-Rate-Limit-Reset:
...
 
 ```yaml
+# in a model
+schemas:
+  properties:
+    name:
+      type: string
+      examples:
+        name:
+          $ref: http://example.org/petapi-examples/OpenApi.json#/components/examples/name-example
```
webron: `openapi.json`

versions/3.0.md
```diff
@@ -938,8 +937,8 @@ Field Name | Type | Description
...
-<a name="parameterExample"></a>example | [Example Object](#exampleObject) \| [Reference Object](#referenceObject) | Example of the content type.  The example object SHOULD be in the correct format as specified in the parameter encoding.  The `example` object is mutually exclusive to the `examples` object.  Furthermore, if referencing a `schema` which contains an example, the `example` value SHALL _override_ the example provided by the schema.
+<a name="parameterExample"></a>example | Any | Example of the media type.  The example object SHOULD be in the correct format as specified in the parameter encoding.  The `example` object is mutually exclusive to the `examples` object.  Furthermore, if referencing a `schema` which contains an example, the `example` value SHALL _override_ the the example provided by the schema. To represent examples of media types that cannot naturally represented in JSON or YAML, a string value can be used to contain the example with escaping where necessary.
```
whitlockjc: To me, _"correct format"_ makes sense but should we be more explicit, something to the effect of _"it should adhere to the schema specified in the parameter encoding"_?

versions/3.0.md
```diff
@@ -938,8 +937,8 @@ Field Name | Type | Description
...
+<a name="parameterExamples"></a>examples | Map[ `string`, [Example Object](#exampleObject) \| [Reference Object](#referenceObject)] | Examples of the media type.  Each example SHOULD contain a value in the correct format as specified in the parameter encoding.  The `examples` object is mutually exclusive to the `example` object.  Furthermore, if referencing a `schema` which contains an example, the `examples` value SHALL _override_ the example provided by the schema.
```
whitlockjc: Same as above.

versions/3.0.md
```diff
@@ -1128,20 +1127,38 @@ A request body with a referenced model definition.
       "schema": {
         "$ref": "#/components/schemas/User"
       },
-      "examples": [ "http://foo.bar/examples/user-example.json" ]
+      "examples": {
```
whitlockjc: I'm not sure I understand, with `examples` being plural should the value be an array?

versions/3.0.md
```diff
@@ -1128,20 +1127,38 @@ A request body with a referenced model definition.
...
+          "user" : {
```
whitlockjc: Shouldn't the key be a mime-type?

versions/3.0.md
whitlockjc: I don't understand the value.  Since `examples` is no longer an array if items, what is the significance of the map key?

whitlockjc: It appears to be a free-form description/label/name but the intent is not explained in the _"Examples Object"_ description.

versions/3.0.md
```diff
@@ -1822,30 +1893,80 @@ X-Rate-Limit-Reset:
...
+# in a model
+schemas:
+  properties:
+    name:
+      type: string
+      examples:
+        name:
+          $ref: http://example.org/petapi-examples/OpenApi.json#/components/examples/name-example
+
+# in a request body, note the plural `examples` as the Content-Type is set to `*`:
```
whitlockjc: Shouldn't it be `*/*`?

versions/3.0.md
```diff
@@ -2412,6 +2532,7 @@ schemas:
             $ref: http://foo.bar#/examples/address-example.json
 ```
 
+>>>>>>> refs/remotes/origin/OpenAPI.next
```
whitlockjc: Great job merging your conflict @darrelmiller.  🥇 _(joke of course)_

versions/3.0.md
```diff
@@ -2689,14 +2809,22 @@ example:
...
   ],
-  "examples": [
-  {
-    "name": "Puma",
-    "id": 1
-  }, {
-    "name": "Ferguson",
-    "id": 2
-  }]
+  "examples": {
```
whitlockjc: Why would we support `example` and `examples` elsewhere but not here?  Was removing this intentional?  It doesn't seem to be because on line 2812 you use `examples`.

whitlockjc:
I forgot to start a review, my bad.  Review done, comments above.

versions/3.0.md
webron: It should be. The comment itself doesn't seem in place though because nothing in the example describes a wildcard content type.

versions/3.0.md
webron: The idea behind removing `examples` was that unlike other places, there's no `content` here and no media types described. Providing multiple examples for a model is out of context and doesn't serve a lot of value. You'd need to combine the schema with a media type to define how it's actually serialized. You could argue that even the `example` should be dropped for that reason.

versions/3.0.md
DavidBiesack: The key allows referencing/reusing the specific example in a $ref

versions/3.0.md
darrelmiller: Done

versions/3.0.md
darrelmiller: There are two issues here.  If the parameter encoding says XML, then the example needs to be XML.  That's what I meant by correct format.  Saying that the example should conform to the schema is a different requirement.  Let me try and come up with words that say both.

webron: "it should make sense, don't do anything crazy"

PR #1040, opened 2017-04-12
Explanation for links confusion, fixes

fehguy:
In the Link Object:
* Changed `href` to `operationRef` to avoid confusion between a reference to an operation vs.  a server
* Added a `server` object to the Link Object to allow for an alternative solution from what was proposed in https://github.com/OAI/OpenAPI-Specification/commit/edc99b3f4099af2878435cef9aeffb5fbad499d4

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -1867,17 +1867,19 @@ The `Link Object` is responsible for defining a possible operation based on a si
 
 Field Name  |  Type  | Description
 ---|:---:|---
-href        | `string`    | a relative or absolute URL to a linked resource. This field is mutually exclusive with the `operationId` field.
+operationRef | `string` | a relative or reference to an OAS operation. This field is mutually exclusive with the `operationId` field.
 operationId | `string` | the name of an _existing_, resolvable OAS operation, as defined with a unique `operationId`.  This field is mutually exclusive with the `href` field.  Relative `href` values MAY be used to locate an existing [Operation Object](#operationObject) in the OAS.
 parameters  | [Link Parameters Object](#linkParametersObject) | an object representing parameters to pass to an operation as specified with `operationId` or identified via `href`.
 headers     | [Headers Object](#headersObject)    | an object representing headers to pass to the linked resource. Where conflicts occur between these headers, and those defined in the related operation, these headers override.
 description | `string` | a description of the link, supports [CommonMark syntax](http://spec.commonmark.org/).
+server      | [Server Object](#serverObject) | a server object to be used by the target operation
 
 This object can be extended with [Specification Extensions](#specificationExtensions). 
 
-Locating a linked resource MAY be performed by either a `href` or `operationId`.
+Locating a linked resource MAY be performed by either a `operationRef` or `operationId`.
 In the case of an `operationId`, it MUST be unique and resolved in the scope of the OAS document.
-Because of the potential for name clashes, consider the `href` syntax as the preferred method for specifications with external references.
+Because of the potential for name clashes, consider the `operationRef` syntax as the preferred 
+method for specifications with external references.
 
 ##### <a name="responsePayload"></a>Response Payload Values
 
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -1868,11 +1868,11 @@ The `Link Object` is responsible for defining a possible operation based on a si
 Field Name  |  Type  | Description
 ---|:---:|---
 operationRef | `string` | a relative or absolute reference to an OAS operation. This field is mutually exclusive with the `operationId` field, and must point to the fragment of a valid OAS definition
-operationId | `string` | the name of an _existing_, resolvable OAS operation, as defined with a unique `operationId`.  This field is mutually exclusive with the `href` field.  Relative `href` values MAY be used to locate an existing [Operation Object](#operationObject) in the OAS.
-parameters  | [Link Parameters Object](#linkParametersObject) | an object representing parameters to pass to an operation as specified with `operationId` or identified via `href`.
-headers     | [Headers Object](#headersObject)    | an object representing headers to pass to the linked resource. Where conflicts occur between these headers, and those defined in the related operation, these headers override.
-description | `string` | a description of the link, supports [CommonMark syntax](http://spec.commonmark.org/).
-server      | [Server Object](#serverObject) | a server object to be used by the target operation
+operationId  | `string` | the name of an _existing_, resolvable OAS operation, as defined with a unique `operationId`.  This field is mutually exclusive with the `operationRef` field.  Relative `operationRef` values MAY be used to locate an existing [Operation Object](#operationObject) in the OAS.
+parameters   | [Link Parameters Object](#linkParametersObject) | an object representing parameters to pass to an operation as specified with `operationId` or identified via `operationRef`.
+headers      | [Headers Object](#headersObject)    | an object representing headers to pass to the linked resource. Where conflicts occur between these headers, and those defined in the related operation, these headers override.
+description  | `string` | a description of the link, supports [CommonMark syntax](http://spec.commonmark.org/).
+server       | [Server Object](#serverObject) | a server object to be used by the target operation
 
 This object can be extended with [Specification Extensions](#specificationExtensions). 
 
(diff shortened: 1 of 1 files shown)


fehguy:
@OAI/tdc let's discuss

dilipkrish:
Disclaimer and apologies as I haven't been keeping up with all the changes to the spec in detail.
As I try and get back to implementing 3.0 of the spec for springfox, I thought I would chime in on this.

While its convenient to use `operationId`, Im not really sure we need to have `operationId` at all. Let me explain: 
- `operationId` as such is not guaranteed to be supplied by the the spec creators, nor is it guaranteed to be unique across spec definitions.
- The intent of operation id is easily satisfied by linking (inline or external) to an operation. Just like we would link to an operation directly  e.g. if we have a path like below

```json
"/pets": {
    "get": {
         "operationId": "petsGET"
     },
    "post": {
    }
}
```

We could easily reference the operation using `#/pets/get` using the json ref notation. If it is hosted in an external server, that would be similar to how we reference external model definitions like `http://example.com/api.json#/path/get`. Also may be worth clarifying that if `server` is supplied, the `$ref` must be relative. The example for the three variants below

```yaml
{
links:
    #Internal Reference
    searchPets:
      $ref: #/pets/get
      parameters:
        name: 'test'
   # With External Reference
    searchPets:
     $ref: http://example.com/api.json#/pets/get
      parameters:
        name: 'test'
   # With With External Reference Server Specified
    searchPets:
     $ref: #/pets/get
      parameters:
        name: 'test'
      server: 
        - url: http://example.com/api.json
           description: Development server
}
```

What about `server` references that are defined in the specification? Should it also allow references to server objects as well?

I know the intent of the original `href` property was to allow to an inlined Uri template for e.g. "href" : "http://example.com/pets{?name,category}" that would resolve using the parameters specified. Is that no longer a use case?

Open questions I have that I'd like addressed are
1. Are inlined links definitions going away? IOW, we only support links to operations defined in OAI specification.
2. Should the `server` be a `oneOf` {`server`, `serverRef`}?

dilipkrish:
One catch to using a operation reference to the operation is that for certain paths it will look weird e.g. for path `/pets/{id}` the reference would be `#/path/{id}/get`. An alternate way to describe it is to use a path reference and method combination. e.g.

```yaml
{
  links:
    #With Operation Reference. Implicit http method as the last path segment
    pet:
      $operationRef: #/pets/{id}/get
      parameters:
      # ...
    # With Path Reference and HTTP Method
    pet:
      $pathRef: #/pets/{id}
      method: GET
      parameters:
      # ...

}
````

**Comment by darrelmiller** [comment]
@dilipkrish A few clarifications and comments:

a) The suggested property is `operationRef` not `$operationRef`
b) The previous iterations of the spec used `href` which contained the path item key as an identifier of the operation.  It was not a JSON Pointer.
c) If `operationRef` is to be a JSON Pointer then the correct JSON Pointer for your example would be `#/paths//pets/get`.  The double-slash being particularly ugly.
d) I don't understand the concern you raise about $ref needing to be relative if `server` is supplied.  `server` is used to change the target of the link.  It is not related to the location of the target operation definition.
e) I don't believe there is any need for the `server` property to allow a reference object.  The `server` property is only used to override the global `servers` with a specific alternative.  We also currently don't have a `servers` map in `components` so there is no place to declare reusable `servers`.

**Comment by darrelmiller** [comment]
@fehguy I really like the `server` idea for allowing links to point to external resources.  I had made it half-way to the same conclusion in that we already allow accessing arbitrary external resources by overriding the `server` in operations.  I just hadn't thought of being consistent and using `server` in links.

I think `operationRef` is a better name than `href`.  However, I'm not exactly sure what the syntax of that property should be.  Previously I had understood `href` to be just the `key` of the path item object.  e.g.  `/pet` which leads us to the problem of not knowing which operation is desired.    Using a full JSON pointer value gives us `#/paths//pet/get` which is it's own kind of crazy as we would probably have to escape the second slash for JSON Pointer libraries to work.  Do we invent some new custom "operationRef" syntax?

**Comment by MikeRalphson** [comment]
> Using a full JSON pointer value gives us #/paths//pet/get which is it's own kind of crazy as we would probably have to escape the second slash for JSON Pointer libraries to work

It would indeed need to be escaped as `#/paths/~1pet/get` according to the [JSON Pointer spec](https://tools.ietf.org/html/rfc6901#section-3). It removes the ugliness of the double-slash, with it's own particular charm.

**Comment by darrelmiller** [comment]
@MikeRalphson I'd forgot about the whole `~0` `~1` shenanigans.

**Comment by dilipkrish** [comment]
> If operationRef is to be a JSON Pointer then the correct JSON Pointer for your example would be #/paths//pets/get. The double-slash being particularly ugly.

Oh yeah. Oversight. And yes to the ugly :)

> I don't understand the concern you raise about $ref needing to be relative if server is supplied. server is used to change the target of the link. It is not related to the location of the target operation definition.

Just to clarify, if we have the following `operationRef` => [`#/pet/{id}`, `http://example.com/apis.json#/pet/{id}`]
what effect would `server` => `{ "url": "http://example.com/api.json", "description": "Development"}` on each of them? How would the operationRefs be resolved in each of the cases? As an corollary, is the intention for `operationRef` to always be relative?

> e) I don't believe there is any need for the server property to allow a reference object. The server property is only used to override the global servers with a specific alternative. We also currently don't have a servers map in components so there is no place to declare reusable servers.

Would it be a good idea to make servers part of components as well? Seems like we'd be repeating servers for every link.

**Comment by darrelmiller** [comment]
@dilipkrish No `operationRef` is not intended to always be relative.  It could be relative or absolute.  `operationRef` is used to locate the __definition__ of the target operation, whereas `server` is intended to change the runtime location of the target __resource__.  You would mainly use the `server` object if the resource you are linking to is on a different server than the API that you are describing.  For example, you have some kind of login resource.  It could have a link to the OpenId Connect discovery resource provided by your preferred identity provider.  That resource is most likely going to be on a completely different server.

**Comment by fehguy** [comment]
@dilipkrish you can use a relative or absolute `operationRef`

@MikeRalphson is correct about that nasty syntax to get to the operation via json reference.  It's ugly, but honestly I'm expecting most OAS docs are going to be written by tooling so it doesn't matter a ton.

File: versions/3.0.md [hunk 1]
```diff
@@ -1867,7 +1867,7 @@ The `Link Object` is responsible for defining a possible operation based on a si
 
 Field Name  |  Type  | Description
 ---|:---:|---
-operationRef | `string` | a relative or reference to an OAS operation. This field is mutually exclusive with the `operationId` field.
+operationRef | `string` | a relative or reference to an OAS operation. This field is mutually exclusive with the `operationId` field, and must point to the fragment of a valid OAS definition
```
darrelmiller: "a relative or absolute reference"

fehguy: thanks @darrelmiller, I've added that

versions/3.0.md
```diff
@@ -1867,17 +1867,19 @@ The `Link Object` is responsible for defining a possible operation based on a si
 
 Field Name  |  Type  | Description
 ---|:---:|---
-href        | `string`    | a relative or absolute URL to a linked resource. This field is mutually exclusive with the `operationId` field.
+operationRef | `string` | a relative or absolute reference to an OAS operation. This field is mutually exclusive with the `operationId` field, and must point to the fragment of a valid OAS definition
 operationId | `string` | the name of an _existing_, resolvable OAS operation, as defined with a unique `operationId`.  This field is mutually exclusive with the `href` field.  Relative `href` values MAY be used to locate an existing [Operation Object](#operationObject) in the OAS.
```
webron: The `href` here needs to be updated as well (to `operationRef`). The last sentence here is unclear. Why does the description of the `operationId` explain about relative values for `href`/`operationRef`?

versions/3.0.md
```diff
@@ -1867,17 +1867,19 @@ The `Link Object` is responsible for defining a possible operation based on a si
...
 parameters  | [Link Parameters Object](#linkParametersObject) | an object representing parameters to pass to an operation as specified with `operationId` or identified via `href`.
```
webron: Again, `href`->`operationRef`

versions/3.0.md
```diff
@@ -1867,17 +1867,19 @@ The `Link Object` is responsible for defining a possible operation based on a si
...
 headers     | [Headers Object](#headersObject)    | an object representing headers to pass to the linked resource. Where conflicts occur between these headers, and those defined in the related operation, these headers override.
 description | `string` | a description of the link, supports [CommonMark syntax](http://spec.commonmark.org/).
+server      | [Server Object](#serverObject) | a server object to be used by the target operation
```
webron: Saw @darrelmiller's comment, but I'm still not sure what this means.

fehguy:
I will update the `href` examples and then after that, merge, per @OAI/tdc

PR #1049, opened 2017-04-23
Added CommonMark support to all description

### LINKED ISSUE #879: 3.0.md description field cleanup
While reviewing the spec, I've found some inconsistencies in regards to which of the various `description` fields allow markdown syntax. It seems some do, some done and I don't really see a pattern. For the sake of consistency does it make sense that we allow markdown for all `description` fields? The other thing that jumps out at me is that some descriptions state that they should be short, others verbose. Is that an artifact of the author of a given section, or has there been some thought put into?

Along those same lines, should we allow markdown for all summery and fields? (Perhaps some others)

Let me know if this seems reasonable and I can get a PR rolling.

Example:
`info.description` : A short description of the application. CommonMark syntax can be used for rich text representation.
`server.description` : An optional string describing the host designated by the URL.
`templates.description` : An optional description for the template parameter
`path.description` : An optional, string description, intended to apply to all operations in this path.
`operation.description` : A verbose explanation of the operation behavior. CommonMark syntax can be used for rich text representation.
`externalDocumentation.description` : A short description of the target documentation. CommonMark syntax can be used for rich text representation.
`parameter.description` : A brief description of the parameter. This could contain examples of use. CommonMark syntax can be used for rich text representation.
....

webron:
Fixes #879

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -324,7 +324,7 @@ An object representing a Server.
 Field Name | Type | Description
 ---|:---:|---
 <a name="serverUrl"></a>url | `string` | A URL to the target host.  This URL supports Server Variables and may be relative, to indicate that the host location is relative to the location where the OpenAPI definition is being served. Variable substitutions will be made when a variable is named in `{`brackets`}`.
-<a name="serverDescription"></a>description | `string` | An optional string describing the host designated by the URL.
+<a name="serverDescription"></a>description | `string` | An optional string describing the host designated by the URL. [CommonMark syntax](http://spec.commonmark.org/) can be used for rich text representation.
 <a name="serverVariables"></a>variables | [Server Variables Object](#serverVariablesObject) | An object holding variables for substitution in the URL template.
 
 This object can be extended with [Specification Extensions](#specificationExtensions). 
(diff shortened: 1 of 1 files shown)


versions/3.0.md
```diff
@@ -324,7 +324,7 @@ An object representing a Server.
 Field Name | Type | Description
 ---|:---:|---
 <a name="serverUrl"></a>url | `string` | A URL to the target host.  This URL supports Server Variables and may be relative, to indicate that the host location is relative to the location where the OpenAPI definition is being served. Variable substitutions will be made when a variable is named in `{`brackets`}`.
-<a name="serverDescription"></a>description | `string` | An optional string describing the host designated by the URL.
+<a name="serverDescription"></a>description | `string` | An optional string describing the host designated by the URL. [CommonMark syntax](http://spec.commonmark.org/) can be used for rich text representation.
```
DavidBiesack: This implies CommonMark is optional, but it is not. 

It is more correct to say, "Description strings are [CommonMark](http://spec.commonmark.org/) text." here and in the remaining changes.

webron: Going to keep it as-is for now. Will submit a separate PR to remove the word 'optional' from a lot of property descriptions instead as there's no need to mention that.

DavidBiesack: I think it is ambiguous and can/should be simplified (more concise, not ambiguous). I'm not referring to the use of the word "optional", but the phrase "can be used", which implies something other than CommonMark can be used if the author does not want rich text. This *is* CommonMark.

whitlockjc:
My worry is that _"Markdown"_ is a well-known _language_ and calling it _"CommonMark"_ is confusing to me.  If it's not to others, this should not stop us from shipping it.

ePaul:
Maybe "This string is Markdown in its [CommonMark variant](http://spec.commonmark.org/)." would be better?

MikeRalphson:
Bikeshedding "This string SHOULD be interpreted as Markdown according to the [CommonMark](http://spec.commonmark.org/) specification."?

webron:
This PR was meant to make sure CommonMark can be used in all the descriptions, not to tackle the wording of the `description` field. We can tackle that separately.

PR #1050, opened 2017-04-23
More details on 'pattern'

### LINKED ISSUE #880: clarify regex format

webron:
Resolves #880 

Copied the description from JSON Schema.

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -2617,7 +2617,7 @@ The following properties are taken directly from the JSON Schema definition and 
 - exclusiveMinimum
 - maxLength
 - minLength
-- pattern (This string SHOULD be a valid regular expression, according to the ECMA 262 regular expression dialect)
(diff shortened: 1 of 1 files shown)


versions/3.0.md
```diff
@@ -2617,7 +2617,7 @@ The following properties are taken directly from the JSON Schema definition and
 - exclusiveMinimum
 - maxLength
 - minLength
-- pattern
+- pattern (This string SHOULD be a valid regular expression, according to the ECMA 262 regular expression dialect)
```
DavidBiesack: Please add a link: `[ECMA 262 regular expression](https://www.ecma-international.org/ecma-262/5.1/#sec-7.8.5)`

whitlockjc:
:shipit:

Issue #963, opened 2017-03-06
3.0.0-rc0 Request Body Object has no required fields?


willdady:
Similar to #962, [Request Body Object](https://github.com/OAI/OpenAPI-Specification/blob/3.0.0-rc0/versions/3.0.md#request-body-object) has no required fields. Shouldn't `content` be required?

ePaul:
Hmm, interesting question. What could be the meaning of a Request body object without a content field?

* Clients can send anything, we don't document what is done with it?
    * I guess this should be done via a `*/*` entry in the Contend object, with documentation.
* Clients should not send anything (i.e. no/empty body)?
    * I guess in this case we omit the requestBody object?

So it seems that making it required is a good idea.

PR #1051, opened 2017-04-23
Added required fields

### LINKED ISSUE #962: 3.0.0-rc0 ServerObject has no required fields?
The `url` field for [ServerObject](https://github.com/OAI/OpenAPI-Specification/blob/3.0.0-rc0/versions/3.0.md#serverObject) isn't explicitly marked as required? Shouldn't it be? Presumably all servers have URL's and the `variables` field is dependant on it being defined. You wouldn't (couldn't) define `variables` but not `url`.

webron:
Resolves #962, #963.

Added required `url` for Server, and `content` for Request Body.

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -323,7 +323,7 @@ An object representing a Server.
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="serverUrl"></a>url | `string` | A URL to the target host.  This URL supports Server Variables and may be relative, to indicate that the host location is relative to the location where the OpenAPI definition is being served. Variable substitutions will be made when a variable is named in `{`brackets`}`.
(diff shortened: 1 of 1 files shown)


RobDolinMS:
@webron Would you please elaborate on why requestBodyContent is required ?

(I didn't see a clear reason for this in #963)

whitlockjc:
I'm with @RobDolinMS, I'd like to understand why `requestBodyContent` is required.

webron:
The Request Body Object currently has no required fields. That means you can specify a request body without specifying anything. That has no semantic meaning. If you define a request body, you should define what it represents. Even if it's any value, it should be specified, meaning, you have to have `content` in it.

whitlockjc:
I thought all `content` properties were optional, which is why this one being required surprised me.  I'll go back through the document to see if `content` being required makes sense.

whitlockjc:
:shipit:

After talking with @webron about this offline, it makes sense to me now.

PR #1053, opened 2017-04-23
Fixed servers type

### LINKED ISSUE #973: 3.0.0-rc0 inconsistent 'servers' fields
[Operation Object](https://github.com/OAI/OpenAPI-Specification/blob/3.0.0-rc0/versions/3.0.md#operationObject) has a fixed field called `servers` which expects a _single_ [Server Object](https://github.com/OAI/OpenAPI-Specification/blob/3.0.0-rc0/versions/3.0.md#serverObject) as it's value. It's description is a little contradictory.

> An alternative server **array** to service this operation. If an alternative server **object** is specified at the Path Item Object or Root level, it will be overridden by this value.

[Path Item](https://github.com/OAI/OpenAPI-Specification/blob/3.0.0-rc0/versions/3.0.md#pathItemObject) is the same as it also has a fixed field named `servers`, but is shown as a single object.

Shouldn't both Operation Object and Path Item Object have their `servers` field denoted as `[Server Object]`?

Assuming servers are supposed to implicitly _inherit_ from their parent object if not defined I think the descriptions could be made more consistent. Something like:

OpenAPI Object
> An optional array of Server Objects which provide connectivity information to a target server.

Path Item Object
> An optional array of Server Objects which provide connectivity information to a target server.
If a server array is specified at the OpenAPI Object level, it will be overridden by this value.

Operation Object
> An optional array of Server Objects which provide connectivity information to a target server.
If a server array is specified at the Path Item Object or OpenAPI Object level, it will be overridden by this value.

webron:
Resolved #973

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -661,7 +661,7 @@ Field Name | Type | Description
 <a name="pathItemHead"></a>head | [Operation Object](#operationObject) | A definition of a HEAD operation on this path.
 <a name="pathItemPatch"></a>patch | [Operation Object](#operationObject) | A definition of a PATCH operation on this path.
 <a name="pathItemTrace"></a>trace | [Operation Object](#operationObject) | A definition of a TRACE operation on this path.
-<a name="pathItemServers"></a>servers | [Server Object](#serverObject) | An alternative `server` array to service all operations in this path.  
+<a name="pathItemServers"></a>servers | [[Server Object](#serverObject)] | An alternative `server` array to service all operations in this path.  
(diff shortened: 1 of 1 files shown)


RobDolinMS:
@darrelmiller Would you please give this a quick review?  It's a two-line change.  Thanks.

whitlockjc:
:shipit:

darrelmiller:
Whoa.  I was not expecting this change.  I thought the `servers` was the typo and we were only going to allow a single server overload.

darrelmiller:
Ok, I withdraw my concerns.  Even with one server, a spec could define different variables at the path item/operation level and require the user to provide different input.  So, let's just be consistent.

PR #1056, opened 2017-04-23
Grammar nit

MikeRalphson:
s/as well as supports/as well as supporting

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -1464,7 +1464,7 @@ When passing complex objects in the `x-www-form-urlencoded` content type, the de
 
 ##### Special Considerations for `multipart` content
 
-It is common to use `multipart/form-data` as a `Content-Type` when transferring request bodies to operations.  In contrast to 2.0, a `schema` is required to define the input parameters to the operation when using `multipart` content.  This allows complex structures as well as supports mechanisms for multiple file uploads.
+It is common to use `multipart/form-data` as a `Content-Type` when transferring request bodies to operations.  In contrast to 2.0, a `schema` is required to define the input parameters to the operation when using `multipart` content.  This allows complex structures as well as supporting mechanisms for multiple file uploads.
 
 When passing in `multipart` types, boundaries MAY be used to separate sections of the content being transferred--thus, the following default `Content-Type`s are defined for `multipart/*`:
 


versions/3.0.md
```diff
@@ -1464,7 +1464,7 @@ When passing complex objects in the `x-www-form-urlencoded` content type, the de
 
 ##### Special Considerations for `multipart` content
 
-It is common to use `multipart/form-data` as a `Content-Type` when transferring request bodies to operations.  In contrast to 2.0, a `schema` is required to define the input parameters to the operation when using `multipart` content.  This allows complex structures as well as supports mechanisms for multiple file uploads.
+It is common to use `multipart/form-data` as a `Content-Type` when transferring request bodies to operations.  In contrast to 2.0, a `schema` is required to define the input parameters to the operation when using `multipart` content.  This allows complex structures as well as supporting mechanisms for multiple file uploads.
```
DavidBiesack: The original intent is not clear:
a) "This supports mechanisms for multiple file uploads", or
b) "This allows [complex structures and/or supporting mechanisms] ..." where "supporting mechanism" is a noun phrase.

It could be simpler and more concise instead: "This allows for multiple file uploads." as it, it begs the questions "what complex structures? What supporting mechanisms?" Need to find out what the original intent was.

RobDolinMS:
@fehguy According to GitHub's "Blame" feature, you were the last person to edit this line.  Would you please take a look?

fehguy:
It seems clear to me already, so i'll explain the intent and someone with better documentation skills can make it better for the masses and take the spoils.

The intent is that in 3.0.x, you use a `schema` object.  That allows for complex objects (previously the spec only supported primitives or arrays of primitives).  You can now describe a "User Object" schema if you like.  In addition to now allowing complex objects, you can also describe multiple file uploads.  There are schema examples of doing so as well in the spec documentation.

versions/3.0.md
darrelmiller: "What complex structures?"  The same complex stuff we use schemas in parameters/request body and response bodies.  e.g. JSON Objects, JSON Arrays of Objects.
"what supporting mechanisms" Encoding Object

darrelmiller:
If may be one of those cases where you need to spend some time working with the multipart stuff to understand the paragraph.  I don't have any issue with it.  But perhaps there is a way to say it more clearly.

PR #1059, opened 2017-04-25
Explicit code over range clarification

### LINKED ISSUE #983: 4xx default with specific overrides
Request to clarify if 4xx examples can be overridden in specific cases such that a 422 can have its own example but all other errors us the default.

webron:
Resolves #983.


whitlockjc:
:shipit:

versions/3.0.md
```diff
@@ -1587,7 +1587,7 @@ Field Name | Type | Description
...
+<a name="responsesCode"></a>[HTTP Status Code](#httpCodes) | [Response Object](#responseObject) \| [Reference Object](#referenceObject) | Any [HTTP status code](#httpCodes) can be used as the property name (one property per HTTP status code). Describes the expected response for that HTTP status code.  [Reference Object](#referenceObject) can be used to link to a response that is defined at the [OpenAPI Object's responses](#oasResponses) section. This field MUST be quoted for compatibility between JSON and YAML (i.e. '200'), and MAY contain the uppercase character, `X` to designate a wildcard, such as `2XX` to represent all response codes between `[200-299]`. If a response range is defined, and an explicit code within that range is defined as well - the explicit code definition takes precedence over the range definition for that code.
```
darrelmiller: I'd use a comma instead of the hyphen.  "...range is defined as well, the explicit code definition..."

PR #1060, opened 2017-04-25
Change requestBody default required to false

### LINKED ISSUE #986: Request Body is mandatory by default
Looks like the Request Body Object has `default` set to `true` by default. Not sure if that's intentional, but it seems wrong.

webron:
Resolves #986


whitlockjc:
:shipit:

PR #1061, opened 2017-04-25
Added default value to Schema Object's deprecated

### LINKED ISSUE #990: deprecated in the Schema Object is missing a default value
should be `false`.

webron:
Resolves #990


whitlockjc:
:shipit:

PR #1062, opened 2017-04-25
clarified `null`

### LINKED ISSUE #1025: Clarify usage of null
In the data types section (https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#data-types), the following sentence:

"null is not supported as a value"

should be:

"null is not supported as a type"

webron:
Resolves #1025


whitlockjc:
:shipit:

darrelmiller:
:shipit:

PR #1064, opened 2017-04-26
Update enum and default type for server variables

### LINKED ISSUE #976: Consider changing the enum type under Server Variable
Right now the type is an array of `primitive`. Variables are not typed, we were just looking to restrict values to non-complex objects. 

We can change it to `string` - which would simplify tooling. This will affect how the following example would be though:
```yaml
    port:
      enum:
        - 8443
        - 443
      default: 8443
```

That would become:
```yaml
    port:
      enum:
        - '8443'
        - '443'
      default: '8443'
```

which is a bit more annoying to write.

Changing the type of `enum` would also change the type of `default`.

webron:
Resolves #976

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -388,8 +388,8 @@ An object representing a Server Variable for server URL template substituion.
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="serverVariableEnum"></a>enum | [`primitive`] | An enumeration of primitive type values to be used if the substitution options are from a limited set.
-<a name="serverVariableDefault"></a>default | `primitive` |  **Required.** The default value to use for substitution if an alternate value is not specified, and will be sent if an alternative value is _not_ supplied. Unlike the [Schema Object's](#schemaObject) `default`, this value MUST be provided by the consumer.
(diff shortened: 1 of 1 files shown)


versions/3.0.md
```diff
@@ -388,8 +388,8 @@ An object representing a Server Variable for server URL template substituion.
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="serverVariableEnum"></a>enum | [`primitive`] | An enumeration of primitive type values to be used if the substitution options are from a limited set.
-<a name="serverVariableDefault"></a>default | `primitive` |  **Required.** The default value to use for substitution if an alternate value is not specified, and will be sent if an alternative value is _not_ supplied. Unlike the [Schema Object's](#schemaObject) `default`, this value MUST be provided by the consumer.
+<a name="serverVariableEnum"></a>enum | [`string`] | An enumeration of primitive type values to be used if the substitution options are from a limited set.
+<a name="serverVariableDefault"></a>default | `string` |  **Required.** The default value to use for substitution if an alternate value is not specified, and will be used if an alternative value is _not_ supplied. Unlike the [Schema Object's](#schemaObject) `default`, this value MUST be provided by the consumer.
 <a name="serverVariableDescription"></a>description | `string` | An optional description for the server variable.
```
darrelmiller: s/will/MUST ?

Issue #1078, opened 2017-04-28
CommonMark allows HTML tags, should some be prohibited/ignored?


darrelmiller:
CommonMark allows embedded HTML.  What do we need to do ensure OpenAPI definitions are "safe".  

Long and painful thread here https://talk.commonmark.org/t/make-commonmark-safe-by-default/1265/52

DavidBiesack:
I don't think we need to manage this other than to mention it in the new paragraph about rich text/CommonMark... a note to tool writers to be careful about `<script>` tags is probably enough.

Interpretation of such text is entirely up to the _execution context_ of the tool(s), and there are many such contexts, not all of which are HTML or browsers. We should note that such tools may wish to escape such description strings, but having worked on `swagger-codegen generate -l html` I can say that overzealous escaping of the text gets in the way. (In swagger-codegen it was done both in the html engine **and** in the Markdown library, so even simple and "safe" tags like `<strong>` were escaped.)

PR #1090, opened 2017-05-07
Permit tooling to limit commonmark usage

darrelmiller:
Addresses issue #1078

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -168,7 +168,7 @@ password | `string` | `password` | Used to hint UIs the input needs to be obscur
 
 ### <a name="richText"></a>Rich Text Formatting
 Throughout the specification `description` fields are noted as supporting CommonMark markdown formatting.
-OpenAPI compliant tooling MUST support markdown that is compliant with [CommonMark 0.27](http://spec.commonmark.org/0.27/). Tooling MAY limit some CommonMark features to address security concerns. 
+Where OpenAPI tooling renders rich text it MUST support, at a minimum, markdown syntax as described by [CommonMark 0.27](http://spec.commonmark.org/0.27/). Tooling MAY choose to ignore some CommonMark features to address security concerns. 
 
 ### <a name="relativeReferences"></a>Relative References in URLs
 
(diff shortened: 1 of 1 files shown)


RobDolinMS:
LGTM.  Is this ready to merge?

fehguy:
I want a pony

whitlockjc:
When I read this wording, I read it as if I have to render CommonMark a certain way and I'm not sure that's how we want to do it.  I think for us to mark things as compliant, we're saying that we should treat values that are valid CommonMark 0.27 as valid whether or not you render it, or how you render it.  But I do get what you're proposing from the security perspective where those choosing to render HTML can choose to limit some features of CommonMark 0.27 for security purposes.

I'll give it some thought and see if I can come up with any suggestions.

webron:
Covered on the call - suggestion was to limit the wording to the impact on rendering. The wording right now implies that a command-line tool (for example) would require to support CommonMark in order to be compliant, however, it has no real impact on the tool.

RobDolinMS:
I was thinking about @whitlockjc's comments on this from today's TDC and wondering if something like this would address people's concerns:

OpenAPI compliant tooling MUST NOT break if markdown that is compliant with CommonMark 0.27 is used as the value of the description field.

Tool authors MAY choose to do any of:
* Implement rendering for ALL of CommonMark 0.27
* Implement rendering for SOME of CommonMark 0.27 (ex: ignore <script> tags)
* Implement rendering for NONE of CommonMark 0.27 (ex: ignore any description value that is not plain text)

/cc @darrelmiller @fehguy

DavidBiesack:
I don't think OAS should overly specify the behavior (i.e. only provide only those three options. There are gradations of support - i.e. strip all HTML / CommonMark markup and leave just the text would not be allowed if phrased as above. I prefer the existing language in the PR, "Tooling MAY limit some CommonMark features to address security concerns." (but was unable to participate in the TDC call so I don't know how the discussion went) although as far as a specification is concerned, a small change might be to state "Tooling MAY limit some or all CommonMark features to address security concerns."

darrelmiller:
I have updated the text to try and address concerns raised in the TDC call.

- Clarified that compatibility with CommonMark is only relevant for tooling that renders rich text.  
- Clarified that tooling is not limited to only supporting CommonMark .27 features. Other newer features can be supported
- Clarified that tooling should only ignore and not fail when features with security concerns are used.

/cc @OAI/tdc

webron:
:shipit:

earth2marsh:
LGTM

PR #1093, opened 2017-05-08
documented the discriminator property

### LINKED ISSUE #1085: document discriminator behavior

fehguy:
Fixes #1085

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -2663,7 +2663,7 @@ Other than the JSON Schema subset fields, the following fields MAY be used for f
 Field Name | Type | Description
 ---|:---:|---
 <a name="schemaNullable"></a>nullable | `boolean` | Allows sending a `null` value for the defined schema. Default value is `false`.
-<a name="schemaDiscriminator"></a>discriminator | [Discriminator Object](#discriminatorObject) | Adds support for polymorphism. The discriminator is the schema property name that is used to differentiate between other schema that inherit this schema. The property name used MUST be defined at this schema and it MUST be in the `required` property list. When used, the value MUST be the name of this schema or any schema that inherits it. See [Composition and Inheritance](#schemaComposition) for more details.
+<a name="schemaDiscriminator"></a>discriminator | [Discriminator Object](#discriminatorObject) | Adds support for polymorphism. The discriminator is an object name that is used to differentiate between other schemas which may satisfy the payload description. See [Composition and Inheritance](#schemaComposition) for more details.
 <a name="schemaReadOnly"></a>readOnly | `boolean` | Relevant only for Schema `"properties"` definitions. Declares the property as "read only". This means that it MAY be sent as part of a response but SHOULD NOT be sent as part of the request. If property is marked as `readOnly` being `true` and is in the `required` list, the `required` will take effect on the response only. A property MUST NOT be marked as both `readOnly` and `writeOnly` being `true`. Default value is `false`.
 <a name="schemaWriteOnly"></a>writeOnly | `boolean` | Relevant only for Schema `"properties"` definitions. Declares the property as "write only". This means that it MAY be sent as part of a request but SHOULD NOT be sent as part of the response. If property is marked as `writeOnly` being `true` and is in the `required` list, the `required` will take effect on the request only. A property MUST NOT be marked as both `readOnly` and `writeOnly` being `true`. Default value is `false`.
 <a name="schemaXml"></a>xml | [XML Object](#xmlObject) | This MAY be used only on properties schemas. It has no effect on root schemas. Adds Additional metadata to describe the XML representation format of this property.
@@ -3020,7 +3020,9 @@ components:
 
 #### <a name="discriminatorObject"></a>Discriminator Object
 
-When request bodies or response payloads may be one of a number of different schemas, a `discriminator` property can be used to aid in serialization, deserialization, and validation.  The discriminator is a specific property in an object which is used to inform the consumer of the specification of an alternative schema based on the value associated with it.
+When request bodies or response payloads may be one of a number of different schemas, a `discriminator` object can be used to aid in serialization, deserialization, and validation.  The discriminator is a specific object in a schema which is used to inform the consumer of the specification of an alternative schema based on the value associated with it.
+
+Note, when using the discriminator, _inline_ schemas will not be considered when using the discriminator.
 
 ##### Fixed Fields
 Field Name | Type | Description
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -2663,7 +2663,7 @@ Other than the JSON Schema subset fields, the following fields MAY be used for f
 Field Name | Type | Description
 ---|:---:|---
 <a name="schemaNullable"></a>nullable | `boolean` | Allows sending a `null` value for the defined schema. Default value is `false`.
-<a name="schemaDiscriminator"></a>discriminator | `string` | Adds support for polymorphism. The discriminator is the schema property name that is used to differentiate between other schema that inherit this schema. The property name used MUST be defined at this schema and it MUST be in the `required` property list. When used, the value MUST be the name of this schema or any schema that inherits it. See [Composition and Inheritance](#schemaComposition) for more details.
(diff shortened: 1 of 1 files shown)


versions/3.0.md
```diff
@@ -3012,6 +3012,125 @@ schemas:
...
+```
+
+When used in conjunction with the `anyOf` construct, the use of the discriminator can avoid ambiguity where multiple schemas may satisfy a single payload.
+
+In both the `oneOf` and `anyOf` use cases, all possible schemas must be listed explicitly.  To avoid redundancy, the discriminator may be added to a parent schema definition, and all schemas composing the parent schema in an `allOf` construct may be used as an alternate schema.
+
+For example:
+
+```
+components:
```
DavidBiesack: remove `components:` and `schemas:` here since they are not used in the other examples.

versions/3.0.md
```diff
@@ -3012,6 +3012,125 @@ schemas:
...
+```
+{
+  "pet_type": "cachorro",
+  "bark": "soft"
+}
+```
+
+will map to `Dog` because of the definition in the `mappings` element.
+
+
```
DavidBiesack: I suggest also adding a note:

Some OpenAPI definitions use external, reusable schema references such as
```
      properties:
        pet_type:
           type: string
           enum:
             - cat
             - cachorro
             - lizard
      oneOf:
      - $ref: 'http://www.example.com/schemas/Cat/schema.json'
      - $ref: 'http://www.example.com/schemas/Doc/schema.json'
      - $ref: 'http://www.example.com/schemas/Lizard/schema.json'
      discriminator:
        propertyName: pet_type
        mapping:
          cat: 'http://www.example.com/schemas/Cat/schema.json'
          cachorro: 'http://www.example.com/schemas/Doc/schema.json'
          lizard: 'http://www.example.com/schemas/Lizard/schema.json'
```
and an implicit name mapping based on schema names is not possible.
For such cases, a `mapping` is **required**.

The above `$ref` values are JSON Schema files, not OpenAPI documents. Schema may
also be referenced from schema definitions from an external OpenAPI definition:
```
      properties:
        pet_type:
           type: string
           enum:
             - cat
             - cachorro
             - lizard
      oneOf:
      - $ref: 'http://www.example.com/library.json#/components/schema/Cat.json'
      - $ref: 'http://www.example.com/library.json#/components/schema/Doc.json'
      - $ref: 'http://www.example.com/library.json#/components/schema/Lizard.json'
      discriminator:
        propertyName: pet_type
        mapping:
          cat: $ref: 'http://www.example.com/library.json#/components/schema/Cat.json'
          cachorro: $ref: 'http://www.example.com/library.json#/components/schema/Doc.json'
          lizard: $ref: 'http://www.example.com/library.json#/components/schema/Lizard.json'
```

[in particular, citing external JSON Schema,  and not just `#/definitions/SchemaName` from other OpenAPI 2.0 documents, worked in 2.0 and we need to ensure this still works in 2.0, whether this involves discriminators or not.]

versions/3.0.md
```diff
@@ -3012,6 +3012,125 @@ schemas:
...
 ```
 
+#### <a name="discriminatorObject"></a>Discriminator Object
+
+When request bodies or response payloads may be one of a number of different schemas, a `discriminator` property can be used to aid in serialization, deserialization, and validation.  The discriminator is a specific property in an object which is used to inform the consumer of the specification of an alternative schema based on the value associated with it.
+
+##### Fixed Fields
+---|:---:|---
+<a name="propertyName"></a>propertyName | `string` | **required** the name of the property in the payload that will hold the discriminator value.
```
ePaul: The examples are using `property`, here it is `propertyName`.

versions/3.0.md
```diff
@@ -3012,6 +3012,125 @@ schemas:
...
+
+```
+MyResponseType:
+  oneOf:
+  - $ref: '#/components/schemas/Cat'
+  - $ref: '#/components/schemas/Dog'
+  - $ref: '#/components/schemas/Lizard'
+```
+
+which means the paylod _must_, by validation, match exactly one of the schemas described by `Cat`, `Dog`, or `Lizard`.  In this case, a discriminator will act as a "hint" to bypass validation and selection of the matching schema which may be a costly operation, depending on the complexity of the schema. We can then describe exactly which field tells us which schema to use:
```
whitlockjc: Are we sure _"bypass validation"_ is the right phrase here?

versions/3.0.md
```diff
@@ -3012,6 +3012,125 @@ schemas:
...
+```
+{
+  "id": 12345,
+  "pet_type": "Cat"
+}
+```
+
+Will indicate that the `Cat` schema be used in conjunction with this payload.
+
+In scenarios where the value of the discriminator field does not match the schema name, an optional `mapping` definition may be used:
```
whitlockjc: What happens if the discriminator value does not match a schema or mapping?  We should probably document this as it will impact tooling.

versions/3.0.md
```diff
@@ -3012,6 +3012,125 @@ schemas:
...
+```
+MyResponseType:
+  oneOf:
+  - $ref: '#/components/schemas/Cat'
+  - $ref: '#/components/schemas/Dog'
+  - $ref: '#/components/schemas/Lizard'
+  discriminator:
+    property: pet_type
+    mapping:
+      cat: '#/components/schemas/Cat'
```
fehguy: @DavidBiesack you are correct, I will update the example.

versions/3.0.md
fehguy: Actually I believe the `components/schemas` is important here, since we're showing references inside the example.  I'll ensure the composition examples are consistent.

versions/3.0.md
```diff
@@ -3012,6 +3012,125 @@ schemas:
...
+```
+components:
+  schemas:
+    Pet:
+      type: object
+      required:
+      - pet_type
+      properties:
+        pet_type:
+          type: string
```
fehguy: That would defeat the goal of this example--here, we're showing that the parent object `Pet` does _not_ have knowledge of all of the composed types.  Thus, **any** object which extends `Pet` with the `allOf` construct is a candidate for use

versions/3.0.md
```diff
@@ -3012,6 +3012,125 @@ schemas:
...
+    Dog:
+      allOf:
+      - $ref: '#/components/schemas/Pet'
+      - type: object
+        # all other properties specific to a `Dog`
+        properties:
+          bark:
+            type: string
+```
+
```
fehguy: sure, I think Lizards are cool

versions/3.0.md
fehguy: good point, I'll add language around this 👍

versions/3.0.md
fehguy: Thanks, will fix

versions/3.0.md
fehguy: yes, I'll rephrase.  `will` ==> `may`, `bypass` ==> `shortcut`

fehguy:
@OAI/tdc updated per feedback above.  Thanks!

fehguy:
@OAI/tdc please review and approve--changes from call have been merged.

versions/3.0.md
```diff
@@ -2826,188 +2826,329 @@ example:
...
+            type: integer
+            format: int32
+            description: the size of the pack the dog is from
+            default: 0
+            minimum: 0
+        required:
+        - packSize
+```
+
+#### <a name="discriminatorObject"></a>Discriminator Object
```
DavidBiesack: I suggest renaming this anchor as discriminatorProperty / Discriminator Property so that OAS only uses "Object" when referring to a JSON object.

ePaul: The discriminator object described here actually is a JSON (or YAML) object in the API definition (In OpenAPI 2.0 it was a simple string), though it describes a single (string) property of the JSON payload.

I would say the name is fine.

fehguy: @DavidBiesack you are correct.  Using `title` is a bad idea, and I've disallowed _inline_ schemas for consideration with the `discriminator`.

DavidBiesack: thanks, Tony

RobDolinMS:
#TDC: Merge it!

PR #1009, opened 2017-03-27
Add guidance on length of Path Item summary field

MikeRalphson:
As per Operation Object summary field

PR #1094, opened 2017-05-08
Remove guidance on summary length

earth2marsh:
Per TDC discussion on PR #1009, this removes the guidance that had advised keeping length under 120 characters guidance (from v2)

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -763,7 +763,7 @@ Describes a single API operation on a path.
 Field Name | Type | Description
 ---|:---:|---
 <a name="operationTags"></a>tags | [`string`] | A list of tags for API documentation control. Tags can be used for logical grouping of operations by resources or any other qualifier.
-<a name="operationSummary"></a>summary | `string` | A short summary of what the operation does. For maximum readability in editing or documentation generation tools, this field SHOULD be less than 120 characters.
+<a name="operationSummary"></a>summary | `string` | A short summary of what the operation does.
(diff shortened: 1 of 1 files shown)


versions/3.0.md
```diff
@@ -763,7 +763,7 @@ Describes a single API operation on a path.
 Field Name | Type | Description
 ---|:---:|---
 <a name="operationTags"></a>tags | [`string`] | A list of tags for API documentation control. Tags can be used for logical grouping of operations by resources or any other qualifier.
-<a name="operationSummary"></a>summary | `string` | A short summary of what the operation does. For maximum readability in editing or documentation generation tools, this field SHOULD be less than 120 characters.
```
DavidBiesack: I'm OK with not limiting this to 120, but I think we should mention that many tools (including Swagger UI, swagger-codegen HTML output) render the summary strings as *titles* and long summaries result in a degraded experience. (Ditto for title in schemas/models) So some context explaining *why* one should keep summary *short* (a single line) would be helpful.

webron: @earth2marsh we did discuss clarifying the difference between `summary` and `description`. I don't think we should mention tooling here, just clarify the difference in goals.

earth2marsh: @webron @DavidBiesack during the TDC call I was asked to make this a 2-part PR, one to remove the character limitation and the second to improve the guidance, esp to better clarify the difference btw `summary` and `description`. Please find that here: https://github.com/OAI/OpenAPI-Specification/pull/1095, and thanks for the comments.

webron: @earth2marsh apologies, completely missed that PR.

PR #1095, opened 2017-05-08
Better guidance around summary properties

earth2marsh:
This is the second of two PRs to resolve issues discussed in PR #1009.


webron:
:shipit:

PR #1098, opened 2017-05-10
Default server value

### LINKED ISSUE #1068: Server Object must define default API location if not present
If Server Object is not provided then we should consider the `url` to be effectively `\`

webron:
Resolved #1068.


webron:
will change wording to add `property` and merge.

PR #1102, opened 2017-05-11
added example, type clarification for encoding headers

### LINKED ISSUE #987: headers under Encoding Property needs clarification
The type and description are unclear, and there are no examples.

fehguy:
Fixes #987


RobDolinMS:
#TDC Approves but @webron  please hyperlink the object.

PR #1103, opened 2017-05-12
OperationRef changes and more

darrelmiller:
Notable changes:
- The term _variable substitution_ has been replaced by _runtime expression_
- Headers object has been removed from the Link object
- The section on "Response Payload Values" that describes how body values are implicitly mapped to parameters of linked operations has been removed.  It is now required to enumerate the list of parameters that are passed to the linked operation and provide a runtime expression to determine the value.
- The variable substitution examples that were in the Link Object section have been moved to their own Runtime Expression section.
- Properties that accept runtime expressions do not need to be surrounded by curly braces.
- curly braces are used to embed the value of runtime expressions inside strings.
- moved large bitbucket repo example into separate example file
- attempted to clean up examples.

I'd better stop making changes before this PR gets even bigger.  /cc @OAI/tdc

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -1737,7 +1737,7 @@ This object can be extended with [Specification Extensions](#specificationExtens
 
 ##### Key Expression
 
-The key used to identify the [Path Item Object](#pathItemObject) is a variable expression that can be evaluated in the context of a runtime HTTP request/response to identify the URL to be used for the callback request.
+The key used to identify the [Path Item Object](#pathItemObject) is a [runtime expression](#runtimeExpression) that can be evaluated in the context of a runtime HTTP request/response to identify the URL to be used for the callback request.
 A simple example might be `$request.body#/url`.
 However, using a [runtime expression](#runtimeExpression) the complete HTTP message can be accessed.
 This includes accessing any part of a body that can be accessed using a JSON Pointer [RFC6901](https://tools.ietf.org/html/rfc6901). 
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -1752,7 +1752,7 @@ Content-Length: 123
 
 {
   "failedUrl" : "http://clientdomain.com/failed"
-  "successUrls : [
+  "successUrls" : [
     "http://clientdomain.com/fast",
     "http://clientdomain.com/medium",
     "http://clientdomain.com/slow"
@@ -1779,12 +1779,12 @@ $response.header.Location | http://example.org/subscription/1
 
 ##### Callback Object Example
 
-A callback to the URL specified by the `url` property in the request body.
+A callback to the URL specified by the `id` and `email` property in the request body.
 
 
 ```yaml
 myWebhook:
-  'http://notificationServer.com?transactionId={$request.body#/id}&email={$request.body#/email}}':
+  'http://notificationServer.com?transactionId={$request.body#/id}&email={$request.body#/email}':
     post:
       requestBody:
         description: Callback payload
@@ -1884,15 +1884,15 @@ The presence of a link does not guarantee the caller's ability to successfully i
 
 As opposed to _dynamic_ links (i.e. links provided **in** the response payload), the OAS linking mechanism does not require that link information be provided in a specific response format at runtime.
 
-For computing links, and providing instructions to execute them, a [runtime expression](#runtimeExpression) is used for accessing values in a response and using them as parameters while invoking the linked operation.
+For computing links, and providing instructions to execute them, a [runtime expression](#runtimeExpression) is used for accessing values in an operation and using them as parameters while invoking the linked operation.
 
 ##### Fixed Fields
 
 Field Name  |  Type  | Description
 ---|:---:|---
-<a name="linkOperationRef"></a>operationRef | `string` | a relative or absolute reference to an OAS operation. This field is mutually exclusive with the `operationId` field, and must point to the fragment of a valid OAS definition.
+<a name="linkOperationRef"></a>operationRef | `string` | a relative or absolute reference to an OAS operation. This field is mutually exclusive with the `operationId` field, and MUST point to an [Operation Object](#operationObject).
 <a name="linkOperationId"></a>operationId  | `string` | the name of an _existing_, resolvable OAS operation, as defined with a unique `operationId`.  This field is mutually exclusive with the `operationRef` field.  Relative `operationRef` values MAY be used to locate an existing [Operation Object](#operationObject) in the OAS.
-<a name="linkParameters"></a>parameters   | Map[`string` \| Any \| [{expression}](#runtimeExpression)] | A map representing parameters to pass to an operation as specified with `operationId` or identified via `operationRef`. The key is the parameter name to be used, whereas the value can be a constant or an expression to be evaluated and passed to the linked operation.
+<a name="linkParameters"></a>parameters   | Map[`string`, Any \| [{expression}](#runtimeExpression)] | A map representing parameters to pass to an operation as specified with `operationId` or identified via `operationRef`. The key is the parameter name to be used, whereas the value can be a constant or an expression to be evaluated and passed to the linked operation.
 <a name="linkDescription"></a>description  | `string` | a description of the link, supports [CommonMark syntax](http://spec.commonmark.org/).
 <a name="linkServer"></a>server       | [Server Object](#serverObject) | a server object to be used by the target operation.
 
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -1853,8 +1853,7 @@ A callback to the URL specified by the `url` property in the request body.
 
 ```yaml
 myWebhook:
-  '$request.body#/url':
-  'http://notificationServer.com?transactionId={$request#/body}&email={$request.body#/email}}'
+  'http://notificationServer.com?transactionId={$request.body#/id}&email={$request.body#/email}}':
     post:
       requestBody:
         description: Callback payload
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -1915,27 +1915,43 @@ paths:
     - name: id
       in: path
       required: true
-      description: the user identifier, as userId or username
+      description: the user identifier, as userId 
       schema:
         type: string
-    responses:
-      '200':
-        description: the user being returned
-        content:
-          application/json:
-            schema:
-              type: object
-              properties:
-                uuid: the unique user id
-                  type: string
-                  format: uuid
+    get:
+      responses:
+        '200':
+          description: the user being returned
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  uuid: the unique user id
+                    type: string
+                    format: uuid
         links:
-          Address:
+          address:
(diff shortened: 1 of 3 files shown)


versions/3.0.md
```diff
@@ -1998,293 +1915,136 @@ paths:
...
+
+##### <a name="runtimeExpression"></a>Runtime Expressions
+
+Runtime expressions allow defining values based on information that will only be available within the HTTP message in an actual API call.
+This mechanism is used by [Link Objects](#linkObject) and [Callback Objects](#callbackObject).
+
+The runtime expression is defined by the following [ABNF](https://tools.ietf.org/html/rfc5234) syntax
+
+```
+      expression = ( "$url" | "$method" | "$request." [ source ] | "$response." [ source ])
```
DavidBiesack: This list should include `$response.statusCode` - the HTTP status code of the response

i.e. `... "$response." [ source | "statusCode" ]`

versions/3.0.md
```diff
@@ -1998,293 +1915,136 @@ paths:
...
+        operationId: getUserAddress
+        responses:
+          '200':
+            description: the user's address
 ```
 
-Where the `$request.path.id` is the value passed in the request to `/users/{id}`.
-
-##### Response Payload Example
+When a runtime expression evaluates to null, no parameter value is passed to the target operation.
```
DavidBiesack: "no parameter value is passed" is not well-defined and probably should be "the expression resolves to an empty string"

DavidBiesack: I think this merits discussion. null is a valid value in a JSON object, so for the response 
```
{ "foo" : { "bar" : null, "zip" : "" } }
```
a expression `$response.body#/foo/bar`  should resolve to `null` - that may be significant. 

Note that the expression `$response.body#/foo/zip`  will result in an empty string.

versions/3.0.md
```diff
@@ -1905,91 +1903,10 @@ In the case of an `operationId`, it MUST be unique and resolved in the scope of
...
-      query-reference = "query." name  
-      path-reference = "path." name
-      body-reference = "body#" fragment
-      fragment = a JSON Pointer [RFC6901](https://tools.ietf.org/html/rfc6901)
-      name = *( char )
-      char = as per [RFC7159](https://tools.ietf.org/html/rfc7159#section-7)
-      token = as per [RFC7230](https://tools.ietf.org/html/rfc7230#section-3.2.6)
-```
-
-The `name` identifier is case-sensitive, whereas `token` is not.
```
DavidBiesack: What is the behavior when an expression/path resolves to something that is not found?
 for the response 
```
{ "foo" : { "bar" : null, "zip" : "" } }
```
a expression `$response.body#/foo/xyz` results in an empty string?
 
What about headers that are not found?

versions/3.0.md
```diff
@@ -1998,293 +1915,136 @@ paths:
...
-      operationRef: '#paths~12.0~1repositories~1{$response.body#/username}'
+links:
+  UserRepositories:
+    # returns array of '#/components/schemas/repository'
+    operationRef: '#paths~12.0~1repositories~1/{username}'
+    parameters:
+      username: $response.body#/username
 ```
 
 or an absolute `operationRef`:
```
DavidBiesack: Link back to `#linkOperationId`, or if you add a section for **Operation Reference**, link to that.

versions/3.0.md
```diff
@@ -1883,18 +1882,17 @@ schemas:
...
+As opposed to _dynamic_ links (i.e. links provided **in** the response payload), the OAS linking mechanism does not require that link information be provided in a specific response format at runtime.
 
-Many operations require parameters to be passed, and these MAY be dynamic depending on the response itself. For computing links, and providing instructions to execute them, [variable substitution](#variableSubstitution) is used for accessing values in a response and using them as values while invoking the linked operation.
+For computing links, and providing instructions to execute them, a [runtime expression](#runtimeExpression) is used for accessing values in a response and using them as parameters while invoking the linked operation.
 
 ##### Fixed Fields
 
 Field Name  |  Type  | Description
 ---|:---:|---
 <a name="linkOperationRef"></a>operationRef | `string` | a relative or absolute reference to an OAS operation. This field is mutually exclusive with the `operationId` field, and must point to the fragment of a valid OAS definition.
```
DavidBiesack: "the fragment of a valid OAS definition" should be more precise:

.. and MUST point to a [Path Item Object](#path-item-object) in an OAS definition.

RobDolinMS:
#TDC: @darrelmiller to review additional feedback from @DavidBiesack and @fehguy and @usarid to give this a thorough read.

versions/3.0.md
darrelmiller: I suspect the behavior of an unresolved expression should be determined based on its usage.  For links, I think that parameter values should not be provided and for callbacks the callback should be disabled.  However, I feel we need more real world experience to be able to make this call.

versions/3.0.md
darrelmiller: It actually must point to an operation.  And there are examples that need to be fixed.

versions/3.0.md
```diff
@@ -1883,18 +1882,17 @@ schemas:
...
-<a name="linkHeaders"></a>headers      | Map[`string`, [Header Object](#headerObject)  \| [Reference Object](#referenceObject)]    | Maps a header name to its definition. Note that [RFC7230](https://tools.ietf.org/html/rfc7230#page-22) states header names are case insensitive. This represents the headers to pass to the linked resource. Where conflicts occur between these headers, and those defined in the related operation, these headers override.
+<a name="linkParameters"></a>parameters   | Map[`string` \| Any \| [{expression}](#runtimeExpression)] | A map representing parameters to pass to an operation as specified with `operationId` or identified via `operationRef`. The key is the parameter name to be used, whereas the value can be a constant or an expression to be evaluated and passed to the linked operation.
```
darrelmiller: It is really just a JSON Reference that has to point to something specific.  I'm not sure it needs much more description.

versions/3.0.md
darrelmiller: But that means that the target operation would not use the default value from an optional parameter.  That would not be the expected behavior.

darrelmiller: Yes, in both those cases the JSON pointer successfully resolved and therefore the value should be passed.

versions/3.0.md
darrelmiller: We should have access to status code.  Not sure if it should be with $response.statusCode or just simply $statusCode

versions/3.0.md
```diff
@@ -1998,293 +1915,136 @@ paths:
...
+HTTP Method            | `$method`         | The allowable values for the `$method` will be those for the HTTP operation 
+Requested media type | `$request.header.accept`        |  
+Request parameter      | `$request.path.id`        | Request parameters MUST be declared in the `parameters` section for the parent operation or they cannot be evaluated.  This includes request headers.
+Request body property   | `$request.body#/user/uuid`   | For operations which accept payloads, references may be made to portions of the `requestBody` or the entire body itself
+Request URL            | `$url`            |  
+Response value         | `$response.body#/status`       |  Only the payload in the response can be accessed with the `$response` syntax.  
+Response header        | `$response.header.Server` |  Single header values only are available
+
+Runtime expressions preserve the type of the referenced value.
+Expression values can be embeded into string values by surrounding the expression with `{}` curly braces.
```
webron: Do we assume that in case of URLs they would be URL encoded and there's no need to escape curly braces if they are part of them as they're already encoded (including in query parameters)?

versions/3.0.md
DavidBiesack: As we open the Call for Comments period on OAS 3.0.0, we should raise this. Implementers will need guidance, and I don't like a specification with open-ended, undefined behavior.

darrelmiller: Have you read the HTTP spec? ;-)

versions/3.0.md
DavidBiesack: yes, sorry - wrong reference; s/b `[Operation Object](#operationObject)`

versions/3.0.md
DavidBiesack: I see the need to allow falling back to the default and think this should be specified here (I assume you are referring to the `default` value define in a JSON schema for an operation parameter - please confirm.)  

However, there is a difference between the value at the JSON pointer being null (valid JSON) compared to the item the JSON path refers to not existing. If the JSON pointer value is null, I think that should be passed; it may be significant (see my other comment about null).

DavidBiesack: So "When a runtime expression evaluates to null," should be "When a runtime expression references a value that does not exist,"

versions/3.0.md
DavidBiesack: Good suggestion; I think just `$statusCode` is cleaner, since it does not complicate  `"$response." [ source ] `

DavidBiesack: Please confirm - this ABNF states that source is optional with `$request.` and `$response.` which implies that naked `$response.` is a valid expression - it's not clear if that is your intent or what that means. Should these omit the brackets: 
`expression = ( "$url" | "$method" | "$request." source | "$response."  source )`

versions/3.0.md
```diff
@@ -1998,293 +1915,136 @@ paths:
...
 ```yaml
-components:
-  links:
-    UserRepositories:
-      # returns array of '#/components/schemas/repository'
-      operationRef: '#paths~12.0~1repositories~1{$response.body#/username}'
+links:
+  UserRepositories:
+    # returns array of '#/components/schemas/repository'
+    operationRef: '#paths~12.0~1repositories~1/{username}'
```
DavidBiesack: Should this be `#/paths` ?

versions/3.0.md
darrelmiller: Hadn't caught the ambiguous name issue and I'd forgot about the request body thing.

versions/3.0.md
```diff
@@ -1761,7 +1761,6 @@ Content-Length: 123
```
RobDolinMS: #TDC: Done :)

versions/3.0.md
RobDolinMS: #TDC Good with adding $statusCode.  @DavidBiesack suggests this may be used in auditing cases.

versions/3.0.md
```diff
@@ -1998,293 +1915,136 @@ paths:
...
+      token = as per RFC [7230](https://tools.ietf.org/html/rfc7230#section-3.2.6)
+```
+
+The `name` identifier is case-sensitive, whereas `token` is not. 
+
+The table below provides examples of runtime expressions and examples of their use in a value:
+
+##### <a name="runtimeExpressionExamples"></a>Examples
+
+Source Location | example expression  | notes
```
RobDolinMS: @darrelmiller to fix

RobDolinMS:
#TDC: Good with curly braces
#TDC: Approved for merge AFTER:
* Conflicts resolved
* Wording changes from Kris
* Update callback example

versions/3.0.md
darrelmiller: No need to escape curly braces.

PR #1112, opened 2017-05-18
Fix typo in spec example: `securitySchemas`->`securitySchemes`

tfesenko:
Fix typo in spec example: `securitySchemas`->`securitySchemes`
JSON example is correct.


webron:
Thanks for the PR.

PR #1113, opened 2017-05-18
Add specification references for securitySchemes

MikeRalphson:
This is the first of two commits in response to the request in #1080 [here](https://github.com/OAI/OpenAPI-Specification/pull/1080#issuecomment-300013500) to add reference links to the specifications of existing OAS3 security schemes. This PR does not depend on #1080 being merged.

Please check I have the right RFC for OpenId Connect - I went with the discovery specification, http://openid.net/specs/openid-connect-discovery-1_0.html


RobDolinMS:
#TDC: @RobDolinMS to reach-out to MSFT and GOOG contacts to get input on what is ideal to reference.

RobDolinMS:
Per previous, I have reached-out to MSFT and GOOG contacts.
(CC'd @earth2marsh on the GOOG outreach)

versions/3.0.md
```diff
@@ -3368,7 +3368,7 @@ animals:
 #### <a name="securitySchemeObject"></a>Security Scheme Object
 
 Allows the definition of a security scheme that can be used by the operations.
-Supported schemes are HTTP authentication, an API key (either as a header or as a query parameter) and OAuth2's common flows (implicit, password, application and access code).
+Supported schemes are HTTP authentication, an API key (either as a header or as a query parameter), OAuth2's common flows (implicit, password, application and access code) as defined in [RFC6749](https://tools.ietf.org/html/rfc6749), and [OpenID Connect](http://openid.net/specs/openid-connect-discovery-1_0.html).
```
WilliamDenniss: Did you mean `http://openid.net/specs/openid-connect-core-1_0.html` not `http://openid.net/specs/openid-connect-discovery-1_0.html`?

RobDolinMS:
@MikeRalphson Would you please update the OpenID Connect reference per @WilliamDenniss's suggestion?  At that point, this should be ready for #TDC to merge.

MikeRalphson:
@RobDolinMS done. Thanks @WilliamDenniss

webron:
@WilliamDenniss Thanks for taking the time to look into this.

A couple of questions, if you don't mind, as I'm completely clueless when it comes to OIC.
- Our only field related to OIC definition is defined as such:
`openIdConnectUrl` - OpenId Connect URL to discover OAuth2 configuration values. This MUST be in the form of a URL. 
Is the discovery URL described in the `core` or the `discovery` doc?
- Do you know if the URL docs are 'fixed'? We've had issues in the past with websites updating URLs or content of URLs, breaking our links in the spec. Is there an alternative URL we should be using?

darrelmiller:
I have resolved the conflict and changed the OpenId Connect Discovery link to point to the draft-06 of the document on the IETF site.  The draft document has just passed Last Call so the chance of breaking changes are low.  However, I still think it is better to point to a draft page that is guaranteed to be stable than a page on openid.net that may change over time.  This is consistent with what we have done for references to JSON Schema.

Assuming there are no functional changes in the final version of the IETF RFC we can update the URL in a patch release of this spec.

RobDolinMS:
#TDC: No objections to pointing to IETF Draft (for now)

PR #1115, opened 2017-05-18
Converted Objects to Maps

### LINKED ISSUE #975: Standardize maps and extensions
Before releasing 3.0.0-RC0 we introduced a new type construct to the spec in the Components Object. For examples, `schemas` has the type of `Map[string, Schema Object]`.

This was done to avoid creating a Schemas Object, Responses Object and so on, which would simply be those maps. I'd recommend we drop existing map objects and use the same construct for those, including Server Variables Object, Scopes Object and so on.

The exception would be the Paths Object as it imposes a pattern to the named field.

Following that, suggest we drop support for extensions in maps (except Paths Object) - there's no way to tell the difference between a name in the map and an extension. If someone wants to name a server variable `x-y-z`, it would be unclear whether it's an extension or a name of a variable.

webron:
Resolved #975.

The following objects have been removed and converted to maps:
- Server Variables Object
  Map[{name}, Server Variable Object]
  No extensions
- Encoding Object
  Map[{property}, Encoding Property Object]
  No extensions
  Change Encoding Property Object to Encoding Object *
- Callbacks Object
  Map[string, Callback Object | Reference Object]
  No extensions *
- Headers Object
  Map[string, Header Object | Reference Object]
  No extensions
- Links Object
  Map[string, Link Object | Reference Object]
  No extensions
- Link Parameters Object
  Map[string, Any | {expression}]
  No extensions
- Scopes Object
  Map[string, string]
  No extensions *
- Security Requirement Object
  Map[string, [string]]
  No extensions
- Content Object
  Map[{mediaType}, Media Type Object]
  No extensions

A few points of consideration:
- `callbacks` and `links` are both maps now, but we should consider converting them to arrays, and pulling the friendly name into their respective objects.
- With the removal of the `Link Parameters Object`, there's no example with a constant parameter value.


DavidBiesack:
Leaving links as maps allows them to be referenced (`$ref`) by name (especially in `/components/links`); making it an array decreases that ability. Thus, it makes some sense to also leave it as a map in the response object, rather than have two different representations of link collections. Ditto for callbacks.

webron:
`parameters` is also an array which allows both inline parameter definitions and references. It's the same structure.

ePaul:
Do we want to have a paragraph in one of the introducing sections about the Map[...] notation?

RobDolinMS:
#TDC @webron is asking if Callbacks and Links should be maps or arrays.
#TDC @darrelmiller is hesitant on these as arrays and will look for a reason.
Parameters is an array because `name` is not unique (`name`+`in` is unique)
Link name can be used similar to link rel (from HTML) to describe the type of link
#TDC @earth2marsh likes the PR as-is and recommends considering maps vs. arrays as a separate PR.

#TDC: We'll merge this PR as-is (if LGTM'd from @darrelmiller and @earth2marsh)

earth2marsh:
LGTM

versions/3.0.md
```diff
@@ -770,7 +752,8 @@ Field Name | Type | Description
...
 <a name="operationResponses"></a>responses | [Responses Object](#responsesObject) | **Required.** The list of possible responses as they are returned from executing this operation.
-<a name="operationCallbacks"></a>callbacks | [Callbacks Object](#callbacksObject) | The list of possible callbacks as they are returned from executing this operation.
+<a name="operationCallbacks"></a>callbacks | Map[`string`, [Callback Object](#callbackObject) \| [Reference Object](#referenceObject)] | A map of possible out-of band callbacks related to the parent operation. Each value in the map is a [Callback Object](#callbackObject) that describes a request that may be initiated by the API provider and the expected responses.
+The key value used to identify the callback object is an expression, evaluated at runtime, that identifies a URL to use for the callback operation.
```
darrelmiller: The key value here is a simple identifier, not a runtime expression.  The runtime expression is for the keys inside the callback object.

versions/3.0.md
```diff
@@ -947,15 +930,14 @@ Field Name | Type | Description
...
 <a name="parameterExamples"></a>examples | Map[ `string`, [Example Object](#exampleObject) \| [Reference Object](#referenceObject)] | Examples of the media type.  Each example SHOULD contain a value in the correct format as specified in the parameter encoding.  The `examples` object is mutually exclusive to the `example` object.  Furthermore, if referencing a `schema` which contains an example, the `examples` value SHALL _override_ the example provided by the schema.
 
-For more complex scenarios a [Content Object](#contentObject) can be used to define the media type and schema of the parameter.
+For more complex scenarios the [`content`](#parameterContent) can be used to define the media type and schema of the parameter.
```
darrelmiller: I think this should say "the content property can be used"

PR #1117, opened 2017-05-19
Fix spec example: Links

tfesenko:
Fix [Links example](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#example-1):
* Fix indents
* Remove the type property from parameter, as it's already defined in the schema property
* Add a name to the link


webron:
Thanks for the PR!

PR #1126, opened 2017-05-24
Encoding Property Object: headers property should me a map of [String -> Header (not Header*s*)]

tfesenko:
I think using the "Header*s*" was just a typo and was not intentional:

1. [Encoding Property Object](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#encoding-property-object)#headers	is currently defined as a Map of [String -> Header*s* Object].
2. But [Headers Object](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#headers-object) is already a map of [{name} -> Header Object].

So, current description is equivalent to a map whose values are maps : [String -> [{name} -> Header Object]]. It contradicts to [this Encoding Object example](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#encoding-object-example)

@timburks and I found this issue while working on [JSON Schema for RC2](https://github.com/googleapis/gnostic/commit/02acf4809ebe82fe6280d952ec7b6467635e5a50#commitcomment-22252403)

Related question: should it also allow reference: [String -> Header Object | **Reference Object**] ?


timburks:
Mentioned also in the comments [here](https://github.com/OAI/OpenAPI-Specification/issues/987).

tfesenko:
Oops, I somehow missed that discussion thread

timburks:
@tfesenko I have this same question:
> should it also allow reference: [String -> Header Object | Reference Object] ?

darrelmiller:
I don't see a downside of allowing the header object be a reference object.

webron:
Thanks @tfesenko! If you'd like to submit a separate PR to allow a ref to the Header Object as well, per your suggestion, we'll merge that too.

MikeRalphson:
@tfesenko do you have time to submit the other PR to allow $refs in the encoding headers map?

PR #1131, opened 2017-05-27
Added Server Object JSON examples.

ckriley:
Enhanced Server Object YAML examples with JSON examples to be consistent with remainder of document.

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -338,12 +338,38 @@ This object can be extended with [Specification Extensions](#specificationExtens
 
 A single server would be described as:
 
+```json
+{
+  "url": "https://development.gigantic-server.com/v1",
+  "description": "Development server"
+}
+```
+
 ```yaml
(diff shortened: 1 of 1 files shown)


versions/3.0.md
```diff
@@ -338,13 +338,39 @@ This object can be extended with [Specification Extensions](#specificationExtens
...
+
 ```yaml
 url: https://development.gigantic-server.com/v1
 description: Development server
 ```
 
 The following shows how multiple servers can be described, for example, at the OpenAPI Object's [`servers`](#oasServers):
 
+```json
+{
```
darrelmiller: The extra curly braces surrounding the `servers` property are unnecessary.  There is some inconsistency between how we show examples of objects under the "OpenAPI object".  The components has "components" at the root, but path items doesn't have "paths".  We just decide what is the standard. /cc @OAI/tdc

webron: I tend to agree.

PR #1135, opened 2017-05-30
#1133 - replace commaDelimited by simple.

### LINKED ISSUE #1133: Where is commaDelimited in styles?
Hi,
I've found in [Parameters Examples](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#parameter-object-examples) this: `"style": "commaDelimited"`
Where is this style defined?

ePaul:
Also, in one example there was `format: form` for a parameter, which is totally bogus.

This fixes #1133 (even though that one is already closed).

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -717,7 +717,7 @@ This object can be extended with [Specification Extensions](#specificationExtens
       "items": {
         "type": "string"
       },
-      "style": "commaDelimited"
+      "style": "simple"
     }
   ]
 }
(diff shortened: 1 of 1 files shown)


darrelmiller:
The example with the `format: form` is even more broken.  It requires a `schema` object.  If you could make that tweak also, that would be awesome.

ePaul:
@darrelmiller done (I hope).

Issue #1086, opened 2017-05-05
Parameters with contents


RobDolinMS:
A parameter today has either a content or a regular type.  

@darrelmiller suggested we should not allow for more than one media type for parameters.

@earth2marsh asks if there are examples of this

@fehguy and @darrelmiller - Please discuss

fehguy:
I talked with @webron about this, I'll retract my concern about limiting the acceptable types to just one.  And since @webron and @darrelmiller have some crazy cosmic communication mechanism, I am pretty sure they're on the same page.

darrelmiller:
@OAI/tdc Add text to limit media type map in parameters to just one entry

webron:
Fixed and merged.

PR #1142, opened 2017-06-01
Limit parameter content object to one media type object

darrelmiller:


RobDolinMS:
#TDC: Looks good

ePaul:
This fixes #1086. (Just so we have the link.)

PR #1144, opened 2017-06-04
Fix schemas/Error code property in Uber example

OJFord:
The [live API returns][Uber products GET] e.g.:

    {
        "message": "No authentication provided.",
        "code": "unauthorized"
    }

This commit corrects the schema for `code` to be of `type: string`.

[Uber products GET]: https://api.uber.com/v1/products

### DIFF
--- a/examples/v3.0/uber.yaml
+++ b/examples/v3.0/uber.yaml
@@ -289,8 +289,7 @@ components:
     Error:
       properties:
         code:
-          type: integer
-          format: int32
+          type: string
         message:
           type: string
         fields:


webron:
Aha! So I'm not sure we're actually looking to accurately describe the Uber API but! Are you sure it's always a string?

OJFord:
> I'm not sure we're actually looking to accurately describe the Uber API

Ah, sorry, I should have guessed it wasn't intended to be a 'real' example from the fact that the 'petstore' example isn't live at all. (I initially assumed that it was a live demo.)

> Are you sure it's always a string?

No, I'm not sure. But I don't believe sum types are supported? Which I suppose doesn't matter if this isn't intended to be a live example.

MikeRalphson:
> Ah, sorry, I should have guessed it wasn't intended to be a 'real' example from the fact that the 'petstore' example isn't live at all. (I initially assumed that it was a live demo.)

Petstore is live at petstore.swagger.io/v2 - is there a link which is not working?

OJFord:
Ah! All (v2/3 YAML/JSON) schemata point to /v2.

RobDolinMS:
If this is intended as an illustrative example, I would think that the `code` property would be an integer and the `message` property would be a string.

RobDolinMS:
Thank you @OJFord for this PR.

Replying to @webron's question: 
> Are you sure it's always a string?

Uber's API documentation at: https://developer.uber.com/docs/riders/guides/errors#error-format says that `code` is a string.

LGTM

RobDolinMS:
#TDC: Thank you @OJFord for this.

PR #1146, opened 2017-06-05
minor editorial changes

krishahn:


RobDolinMS:
#TDC: Thumbs-up from @emckean and @webron

PR #1149, opened 2017-06-07
Captitalise key words, update guidance to rfc8174

MikeRalphson:
More nit-picking / doc cleanup ;)

* Update the guidance text on key words to clarify only uppercase instances are to be interpreted as key words - this helps with ambguity over 'may' meaning 'might' and 'MAY' meaning 'it is permitted', 'not all ... must be' etc.
* Update guidance to reference RFC8174
* Uppercase key words where apparently needed
* Make use of **Required.** / **Required**. / Required consistent => **REQUIRED**
* Change one \* to \\* to fix an ambiguity over italic text and aid editing the spec in Vim's markdown syntax highlighting mode (other editors are probably available)
* Erroneous whitespace has been removed only where it was on a line already being modified.

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -2,7 +2,7 @@
 
 #### Version 3.0.0-rc2
 
-The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC2119](http://www.ietf.org/rfc/rfc2119.txt).
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [BCP 14](https://tools.ietf.org/html/bcp14) [RFC2119](https://tools.ietf.org/html/rfc2119) [RFC8174](https://tools.ietf.org/html/rfc8174) when, and only when, they appear in all capitals, as shown here.
 
 This document is licensed under [The Apache License, Version 2.0](http://www.apache.org/licenses/LICENSE-2.0.html).
 
@@ -115,7 +115,7 @@ The schema exposes two types of fields: Fixed fields, which have a declared name
 
 Patterned fields can have multiple occurrences as long as each has a unique name. 
 
-In order to preserve the ability to round-trip between YAML and JSON formats, YAML version [1.2](http://www.yaml.org/spec/1.2/spec.html) is recommended along with some additional constraints:
+In order to preserve the ability to round-trip between YAML and JSON formats, YAML version [1.2](http://www.yaml.org/spec/1.2/spec.html) is RECOMMENDED along with some additional constraints:
 
 - Tags MUST be limited to those allowed by the [JSON Schema ruleset](http://www.yaml.org/spec/1.2/spec.html#id2803231)
 - Keys used in YAML maps MUST be limited to a scalar string, as defined by the [YAML Failsafe schema ruleset](http://yaml.org/spec/1.2/spec.html#id2802346)
@@ -171,7 +171,7 @@ Relative references used in `$ref` are processed as per [JSON Reference](https:/
 
 ### Schema
 
-In the following description, if a field is not explicitly **Required** or described with a MUST or SHALL, it can be considered OPTIONAL. 
+In the following description, if a field is not explicitly **REQUIRED** or described with a MUST or SHALL, it can be considered OPTIONAL.
 
 #### <a name="oasObject"></a>OpenAPI Object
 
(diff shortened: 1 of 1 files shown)


versions/3.0.md
```diff
@@ -2,7 +2,7 @@
 
 #### Version 3.0.0-rc2
 
-The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC2119](http://www.ietf.org/rfc/rfc2119.txt).
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC8174](http://www.ietf.org/rfc/rfc8174.txt) when, and only when, they appear in all capitals, as shown here.
```
ePaul: Instead of linking the .txt version, I would recommend the [HTML one](https://tools.ietf.org/html/rfc8174).

ePaul: The [template in RFC8174](https://tools.ietf.org/html/rfc8174#section-2) says to actually mention both RFCs:

      The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL
      NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED",
      "MAY", and "OPTIONAL" in this document are to be interpreted as
      described in BCP 14 [RFC2119] [RFC8174] when, and only when, they
      appear in all capitals, as shown here.

MikeRalphson: Yes, I hesitated on that and went with brevity over having the three links, but can amend.

versions/3.0.md
```diff
@@ -182,10 +182,10 @@ It combines what previously was the Resource Listing and API Declaration (versio
...
-<a name="oasInfo"></a>info | [Info Object](#infoObject) | **Required.** Provides metadata about the API. The metadata can be used by the clients if needed.
+<a name="oasVersion"></a>openapi | `string` | **REQUIRED**. Specifies the OpenAPI Specification version being used. It can be used by tooling Specifications and clients to interpret the version. The structure SHALL be `major`.`minor`.`patch`, where `patch` versions MUST be compatible with the existing `major`.`minor` tooling. Typically patch versions will be introduced to address errors in the documentation, and tooling SHOULD typically be compatible with the corresponding `major`.`minor` (3.0.\*). Patch versions will correspond to patches of this document.
```
ePaul: I'm not sure if the word "**REQUIRED**" here actually conforms to RFC 2119, as it doesn't prescribe any behavior.
Maybe the initial paragraph referencing the RFCs and listing the keywords, or the first paragraph under "Schema" should get an additional sentence like this:

> A sentence consisting of just the word REQUIRED in a field definition means that including the field is REQUIRED when sending the containing object.

versions/3.0.md
```diff
@@ -2,7 +2,7 @@
 
 #### Version 3.0.0-rc2
 
-The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC2119](http://www.ietf.org/rfc/rfc2119.txt).
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC8174](https://tools.ietf.org/html/rfc8174) when, and only when, they appear in all capitals, as shown here.
```
RobDolinMS: According to RFC8174, should the full language be: 

      The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL
      NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED",
      "MAY", and "OPTIONAL" in this document are to be interpreted as
      described in BCP 14 [RFC2119] [RFC8174] when, and only when, they
      appear in all capitals, as shown here.

versions/3.0.md
RobDolinMS: While you're editing this line, would you want to also make it one sentence per line?

RobDolinMS:
Thank you @MikeRalphson for this PR.  I'm a fan of the clear use of CAPITALIZATION to indicate BCP14 words.  I added two suggestions on specific lines.

versions/3.0.md
MikeRalphson: I think elsewhere table cells are always written on one line?

versions/3.0.md
MikeRalphson: Will update.

webron:
I've yet had a chance to review the PR but something did catch my eye. The 'Required.' in field descriptions is our indication of a required field. It should not be capitalized.

MikeRalphson:
> The 'Required.' in field descriptions is our indication of a required field. 

Well, yes, obviously. ;) 

> It should not be capitalized.

Happy to revert, but in what sense does **Required.** not "mean that the definition is an absolute requirement of the specification." ?

I specifically wanted to avoid *increasing* ambiguity by saying REQUIRED was only a key word when capitalised, then having it **not** capitalised where it does in fact mandate behaviour.

We could add extra wording?

webron:
Perhaps I shouldn't have jumped before reading the entire PR and conversation - I just didn't want extra work to be put into it for nothing. Unfortunately, I can't review it all right now.

versions/3.0.md
darrelmiller: The table cells have to be one line.  At least in the past that was the case.

RobDolinMS:
#TDC @robdolinms and @fehguy  are both thumbs-up.

PR #1153, opened 2017-06-08
Fix version in path for petstore server

OJFord:
This PR fixes the petstore example to use /v2 in the path rather than /v1, that at the time of writing 404s every request.

c.f. #1144


OJFord:
Actually, nevermind. The schema here is way out of sync with the live API ([see docs here](http://petstore.swagger.io/)) - so I can't see that it's of benefit to anyone to have the correct URL on an otherwise incorrect schema. (Even the most straightforward `GET /pets` is no longer functional.)

PR #1159, opened 2017-06-10
[Implementations] Added editing tools section with known OpenAPI 3.0 Editors

tedepstein:
Not sure if you want to include commercial tools, or just open source components.  If this is only for open source tooling, you can remove the RepreZen API Studio row from the table.  (Or comment on the PR, and I'll remove it.)

### DIFF
--- a/IMPLEMENTATIONS.md
+++ b/IMPLEMENTATIONS.md
@@ -20,6 +20,15 @@ These tools are not necessarily endorsed by the OAI.
 | Tavis.OpenApi | [GitHub/tavis-sofware/Tavis.OpenApi](https://github.com/tavis-software/Tavis.OpenApi/) | dotnet | C# based parser with definition validation and migration support from V2 |
 
 
+#### Editors
+
+| Title          | Project Link | Language |Description                          |
+|----------------|--------------|----------|---------------------|
+| KaiZen OpenAPI Editor | [GitHub/RepreZen/KaiZen-OpenAPI-Editor](https://github.com/RepreZen/KaiZen-OpenAPI-Editor) | Java | Eclipse Editor for OpenAPI 2.0 and 3.0 |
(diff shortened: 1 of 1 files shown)


IMPLEMENTATIONS.md
```diff
@@ -20,6 +20,15 @@ These tools are not necessarily endorsed by the OAI.
 | Tavis.OpenApi | [GitHub/tavis-sofware/Tavis.OpenApi](https://github.com/tavis-software/Tavis.OpenApi/) | dotnet | C# based parser with definition validation and migration support from V2 |
 
 
+#### Editors
+
+| Title          | Project Link | Language |Description                          |
```
ePaul: Is the language in which it is implemented actually relevant for an editor?

MikeRalphson: Possibly more useful would be a `platform` column with e.g. `Eclipse` or `web` in these cases.

RobDolinMS:
I wonder if the whole implementations.md file should be moved to the OpenAPI website instead of being a Markdown file on GitHub.

fehguy:
I always considered it a temporary solution to try to get more eyeballs on the implementations.  I don't think it belongs on the OAI site because (a) it is too slow to update and (b) it may be construed as favoring a particular project.  Where it is now, users (like @tedepstein did) can send a PR to get their project added.

RobDolinMS:
#TDC: We want to have something like this in GitHub so that TDC can add/edit (instead of public website where it may be harder or more expensive to edit.)

RobDolinMS:
#TDC: We want to include ALL tools (OSS + closed source, free + paid, etc.)

#TDC: It would be nice if the OAI website could be driven from GitHub.  (RAML.org does this)

#TDC: It would also be nice to have a semi-official list of data types; maybe as a separate repo.

See #1246 for more discussion.

PR #1163, opened 2017-06-11
edit of Fixed Fields

krishahn:


versions/3.0.md
```diff
@@ -180,9 +180,9 @@ It combines what previously was the Resource Listing and API Declaration (versio
...
 <a name="oasInfo"></a>info | [Info Object](#infoObject) | **REQUIRED**. Provides metadata about the API. The metadata can be used by the clients if needed.
-<a name="oasServers"></a>servers | [[Server Object](#serverObject)] | An array of Server Objects which provide connectivity information to a target server. If the `servers` property is not provided, or is an empty array, the default value would be a [Server Object](#serverObject) with a [url](#serverUrl) value of `/`.
+<a name="oasServers"></a>servers | [[Server Object](#serverObject)] | An array of Server Objects that provides connectivity information to a target server. If the `servers` property is not provided, or is an empty array, the default value would be a [Server Object](#serverObject) with a [url](#serverUrl) value of `/`.
```
MikeRalphson: It is the server objectS which provide the connectivity, not the array, so I think 'which provide' is correct.

krishahn: ok, tks. Corrected PR to make the verb match Server Objects and added a comma before which to clarify things.

versions/3.0.md
```diff
@@ -180,9 +180,9 @@ It combines what previously was the Resource Listing and API Declaration (versio
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="oasVersion"></a>openapi | `string` | **REQUIRED**. Specifies the OpenAPI Specification version being used. It can be used by tooling Specifications and clients to interpret the version. The structure SHALL be `major`.`minor`.`patch`, where `patch` versions MUST be compatible with the existing `major`.`minor` tooling. Typically patch versions will be introduced to address errors in the documentation, and tooling SHOULD typically be compatible with the corresponding `major`.`minor` (3.0.\*). Patch versions will correspond to patches of this document.
+<a name="oasVersion"></a>openapi | `string` | **REQUIRED**. Specifies the OpenAPI Specification version being used. It can be used by tooling specifications and clients to interpret the version. The structure SHALL be `major`.`minor`.`patch`, where `patch` versions MUST be compatible with the existing `major`.`minor` tooling. Typically patch versions will be introduced to address errors in the documentation, and tooling SHOULD typically be compatible with the corresponding `major`.`minor` (3.0.\*). Patch versions will correspond to patches of this document.
```
DavidBiesack: Should we specify that the OpenAPI Specification version is a [Semantic Version 2.0.0 version number](http://semver.org/spec/v2.0.0.html)?

DavidBiesack: This is inconsistent with #1164 for "OpenAPI Version String" which adds, "The patch MAY be suffixed by a hyphen and extra alphanumeric characters." Remove redundancy; I think it is sufficient here to here note that this string is the `[OpenAPI Version String](#oasVersionString)` without repeating the major.minor.patch syntax (i.e. link to the immediately following section).

I think the information here about compatibility and tooling belongs in the **OpenAPI Version String** section below, or better in a higher level section which discusses OAS versioning (not just the version of the OpenAP definition file). The last two paragraphs are also more about the OAS versioning and not really the OpenAPI Version String in the Info Object.

Also, if we do **not** say that OAS uses semver, then we should be more specific here and should say that `major`, `minor` and `patch` are non-negative integers and `patch` may have a suffix consisting of a hyphen `-` and additional alphanumeric characters. (But I suggest we instead say we use semver so we don't have to define all that.)

**Note** the current 3.0.0-rc2 is a *prerelease* version which does **not** conform to this syntax here (unless we diverge from semver and claim that `patch == 0.rc2` but I do not think we should do that.

webron:
Approved by @OAI/tdc.

PR #1168, opened 2017-06-11
edit Parameter Locations and its Fixed Fields

krishahn:
Is something missing in the last sentence of allowEmptyValue? If [`style`](#parameterStyle) is used, if behavior is `n/a`, the value of `allowEmptyValue` SHALL be ignored.

... and if behavior is n/a, ... ?

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -969,7 +969,7 @@ Field Name | Type | Description
 <a name="parameterDescription"></a>description | `string` | A brief description of the parameter. This could contain examples of use.  [CommonMark syntax](http://spec.commonmark.org/) can be used for rich text representation.
 <a name="parameterRequired"></a>required | `boolean` | Determines whether this parameter is mandatory. If the [parameter location](#parameterIn) is "path", this property is **REQUIRED** and its value MUST be `true`. Otherwise, the property MAY be included and its default value is `false`.
 <a name="parameterDeprecated"></a> deprecated | `boolean` | Specifies that a parameter is deprecated and SHOULD be transitioned out of usage.
-<a name="parameterAllowEmptyValue"></a> allowEmptyValue | `boolean` | Sets the ability to pass empty-valued parameters. This is valid only for `query` parameters and allows sending a parameter with an empty value. Default value is `false`. If [`style`](#parameterStyle) is used, if behavior is `n/a`, the value of `allowEmptyValue` SHALL be ignored.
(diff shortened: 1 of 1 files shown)


versions/3.0.md
```diff
@@ -972,25 +972,25 @@ Field Name | Type | Description
 <a name="parameterAllowEmptyValue"></a> allowEmptyValue | `boolean` | Sets the ability to pass empty-valued parameters. This is valid only for `query` parameters and allows sending a parameter with an empty value. Default value is `false`. If [`style`](#parameterStyle) is used, if behavior is `n/a`, the value of `allowEmptyValue` SHALL be ignored.
 
 The rules for serialization of the parameter are specified in one of two ways.
-For simpler scenarios, a [`schema`](#parameterSchema) and [`style`](#parameterStyle) can be used to describe the structure and syntax of the parameter.
+For simplicity, a [`schema`](#parameterSchema) and [`style`](#parameterStyle) can describe the structure and syntax of the parameter.
```
MikeRalphson: Prefer previous wording.

krishahn: Tks for letting me know. I'll stop removing "be used to", assuming that's what you're referring to.  Using "simpler scenarios" begs the question, simpler than what?

MikeRalphson: No, it was the 'simpler scenarios' I preferred, it indicates there are **two** different scenarios, a simpler one and one described in the subsequent paragraph ('For more complex scenarios...'), rather than a simple and complex way of doing the same thing. It is effectively a run-on sentence with a table in the middle of it. The 'complex scenario' does still need an example.

krishahn: ah, ok fixing. tks.

krishahn:
Behavior is mentioned again in the Fixed Fields of the Encoding Object, "The behavior follows the same values allowed for `query` parameters, including default values." Is behavior in Parameter Object and Encoding Object referring to serialization behavior?

webron:
The behavior refers to the serialization. If it's `n\a` in the [table](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#style-examples) - it means it cannot be serialized. 

That table has an `empty` column which is related to the `allowEmpty` value. If when there's an empty value, there's no serialization defined for it, then the `allowEmpty` value has no meaning. Instead of saying it's not allowed to specify it, we just say that it will be ignored if it exists.

PR #1173, opened 2017-06-12
edit 'Considerations for file uploads' and 'Support for x-www-form-ur…

krishahn:
…lencoded request bodies'

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -1343,7 +1343,7 @@ application/json:
       $ref: "#/components/examples/frog-example"
 ```
 
-##### Considerations for file uploads
+##### Considerations for File Uploads
 
 In contrast with the 2.0 specification, `file` input/output content in OpenAPI is 
 described with the same semantics as any other schema type. Specifically:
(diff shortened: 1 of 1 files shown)


versions/3.0.md
```diff
@@ -1409,7 +1409,7 @@ requestBody:
             properties: {}
 ```
 
-Note that in the above example, the contents in the `requestBody` MUST be stringified per [RFC1866](https://tools.ietf.org/html/rfc1866/) when being passed to the server.  In addition, the `address` field complex object will be stringified as well.
+In this example, the contents in the `requestBody` MUST be stringified per [RFC1866](https://tools.ietf.org/html/rfc1866/) when being passed to the server.  In addition, the `address` field complex object will be stringified.
```
DavidBiesack: consider dropping "being" as in "when passed to the server".

versions/3.0.md
```diff
@@ -1360,9 +1360,9 @@ schema:
   format: binary
 ```
 
-Note that the above examples apply to either input payloads (i.e. file uploads) or response payloads.
+These examples apply to either input payloads of file uploads or response payloads.
 
-A `requestBody` example for submitting a file in a `POST` operation therefore may look like the following:
+A `requestBody` for submitting a file in a `POST` operation, however, may look like the following example:
```
DavidBiesack: I this this should drop the "however".

versions/3.0.md
krishahn: +1

versions/3.0.md
krishahn: +1

PR #1177, opened 2017-06-13
edit Patterned Fields in Responses Object

krishahn:


versions/3.0.md
```diff
@@ -1534,7 +1534,7 @@ Field Name | Type | Description
...
+<a name="responsesCode"></a>[HTTP Status Code](#httpCodes) | [Response Object](#responseObject) \| [Reference Object](#referenceObject) | Any [HTTP status code](#httpCodes) can be used as the property name, but only one property per code, to describe the expected response for that HTTP status code.  A [Reference Object](#referenceObject) can link to a response that is defined in the [OpenAPI Object's components/responses](#componentsResponses) section. This field MUST be enclosed in quotation marks for compatibility between JSON and YAML (i.e. "200"), and MAY contain the uppercase character, `X` to designate a wildcard, such as `2XX` to represent all response codes between `[200-299]`. If a response range is defined, and an explicit code within that range is defined, the explicit code definition takes precedence over the range definition for that code.
```
DavidBiesack: TDC should review this - does the OpenAPI Specification intent to allow just `1XX, 2XX, 3XX, 4XX, 5XX` (which is stricter than what is described here? If so, this may need tightening up to say:

*`XX` can replace the last two digits of the HTTTP status code...*

**Or** should the spec allow `XXX` (which may be possible but I expect pretty rare)? 

Or allow `X04` or `20X` or 'X3X` and other variation which the above sentence allows but which really don't make much sense?

This impacts how tools would work to support this, and it may be overly complex to support all these wildcard formats.

PR #1187, opened 2017-06-13
edit Response Payload Example

krishahn:


versions/3.0.md
```diff
@@ -2046,15 +2046,15 @@ ColorSelection:
 
 Would produce three links with the `colorName` of `red`, `green`, and `blue`:
 
-As with all links, it is at the clients' discretion to follow them, neither 
-permissions nor the ability to make a successful call to that link is guaranteed 
+Clients follow all links, including these, at their own discretion. Neither
```
DavidBiesack: Can drop ', including these," here. I think.

versions/3.0.md
```diff
@@ -2280,7 +2280,7 @@ components:
       href: 'https://na2.gigantic-server.com/#/paths/~12.0~1repositories~1{$response.body#/username}'
 ```
 
-Note that in the use of `operationRef`, the _escaped forward-slash_ is necessary when 
+Prefixing `operationRef` using the _escaped forward-slash_ is necessary when
```
DavidBiesack: This needs clarification because the URL includes both escaped and unescaped forward slashes. 

The URL (right hand side) here is a reference to the path `/2.0/repositories/` and the `/` characters in that path are escaped by converting them to the character sequence `~1`. (that is a tilde and the numeral one), yielding the JSON path element `~12.0~1repositories~1` . These forward slashes are escaped because this part of the JSON reference refers to an element in the JSON document with the key `/2.0/repositories/` -- a child of the `paths` object in that OpenAPI document.

The `/` characters in the other part of the href, namely `https://na2.gigantic-server.com/#/paths/` should **not** be escaped because the first part is the location of an OpenAPI document and `#/paths/` is the path to the children of the `paths` object.

The final forward slash in `{$response.body#/username}` is **not** escaped because that is a second (nested) JSON reference which refers to the top-level `username` attribute in the runtime response body, such as
```json
{
   "id" : "u284230",
  "username" : "darrelmiller"
}
```
Forward slashes would be necessary here if and only if items in the response body themselves contain forward slashes, such as accessing the members in  the the following JSON response body:
```json
{
  "codeNames" : {
     "mi/0" : "darrelmiller",
     "mi/6" : "u284230",
   }
}
```
If one wanted to insert the value associated with `"mi/6"` into the link URL, that forward slash would have to be escaped: `{$response.body#/codeNames/mi~16}`

Again, @darrelmiller  please confirm. Perhaps this level of detail is not necessary, but some explanation of slash escaping should be here.

versions/3.0.md
```diff
@@ -2280,7 +2280,7 @@ components:
       href: 'https://na2.gigantic-server.com/#/paths/~12.0~1repositories~1{$response.body#/username}'
```
DavidBiesack: I believe this should be `operationRef:` not `href:` - @darrelmiller please confirm.

darrelmiller:
Much of this has changed in in my #1103 PR

PR #1195, opened 2017-06-14
edit Composition and Inheritance

krishahn:


krishahn:
Does "Override the schema name by overriding the property with a new value," mean override the schema name by giving the property a new value? If so, it might be better not to use "overriding" again.

PR #1199, opened 2017-06-14
Semantic Versioning

DavidBiesack:
Document how OAS uses semantic versioning (semver).
    
    The specification itself is versioned using semver, so a new top level section
    describes specification versioning, what OAS 3.0 means, how patch levels are used,
    and how tools use the version.
    
    The `openapi` field in the root object is a semver string that identifies
    which OAS the definition file uses and conforms to.
    
    Added a definition of "OpenAPI definition file" to the Definitions section.
    
    Removed the "OpenAPI Version String" section as it is not needed:
    info about the `openapi` field is in the table,
     and info about the spec version is in the OpenAPI Version section.

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -20,10 +20,12 @@ Additional utilities, such as testing tools, can also use the files.
 <!-- TOC depthFrom:1 depthTo:3 withLinks:1 updateOnSave:1 orderedList:0 -->
 
 - [Definitions](#definitions)
+	- [OpenAPI Definition File](#oasDefinitionFile)
 	- [Path Templating](#pathTemplating)
 	- [Media Types](#mediaTypes)
 	- [HTTP Status Codes](#httpCodes)
 - [Specification](#specification)
+	- [Versions](#versions)
 	- [Format](#format)
 	- [File Structure](#fileStructure)
 	- [Data Types](#dataTypes)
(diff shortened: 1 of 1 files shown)


versions/3.0.md
```diff
@@ -94,6 +99,17 @@ The available status codes are defined by [RFC7231](http://tools.ietf.org/html/r
...
+This document specifies OpenAPI Specification version `3.0.0-rc2`.
+
+The `major`.`minor` portion of the semver (for example `3.0`) SHALL designate the OAS feature set. Typically, *`.patch`* versions address errors in this document, not the feature set. Tooling which supports OAS 3.0 SHOULD be compatible with all OAS 3.0.\* versions. The patch version SHOULD NOT be considered by tooling, making no distinction between `3.0.0` and `3.0.1` for example.
+
+Subsequent minor version releases of the OpenAPI Specification (incrementing the `minor` version number) SHOULD NOT interfere with tooling developed to a lower minor version and same major version.  Thus a hypothetical `3.1.0` specification SHOULD be usable with tooling designed for `3.0.0`.
+
+An OpenAPI definition file compatible with OAS 3.\*.\* contains a required [`openapi`](#oasVersion) field which designates the semantic version of the OAS that it uses. (OAS 2.0 definition document contain a top-level version field named [`swagger`](http://swagger.io/specification/#swaggerObject) and value `"2.0"`.)
+
```
darrelmiller: If we are adding a reference to the semver spec in description of the "openapi" property I see no reason to try and summarize what semver means here.

RobDolinMS:
#TDC: Please take a look and comment.  Let's discuss on Fri, 16 June.

versions/3.0.md
fehguy: Agree, but it doesn't hurt, given all the lazy readers out there...

DavidBiesack:
I've removed the
> This document specifies OpenAPI Specification version `3.0.0-rc2`.

statement

PR #1201, opened 2017-06-15
Update link to RFC7235 section in Security Scheme `scheme` field description

brendo:
The existing link goes to [Section 4.2](https://tools.ietf.org/html/rfc7235#section-4.2) of RFC 7235 which is useful, but didn't contain the content that I expected after reading the description of the `scheme` field.

I expected to see details that [Section 5.1](https://tools.ietf.org/html/rfc7235#section-5.1) contained, or more correctly, the link within Section 5.1 to the [Authentication Scheme Registry](https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml).

Is my understanding of `scheme` is correct, the valid values are actually those listed in Authentication Scheme Registry (`Bearer`, `Basic`, `Digest` etc.)?

If so, I believe this patch will help others arrive at that conclusion too 👍


darrelmiller:
Makes sense to me.  Thanks @brendo

RobDolinMS:
Thanks @brendo

brendo:
You are welcome!

PR #1202, opened 2017-06-15
Schema refs begone

### LINKED ISSUE #1190: Doc fix - Schema and References
The Schema Object contains the Reference Object. As such, there's no need to explicitly say that a value can be either a Schema Object or a Reference Object (see https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#mediaTypeSchema for example).

webron:
Resolves #1190.

This is more of a proposal, following the previous version of the spec. Since the Schema Object actually contains the Reference Object, there's no need to explicitly say when a Schema Object can be used, a Reference Object can be used as well - it's redundant. This is just a doc cleanup.

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -443,7 +443,7 @@ All objects defined within the components object will have no effect on the API 
 
 Field Name | Type | Description
 ---|:---|---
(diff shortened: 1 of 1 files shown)


darrelmiller:
I can't say I'm a fan of this change.

RobDolinMS:
#TDC: @darrelmiller suggests we not make this change as it would be inconsistent with other references
#TDC: Closing

PR #1204, opened 2017-06-15
Clarified path matching

### LINKED ISSUE #1127: Clarification on path templates that overlap another path
Apologies if I'm missing something obvious, but I was trying to find a definitive statement that describes whether/how overlapping paths should work.  For example, consider two paths like:
* /one
* /{template}

I browsed the spec but didn't see it.  I browsed through Issue #574 and related issues, but those seem to cover more advanced cases.

I think its generally understood that this is supported (just look at http://petstore.swagger.io/) and that the concrete/fixed path(s) takes 'precedence' over the path template or path wildcard, but I was just looking for some re-assurance from the spec.  Hopefully something I just missed?

webron:
Resolves #1127.

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -632,9 +632,32 @@ The path is appended to the URL from the [`Server Object`](#serverObject) in ord
 
 Field Pattern | Type | Description
 ---|:---:|---
-<a name="pathsPath"></a>/{path} | [Path Item Object](#pathItemObject) | A relative path to an individual endpoint. The field name MUST begin with a slash. The path is **appended** (no relative URL resolution) to the expanded URL from the [`Server Object`](#serverObject)'s `url` field in order to construct the full URL. [Path templating](#pathTemplating) is allowed.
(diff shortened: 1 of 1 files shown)


versions/3.0.md
```diff
@@ -632,10 +632,33 @@ The path is appended to the URL from the [`Server Object`](#serverObject) in ord
...
+```
+  /pets/{petId}
+  /pets/mine
+```
+
+The following paths are considered identical and invalid:
+
+```
+  /pets/{petId}
+  /pets/{name}
```
DavidBiesack: Some APIs may use unique `pattern` regular expressions for the path parameter schema, in which case these are not identical or ambiguous. (Spring [`@RequestMapping`](https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/web/bind/annotation/RequestMapping.html) supports patterns in `@PathVariable`). 

Thus, I suggest marking this as potentially ambiguous, but not invalid.

webron: It _is_ invalid today. We never discussed differentiation based on the parameter itself.

darrelmiller: I don't like the idea of being able to disambiguate operations based on parameter values.  ASP.NET allows it too and I find it unnecessary complexity.

PR #1208, opened 2017-06-15
Clarified multi media types - for encoding object only

### LINKED ISSUE #1124: Content media type should allow.... multiple media types?
Today under the `Encoding Object` we can specify:
`contentType: image/png, image/jpeg`

It seems like it would almost make sense to support multi-media-types when it comes to `content` definitions (which is more than then wildcards we support today).

webron:
~~Resolved #1124.~~

This PR does 3 things:
- Adds a description to Encoding Object's `contentType` explaining the ability to use wildcards and multi-types.
- ~~Changes the description of the `content` property in the Parameter Object, Response Object and Request Body Object to allow similar keys (this is a functionality change).~~
- Fixes an HTML anchor in for one of the properties.

Update: After the TDC call we've decided to not make the change to the different `content` maps. We'll revisit it based on user feedback. For now, this PR just adds the clarifications and fixes mentioned above.


versions/3.0.md
```diff
@@ -989,7 +989,7 @@ When `example` or `examples` are provided in conjunction with the `schema` objec
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="parameterContent"></a>content | Map[`string`, [Media Type Object](#mediaTypeObject)] | A map containing the representations for the parameter. The key is the media type and the value describes it.  The map MUST only contain one entry.
+<a name="parameterContent"></a>content | Map[`string`, [Media Type Object](#mediaTypeObject)] | A map containing the representations for the parameter. The key is the media type and the value describes it.  The key can be a single media type (e.g. `application/json`), a wildcard media type (e.g. `image/*`), or a comma-separated list of the two types. The map MUST only contain one entry.
```
brendo: The wording specifically refers to `two types`. Is that to say `image/png, image/gif, image/jpeg` is invalid (3 types?). Would "comma-separated list of types" be a more generic way to document multi types?

brendo: Or is this wording meant to mean "you can use a mixture of media types or wildcard media types", eg. "image/*, application/json"?

webron: The latter. If you have a different wording suggestion, I would be happy to modify.

brendo: Hmm, my suggestion is "The key can be singular (e.g. `application/json`), a wildcard (e.g. `image/*`), or a comma-separated list of both (e.g. `image/*, application/json, text/plain`).".

I'm not sure it's much better though!

webron: How's the new wording?

brendo: Better! Amazing how one word makes a difference!

evanread: How about:

The key can be one of the following:

- A media specific type/subtype pair. For example, `application/json`.
- A media type with a wildcard. For example, `image/*` or `*`.
- A comma separated list combining media types together. For example, `application/json, image/*`.

whitlockjc:
:shipit:

PR #1209, opened 2017-06-16
multipart clarifications

### LINKED ISSUE #954: Two references to 'parameterContent' seem incomplete /  left-over ?
> When passing complex objects in the x-www-form-urlencoded content type, the default serialization strategy of such properties is described in the parameterContent section as form.

> The Content-Type to use for encoding a specific property. See (#parameterContent) for details on the style property.

Also what looks like another broken link:

> The following shows how multiple servers can be described, for example, at the OpenAPI Object's servers(#oasServers):

webron:
Summary of changes:
- Mentioned `encoding` applies to `x-www-form-urlencoded` as well, changed `SHOULD` to `SHALL`.
- Fixed wording of how complex objects work with `x-www-form-urlencoded` (Resolves #954).
- Added an example of multi-file upload.
- Under the Encoding Object, added restrictions to the properties that can only be applied to certain request body media types. Wording was chosen to make validation simpler.


versions/3.0.md
```diff
@@ -1298,7 +1298,7 @@ Field Name | Type | Description
...
 <a name="mediaTypeExamples"></a>examples | Map[ `string`, [Example Object](#exampleObject) \| [Reference Object](#referenceObject)] | Examples of the media type.  Each example object SHOULD  match the media type and specified schema if present.  The `examples` object is mutually exclusive of the `example` object.  Furthermore, if referencing a `schema` which contains an example, the `examples` value SHALL _override_ the example provided by the schema.
-<a name="mediaTypeEncoding"></a>encoding | Map[`string`, [Encoding Object](#encodingObject)] | A map between a property name and its encoding information. The key, being the property name, MUST exist in the schema as a property. The encoding object SHOULD only apply to `requestBody` objects when the content type is `multipart`.
+<a name="mediaTypeEncoding"></a>encoding | Map[`string`, [Encoding Object](#encodingObject)] | A map between a property name and its encoding information. The key, being the property name, MUST exist in the schema as a property. The encoding object SHALL only apply to `requestBody` objects when the content type is `multipart` or `x-www-form-urlencoded`.
```
webron: Fixed!

webron: I agree it should say `application/x-www-form-urlencoded`. I'll find the example you mentioned and fix that too. 

My concern with `multipart` vs `multipart/*` is that neither is accurate. My hope is that with the first it would be easier for people to infer that it applies to all multipart options, whereas the latter might make people understand that it only applies to the wildcard. Additional wording seems to complicate the sentence too much. I'll fix the others first, and wait for additional feedback.

webron:
I'll convert all `multipart/*` to `multipart` and merge.

PR #1217, opened 2017-06-16
Fixed ambuiguity with link parameters

darrelmiller:
This implementation isn't quite how it was discussed but I think it resolves some of the concerns that were raised.

Here is an example:
```yaml
links:
  postLink:
    operationId: SomePostOperation
    parameters:
      path.id: $request.path.id
    requestBody:  $response.body#/some/inner/content

```
Link parameter keys are now defined as `[{in}.]{name}`.  So the target parameter name can be prefixed by one of the values defined by the parameter `in` property.
Specifying the request body to pass to the target operation is in its own property because, it is no longer a 'parameter' so I don't want to confuse the issue by pretending it is.


versions/3.0.md
```diff
@@ -1978,8 +1978,9 @@ For computing links, and providing instructions to execute them, a [runtime expr
...
+<a name="linkOperationId"></a>operationId  | `string` | the name of an _existing_, resolvable OAS operation, as defined with a unique `operationId`.  This field is mutually exclusive with the `operationRef` field.  Relative `operationRef` values MAY be used to locate an existing [Operation Object](#operationObject) in the OAS. 
+<a name="linkParameters"></a>parameters   | Map[`string`, Any \| [{expression}](#runtimeExpression)] | A map representing parameters to pass to an operation as specified with `operationId` or identified via `operationRef`. The key is the parameter name to be used, whereas the value can be a constant or an expression to be evaluated and passed to the linked operation. The parameter name can be qualified using the [parameter location](#parameterIn) `[{in}.]{name}` for operations that use the same parameter name in different locations (e.g. path.id).
+<a name="linkRequestBody"></a>requestBody | Any \| [{expression}](#runtimeExpression) | A literal value or [{expression}](#runtimeExpression)] to use as a request body when calling the target operation.
```
earth2marsh: Looks like a spurious `]` after the 2nd `runtimeExpression`. Otherwise LGTM.

darrelmiller: Thanks @earth2marsh.  I fixed it in the redo-grammar PR that is still open.

PR #1223, opened 2017-06-18
Make callback example link relative

MikeRalphson:
Saves having to remember to update it later.

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -1820,7 +1820,7 @@ The key value used to identify the callback object is an expression, evaluated a
 ##### Patterned Fields
 Field Pattern | Type | Description
 ---|:---:|---
-<a name="callbackExpression"></a>{expression} | [Path Item Object](#pathItemObject) | A Path Item Object used to define a callback request and expected responses.  A [complete example](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/examples/v3.0/callback-example.yaml) is available.
+<a name="callbackExpression"></a>{expression} | [Path Item Object](#pathItemObject) | A Path Item Object used to define a callback request and expected responses.  A [complete example](../examples/v3.0/callback-example.yaml) is available.
 
 This object can be extended with [Specification Extensions](#specificationExtensions). 
 


versions/3.0.md
```diff
@@ -1820,7 +1820,7 @@ The key value used to identify the callback object is an expression, evaluated a
 ##### Patterned Fields
 Field Pattern | Type | Description
 ---|:---:|---
-<a name="callbackExpression"></a>{expression} | [Path Item Object](#pathItemObject) | A Path Item Object used to define a callback request and expected responses.  A [complete example](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/examples/v3.0/callback-example.yaml) is available.
+<a name="callbackExpression"></a>{expression} | [Path Item Object](#pathItemObject) | A Path Item Object used to define a callback request and expected responses.  A [complete example](/examples/v3.0/callback-example.yaml) is available.
```
darrelmiller: Doesn't that need a `..` at the front of that URL?

MikeRalphson: Seems to work ok in my fork, i.e. it's a repo-root relative url.

ePaul: I guess the difference is that with the `..` it points to the example in the current branch, without it points to master (where this file doesn't exist yet).

MikeRalphson: The link in my fork points to https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/examples/v3.0/callback-example.yaml - not the master branch? Viewing the changes I see a link to the [commit blob](https://github.com/MikeRalphson/OpenAPI-Specification/blob/7dd9891b20ceb7232c9704dbc7b3f7bbff30204e/examples/v3.0/callback-example.yaml)

ePaul: Hmm, I looked at it in the [differences view](https://github.com/OAI/OpenAPI-Specification/pull/1223/files?short_path=4a97aef#diff-4a97aef5bba9ddb7fd8609f48f4bdc30), and there it pointed to https://github.com/OAI/OpenAPI-Specification/blob/master/examples/v3.0/callback-example.yaml. Strangely, if I open the same address again in a new browser tab, it now points to https://github.com/OAI/OpenAPI-Specification/blob/rel_link/examples/v3.0/callback-example.yaml.  (Both links give 404s currently, the second one uses your branch name with the main fork's repository name.)

I guess Github it not that predictable with the link expansion.

MikeRalphson: Thanks @ePaul for looking at it again. Let's try it with the ..

PR #1228, opened 2017-06-19
Fix callback-example.yaml

tfesenko:
The runtime expression [must be quoted](http://www.yaml.org/spec/1.2/spec.html#id2788859) as it starts with a '{' - an [indicator character](http://www.yaml.org/spec/1.2/spec.html#c-indicator), otherwise it's invalid YAML.

I also added the 'openapi' and 'info' properties because they are required and, thus, the spec file is invalid without them.

### DIFF
--- a/examples/v3.0/callback-example.yaml
+++ b/examples/v3.0/callback-example.yaml
@@ -1,3 +1,7 @@
+openapi: 3.0.0
+info: 
+  title: Callback Example
+  version: 1.0.0
 paths:
   /streams:
     post:
@@ -32,7 +36,7 @@ paths:
         onData:
           # when data is sent, it will be sent to the `callbackUrl` provided
           # when making the subscription PLUS the suffix `/data`
-          {$request.query.callbackUrl}/data:
+          '{$request.query.callbackUrl}/data':
             post:
               requestBody:
                 description: subscription payload
--- a/examples/v3.0/callback-example.yaml
+++ b/examples/v3.0/callback-example.yaml
@@ -1,7 +1,4 @@
-openapi: 3.0.0
-info: 
-  title: Callback Example
-  version: 1.0.0
+# ...
 paths:
   /streams:
     post:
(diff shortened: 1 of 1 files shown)


MikeRalphson:
Was this example ever intended to be complete / self-contained? It also has a missing definition of `#/components/schemas/UserLogData`

RobDolinMS:
#TDC: @tfesenko Thank you for submitting this PR.  Would you please update per @MikeRalphson's comment so this is a complete example.  

You may want to align with what @whitlockjc is drafting for Issue #1245

whitlockjc:
@tfesenko If you would like to get this addressed ASAP and not wait for #1245, I would just make the following changes to your PR as it will align with our plans:

1. Replaces lines 1-4 with `# ...` _(this will be our way to indicate "omitted for brevity")_
2. Resolve the `$ref` on line 51:
    a. Add `# ...` as the last line of the document
    b. Replace `$ref` with an inline schema

I guess 2.a could be documented in #1245 but for now, being explicit might not be a bad thing.

RobDolinMS:
ACTION: @tedepstein to reach-out to @tfesenko when she's back (likely Mon, 7/10)

tfesenko:
@whitlockjc , I addressed your suggestions in commit ccf0169. Can you please have a look? Thanks!

whitlockjc:
:shipit:

whitlockjc:
Since this is a non-spec change, we should be able to branch/merge into `master` instead of a release-specific branch.  I will have to merge this manually.

whitlockjc:
Merged manually 542fc445ced49073d277b0db526bc0bcccc3234b

whitlockjc:
Weird, it didn't keep your commits but the commit message clearly points to your commit.

whitlockjc:
@tfesenko I apologize the merged commit didn't retain your authorship of the two commits.  I did it this way to avoid asking you to delete this PR and recreate a new one from `master`.  Next time, I'll just ask you to create a new PR so your details in up in the commit log.  I apologize.

PR #1231, opened 2017-06-23
Fixed JSON request example in "Key Expression"

hkosova:
This PR fixes the sample request in the "Key Expression" section:
* added missing comma between the JSON fields,
* updated `Content-Length` to match the actual payload size.


earth2marsh:
LGTM

whitlockjc:
:shipit:

RobDolinMS:
#TDC: We may choose not to fix all of the content length issues; but thank you for fixing this one.  👍

whitlockjc:
I think that since this `Content-Length` fix was with a very valid JSON fix, it made sense.  ;)

PR #1233, opened 2017-06-26
Fixed an invalid example (Encoding Object Example)

EricWittmann:
This is a fix for an invalid sample here:

https://github.com/OAI/OpenAPI-Specification/blob/3.0.0-rc2/versions/3.0.md#encoding-object-example

The tabbing was a bit off and it needed a "schema".


earth2marsh:
LGTM, thanks Eric

whitlockjc:
:shipit:

PR #1236, opened 2017-06-28
OpenAPI v3 JSON Schema

timburks:
From the README:

This PR contains a JSON Schema for OpenAPI v3.

It was generated using a tool that scraped the Markdown text of the OpenAPI Specification and made several semi-automated refinements. This tool is part of the [gnostic project](https://github.com/googleapis/gnostic), which generates type-safe models of OpenAPI v2 and v3 directly from their JSON Schemas.

The v3 JSON Schema was significantly reviewed and improved by [Mike Ralphson](https://github.com/mikeralphson) and [Tanya Fesenko](https://github.com/tfesenko) and has been tested with automated schema validators.

The source for this schema is in the [gnostic project](https://github.com/googleapis/gnostic) in [openapi-3.0.json](https://github.com/googleapis/gnostic/blob/master/OpenAPIv3/openapi-3.0.json) which will also contain the latest revision.


versions/3.0.md
```diff
@@ -1979,7 +1979,7 @@ Field Name  |  Type  | Description
...
-<a name="linkParameters"></a>parameters   | Map[`string`, Any \| [{expression}](#runtimeExpression) | A map representing parameters to pass to an operation as specified with `operationId` or identified via `operationRef`. The key is the parameter name to be used, whereas the value can be a constant or an expression to be evaluated and passed to the linked operation.  The parameter name can be qualified using the [parameter location](#parameterIn) `[{in}.]{name}` for operations that use the same parameter name in different locations (e.g. path.id).
+<a name="linkParameters"></a>parameters   | Map[`string`, Any \| [{expression}](#runtimeExpression)] | A map representing parameters to pass to an operation as specified with `operationId` or identified via `operationRef`. The key is the parameter name to be used, whereas the value can be a constant or an expression to be evaluated and passed to the linked operation.  The parameter name can be qualified using the [parameter location](#parameterIn) `[{in}.]{name}` for operations that use the same parameter name in different locations (e.g. path.id).
```
RobDolinMS: @timburks Can the bonus character be separated into its own PR?

RobDolinMS:
#TDC: @darrelmiller Would like to see some wording to clearly state that:
* Schema only provides partial validation (validation against schema is necessary but not sufficient)
* Source of truth is the spec, not the schema
* Maybe add this in the ReadMe in the Schema folder

#TDC: Additional discussion:
* Should people recognized in the Schema/ReadMe.md just get recognized in the Contributors file?
* A few options for Schema:
  - Put into Examples folder
  - Put into its own repo

versions/3.0.md
timburks: That commit is in a PR that's a head of this one and accidentally got included here. If you need a PR without it, I'll try to revise this one.

timburks:
@RobDolinMS I'm happy to revise this or file a new PR based on these and other suggested changes. Please let me know when you feel there's enough feedback to clarify what's needed in that update.

RobDolinMS:
ACTION: @webron and @whitlockjc, please review the schema.

#TDC: We agree this is not a blocker for releasing the spec; but it would be really, really nice to have at launch.

schemas/v3.0/schema.json
```diff
@@ -0,0 +1,1254 @@
...
+    },
+    "info": {
+      "$ref": "#/definitions/info"
+    },
+    "servers": {
+      "type": "array",
+      "items": {
+        "$ref": "#/definitions/server"
+      },
+      "uniqueItems": true
```
darrelmiller: The uniqueItems constraint brings up an interesting set of issues.  This constraint isn't in the spec. What is the reason for adding it in the Schema?  It seems like it is common sense that the URLs should be unique.  Should that wording be added to the spec?  Can JSON schema validate that the just URLs are unique?

schemas/v3.0/schema.json
```diff
@@ -0,0 +1,1254 @@
...
+      },
+      "properties": {
+        "title": {
+          "type": "string"
+        },
+        "description": {
+          "type": "string"
+        },
+        "termsOfService": {
+          "type": "string"
```
darrelmiller: Should format: uriref be added to termsOfService?

schemas/v3.0/schema.json
timburks: When I was writing the schema generator, I found that the uniqueItems property was generally set for arrays in the v2 schema, so I applied it by default to arrays in the generated 3.0 schema. So in this case, it's present by accident for the "servers" array. 

I'd like to continue to be able to generate the schema automatically from the 3.0 text, so the simplest solution would be to drop "uniqueItems" for all arrays, but I see that the text description for the tags field of the root object includes the phrase "MUST be unique". Based on the presence of that phrase, I could add the uniqueItems property for the array value of the tags field, but no others.

schemas/v3.0/schema.json
timburks: There are four instances of the phrase "MUST be in the format of a URL" in the 3.0 spec. I can add format: uriref for all of them - is "uriref" the proper value to use for this? I don't find that word specifically used anywhere in the 3.0 spec. I do find it in [this document](http://json-schema.org/latest/json-schema-validation.html), but there it seems to have been deprecated in favor of uri-ref.

darrelmiller: The OpenAPI V3.0.0 spec is currently pinned to the [draft 00](https://tools.ietf.org/html/draft-wright-json-schema-validation-00#section-7.3.7) version of JSON Schema, so we are stuck with using the uriref term.

handrews: You might want to consider using `uri-reference` which is what `uriref` became in the next draft.  Since formats are extensible, you would not be violating draft-wright-00 (a.k.a. draft-05) by doing this, and would be future-proofing your approach.  While we are seeing implementations of draft-wright-01 (draft-06) in the wild, I'm not aware of any validators implementing `uriref` from draft-05.  `uriref` did not fit the pattern for format names which is why it was changed.

KES777:
When it will be officially released?

handrews:
Note that I have started building on PR #1270, beginning with PR #1744 with more to come.  Perhaps we should close this so that there is only one line of open PRs?

It's my understanding that #1270 already incorporates ideas from here- @timburks please let me know if you think there are things that I should incorporate that are present here but not in #1270.

webron:
Closing as #1270 was merged. Thanks for the help, @timburks!

Issue #1032, opened 2017-04-07
JSON Schema for the OpenAPI 3.0 Specification


tedepstein:
It looks like we don't yet have a JSON Schema to describe the OpenAPI 3.0 spec.  The [schemas folder](https://github.com/OAI/OpenAPI-Specification/tree/master/schemas) has 1.2 and 2.0 subfolders, but nothing for 3.0 yet.

We'd like to contribute to this.  Is anyone working on a JSON Schema for the 3.0 spec yet?

MikeRalphson:
@timburks and I are [working on one](https://github.com/googleapis/gnostic/blob/master/OpenAPIv3/openapi-3.0.json) over at Google [Gnostic](https://github.com/googleapis/gnostic), which is generated from the spec itself.

My latest version is [here](https://github.com/Mermade/swagger2openapi/blob/master/schemas/openapi-3.json), which incorporates a few manual fixes. Help improving and testing it would be gratefully received.

tedepstein:
@MikeRalphson , @timburks, great idea to auto-generate a schema from the written spec. I haven't looked at it carefully yet, but at first glance it seems like at least a good starting point.

@earth2marsh, you mentioned in your [Google Cloud Next talk](https://www.youtube.com/watch?v=4lBMQteMd6Y) that schema development was one of the next items on your agenda.  Does the TDC or the Apigee team have an OpenAPI v3 JSON schema in some stage of development...?

brandur:
I hope this doesn't seem too much like a content-free +1, but I'd also love to see a JSON schema along with the specification.

I don't really trust myself to be able to accurately read the spec in its entirely, so for 2.0 I fell back heavily on using the included schemas to verify that what I'm generating is actually intelligible (and it worked, they caught many problems).

timburks:
My latest draft is [here](https://github.com/googleapis/gnostic/blob/openapi-v3.0.0-rc2/OpenAPIv3/openapi-3.0.json), semi-automatically generated from the 3.0.md text (v3.0.0-rc2).

brandur:
@timburks @MikeRalphson That's awesome. Thanks for the continued work on that! Just out of curiosity because I don't know the project dynamics here, is there any chance that your JSON schema can be sent upstream and be officially supported as part of the 3.0 release? I haven't seen or heard much from the project maintainers, but I figure that you may have spoken to them through another channel.

timburks:
[...]

PR #1239, opened 2017-06-30
Adding API Version for an operation in Swagger 2.0 Schema

Adrael:
This new **version** property allows us to specify what version of the operation is behind the route.

BTW, thanks for all the work pals!

### DIFF
--- a/schemas/v2.0/schema.json
+++ b/schemas/v2.0/schema.json
@@ -254,6 +254,10 @@
           },
           "uniqueItems": true
         },
+        "version": {
+          "type": "string",
+          "description": "The API version of the operation"
+        },
         "summary": {
           "type": "string",
           "description": "A brief summary of the operation."


MikeRalphson:
* Version 2.0 of the spec is fixed and unchanging, so proposed changes should be against the OpenAPI.next branch
* OpenAPI 3.0.0 is feature complete, so this would be a 3.1.0 change at the earliest
* Changing the schema without changing the markdown spec itself (the source of truth) is not a good idea
* If you need this feature now against version 2.0, consider using a vendor extension such as `x-version`

Adrael:
@MikeRalphson Thanks for the tip.
So basically I have to adapt my changes to the OpenAPI.next branch while also updating the md doc (my bad didn't think of it first time) and you guys will eventually ship it on 3.1.0 release?

MikeRalphson:
> So basically I have to adapt my changes to the OpenAPI.next branch while also updating the md doc (my bad didn't think of it first time) and you guys will eventually ship it on 3.1.0 release?

Yes*, yes and not necessarily. :smile: Not all proposed features make it into the spec. It might help if you could demonstrate a real-world need for this feature.

\* The 3.0.0 schema has not yet been merged (see PR #1032) and I don't think it's currently clear as to whether manual edits will be accepted or the schema will be automatically regenerated as needed.

Adrael:
Ok so I've juste submitted a PR on the OpenAPI.next branch with the updated doc, not changing anything else.
I will provide an example on the PR (https://github.com/OAI/OpenAPI-Specification/pull/1240) directly, closing this one.

Thanks! :)

PR #1240, opened 2017-06-30
Adding support for operation versioning

Adrael:
This new **version** property allows us to specify what version of the operation is behind the route.

BTW, thanks for all the work pals!

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -826,6 +826,7 @@ Describes a single API operation on a path.
 Field Name | Type | Description
 ---|:---:|---
 <a name="operationTags"></a>tags | [`string`] | A list of tags for API documentation control. Tags can be used for logical grouping of operations by resources or any other qualifier.
+<a name="operationVersion"></a>version | `string` | The version of the operation using [Semantic Versioning 2.0.0](http://semver.org/spec/v2.0.0.html) (semver).
 <a name="operationSummary"></a>summary | `string` | A short summary of what the operation does.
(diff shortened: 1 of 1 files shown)


Adrael:
As requested, here is an example of a real-world need for this PR.
Feel free to add your own folks!

As of API versioning, one can choose from multiple solutions to implement and deploy versions.
For example:
- Path - `/api/v1/pets`
- Path - `/pets-api/v1`
- Content-type - `Accept: application/vnd.pets.v1+json`
- Custom header - `api-version: 1`

Obviously, the ones involving URL versioning are simple, you can simply look at the resource to know exactly where you're going.

My case is much more like the last one.
My team and I are using the amazing [swagger-ui](https://github.com/swagger-api/swagger-ui), and we use Header versioning. This offers the ability to version each resources independently.

It would be great to have the ability to choose which resource to use at a glance, like this:
image

RobDolinMS:
@Adrael Do you have concerns with moving this to after v3.0.0 ?

@darrelmiller may have an alternate approach.

Adrael:
@RobDolinMS As long as the idea of this PR gets checked and shipped, I don't see any concerns as of a 3.0.1 or even a 3.1.0 👍

fehguy:
I don't think that moving it to post 3.0 is a guarantee that it'll be moved in.  The point is the discussion will move to after the release.

Adrael:
@fehguy Yes I understand that, no problem!
Thanks guys!

Just curious, do you have any idea about a release date?

PR #1261, opened 2017-07-09
Allow $refs in the EncodingObject#headers map

tfesenko:
As discussed in PR #1126


versions/3.0.md
```diff
@@ -1553,7 +1553,7 @@ A single encoding definition applied to a single schema property.
...
-<a name="encodingHeaders"></a>headers | Map[`string`, [Header Object](#headerObject)] | A string map allowing additional information to be provided as headers, for example `Content-Disposition`.  `Content-Type` is described separately and SHALL be ignored in this section. This property SHALL be ignored if the request body media type is not a `multipart`.
+<a name="encodingHeaders"></a>headers | Map[`string`, [Header Object](#headerObject) \| [Reference Object](#referenceObject)] | A string map allowing additional information to be provided as headers, for example `Content-Disposition`.  `Content-Type` is described separately and SHALL be ignored in this section. This property SHALL be ignored if the request body media type is not a `multipart`.
```
DavidBiesack: Not sure if "A string map" is accurate anymore. It's a map of String -> [Header Object | $ref ]

tfesenko: A good point, @DavidBiesack !
So, the current description is
> A string map allowing additional information to be provided as headers, <...>

Would it be better if I remove the "string" in "a string map":
>  A map allowing additional information to be provided as headers, <...>

DavidBiesack: yes, just say "a map ..."

tfesenko: 'a string map' -> 'a map' - done in de6ec79

Issue #229, opened 2014-12-20
add "null value support" in spec?


fehguy:
Per https://github.com/swagger-api/swagger-core/issues/459 there are cases where passing null is important.  For example:

```
[ null ]
```

will be quite different than `[]`.  See if this can be an option

enkafan:
Love to see this. Especially helpful for us with some of the datetime requirements we have. Right now I'm resorting to `"x-isnullable" : true`

jphastings:
I'd like to see this too. I'm trying to define a `PATCH` endpoint which merges incoming fields of a JSON object over the existing one.

Accepting a document with `age: null` is the way I intend to specify that the value should be overwritten, rather than left as it currently is (which is what would happen when omitting the `age` key from the request).

z0r1k:
Having ```null``` as type would be handy.
Otherwise in Response Class (Status 200) Model Schema you get 
```
{
  "payload": {},
  "status": "ok"
}
```
instead of 
```
{
  "payload": null,
  "status": "ok"
}
```

webron:
@z0r1k - `null` normally makes sense in requests but less so in responses (though I can think of use cases it's useful for responses as well).

In your example, wouldn't it make more sense to just have the following if there's no payload?
```js
{
  "status": "ok"
}
```

To me, `null` has the meaning of "this property existed before, and I want to nullify its value". If there's *no* value, or the value needs not change (for requests, for example), it just shouldn't be sent.

tuukkamustonen:
I'm in similar situation with jphastings (and chimmelb). Allowing null values is required for `PATCH` as I don't want to support empty strings for that. Support for null would be needed.

bpicolo:
+1. Nullability is a sort of awkward thing to not have. It's definitely a distinct case from required. One is KEY must be present. One is VALUE of key must be present.

willangenent:
+1. We're using the good suggestion by @enkafan too.

jasonh-n-austin:
+1 ran into this today, trying to validate responses in tests against Swagger schema.

A suggestion for a non-breaking change: `nullable: true`. This would avoid the problem of API developers being silly and implementing polymorphic fields (although I could see a Postel's Law argument on the request side).

webron:
If you're suggesting `x-nullable` \*cough\* then sure, go ahead and use it for now :)

wking:
[...]

PR #1270, opened 2017-07-14
[WIP] Alternative OAS3 JSON Schema

webron:
This is a proposal for an alternative JSON Schema provided by #1236.

The work done in #1236 is good but:
- It is too permissive - it allows things that should not validate, for example, a parameter with both `content` and `schema` (and a few other cases).
- It doesn't cover types properly.
- As far as I understand, it adds a couple of restrictions that are not provided in the spec itself.

The reasons for creating a new PR instead of suggesting fixes to #1236:
- Editing JSON Schemas can be a pain, and trying to navigate through different styles is not easy.
- The amount of changes to make it less permissive is quite large.
- I had about 80% of the schema ready...

Odd things about this version:
- It's a JSON Schema written in YAML.  It's just easier to write and edit and.. read. Easy enough to convert to JSON if this PR is accepted.
- To make it more restrictive, there are somewhat a lot of duplication in the schema. Using `allOf` to reuse things didn't work for me due to `additionalProperties: false`. If anyone wants to take a crack and minimizing it, by all means... I just wanted to get a working version.

What's missing:
- It doesn't cover checking of `style`/`explode` inside Request Bodies where it's multipart. That was just too much for me for now.
- Testing. I'm willing to bet there are bugs in it.
- JSON Schema decorations (version, id, $schema and so on).

Pinging @darrelmiller as he asked to be pinged.

Pinging @MikeRalphson for assistance in testing - would really appreciate if you can take a crack at it.

I'm sure converting the YAML to JSON is going to be easy enough.

### DIFF
--- a/schemas/v3.0/schema.yaml
+++ b/schemas/v3.0/schema.yaml
@@ -634,7 +634,9 @@ definitions:
       callbacks:
         type: object
         additionalProperties:
-          $ref: '#/definitions/Callback'
+          oneOf:
+            - $ref: '#/definitions/Callback'
+            - $ref: '#/definitions/Reference'
       deprecated:
         type: boolean
         default: false
--- a/schemas/v3.0/schema.yaml
+++ b/schemas/v3.0/schema.yaml
@@ -1315,7 +1315,7 @@ definitions:
     type: object
     required:
       - type
-      - openIdConnect
+      - openIdConnectUrl
     properties:
       type:
         type: string
@@ -1472,7 +1472,6 @@ definitions:
       $ref: '#/definitions/PathItem'
     patternProperties:
       '^x-': {}
-    additionalProperties: false
 
   Encoding:
     type: object
--- a/schemas/v3.0/schema.yaml
+++ b/schemas/v3.0/schema.yaml
@@ -1,0 +1,1400 @@
+type: object
+required:
+  - openapi
+  - info
+  - paths
+properties:
+  openapi:
+    type: string
+    enum:
+      - 3.0.0
+  info:
+    $ref: '#/definitions/Info'
+  externalDocs:
+    $ref: '#/definitions/ExternalDocumentation'
+  servers:
+    type: array
+    items:
+      $ref: '#/definitions/Server'
+  security:
+    type: array
+    items:
+      $ref: '#/definitions/SecurityRequirement'
+  tags:
+    type: array
+    items:
+      $ref: '#/definitions/Tag'
+  paths:
+    $ref: '#/definitions/Paths'
+  components:
+    $ref: '#/definitions/Components'
+patternProperties:
+  '^x-': {}
+additionalProperties: false
+definitions:
+  Reference:
+    type: object
+    properties:
+      $ref:
+        type: string
+        format: uri-ref
+  Info:
+    type: object
+    required:
+      - title
+      - version
+    properties:
+      title:
+        type: string
+      description:
+        type: string
+      termsOfService:
+        type: string
+        format: uri-ref
+      contact:
+        $ref: '#/definitions/Contact'
+      license:
+        $ref: '#/definitions/License'
+      version:
+        type: string
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false
+
+
+  Contact:
+    type: object
+    properties:
+      name:
+        type: string
+      url:
+        type: string
+        format: uri-ref
+      email:
+        type: string
+        format: email
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false    
+
+  License:
+    type: object
+    required:
+      - name
+    properties:
+      name:
+        type: string
+      url:
+        type: string
+        format: uri-ref
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false
+
+  Server:
+    type: object
+    required:
+      - url
+    properties:
+      url:
+        type: string
+        format: uri-ref
+      description:
+        type: string
+      variables:
+        type: object
+        additionalProperties:
+          $ref: '#/definitions/ServerVariable'
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false
+
+  ServerVariable:
+    type: object
+    required:
+      - default
+    properties:
+      enum:
+        type: string
+      default:
+        type: string
+      description:
+        type: string
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false
+
+  Components:
+    type: object
+    properties:
+      schemas:
+        type: object
+        patternProperties:
+          '^[a-zA-Z0-9\.\-_]+$':
+            oneOf:
+              - $ref: '#/definitions/Reference'
+              - $ref: '#/definitions/Schema'
+      responses:
+        type: object
+        patternProperties:
+          '^[a-zA-Z0-9\.\-_]+$':
+            oneOf:
+              - $ref: '#/definitions/Reference'
+              - $ref: '#/definitions/Response'
+      parameters:
+        type: object
+        patternProperties:
+          '^[a-zA-Z0-9\.\-_]+$':
+            oneOf:
+              - $ref: '#/definitions/Reference'
+              - $ref: '#/definitions/Parameter'
+      examples:
+        type: object
+        patternProperties:
+          '^[a-zA-Z0-9\.\-_]+$':
+            oneOf:
+              - $ref: '#/definitions/Reference'
+              - $ref: '#/definitions/Example'
+      requestBodies:
+        type: object
+        patternProperties:
+          '^[a-zA-Z0-9\.\-_]+$':
+            oneOf:
+              - $ref: '#/definitions/Reference'
+              - $ref: '#/definitions/RequestBody'
+      headers:
+        type: object
+        patternProperties:
+          '^[a-zA-Z0-9\.\-_]+$':
+            oneOf:
+              - $ref: '#/definitions/Reference'
+              - $ref: '#/definitions/Header'
+      securitySchemes:
+        type: object
+        patternProperties:
+          '^[a-zA-Z0-9\.\-_]+$':
+            oneOf:
+              - $ref: '#/definitions/Reference'
+              - $ref: '#/definitions/SecurityScheme'
+      links:
+        type: object
+        patternProperties:
+          '^[a-zA-Z0-9\.\-_]+$':
+            oneOf:
+              - $ref: '#/definitions/Reference'
+              - $ref: '#/definitions/Link'
+      callbacks:
+        type: object
+        patternProperties:
+          '^[a-zA-Z0-9\.\-_]+$':
+            oneOf:
+              - $ref: '#/definitions/Reference'
+              - $ref: '#/definitions/Callback'
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false
+
+  Schema:
+    type: object
+    properties:
+      title:
+        type: string
+      multipleOf:
+        type: number
+        minimum: 0
+        exclusiveMinimum: true
+      maximum:
+        type: number
+      exclusiveMaximum:
+        type: boolean
+        default: false
+      minimum:
+        type: number
+      exclusiveMinimum:
+        type: boolean
+        default: false
+      maxLength:
+        type: integer
+        minimum: 0
+      minLength:
+        type: integer
+        minimum: 0
+        default: 0
+      pattern:
+        type: string
+        format: regex
+      maxItems:
+        type: integer
+        minimum: 0
+      minItems:
+        type: integer
+        minimum: 0
+        default: 0
+      uniqueItems:
+        type: boolean
+        default: false
+      maxProperties:
+        type: integer
+        minimum: 0
+      minProperties:
+        type: integer
+        minimum: 0
+        default: 0
+      required:
+        type: array
+        items:
+          type: string
+        minItems: 1
+        uniqueItems: true
+      enum:
+        type: array
+        items: {}
+        minItems: 1
+        uniqueItems: true
+      type:
+        type: string
+        enum:
+         - array
+         - boolean
+         - integer
+         - number
+         - object
+         - string
+      not:
+        $ref: '#/definitions/Schema'
+      properties:
+        type: object
+        additionalProperties:
+          $ref: '#/definitions/Schema'
+      additionalProperties:
+        oneOf:
+          - $ref: '#/definitions/Schema'
+          - type: boolean
+        default: true
+      description:
+        type: string
+      format:
+        type: string
+      default: {}
+      $ref:
+        type: string
+        format: uri-ref
+      nullable:
+        type: boolean
+        default: false
+      discriminator:
+        $ref: '#/definitions/Discriminator'
+      readOnly:
+        type: boolean
+        default: false
+      writeOnly:
+        type: boolean
+        default: false
+      example: {}
+      externalDocs:
+        $ref: '#/definitions/ExternalDocumentation'
+      deprecated:
+        type: boolean
+        default: false
+      xml:
+        $ref: '#/definitions/XML'
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false        
+  
+  Discriminator:
+    type: object
+    required:
+      - propertyName
+    properties:
+      propertyName:
+        type: string
+      mapping:
+        type: object
+        additionalProperties:
+          type: string
+
+  XML:
+    type: object
+    properties:
+      name:
+        type: string
+      namespace:
+        type: string
+        format: url
+      prefix:
+        type: string
+      attribute:
+        type: boolean
+        default: false
+      wrapperd:
+        type: boolean
+        default: false
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false
+
+  Response:
+    type: object
+    required:
+      - description
+    properties:
+      description:
+        type: string
+      headers:
+        additionalProperties:
+          oneOf:
+            - $ref: '#/definitions/Header'
+            - $ref: '#/definitions/Reference'
+      content:
+        type: object
+        additionalProperties:
+          $ref: '#/definitions/MediaType'
+      links:
+        type: object
+        additionalProperties:
+          oneOf:
+            - $ref: '#/definitions/Link'
+            - $ref: '#/definitions/Reference'
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false
+
+  MediaType:
+    oneOf:
+      - $ref: '#/definitions/MediaTypeWithExample'
+      - $ref: '#/definitions/MediaTypeWithExamples'
+
+  MediaTypeWithExample:
+    type: object
+    properties:
+      schema:
+        oneOf:
+          - $ref: '#/definitions/Schema'
+          - $ref: '#/definitions/Reference'
+      example: {}
+      encoding:
+        type: object
+        additionalProperties:
+          $ref: '#/definitions/Encoding'
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false 
+
+  MediaTypeWithExamples:
+    type: object
+    properties:
+      schema:
+        oneOf:
+          - $ref: '#/definitions/Schema'
+          - $ref: '#/definitions/Reference'
+      examples:
+        type: object
+        additionalProperties:
+          oneOf:
+            - $ref: '#/definitions/Example'
+            - $ref: '#/definitions/Reference'
+      encoding:
+        type: object
+        additionalProperties:
+          $ref: '#/definitions/Encoding'
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false 
+
+  Example:
+    type: object
+    properties:
+      summary:
+        type: string
+      description:
+        type: string
+      value:
+        type: string
+      externalValue:
+        type: string
+        format: uri-ref
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false
+
+  Header:
+    type: object
+    properties:
+      description:
+        type: string
+      required:
+        type: boolean
+        default: false
+      deprecated:
+        type: boolean
+        default: false
+      allowEmptyValue:
+        type: boolean
+        default: false
+      style:
+        type: string
+        enum:
+          - simple
+        default: simple
+      explode:
+        type: boolean
+        enum:
+          - false
+        default: false
+      schema:
+        oneOf:
+          - $ref: '#/definitions/Schema'
+          - $ref: '#/definitions/Reference'
+      examples:
+        type: array
+        items:
+          oneOf:
+            - $ref: '#/definitions/Example'
+            - $ref: '#/definitions/Reference'
+      example: {}
+      content:
+        type: object
+        additionalProperties:
+          $ref: '#/definitions/MediaType'
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false
+
+  Paths:
+    type: object
+    patternProperties:
+      '^\/': 
+        $ref: '#/definitions/PathItem'
+      '^x-': {}
+    additionalProperties: false
+
+  PathItem:
+    type: object
+    properties:
+      $ref:
+        type: string
+      summary:
+        type: string
+      description:
+        type: string
+      get:
+        $ref: '#/definitions/Operation'
+      put:
+        $ref: '#/definitions/Operation'
+      post:
+        $ref: '#/definitions/Operation'
+      delete:
+        $ref: '#/definitions/Operation'
+      options:
+        $ref: '#/definitions/Operation'
+      head:
+        $ref: '#/definitions/Operation'
+      patch:
+        $ref: '#/definitions/Operation'
+      trace:
+        $ref: '#/definitions/Operation'
+      servers:
+        type: array
+        items:
+          $ref: '#/definitions/Server'
+      parameters:
+        type: array
+        items:
+          oneOf:
+            - $ref: '#/definitions/Parameter'
+            - $ref: '#/definitions/Reference'
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false
+
+  Operation:
+    type: object
+    required:
+      - responses
+    properties:
+      tags:
+        type: array
+        items:
+          type: string
+      summary:
+        type: string
+      description:
+        type: string
+      externalDocs:
+        $ref: '#/definitions/ExternalDocumentation'
+      operationId:
+        type: string
+      parameters:
+        type: array
+        items:
+          oneOf:
+            - $ref: '#/definitions/Parameter'
+            - $ref: '#/definitions/Reference'
+      requestBody:
+        oneOf:
+          - $ref: '#/definitions/RequestBody'
+          - $ref: '#/definitions/Reference'
+      responses:
+        $ref: '#/definitions/Responses'
+      callbacks:
+        type: object
+        additionalProperties:
+          $ref: '#/definitions/Callback'
+      deprecated:
+        type: boolean
+        default: false
+      security:
+        type: array
+        items:
+          $ref: '#/definitions/SecurityRequirement'
+      servers:
+        type: array
+        items:
+          $ref: '#/definitions/Server'
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false
+
+  Responses:
+    allOf:
+      - $ref: '#/definitions/Extensions'    
+    type: object
+    properties:
+      default:
+        oneOf:
+          - $ref: '#/definitions/Response'
+          - $ref: '#/definitions/Reference'
+    patternProperties:
+      '[1-5](?:\d{2}|XX)':
+        oneOf:
+          - $ref: '#/definitions/Response'
+          - $ref: '#/definitions/Reference'
+      '^x-': {}
+    minProperties: 1
+    additionalProperties: false
+    not:
+      type: object
+      patternProperties:
+        '^x-': {}
+ 
+
+  SecurityRequirement:
+    type: object
+    additionalProperties:
+      type: array
+      items:
+        type: string
+
+  Tag:
+    type: object
+    required:
+      - name
+    properties:
+      name:
+        type: string
+      description:
+        type: string
+      externalDocs:
+        $ref: '#/definitions/ExternalDocumentation'
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false
+
+  ExternalDocumentation:
+    type: object
+    required:
+      - url
+    properties:
+      description:
+        type: string
+      url:
+        type: string
+        format: uri-ref
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false
+
+  Parameter:
+    oneOf:
+      - $ref: '#/definitions/ParameterWithSchema'
+      - $ref: '#/definitions/ParameterWithContent'
+  
+  ParameterWithSchema:
+    oneOf:
+      - $ref: '#/definitions/ParameterWithSchemaWithExample'
+      - $ref: '#/definitions/ParameterWithSchemaWithExamples'
+
+  ParameterWithSchemaWithExample:
+    oneOf:
+      - $ref: '#/definitions/ParameterWithSchemaWithExampleInPath'
+      - $ref: '#/definitions/ParameterWithSchemaWithExampleInQuery'
+      - $ref: '#/definitions/ParameterWithSchemaWithExampleInHeader'
+      - $ref: '#/definitions/ParameterWithSchemaWithExampleInCookie'
+
+  ParameterWithSchemaWithExampleInPath:
+    type: object
+    required:
+      - name
+      - in
+      - schema
+      - required
+    properties:
+      name:
+        type: string
+      in:
+        type: string
+        enum:
+          - path
+      description:
+        type: string
+      required:
+        type: boolean
+        enum:
+          - true
+      deprecated:
+        type: boolean
+        default: false
+      allowEmptyValue:
+        type: boolean
+        default: false
+      style:
+        type: string
+        enum:
+          - matrix
+          - label
+          - simple
+        default: simple
+      explode:
+        type: boolean
+      allowReserved:
+        type: boolean
+        default: false
+      schema:
+        oneOf:
+          - $ref: '#/definitions/Schema'
+          - $ref: '#/definitions/Reference'
+      example: {}
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false
+
+  ParameterWithSchemaWithExampleInQuery:
+    type: object
+    required:
+      - name
+      - in
+      - schema
+    properties:
+      name:
+        type: string
+      in:
+        type: string
+        enum:
+          - query
+      description:
+        type: string
+      required:
+        type: boolean
+        default: false
+      deprecated:
+        type: boolean
+        default: false
+      allowEmptyValue:
+        type: boolean
+        default: false
+      style:
+        type: string
+        enum:
+          - form
+          - spaceDelimited
+          - pipeDelimited
+          - deepObject
+        default: form
+      explode:
+        type: boolean
+      allowReserved:
+        type: boolean
+        default: false
+      schema:
+        oneOf:
+          - $ref: '#/definitions/Schema'
+          - $ref: '#/definitions/Reference'
+      example: {}
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false    
+
+  ParameterWithSchemaWithExampleInHeader:
+    type: object
+    required:
+      - name
+      - in
+      - schema
+    properties:
+      name:
+        type: string
+      in:
+        type: string
+        enum:
+          - header
+      description:
+        type: string
+      required:
+        type: boolean
+        default: false
+      deprecated:
+        type: boolean
+        default: false
+      allowEmptyValue:
+        type: boolean
+        default: false
+      style:
+        type: string
+        enum:
+          - simple
+        default: simple
+      explode:
+        type: boolean
+      allowReserved:
+        type: boolean
+        default: false
+      schema:
+        oneOf:
+          - $ref: '#/definitions/Schema'
+          - $ref: '#/definitions/Reference'
+      example: {}
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false    
+
+  ParameterWithSchemaWithExampleInCookie:
+    type: object
+    required:
+      - name
+      - in
+      - schema
+    properties:
+      name:
+        type: string
+      in:
+        type: string
+        enum:
+          - cookie
+      description:
+        type: string
+      required:
+        type: boolean
+        default: false
+      deprecated:
+        type: boolean
+        default: false
+      allowEmptyValue:
+        type: boolean
+        default: false
+      style:
+        type: string
+        enum:
+          - form
+        default: form
+      explode:
+        type: boolean
+      allowReserved:
+        type: boolean
+        default: false
+      schema:
+        oneOf:
+          - $ref: '#/definitions/Schema'
+          - $ref: '#/definitions/Reference'
+      example: {}
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false  
+
+  ParameterWithSchemaWithExamples:
+    oneOf:
+      - $ref: '#/definitions/ParameterWithSchemaWithExamplesInPath'
+      - $ref: '#/definitions/ParameterWithSchemaWithExamplesInQuery'
+      - $ref: '#/definitions/ParameterWithSchemaWithExamplesInHeader'
+      - $ref: '#/definitions/ParameterWithSchemaWithExamplesInCookie'
+
+  ParameterWithSchemaWithExamplesInPath:
+    type: object
+    required:
+      - name
+      - in
+      - schema
+      - required
+    properties:
+      name:
+        type: string
+      in:
+        type: string
+        enum:
+          - path
+      description:
+        type: string
+      required:
+        type: boolean
+        enum:
+          - true
+      deprecated:
+        type: boolean
+        default: false
+      allowEmptyValue:
+        type: boolean
+        default: false
+      style:
+        type: string
+        enum:
+          - matrix
+          - label
+          - simple
+        default: simple
+      explode:
+        type: boolean
+      allowReserved:
+        type: boolean
+        default: false
+      schema:
+        oneOf:
+          - $ref: '#/definitions/Schema'
+          - $ref: '#/definitions/Reference'
+      examples:
+        type: object
+        additionalProperties:
+          oneOf:
+            - $ref: '#/definitions/Example'
+            - $ref: '#/definitions/Reference'
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false
+
+  ParameterWithSchemaWithExamplesInQuery:
+    type: object
+    required:
+      - name
+      - in
+      - schema
+    properties:
+      name:
+        type: string
+      in:
+        type: string
+        enum:
+          - query
+      description:
+        type: string
+      required:
+        type: boolean
+        default: false
+      deprecated:
+        type: boolean
+        default: false
+      allowEmptyValue:
+        type: boolean
+        default: false
+      style:
+        type: string
+        enum:
+          - form
+          - spaceDelimited
+          - pipeDelimited
+          - deepObject
+        default: form
+      explode:
+        type: boolean
+      allowReserved:
+        type: boolean
+        default: false
+      schema:
+        oneOf:
+          - $ref: '#/definitions/Schema'
+          - $ref: '#/definitions/Reference'
+      examples:
+        type: object
+        additionalProperties:
+          oneOf:
+            - $ref: '#/definitions/Example'
+            - $ref: '#/definitions/Reference'
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false    
+
+  ParameterWithSchemaWithExamplesInHeader:
+    type: object
+    required:
+      - name
+      - in
+      - schema
+    properties:
+      name:
+        type: string
+      in:
+        type: string
+        enum:
+          - header
+      description:
+        type: string
+      required:
+        type: boolean
+        default: false
+      deprecated:
+        type: boolean
+        default: false
+      allowEmptyValue:
+        type: boolean
+        default: false
+      style:
+        type: string
+        enum:
+          - simple
+        default: simple
+      explode:
+        type: boolean
+      allowReserved:
+        type: boolean
+        default: false
+      schema:
+        oneOf:
+          - $ref: '#/definitions/Schema'
+          - $ref: '#/definitions/Reference'
+      examples:
+        type: object
+        additionalProperties:
+          oneOf:
+            - $ref: '#/definitions/Example'
+            - $ref: '#/definitions/Reference'
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false    
+
+  ParameterWithSchemaWithExamplesInCookie:
+    type: object
+    required:
+      - name
+      - in
+      - schema
+    properties:
+      name:
+        type: string
+      in:
+        type: string
+        enum:
+          - cookie
+      description:
+        type: string
+      required:
+        type: boolean
+        default: false
+      deprecated:
+        type: boolean
+        default: false
+      allowEmptyValue:
+        type: boolean
+        default: false
+      style:
+        type: string
+        enum:
+          - form
+        default: form
+      explode:
+        type: boolean
+      allowReserved:
+        type: boolean
+        default: false
+      schema:
+        oneOf:
+          - $ref: '#/definitions/Schema'
+          - $ref: '#/definitions/Reference'
+      examples:
+        type: object
+        additionalProperties:
+          oneOf:
+            - $ref: '#/definitions/Example'
+            - $ref: '#/definitions/Reference'
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false
+
+  ParameterWithContent:
+    oneOf:
+      - $ref: '#/definitions/ParameterWithContentInPath'
+      - $ref: '#/definitions/ParameterWithContentNotInPath'
+
+  ParameterWithContentInPath:
+    type: object
+    required:
+      - name
+      - in
+      - content
+    properties:
+      name:
+        type: string
+      in:
+        type: string
+        enum:
+          - path
+      description:
+        type: string
+      required:
+        type: boolean
+        enum:
+          - true
+      deprecated:
+        type: boolean
+        default: false
+      allowEmptyValue:
+        type: boolean
+        default: false
+      content:
+        type: object
+        additionalProperties:
+          $ref: '#/definitions/MediaType'
+        minProperties: 1
+        maxProperties: 1
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false
+
+  ParameterWithContentNotInPath:
+    type: object
+    required:
+      - name
+      - in
+      - content
+    properties:
+      name:
+        type: string
+      in:
+        type: string
+        enum:
+          - query
+          - header
+          - cookie
+      description:
+        type: string
+      required:
+        type: boolean
+        default: false
+      deprecated:
+        type: boolean
+        default: false
+      allowEmptyValue:
+        type: boolean
+        default: false
+      content:
+        type: object
+        additionalProperties:
+          $ref: '#/definitions/MediaType'
+        minProperties: 1
+        maxProperties: 1
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false
+
+  RequestBody:
+    type: object
+    required:
+      - content
+    properties:
+      description:
+        type: string
+      content:
+        type: object
+        additionalProperties:
+          $ref: '#/definitions/MediaType'
+      required:
+        type: boolean
+        default: false
+    patternProperties:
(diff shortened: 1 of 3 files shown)


fehguy:
👍

timburks:
Glad to see effort going into this. The schema that I automatically generated from the spec is easy to reexport as YAML, which I'll append for comparison. The most apparent difference to me is that you've flattened out some intermediate structures.

```
title: A JSON Schema for OpenAPI 3.0.
id: http://openapis.org/v3/schema.json#
type: object
description: This is the root document object of the OpenAPI definition.
required:
- openapi
- info
- paths
additionalProperties: false
patternProperties:
  ^x-:
    $ref: '#/definitions/specificationExtension'
properties:
  openapi:
    type: string
  info:
    $ref: '#/definitions/info'
  servers:
    type: array
    items:
      $ref: '#/definitions/server'
    uniqueItems: true
  paths:
    $ref: '#/definitions/paths'
  components:
    $ref: '#/definitions/components'
  security:
    type: array
    items:
      $ref: '#/definitions/securityRequirement'
    uniqueItems: true
  tags:
    type: array
    items:
      $ref: '#/definitions/tag'
    uniqueItems: true
  externalDocs:
    $ref: '#/definitions/externalDocs'
definitions:
  info:
    type: object
    description: The object provides metadata about the API. The metadata can be used
      by the clients if needed, and can be presented in editing or documentation generation
      tools for convenience.
    required:
    - title
    - version
    additionalProperties: false
    patternProperties:
      ^x-:
        $ref: '#/definitions/specificationExtension'
    properties:
      title:
        type: string
      description:
        type: string
      termsOfService:
        type: string
      contact:
        $ref: '#/definitions/contact'
      license:
        $ref: '#/definitions/license'
      version:
        type: string
  contact:
    type: object
    description: Contact information for the exposed API.
    additionalProperties: false
    patternProperties:
      ^x-:
        $ref: '#/definitions/specificationExtension'
    properties:
      name:
        type: string
      url:
        type: string
      email:
        type: string
  license:
    type: object
    description: License information for the exposed API.
    required:
    - name
    additionalProperties: false
    patternProperties:
      ^x-:
        $ref: '#/definitions/specificationExtension'
    properties:
      name:
        type: string
      url:
        type: string
  server:
    type: object
    description: An object representing a Server.
    required:
    - url
    additionalProperties: false
    patternProperties:
      ^x-:
        $ref: '#/definitions/specificationExtension'
    properties:
      url:
        type: string
      description:
        type: string
      variables:
        $ref: '#/definitions/serverVariables'
  serverVariable:
    type: object
    description: An object representing a Server Variable for server URL template
      substitution.
    required:
    - default
    additionalProperties: false
    patternProperties:
      ^x-:
        $ref: '#/definitions/specificationExtension'
    properties:
      enum:
        type: array
        items:
          type: string
        uniqueItems: true
      default:
        type: string
      description:
        type: string
  components:
    type: object
    description: Holds a set of reusable objects for different aspects of the OAS.
      All objects defined within the components object will have no effect on the
      API unless they are explicitly referenced from properties outside the components
      object.
    additionalProperties: false
    patternProperties:
      ^x-:
        $ref: '#/definitions/specificationExtension'
    properties:
      schemas:
        $ref: '#/definitions/schemasOrReferences'
      responses:
        $ref: '#/definitions/responsesOrReferences'
      parameters:
        $ref: '#/definitions/parametersOrReferences'
      examples:
        $ref: '#/definitions/examplesOrReferences'
      requestBodies:
        $ref: '#/definitions/requestBodiesOrReferences'
      headers:
        $ref: '#/definitions/headersOrReferences'
      securitySchemes:
        $ref: '#/definitions/securitySchemesOrReferences'
      links:
        $ref: '#/definitions/linksOrReferences'
      callbacks:
        $ref: '#/definitions/callbacksOrReferences'
  paths:
    type: object
    description: Holds the relative paths to the individual endpoints and their operations.
      The path is appended to the URL from the `Server Object` in order to construct
      the full URL.  The Paths MAY be empty, due to ACL constraints.
    additionalProperties: false
    patternProperties:
      ^/:
        $ref: '#/definitions/pathItem'
      ^x-:
        $ref: '#/definitions/specificationExtension'
  pathItem:
    type: object
    description: Describes the operations available on a single path. A Path Item
      MAY be empty, due to ACL constraints. The path itself is still exposed to the
      documentation viewer but they will not know which operations and parameters
      are available.
    additionalProperties: false
    patternProperties:
      ^x-:
        $ref: '#/definitions/specificationExtension'
    properties:
      $ref:
        type: string
      summary:
        type: string
      description:
        type: string
      get:
        $ref: '#/definitions/operation'
      put:
        $ref: '#/definitions/operation'
      post:
        $ref: '#/definitions/operation'
      delete:
        $ref: '#/definitions/operation'
      options:
        $ref: '#/definitions/operation'
      head:
        $ref: '#/definitions/operation'
      patch:
        $ref: '#/definitions/operation'
      trace:
        $ref: '#/definitions/operation'
      servers:
        type: array
        items:
          $ref: '#/definitions/server'
        uniqueItems: true
      parameters:
        type: array
        items:
          $ref: '#/definitions/parameterOrReference'
        uniqueItems: true
  operation:
    type: object
    description: Describes a single API operation on a path.
    required:
    - responses
    additionalProperties: false
    patternProperties:
      ^x-:
        $ref: '#/definitions/specificationExtension'
    properties:
      tags:
        type: array
        items:
          type: string
        uniqueItems: true
      summary:
        type: string
      description:
        type: string
      externalDocs:
        $ref: '#/definitions/externalDocs'
      operationId:
        type: string
      parameters:
        type: array
        items:
          $ref: '#/definitions/parameterOrReference'
        uniqueItems: true
      requestBody:
        $ref: '#/definitions/requestBodyOrReference'
      responses:
        $ref: '#/definitions/responses'
      callbacks:
        $ref: '#/definitions/callbacksOrReferences'
      deprecated:
        type: boolean
      security:
        type: array
        items:
          $ref: '#/definitions/securityRequirement'
        uniqueItems: true
      servers:
        type: array
        items:
          $ref: '#/definitions/server'
        uniqueItems: true
  externalDocs:
    type: object
    description: Allows referencing an external resource for extended documentation.
    required:
    - url
    additionalProperties: false
    patternProperties:
      ^x-:
        $ref: '#/definitions/specificationExtension'
    properties:
      description:
        type: string
      url:
        type: string
  parameter:
    type: object
    description: Describes a single operation parameter.  A unique parameter is defined
      by a combination of a name and location.
    required:
    - name
    - in
    additionalProperties: false
    patternProperties:
      ^x-:
        $ref: '#/definitions/specificationExtension'
    properties:
      name:
        type: string
      in:
        type: string
      description:
        type: string
      required:
        type: boolean
      deprecated:
        type: boolean
      allowEmptyValue:
        type: boolean
      style:
        type: string
      explode:
        type: boolean
      allowReserved:
        type: boolean
      schema:
        $ref: '#/definitions/schemaOrReference'
      example:
        $ref: '#/definitions/any'
      examples:
        $ref: '#/definitions/examplesOrReferences'
      content:
        $ref: '#/definitions/mediaTypes'
  requestBody:
    type: object
    description: Describes a single request body.
    required:
    - content
    additionalProperties: false
    patternProperties:
      ^x-:
        $ref: '#/definitions/specificationExtension'
    properties:
      description:
        type: string
      content:
        $ref: '#/definitions/mediaTypes'
      required:
        type: boolean
  mediaType:
    type: object
    description: Each Media Type Object provides schema and examples for the media
      type identified by its key.
    additionalProperties: false
    patternProperties:
      ^x-:
        $ref: '#/definitions/specificationExtension'
    properties:
      schema:
        $ref: '#/definitions/schemaOrReference'
      example:
        $ref: '#/definitions/any'
      examples:
        $ref: '#/definitions/examplesOrReferences'
      encoding:
        $ref: '#/definitions/encodings'
  encoding:
    type: object
    description: A single encoding definition applied to a single schema property.
    additionalProperties: false
    patternProperties:
      ^x-:
        $ref: '#/definitions/specificationExtension'
    properties:
      contentType:
        type: string
      headers:
        $ref: '#/definitions/headers'
      style:
        type: string
      explode:
        type: boolean
      allowReserved:
        type: boolean
  responses:
    type: object
    description: A container for the expected responses of an operation. The container
      maps a HTTP response code to the expected response.  The documentation is not
      necessarily expected to cover all possible HTTP response codes because they
      may not be known in advance. However, documentation is expected to cover a successful
      operation response and any known errors.  The `default` MAY be used as a default
      response object for all HTTP codes  that are not covered individually by the
      specification.  The `Responses Object` MUST contain at least one response code,
      and it  SHOULD be the response for a successful operation call.
    additionalProperties: false
    patternProperties:
      ^([0-9X]{3})$:
        $ref: '#/definitions/responseOrReference'
      ^x-:
        $ref: '#/definitions/specificationExtension'
    properties:
      default:
        $ref: '#/definitions/responseOrReference'
  response:
    type: object
    description: Describes a single response from an API Operation, including design-time,
      static  `links` to operations based on the response.
    required:
    - description
    additionalProperties: false
    patternProperties:
      ^x-:
        $ref: '#/definitions/specificationExtension'
    properties:
      description:
        type: string
      headers:
        $ref: '#/definitions/headersOrReferences'
      content:
        $ref: '#/definitions/mediaTypes'
      links:
        $ref: '#/definitions/linksOrReferences'
  callback:
    type: object
    description: A map of possible out-of band callbacks related to the parent operation.
      Each value in the map is a Path Item Object that describes a set of requests
      that may be initiated by the API provider and the expected responses. The key
      value used to identify the callback object is an expression, evaluated at runtime,
      that identifies a URL to use for the callback operation.
    additionalProperties: false
    patternProperties:
      ^:
        $ref: '#/definitions/pathItem'
      ^x-:
        $ref: '#/definitions/specificationExtension'
  example:
    type: object
    description: ""
    additionalProperties: false
    patternProperties:
      ^x-:
        $ref: '#/definitions/specificationExtension'
    properties:
      summary:
        type: string
      description:
        type: string
      value:
        $ref: '#/definitions/any'
      externalValue:
        type: string
  link:
    type: object
    description: The `Link object` represents a possible design-time link for a response.
      The presence of a link does not guarantee the caller's ability to successfully
      invoke it, rather it provides a known relationship and traversal mechanism between
      responses and other operations.  Unlike _dynamic_ links (i.e. links provided
      **in** the response payload), the OAS linking mechanism does not require link
      information in the runtime response.  For computing links, and providing instructions
      to execute them, a runtime expression is used for accessing values in an operation
      and using them as parameters while invoking the linked operation.
    additionalProperties: false
    patternProperties:
      ^x-:
        $ref: '#/definitions/specificationExtension'
    properties:
      operationRef:
        type: string
      operationId:
        type: string
      parameters:
        $ref: '#/definitions/anysOrExpressions'
      requestBody:
        $ref: '#/definitions/anyOrExpression'
      description:
        type: string
      server:
        $ref: '#/definitions/server'
  header:
    type: object
    description: 'The Header Object follows the structure of the Parameter Object
      with the following changes:  1. `name` MUST NOT be specified, it is given in
      the corresponding `headers` map. 1. `in` MUST NOT be specified, it is implicitly
      in `header`. 1. All traits that are affected by the location MUST be applicable
      to a location of `header` (for example, `style`).'
    additionalProperties: false
    patternProperties:
      ^x-:
        $ref: '#/definitions/specificationExtension'
    properties:
      description:
        type: string
      required:
        type: boolean
      deprecated:
        type: boolean
      allowEmptyValue:
        type: boolean
      style:
        type: string
      explode:
        type: boolean
      allowReserved:
        type: boolean
      schema:
        $ref: '#/definitions/schemaOrReference'
      example:
        $ref: '#/definitions/any'
      examples:
        $ref: '#/definitions/examplesOrReferences'
      content:
        $ref: '#/definitions/mediaTypes'
  tag:
    type: object
    description: Adds metadata to a single tag that is used by the Operation Object.
      It is not mandatory to have a Tag Object per tag defined in the Operation Object
      instances.
    required:
    - name
    additionalProperties: false
    patternProperties:
      ^x-:
        $ref: '#/definitions/specificationExtension'
    properties:
      name:
        type: string
      description:
        type: string
      externalDocs:
        $ref: '#/definitions/externalDocs'
  examples:
    type: object
    description: ""
    additionalProperties: false
  reference:
    type: object
    description: A simple object to allow referencing other components in the specification,
      internally and externally.  The Reference Object is defined by JSON Reference
      and follows the same structure, behavior and rules.   For this specification,
      reference resolution is accomplished as defined by the JSON Reference specification
      and not by the JSON Schema specification.
    required:
    - $ref
    additionalProperties: false
    properties:
      $ref:
        type: string
  schema:
    type: object
    description: The Schema Object allows the definition of input and output data
      types. These types can be objects, but also primitives and arrays. This object
      is an extended subset of the JSON Schema Specification Wright Draft 00.  For
      more information about the properties, see JSON Schema Core and JSON Schema
      Validation. Unless stated otherwise, the property definitions follow the JSON
      Schema.
    additionalProperties: false
    patternProperties:
      ^x-:
        $ref: '#/definitions/specificationExtension'
    properties:
      nullable:
        type: boolean
      discriminator:
        $ref: '#/definitions/discriminator'
      readOnly:
        type: boolean
      writeOnly:
        type: boolean
      xml:
        $ref: '#/definitions/xml'
      externalDocs:
        $ref: '#/definitions/externalDocs'
      example:
        $ref: '#/definitions/any'
      deprecated:
        type: boolean
      title:
        $ref: http://json-schema.org/draft-04/schema#/properties/title
      multipleOf:
        $ref: http://json-schema.org/draft-04/schema#/properties/multipleOf
      maximum:
        $ref: http://json-schema.org/draft-04/schema#/properties/maximum
      exclusiveMaximum:
        $ref: http://json-schema.org/draft-04/schema#/properties/exclusiveMaximum
      minimum:
        $ref: http://json-schema.org/draft-04/schema#/properties/minimum
      exclusiveMinimum:
        $ref: http://json-schema.org/draft-04/schema#/properties/exclusiveMinimum
      maxLength:
        $ref: http://json-schema.org/draft-04/schema#/properties/maxLength
      minLength:
        $ref: http://json-schema.org/draft-04/schema#/properties/minLength
      pattern:
        $ref: http://json-schema.org/draft-04/schema#/properties/pattern
      maxItems:
        $ref: http://json-schema.org/draft-04/schema#/properties/maxItems
      minItems:
        $ref: http://json-schema.org/draft-04/schema#/properties/minItems
      uniqueItems:
        $ref: http://json-schema.org/draft-04/schema#/properties/uniqueItems
      maxProperties:
        $ref: http://json-schema.org/draft-04/schema#/properties/maxProperties
      minProperties:
        $ref: http://json-schema.org/draft-04/schema#/properties/minProperties
      required:
        $ref: http://json-schema.org/draft-04/schema#/properties/required
      enum:
        $ref: http://json-schema.org/draft-04/schema#/properties/enum
      type:
        type: string
      allOf:
        type: array
        items:
          $ref: '#/definitions/schemaOrReference'
        minItems: 1
      oneOf:
        type: array
        items:
          $ref: '#/definitions/schemaOrReference'
        minItems: 1
      anyOf:
        type: array
        items:
          $ref: '#/definitions/schemaOrReference'
        minItems: 1
      not:
        $ref: '#/definitions/schema'
      items:
        anyOf:
        - $ref: '#/definitions/schemaOrReference'
        - type: array
          items:
            $ref: '#/definitions/schemaOrReference'
          minItems: 1
      properties:
        type: object
        additionalProperties:
          $ref: '#/definitions/schemaOrReference'
      additionalProperties:
        oneOf:
        - $ref: '#/definitions/schemaOrReference'
        - type: boolean
      default:
        $ref: '#/definitions/defaultType'
      description:
        type: string
      format:
        type: string
  discriminator:
    type: object
    description: When request bodies or response payloads may be one of a number of
      different schemas, a `discriminator` object can be used to aid in serialization,
      deserialization, and validation.  The discriminator is a specific object in
      a schema which is used to inform the consumer of the specification of an alternative
      schema based on the value associated with it.  When using the discriminator,
      _inline_ schemas will not be considered.
    required:
    - propertyName
    additionalProperties: false
    properties:
      propertyName:
        type: string
      mapping:
        $ref: '#/definitions/strings'
  xml:
    type: object
    description: A metadata object that allows for more fine-tuned XML model definitions.  When
      using arrays, XML element names are *not* inferred (for singular/plural forms)
      and the `name` property SHOULD be used to add that information. See examples
      for expected behavior.
    additionalProperties: false
    patternProperties:
      ^x-:
        $ref: '#/definitions/specificationExtension'
    properties:
      name:
        type: string
      namespace:
        type: string
      prefix:
        type: string
      attribute:
        type: boolean
      wrapped:
        type: boolean
  securityScheme:
    type: object
    description: Defines a security scheme that can be used by the operations. Supported
      schemes are HTTP authentication, an API key (either as a header or as a query
      parameter), OAuth2's common flows (implicit, password, application and access
      code) as defined in RFC6749, and OpenID Connect Discovery.
    required:
    - type
    additionalProperties: false
    patternProperties:
      ^x-:
        $ref: '#/definitions/specificationExtension'
    properties:
      type:
        type: string
      description:
        type: string
      name:
        type: string
      in:
        type: string
      scheme:
        type: string
      bearerFormat:
        type: string
      flows:
        $ref: '#/definitions/oauthFlows'
      openIdConnectUrl:
        type: string
  oauthFlows:
    type: object
    description: Allows configuration of the supported OAuth Flows.
    additionalProperties: false
    patternProperties:
      ^x-:
        $ref: '#/definitions/specificationExtension'
    properties:
      implicit:
        $ref: '#/definitions/oauthFlow'
      password:
        $ref: '#/definitions/oauthFlow'
      clientCredentials:
        $ref: '#/definitions/oauthFlow'
      authorizationCode:
        $ref: '#/definitions/oauthFlow'
  oauthFlow:
    type: object
    description: Configuration details for a supported OAuth Flow
    additionalProperties: false
    patternProperties:
      ^x-:
        $ref: '#/definitions/specificationExtension'
    properties:
      authorizationUrl:
        type: string
      tokenUrl:
        type: string
      refreshUrl:
        type: string
      scopes:
        $ref: '#/definitions/strings'
  securityRequirement:
    type: object
    description: Lists the required security schemes to execute this operation. The
      name used for each property MUST correspond to a security scheme declared in
      the Security Schemes under the Components Object.  Security Requirement Objects
      that contain multiple schemes require that all schemes MUST be satisfied for
      a request to be authorized. This enables support for scenarios where multiple
      query parameters or HTTP headers are required to convey security information.  When
      a list of Security Requirement Objects is defined on the Open API object or
      Operation Object, only one of Security Requirement Objects in the list needs
      to be satisfied to authorize the request.
    additionalProperties: false
    patternProperties:
      ^[a-zA-Z0-9\.\-_]+$:
        type: array
        items:
          type: string
        uniqueItems: true
  anyOrExpression:
    oneOf:
    - $ref: '#/definitions/any'
    - $ref: '#/definitions/expression'
  callbackOrReference:
    oneOf:
    - $ref: '#/definitions/callback'
    - $ref: '#/definitions/reference'
  exampleOrReference:
    oneOf:
    - $ref: '#/definitions/example'
    - $ref: '#/definitions/reference'
  headerOrReference:
    oneOf:
    - $ref: '#/definitions/header'
    - $ref: '#/definitions/reference'
  linkOrReference:
    oneOf:
    - $ref: '#/definitions/link'
    - $ref: '#/definitions/reference'
  parameterOrReference:
    oneOf:
    - $ref: '#/definitions/parameter'
    - $ref: '#/definitions/reference'
  requestBodyOrReference:
    oneOf:
    - $ref: '#/definitions/requestBody'
    - $ref: '#/definitions/reference'
  responseOrReference:
    oneOf:
    - $ref: '#/definitions/response'
    - $ref: '#/definitions/reference'
  schemaOrReference:
    oneOf:
    - $ref: '#/definitions/schema'
    - $ref: '#/definitions/reference'
  securitySchemeOrReference:
    oneOf:
    - $ref: '#/definitions/securityScheme'
    - $ref: '#/definitions/reference'
  anysOrExpressions:
    type: object
    additionalProperties:
      $ref: '#/definitions/anyOrExpression'
  callbacksOrReferences:
    type: object
    additionalProperties:
      $ref: '#/definitions/callbackOrReference'
  encodings:
    type: object
    additionalProperties:
      $ref: '#/definitions/encoding'
  examplesOrReferences:
    type: object
    additionalProperties:
      $ref: '#/definitions/exampleOrReference'
  headers:
    type: object
    additionalProperties:
      $ref: '#/definitions/header'
  headersOrReferences:
    type: object
    additionalProperties:
      $ref: '#/definitions/headerOrReference'
  linksOrReferences:
    type: object
    additionalProperties:
      $ref: '#/definitions/linkOrReference'
  mediaTypes:
    type: object
    additionalProperties:
      $ref: '#/definitions/mediaType'
  parametersOrReferences:
    type: object
    additionalProperties:
      $ref: '#/definitions/parameterOrReference'
  requestBodiesOrReferences:
    type: object
    additionalProperties:
      $ref: '#/definitions/requestBodyOrReference'
  responsesOrReferences:
    type: object
    additionalProperties:
      $ref: '#/definitions/responseOrReference'
  schemasOrReferences:
    type: object
    additionalProperties:
      $ref: '#/definitions/schemaOrReference'
  securitySchemesOrReferences:
    type: object
    additionalProperties:
      $ref: '#/definitions/securitySchemeOrReference'
  serverVariables:
    type: object
    additionalProperties:
      $ref: '#/definitions/serverVariable'
  strings:
    type: object
    additionalProperties:
      type: string
  object:
    type: object
    additionalProperties: true
  any:
    additionalProperties: true
  expression:
    type: object
    additionalProperties: true
  specificationExtension:
    description: Any property starting with x- is valid.
    oneOf:
    - type: "null"
    - type: number
    - type: boolean
    - type: string
    - type: object
    - type: array
  defaultType:
    oneOf:
    - type: "null"
    - type: array
    - type: object
    - type: number
    - type: boolean
    - type: string
```

schemas/v3.0/schema.yaml
```diff
@@ -0,0 +1,1400 @@
...
+      schema:
+        oneOf:
+          - $ref: '#/definitions/Schema'
+          - $ref: '#/definitions/Reference'
+      example: {}
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false  
+
+  ParameterWithSchemaWithExamples:
```
MikeRalphson: I seem to be having problems with these - as neither `example` or `examples` is required in `ParameterWithSchemaWithExampleInQuery` or `ParameterWithSchemaWithExamplesInQuery` and there is no `ParameterWithSchemaWithoutExampleInQuery` type, how is the correct type to match the `oneOf` to be determined?

I temporarily changed `examples` to be a required property in `ParameterWithSchemaWithExamplesIn*`.

schemas/v3.0/schema.yaml
```diff
@@ -0,0 +1,1400 @@
...
+            oneOf:
+              - $ref: '#/definitions/Reference'
+              - $ref: '#/definitions/Callback'
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false
+
+  Schema:
+    type: object
+    properties:
```
MikeRalphson: Is the `Schema` object missing the `items` property?

schemas/v3.0/schema.yaml
```diff
@@ -0,0 +1,1400 @@
...
+    additionalProperties: false 
+
+  MediaTypeWithExamples:
+    type: object
+    properties:
+      schema:
+        oneOf:
+          - $ref: '#/definitions/Schema'
+          - $ref: '#/definitions/Reference'
+      examples:
```
MikeRalphson: Similarly, I had to make `examples` a required property to make the `oneOf` detection work correctly.

schemas/v3.0/schema.yaml
```diff
@@ -0,0 +1,1400 @@
...
+        type: string
+      namespace:
+        type: string
+        format: url
+      prefix:
+        type: string
+      attribute:
+        type: boolean
+        default: false
+      wrapperd:
```
MikeRalphson: Typo `wrapperd` for `wrapped`.

schemas/v3.0/schema.yaml
```diff
@@ -0,0 +1,1400 @@
...
+      type:
+        type: string
+        enum:
+         - array
+         - boolean
+         - integer
+         - number
+         - object
+         - string
+      not:
```
MikeRalphson: Are we also missing definitions for the `allOf`, `oneOf' and 'anyOf' arrays?

schemas/v3.0/schema.yaml
```diff
@@ -0,0 +1,1400 @@
...
+          - type: boolean
+        default: true
+      description:
+        type: string
+      format:
+        type: string
+      default: {}
+      $ref:
+        type: string
+        format: uri-ref
```
MikeRalphson: Is this `uriref` in Draft 05?

schemas/v3.0/schema.yaml
```diff
@@ -0,0 +1,1400 @@
...
+        type: boolean
+        enum:
+          - false
+        default: false
+      schema:
+        oneOf:
+          - $ref: '#/definitions/Schema'
+          - $ref: '#/definitions/Reference'
+      examples:
+        type: array
```
MikeRalphson: Is this correct, or is it now a Map as elsewhere?

schemas/v3.0/schema.yaml
```diff
@@ -0,0 +1,1400 @@
...
+
+  Example:
+    type: object
+    properties:
+      summary:
+        type: string
+      description:
+        type: string
+      value:
+        type: string
```
MikeRalphson: Example.value should be an empty schema?

MikeRalphson:
@webron with the changes implied by the above comments, I've now successfully validated 574 converted OpenAPI v2 definitions - with only one minor bug found in my converter (thank you).

schemas/v3.0/schema.yaml
```diff
@@ -0,0 +1,1400 @@
...
+  components:
+    $ref: '#/definitions/Components'
+patternProperties:
+  '^x-': {}
+additionalProperties: false
+definitions:
+  Reference:
+    type: object
+    properties:
+      $ref:
```
webron: Yes, will fix.

schemas/v3.0/schema.yaml
```diff
@@ -0,0 +1,1400 @@
+type: object
+required:
+  - openapi
+  - info
+  - paths
+properties:
+  openapi:
+    type: string
+    enum:
+      - 3.0.0
```
webron: Yeah, this should actually change to a pattern that matches anything starting with 3.0 as according to semver, any patch level changes should not affect validation. That said, I'm embarrassed to say I've avoided learning regex as much as I can 😁

schemas/v3.0/schema.yaml
```diff
@@ -0,0 +1,1400 @@
...
+    additionalProperties: false
+
+  Components:
+    type: object
+    properties:
+      schemas:
+        type: object
+        patternProperties:
+          '^[a-zA-Z0-9\.\-_]+$':
+            oneOf:
```
webron: That's an excellent question. I've revisited the spec. I'll either make it `anyOf` or remove the `$ref` from the `Schema` definition and make sure it's always interchangeable with the `Reference` in the JSON Schema.

schemas/v3.0/schema.yaml
webron: Yeah, will add.

schemas/v3.0/schema.yaml
webron: Can you explain this one?

schemas/v3.0/schema.yaml
webron: Yup, will fix.

schemas/v3.0/schema.yaml
MikeRalphson: Sorry, I made this edit after another comment, but it appears earlier in the schema. I believe this is `MediaTypeWithExample` vs `MediaTypeWithExamples`. A media-type object with neither `example` or `examples` properties matches both definitions. Suggest making `examples` a required property of `MediaTypeWithExamples`.

schemas/v3.0/schema.yaml
```diff
@@ -0,0 +1,1400 @@
...
+          type: string
+
+  XML:
+    type: object
+    properties:
+      name:
+        type: string
+      namespace:
+        type: string
+        format: url
```
MikeRalphson: Should this format be `uri-ref` / `uriref` not `url` ?

schemas/v3.0/schema.yaml
```diff
@@ -0,0 +1,1400 @@
...
+        type: array
+        items:
+          $ref: '#/definitions/Server'
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false
+
+  Responses:
+    allOf:
+      - $ref: '#/definitions/Extensions'
```
webron: That was from a previous life of the schema, and I just forgot to remove it. Will do.

schemas/v3.0/schema.yaml
webron: Yes, added.

schemas/v3.0/schema.yaml
webron: This requires a complete rework.

webron:
@MikeRalphson No, I wasn't referring to adding the `descriptions` as part of the decorations. While important, those are nice to have, and I don't necessarily have the capacity to add those right now all around the schema. If we decide to merge this schema (with it's JSON representation, of course), it'd be great if people helped contributing descriptions.

Right now I'm more focused on the functionality.

schemas/v3.0/schema.yaml
webron: No. We changed the `namespace` definition to be an absolute URI.

schemas/v3.0/schema.yaml
webron: Yes, that's a miss on my part. I actually looked it up before, and went with the wrong one 😕

webron:
@MikeRalphson Updated with your comments a possible a few additional things I've found. Would appreciate a second look.

schemas/v3.0/schema.yaml
```diff
@@ -254,31 +255,57 @@ definitions:
...
+            - $ref: '#/definitions/Schema'
+            - $ref: '#/definitions/Reference'
+      anyOf:
+        type: array
+        items:
+          oneOf:
+            - $ref: '#/definitions/Schema'
+            - $ref: '#/definitions/Reference'
+      items:
+        type: array
```
tfesenko: Wondering why array's item type is `array` and not an `object`. From the [OAS3 spec](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#properties):
> items - Value MUST be an object and not an array. Inline or referenced schema MUST be of a Schema Object and not a standard JSON Schema. <...>

webron: That's me being too tired. Good catch, will fix shortly.

schemas/v3.0/schema.yaml
```diff
@@ -0,0 +1,1500 @@
...
+        type: string
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false
+
+  OpenIdConnectSecurityScheme:
+    type: object
+    required:
+      - type
+      - openIdConnect
```
webron: Indeed.

schemas/v3.0/schema.yaml
```diff
@@ -0,0 +1,1500 @@
...
+      '^x-': {}
+    additionalProperties: false
+
+  Callback:
+    type: object
+    additionalProperties:
+      $ref: '#/definitions/PathItem'
+    patternProperties:
+      '^x-': {}
+    additionalProperties: false
```
webron: Thanks.

schemas/v3.0/schema.yaml
```diff
@@ -0,0 +1,1500 @@
...
+
+  Server:
+    type: object
+    required:
+      - url
+    properties:
+      url:
+        type: string
+        format: uriref
```
webron: I'd rather not get into that. I'm not sure what the difference is between `uriref` and `uri-template`. If it's too permissive, in this case, I'm ok with it. If it's too constrict, we can drop the format altogether. 

I find adding complex regex impossible to read and maintain. We could have added a regex to define media types as well and we didn't (we had a similar discussion about it around 2.0's schema).

MikeRalphson: Understood. I'm seeing that it's too restrictive being a `uriref`, not allowing `{}` characters. Dropping the `format` seems the best option.

webron: Makes sense, will remove.

webron:
@MikeRalphson thanks for the review - any other comments or thoughts?

schemas/v3.0/schema.yaml
```diff
@@ -0,0 +1,1496 @@
...
+      requestBody:
+        oneOf:
+          - $ref: '#/definitions/RequestBody'
+          - $ref: '#/definitions/Reference'
+      responses:
+        $ref: '#/definitions/Responses'
+      callbacks:
+        type: object
+        additionalProperties:
+          $ref: '#/definitions/Callback'
```
tfesenko: Should not it be Callback or **[Reference](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#operationObject)**:
> callbacks | Map[string, Callback Object \| Reference Object]

MikeRalphson:
@webron I'm seeing a rate of about 0.35% errors over 35,000 converted 'valid' OpenAPI 2.0 documents. All the failures appear to be correctly identified - i.e. things which have changed in the spec that the converter can't process automatically or that the 2.0 validators missed. This schema certainly validates more of the spec than the @timburks / #1236 schema, but I do think you may have made a rod for your own back by multiplying out the combinations of (for example) parameters `in` X with/without Y with/without a Z. 

What I'm trying to say is that maintenance of this schema is going to be somewhat harder going forward (depending on what changes come along in 3.1 and subsequent minor versions) as it has more redundancy and a slightly different mental model to grok rather than modelling only those objects that the spec defines. When (if?) we have a test-suite and a reference parser/validator, some of these issues go away.

The schema seems to be performant even on large documents, though the use of `oneOf` means more verbose error messages, at least from the schema validator I'm using.

That said... add the `id` and `schema` properties and :shipit:

schemas/v3.0/schema.yaml
webron: Yup.

webron:
@MikeRalphson Thanks for the feedback. 

Trust me, I don't like the breakdown as it is now, and I agree it's harder to maintain. The issue is derived from the way JSON Schema combines `allOf` with `additionalProperties: false`. If I had a solution to that, it'd make things much much easier. I only need one example of making it work and I can convert the rest.

schemas/v3.0/schema.yaml
```diff
@@ -0,0 +1,1498 @@
...
+      '^x-': {}
+    additionalProperties: false
+
+  ServerVariable:
+    type: object
+    required:
+      - default
+    properties:
+      enum:
+        type: string
```
MikeRalphson: `serverVariables.enum` should be an `array` of `string`.

webron: Thanks, should be fixed now.

hkosova:
@webron In this schema, `ParameterWithContentNotInPath` object seems to be missing the [`allowReserved`](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#user-content-parameterAllowReserved) property. Is this a bug or by design (e.g. `allowReserved` is incompatible with `content`)?

webron:
@hkosova it's by design. A parameter can have either `content` or a `schema`, and `style`, `explode` and so on are only applicable with `schema`.

hkosova:
@webron `openIdConnectUrl` has `format: url` - shouldn't it be `uriref` (relative URL allowed) like `tokenUrl` etc?

schemas/v3.0/schema.yaml
brandur-stripe: Hi @webron, as far as I can tell, `url` isn't a format that's in the official specification. It sounds like you're implying here that it's like a `uri`, but required to be absolute?

brandur: Hi @webron, as far as I can tell, url isn't a format that's in the official specification. It sounds like you're implying here that it's like a uri, but required to be absolute?

schemas/v3.0/schema.yaml
```diff
@@ -0,0 +1,1500 @@
...
+    type: object
+    properties:
+      summary:
+        type: string
+      description:
+        type: string
+      value: {}
+      externalValue:
+        type: string
+        format: uriref
```
brandur: I think that the modern name of a `uriref` [is `uri-reference`](http://json-schema.org/latest/json-schema-validation.html#rfc.section.8.3.7). It might be better to switch over to that in the name of general correctness.

schemas/v3.0/schema.yaml
```diff
@@ -0,0 +1,1500 @@
...
+        uniqueItems: true
+      type:
+        type: string
+        enum:
+          - array
+          - boolean
+          - integer
+          - number
+          - object
+          - string
```
brandur: Would it be possible to get `"null"` added to this list? We return some properties that are present in a response object, but contain a `null` value, and we need to have the type `null` to accurate describe them.

brandur: Ah, I see now that 3.0.0 indicates that it doesn't support the `null` type.

Do you have any feeling as to whether this is set in stone yet? Unless I'm really misunderstanding something, the trouble with `nullable` is that it's not in the JSON schema core, and means that we can no longer use off-the-shelf JSON schema validators to validate the schemas that are generated for OpenAPI 3.0.

brandur: And alright, I just read through https://github.com/OAI/OpenAPI-Specification/issues/229.

It doesn't look like the core team can be persuaded on this point, but oof, one of my perceived advantages of the 3.0 upgrade was not having to deal with the arbitrary restrictions that 2.0 imposed on JSON schemas. It seems that there are now fewer of them for 3.0, but they're still there, and JSON schema generated to be OpenAPI-compatible still cannot be used to accurately describe all types of API resources without additional meta-annotations (which are obviously not portable).

I'm not sure where that leaves us.

schemas/v3.0/schema.yaml
MikeRalphson: OpenAPI 3.0.0 uses JSON schema Draft Wright 00 (aka Draft 5) so [uriref](https://tools.ietf.org/html/draft-wright-json-schema-validation-00#section-7.3.7) is correct, [uri-reference](http://json-schema.org/latest/json-schema-validation.html#rfc.section.8.3.7) is defined in Draft Wright 01 (aka JSON Schema Draft 6).

brandur: Thanks Mike. I'd seen the [change in the changelog](http://json-schema.org/latest/json-schema-validation.html#rfc.appendix.B), but assumed maybe some tooling was outdated. Do you know what the rational for tying the spec and an older JSON schema version is? I guess because draft 6 introduces quite a few changes?

darrelmiller: @brandur ...because tying the spec to "whatever is the current version" would be impossible for tooling to deal with.  We had to lock on one version.

schemas/v3.0/schema.yaml
darrelmiller: V3 leaves us in the position where the Schema object can processed with JSON schema tooling without generating errors (due mainly to type: file).  It is true that not everything in JSON schema is supported.  Hopefully we can get there in the future.

MikeRalphson:
@webron I've ported over the descriptions from @timburks schema to yours. It includes duplicated (but not amended) descriptions for your ...With... objects. It seems to be complete except for around the lower-level security objects.

https://gist.github.com/MikeRalphson/349b3d9208c07a15aa200494ed49e8c4

schemas/v3.0/schema.yaml
handrews: @MikeRalphson it would be better to use `uri-reference` as formats are extensible.  It does not violate Draft 5, and is forward compatible with Draft 6.  There are actual Draft 6 validators in the wild (Draft 5 has never been directly implemented, and probably won't be).

schemas/v3.0/schema.yaml
```diff
@@ -0,0 +1,1500 @@
...
+        type: array
+        items:
+          type: string
+        minItems: 1
+        uniqueItems: true
+      enum:
+        type: array
+        items: {}
+        minItems: 1
+        uniqueItems: true
```
MikeRalphson: [JSON Schema Draft 5](https://tools.ietf.org/html/draft-wright-json-schema-validation-00#section-5.20) specifies that `enum` items only SHOULD be unique, not MUST. @handrews is there a specific reason non-unique values are permitted in `enum`s?

schemas/v3.0/schema.yaml
MikeRalphson: @handrews for better or worse, we're stuck with Draft 5 until at least OpenAPI 3.1.0 (or possibly 4.0 if it's deemed a breaking change). I think it would cause more confusion to use formats from Draft 6, and possibly result in those URI Reference properties not being validated at all. At least `ajv` can be convinced to operate in Draft 5 mode with a little setup. Of course there's no specific reason the schema for OpenAPI 3 has to itself be in Draft 5 just because OpenAPI 3 uses a modified version of Draft 5 internally, but again, it reduces confusion.

handrews: @MikeRalphson I was under the impression that a lot of people use extended formats with OpenAPI?  Most validators that support formats, including Ajv allow registering additional ones.

If this is the only extended (from the draft-05 perspective) format being used, then yes, I agree it makes sense _not_ to use it.  But if it's not, I'd encourage folks to ignore `uriref` and use `uri-reference` instead, just as they would use `byte` or `password` or some other non-standard format.

schemas/v3.0/schema.yaml
handrews: @MikeRalphson IIRC the idea was to treat the enum values like an unordered set, in which duplicates are harmless and ignored.  This might be useful in auto-generated schemas that pull enum values from dynamic sources that may overlap.

JSON Schema tends to err on the side of permissiveness, as it is easy to add restrictions and hard to remove them.  In draft-08 I hope we can make it easier for users of JSON Schema to specify restrictions- adding a `uniqueItems` restriction to `enum` is easy enough in terms extending the meta-schema.

handrews:
@webron 

> To make it more restrictive, there are somewhat a lot of duplication in the schema. Using allOf to reuse things didn't work for me due to `additionalProperties: false`. If anyone wants to take a crack and minimizing it, by all means... I just wanted to get a working version.

We're tackling this in draft-08.  Specifically, https://github.com/json-schema-org/json-schema-spec/issues/515 lays out the main options and how they can be made to fit (or not) into JSON Schema's processing model.  It's a bit of a slog to read, but given the significance of OAI as a user of JSON Schema (both within the spec and to describe the spec), if anyone from here would like to weigh in that would be great.  It's easily the most divisive issue in all of JSON Schema's history, so the more community involvement we get on this topic the more likely the result is to be broadly acceptable.

MikeRalphson:
I've been wondering whether we actually want to have `additionalProperties` false everywhere in the schema, as it leaves us no mechanism for non-breaking minor changes (v3.1.0 etc).

handrews:
@MikeRalphson in general `"additionalProperties": false` makes evolving a schema impossible.  Everything becomes a breaking change.  `"required"` has some similar problems but they're more intuitive- folks get that you can't just make something not-required and expect clients to be OK with it.

But `"additionalProperties": false` is often used to mean "I want to have an error of someone misspells something" or "I want to prevent extensions from outside of the project's control".  Which is not what it's for.  `"additionalProperties"` is for implementing generic map structures with a uniform value type.  Including allowing a few special cases and then making all other keys fit a generic structure.  When you set it to `false` it has the effect of banning all other properties but that's not really what it was designed for as far as I can tell.

schemas/v3.0/schema.yaml
handrews: @MikeRalphson 

You can also make things mutually exclusive by forbidding the other property rather than requiring the property that you have:

```YAML
oneOf:
  - properties:
        example:
            properties:
                ...
        examples:
            not: {}
  - properties:
        example:
            not: {}
        examples:
            patternProperties:
                ...
```

schemas/v3.0/schema.yaml
```diff
@@ -0,0 +1,1500 @@
...
+    additionalProperties: false
+
+  Schema:
+    type: object
+    properties:
+      title:
+        type: string
+      multipleOf:
+        type: number
+        minimum: 0
```
handrews: Why is there a minimum on this?  Negative numbers are valid for `multipleOf`

schemas/v3.0/schema.yaml
handrews: @brandur see also #1389

schemas/v3.0/schema.yaml
handrews: `uri` is the correct format for an absolute URI.
`uriref` (which became `uri-reference` in the next draft so that is really preferred) is for relative URI references.

It would be really great to avoid `url` as having `uri` and `url` is horribly confusing, particularly as RFC 3986 goes out of its way to explain that there is no hard line between `url` and `urn` (e.g. you cannot make the distinction based on the scheme)

schemas/v3.0/schema.yaml
handrews: Wait, never mind, I confused myself.  This is, of course, correct.

deathiop:
As many people, I do have a lot of interest in the changes the OAS3 version brought.
This pull request seems to be stucked. Is there still plans to add this schema? It's the base of all tools we might need, and until it is released, people might continue to redact/generate OAS2 documents.

The request for changed has been set in July.
What can we do to move forward on that topic?

handrews:
@deathiop after talking with @MikeRalphson about this yesterday, I'm going to show how the `"additionalProperties": false` stuff could be reworked with a proposal slated for the next draft of JSON Schema.  I'll make a gist of it and link here so that y'all and @webron can take a look without too much clutter.

Obviously a proposal for the next draft (which I hope to publish before the current draft expires on May 23rd) is not a slam-dunk solution here, but perhaps showing that this approach will eventually be handled better will make it a more palatable short-term solution.  Or OAI could decide that the JSON Schema version used to describe OAS does not need to be the same version as OAS uses within its own specification.  The schema already uses features that are not allowed within OAS even though it is using the same draft version, so it would not be that much more different.

schemas/v3.0/schema.yaml
```diff
@@ -0,0 +1,1400 @@
...
+        oneOf:
+          - $ref: '#/definitions/Response'
+          - $ref: '#/definitions/Reference'
+      '^x-': {}
+    minProperties: 1
+    additionalProperties: false
+    not:
+      type: object
+      patternProperties:
+        '^x-': {}
```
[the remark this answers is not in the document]
handrews: @webron @MikeRalphson why is this `not:` block here?  Filtering out other stuff, you have

```YAML
  Responses:
    type: object
    patternProperties:
      '^x-': {}
    additionalProperties: false
    not:
      type: object
      patternProperties:
        '^x-': {}
      additionalProperties: false
```

which is an impossible schema.  `{"type": "object", "not": {"type": "object"}}` alone makes it impossible.  Any keywords that are the same inside and outside of a `not` by definition are impossible to satisfy.

Am I misreading something here?

handrews:
I have completed a [refactored schema](https://gist.github.com/handrews/6dfebd56ef97328f9e4dc7a47a1e8bc7) using draft-08 proposals.  I'm sure it has bugs as I haven't tested it at all (there are no draft-08 implementations as we haven't even published it yet!)

This obviously won't help anything immediately, but it shows that while we may need to accept a lot of difficult-to-maintain duplication for now, the situation will improve in the near future, at which point we can refactor this or just make more maintainable ones for future OAS versions.

If anyone would like to see more about the `unevaluatedProperties` proposal or provide feedback, it is issue json-schema-org/json-schema-spec#556.

handrews:
While the refactor I linked to in the last comment requires draft-08, I think I can see how to compress some of this down without relying on newer features.  In particular, variations that differ only by enum values, defaults, and/or required-ness can be factored out without breaking the `"additionalProperties": false` keywords.  There are also some tricks that can be done to keep the field definitions in one place while placating `"additionalProperties"`.  I can make a PR against the PR to demonstrate.

schemas/v3.0/schema.yaml
```diff
@@ -0,0 +1,1500 @@
...
+      - type
+      - openIdConnectUrl
+    properties:
+      type:
+        type: string
+        enum:
+          - openIdConnect      
+      openIdConnectUrl:
+        type: string
+        format: url
```
MikeRalphson: @jblazek spotted that this either should be `format: uriref` or there is an omission in the spec requiring an absolute URL.

schemas/v3.0/schema.yaml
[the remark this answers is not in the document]
webron: @MikeRalphson - not that I recall, but it seems that right now, there's no `$ref` property in `Schema` and it stayed `oneOf` wherever a Schema is called. I _think_ that should be fine. Please let me know what you're thinking - if it's fine, I'll create a definition for `SchemaOrReference` and use that across the board instead of repeating the `oneOf` everywhere.

philsturgeon:
On a TSC call in the past I mentioned folks are either struggling to (or refusing to) upgrade their tooling to support v3 until there is a meta Schema. I only had one example offhand but here is another that just came up. https://github.com/apigee-127/sway/issues/128#issuecomment-407769153

robertmassaioli:
@philsturgeon Anecdotally, this is the case at my place of work. Having a schema is critical to allowing the many teams that we have to commit to the spec. We also use Swagger files as part of our automated testing and documentation. So yes, an official JSON Schema would be awesome.

schemas/v3.0/schema.yaml
handrews: @webron @MikeRalphson I'm planning to change this to `uri-reference`, even if we are saying that this schema is a draft-wright-*-00 (a.k.a. "draft-05") schema, and that draft used `uriref`.  I'm doing this because:

1.  Literally no one has ever implemented a validator for draft-wright-*-00.  Ever implementation that I'm aware of skipped from draft-04 (draft-zyp-json-schema-04 + draft-fge-json-schema-validation-00) to draft-06 (draft-wright-*-01).
2.  There is no official meta-schema for draft-05, and there never will be, so there is no programmatic way to inform validators that you are using draft-05
3.  Format is extensible, so using `uri-reference` as a format is entirely valid even in draft-05

Any objections?

handrews: Ideally we could also publish a draft-06 version where `uri-reference` would be recognized, but `format` validation in general is such a hazy weird optional thing that, really, no one should rely on it unless they are using a specific validator and know it's exact behavior.  e.g. some validators try to use a complex regex to validate the `email` format, but at least one popular validator just checks to see if there's an "@" character in it somewhere and doesn't bother with anything else.

handrews:
@MikeRalphson @webron @darrelmiller there was some talk in one of the comments about possibly not using `"additionalProperties": false` in order to allow the same schema to work for 3.1.  I am assuming we've decided against that since folks seem to want the schema to detect misspelled properties.

Please let me know if I'm wrong about that.

darrelmiller:
TSC Action items:
- Add a $schema property (draft-4 metaschema) 
- Track revisions of this document using a "id" property  e.g.  https://spec.openapis.org/oas/3.1/schema/2019-03-21  
- Convert YAML to JSON

MikeRalphson:
@webron @handrews I've rechecked @vearutop's latest schema and have seen no other regressions.

If we are declaring (via `$schema`) that this is a Draft 04 schema, we should (I feel) change each occurrence of the `uri-reference` `format` back to `uriref` so Draft 04 validators **by default** perform validation on those properties.

handrews:
@MikeRalphson neither `uriref` nor `uri-reference` is part of draft-04.  Technically draft-04 only has the `uri` format for full URI (that start with a scheme).

Your options are:
* Don't use a `format` on the grounds that there isn't a draft-04 format for this
* Use `uriref`_as an extension_  which was briefly described in a draft that **is not possible to implement and has never been implemented**
* Use `uri-reference` _as an extension_ which is actually what everyone post-draft-04 uses and will be compatible with that, at least

I recommend either the first or last options, but strongly recommend against `uriref`.  It doesn't really exist in any sort of practical way whatsoever.

MikeRalphson:
Thanks for clarifying @handrews - I obviously misremembered the situation.

webron:
@MikeRalphson, @handrews - I've updated the schema with the missing metadata (`id`, `$schema`, `description`). Can you please run final validation and tests on it? 

If you give the thumbs up, and we get an official @OAI/tsc vote, I'll add a conversion of the schema to JSON, and merge it (at last).

philsturgeon:
🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢

webron:
Added the converted JSON file and a README with some relevant details.

Thanks everyone for making it happen.

philsturgeon:
giphy

Great work everyone, this is awesome.

cebe:
this has been merged into `OpenAPI.next`, what is the process or what are the plans of getting it into master?

webron:
@cebe *sigh* just shows how old this PR was. I'll move the content to `master` as well. Thanks for catching it.

cebe:
hehe, I thought this was part of your workflow :)

Issue #1116, opened 2017-05-19
api-with-examples.yaml: `example` vs `examples` in Media Type Object


tfesenko:
Thank you for merging PR #985 and providing official examples of OpenAPI v3 files! :clap::clap::clap:

Is it possible that `api-with-examples.yaml` has errors?

I am asking because I validated all four  examples/v3.0/ files with [JSON Schema](https://github.com/googleapis/gnostic/blob/openapi-v3.0.0-rc2/OpenAPIv3/openapi-3.0.json) created by @timburks and @MikeRalphson. The JSON Schema validation is OK on `petstore-expanded.yaml`, `petstore.yaml`, and `uber.yaml`, but fails on `api-with-examples.yaml`. 

Validation fails on content `examples` - there are four `examples` in the file, the first of which is defined on [line 16](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/examples/v3.0/api-with-examples.yaml#L16):
```yaml
          content:
            application/json:
              examples: 
                - {
                    "versions": [
                        {<...>}, {<...>}]
                   }           
```

[Media Type Object](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#mediaTypeObject) can define examples in two ways:
* `example` (singular) - inlines example text directly (`Any` in JSON Schema);
* `examples` (plural) - uses the [Example Object](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#exampleObject), which also provides additional information, such as example name, summary, or description. 

The problem is that `api-with-examples.yaml` defines examples with `examples`, but inlines them as if they were defined with `example`. 
In my understanding of the spec, `api-with-examples.yaml` should be corrected:
1. Either `examples` keyword should be replaced by `example`:
```yaml
          content:
            application/json:
              example: 
                  {
                    "versions": [
                        {<...>}, {<...>}]
                   }           
```
2. Or, we should use a real named example object where the example snippet is defined in `value` (or `externalValue`):
```yaml
          content:
            application/json:
              examples: 
                versions-example:
                  summary: Versions Example in JSON
                  value:  {
                    "versions": [
                        {<...>}, {<...>}]
                   }      
```

Can you please clarify the usage of the `examples` keyword in `api-with-examples.yaml`?

darrelmiller:
These examples need to be fixed.  Sorry for the confusion

zbinlin:
[...]

PR #1286, opened 2017-07-21
Fix examples object example and api-with-examples

MikeRalphson:
Fixes a few issues:

* misleading comments
* `examples` is no longer an array anywhere
* use of `$ref`s in `examples` (as opposed to `example`) - have preferred `externalValue` when refererencing whole files (which may not be json/yaml) and `$ref` in `example` which can have a fragment id. Think this is the intent.

Refs #1116


webron:
@darrelmiller will fix more things!

Issue #1292, opened 2017-07-24
Invalid `operationRef` in links example


tfesenko:
[OperationRef Examples](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#operationref-examples) suggests a value with a missing `/` separator after the `paths` in `operationRef`: 
`'#/paths~12.0~1repositories~1/{username}/get'` , but it should be `'#/paths/~12.0~1repositories~1/{username}/get'`  == `'#/paths/<'/2.0/repositories/{username}'.encode()>/get`.

The example of an absolute operationRef correctly uses a JSON Pointer with a `/paths/<JSON Pointer-encoded path item>`.

Related to https://github.com/OAI/OpenAPI-Specification/issues/1152

webron:
@tfesenko thanks. do you have time to submit a PR to fix it? we're getting very close to the actual release.

tfesenko:
@webron wrote:
> do you have time to submit a PR to fix it? 

Sure! I created PR #1293 for this. Just did not know if you wanted to include this fix into the release.

webron:
Thanks, I'll review in just a few minutes. Anything that can be fixed before the actual release is welcome. This means we'll have less patch releases which is always a good thing.

webron:
Closed as fixed, thanks again!

PR #1293, opened 2017-07-24
#1292 Added a missing / separator after the paths in operationRef

tfesenko:
Added a slash (`/`) after the `/paths`, so now it's `#/paths/~12.0~1repositories~1/{username}/get` == `#/paths/<'/2.0/repositories/{username}'.encode()>/get`. See #1292 for details.

There is a PR #1291 which renames 3.0.md to 3.0.0.md . 
In this PR, I am modifying 3.0.md file which is used on the current OpenAPI.next branch. Please let me know if I need to modify 3.0.0.md instead.

### DIFF
--- a/versions/3.0.md
+++ b/versions/3.0.md
@@ -2069,7 +2069,7 @@ value), references MAY also be made through a relative `operationRef`:
 links:
   UserRepositories:
     # returns array of '#/components/schemas/repository'
-    operationRef: '#/paths/~12.0~1repositories~1/{username}/get'
+    operationRef: '#/paths/~12.0~1repositories~1{username}/get'
     parameters:
       username: $response.body#/username
 ```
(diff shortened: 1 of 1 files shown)


versions/3.0.md
```diff
@@ -2069,7 +2069,7 @@ value), references MAY also be made through a relative `operationRef`:
 links:
   UserRepositories:
     # returns array of '#/components/schemas/repository'
-    operationRef: '#/paths~12.0~1repositories~1/{username}/get'
+    operationRef: '#/paths/~12.0~1repositories~1/{username}/get'
```
webron: Looking at it again, it seems this might need to be `#/paths/~12.0~1repositories~1{username}/get`?
That would translate to:

```yaml
paths:
  /2.0/repositories/{username}:
    get:  
```

tfesenko: Oops, missed that one.
Fixed in be6f83d

webron:
Thanks for getting this in!

PR #1298, opened 2017-07-26
Fix callback-example.yaml

hkosova:
This fix adds quotes around `{$request.query.callbackUrl}/data` to make it valid YAML.

`{` is a special character that starts a mapping, so string literals beginning with `{` need to be escaped.

**UPD:** Now that 3.0 is released, should the PR target `master` or `OpenAPI.next`?


hkosova:
Closing, because there's another PR with the same fix - https://github.com/OAI/OpenAPI-Specification/pull/1228.

PR #1299, opened 2017-07-26
Add quotes around response statuses in standalone 3.0 examples

hkosova:
According to the 3.0 Specification, response status codes must be quoted. This PR fixes the examples in the `examples\v3.0` folder.

Single vs double quotes: Some examples use `"` for quoting and others use `'`. I used the same quote character as in the rest of each file.


webron:
@hkosova - for consistency, can you change it all to single quotes?

hkosova:
@webron - Done.

webron:
Thanks @hkosova

PR #1332, opened 2017-08-21
object -> field

webron:
More consistent wording and clearer intention - there's no change to the meaning.


webron:
Need to change a couple of `object`s to `field` and then merge.

PR #1335, opened 2017-08-23
Add link to Open Api Spec Library

faragorn:
There is a way to describe any API with Swagger but there is no single place to share your specifications with others.
I created a repo that will hopefully become a library of specifications for various APIs that are in the wild.
Here is the link: https://github.com/faragorn/open-api-specs
I hope this is the right place to list this initiative.


MikeRalphson:
Just a note re: a similarly focused project with almost 600 OpenAPI definitions https://github.com/APIs-guru/openapi-directory - would be happy to have another collaborator.

faragorn:
Didn't know it exists @MikeRalphson . I will contribute the definitions I have. Cheers!

PR #1338, opened 2017-08-28
Moved 'links' object to response object

semonte:
Updated the link object example. According to the specification, 'links' should be inside of a response object.


darrelmiller:
Thanks for the PR.  That issue was corrected in the. 3.0.1 just a few days ago.  https://github.com/OAI/OpenAPI-Specification/pull/1336

PR #1350, opened 2017-09-15
Correct white space for several code blocks

mikenholt:
There was a slight white space error causing the example code to be rendered outside the code block.


webron:
Thank you for the PR. Existing versions are sealed and cannot be modified anymore.

Only fixes to 3.0.0.md can be accepted to the 3.0.1.md file/branch (which is the next patch version).

PR #1357, opened 2017-09-28
Add json examples for v3 specification

learnercys:


examples/v3.0/json/petstore-expanded.json
```diff
@@ -0,0 +1,239 @@
+{
+  "openapi": "3.0.0",
+  "info": {
+    "version": "1.0.0",
+    "title": "Swagger Petstore",
+    "description": "A sample API that uses a petstore as an example to demonstrate features in the swagger-2.0 specification",
```
MikeRalphson: `swagger-2.0 specification` should be changed to `OpenAPI 3.0 specification` as per PR #1373

darrelmiller:
Thanks for the PR.  We do think having the JSON examples are valuable, however, we don't want to add the additional maintenance burden to keep them up to date.  We plan on adding tooling to the repo to generate these JSON examples.  However, we can't merge this into master because the spec has some links to the 3.0 examples.  We will do this work in the 3.0.1 release. See #1385

PR #1381, opened 2017-10-19
Correct comment on the link object

Bert-R:
The example referencing the response body had the same comment as the one referencing the request path.


Bert-R:
Done.

Bert-R:
Closing this one, as it seems messed up after changing the base.
The change is available in #1386.

PR #1386, opened 2017-10-20
Correct comment on the link object

Bert-R:
The example referencing the response body had the same comment as the one referencing the request path.

This one replaces #1381, which seems messed up after changing the base.


webron:
Merging as discussed on our weekly call.

PR #1419, opened 2017-11-21
Examples; replace Uber with USPTO

MikeRalphson:
This PR replaces the commercial, outdated and oAuth protected Uber example with a public, current and open one from the US Patent & Trademark Office.

Disclaimer: This is a slightly cut-down (bulk download endpoints have been removed for brevity) converted OpenAPI 2.0 definition (thus the Uber example in v2.0 could also be replaced). The response model is not fully defined.

Although it relies on a US Government department, it seems unlikely to come under direct threat from the current administration.


examples/v3.0/uspto.yaml
```diff
@@ -0,0 +1,216 @@
+openapi: 3.0.0
+servers:
+  - url: 'https://developer.uspto.gov/ibd-api'
```
earth2marsh: Is it worth turning the scheme into a variable with enums of `http` and `https`?
If not, what about adding a description to explain the two URLs?

examples/v3.0/uspto.yaml
```diff
@@ -0,0 +1,216 @@
...
+    Bulk data search and download API exposes an API to search published patent
+    grants and applications across various fields.
+  version: 1.0.0
+  title: Bulk data search API
+  contact:
+    name: Open Data Portal
+    url: 'https://developer.uspto.gov'
+    email: developer@uspto.gov
+tags:
+  - name: bulkdata
```
earth2marsh: Consider a description for the tag to show best practices.

examples/v3.0/uspto.yaml
MikeRalphson: I must confess I hadn't thought of using server variables for anything except the host, port or path (where RFC6570 templates are normally used), but happy to go with that, or the description option.

examples/v3.0/uspto.yaml
MikeRalphson: Will amend.

examples/v3.0/uspto.yaml
earth2marsh: I don't have a strong pref, either will improve this slightly. Given that http/https should be somewhat common, it seems like that is self-evident as a variable.

MikeRalphson:
I've replaced the Bulk Data Search API with the Data Set API, as this has a `post` request as well as `get`. Hopefully this addresses the other points raised:

* [x] uses server variable for scheme
* [x] includes tag descriptions
* [x] includes an example response
* [x] has `$ref` to `#/components/schemas`
* [x] has been validated & checked against both the draft schemas

darrelmiller:
Thanks for all your work on this @MikeRalphson !

PR #1422, opened 2017-12-01
[v3.0.1] Prefer https links where available

MikeRalphson:
This updates links to apache.org (including in examples), iana.org, ietf.org and semver.org to https for security and consistency. It leaves http links to json-schema.org, yaml.org and commonmark.org as they either don't provide https or Firefox 57 complains about their https setup.

(ping @handrews)


handrews:
Sigh.  GitHub pages won't allow HTTPS with custom domains (you can do one or the other but not both).  And we don't actually own the DNS account, although one of the former JSON Schema editors has a way to pay for it (how that works I do not know).  We really need to find some way to claim that account or give up and get a new domain name and host it somewhere that will support HTTPS :-/

MikeRalphson:
> Sigh.

:( as long as you're aware.

handrews:
Oh, we are so very aware.  It's the same reason why there's a json-schema/json-schema GitHub account still lying around.  Someone files a bug on all this at least once or twice a year.

PR #1432, opened 2017-12-08
Update 3.0.1.md

nkolban:
Indention errors and missing `type: object` YAML entry


webron:
Thanks for the PR.

We do not accept PRs changing released versions of the spec. 
Early next week we'll have the 3.0.2 branch and you could submit the PR against that branch and version.

nkolban:
Thank you sir.  Will re-check the 3.0.2 and work against that.

PR #1436, opened 2017-12-13
Update Open API object to OpenAPI Object for consistency

PerthCharern:

PR #1444, opened 2017-12-15
Update Open API object to OpenAPI Object in 3.0.2

PerthCharern:
Update Open API object to OpenAPI Object in 3.0.2

(Same change as #1436 but submitted to the correct branch)


PerthCharern:
@darrelmiller Updated the base branch as instructed :D

PR #1457, opened 2018-01-13
add a future specfication 3.0.2.md fix issue/1449

### LINKED ISSUE #1449: The description `Invalid Input` of 405 Responses Object is ambiguous  and leads to misunderstanding
As the document says, [the key of a responses object's Patterned Fields ](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#patterned-fields-1) is  Http Status Code . 

and the document gives  responses object examples  like this at following part.

```
  "responses": {
    "405": {
      "description": "Invalid input",
      "content": {
        "application/json": {},
        "application/xml": {}
      }
    }
  }
```

As it defined in [ rfc7231 Specification](https://tools.ietf.org/html/rfc7231#section-6.5.5) , 405 indicates `Method Not Allowed` rather than `Invalid Input` .   

>  The 405 (Method Not Allowed) status code indicates that the method
   received in the request-line is known by the origin server but not
   supported by the target resource.  

I think we'd better follow the [ rfc7231 Specification](https://tools.ietf.org/html/rfc7231#section-6.5.5) 's definition

weimengxi:
Fixes #1449

### DIFF
--- a/versions/3.0.2.md
+++ b/versions/3.0.2.md
@@ -1,0 +1,3380 @@
+# OpenAPI Specification
+
+#### Version 3.0.1
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [BCP 14](https://tools.ietf.org/html/bcp14) [RFC2119](https://tools.ietf.org/html/rfc2119) [RFC8174](https://tools.ietf.org/html/rfc8174) when, and only when, they appear in all capitals, as shown here.
+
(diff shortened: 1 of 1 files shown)


MikeRalphson:
@weimengxi thanks for amending the PR to touch the `3.0.2.md` file only, but could you also change the base branch to `v3.0.2-dev` from `master`? To clarify, PRs against future versions of the spec. should be in the form of an edit, not creating a new version file. Please let us know if [the development guidelines](https://github.com/OAI/OpenAPI-Specification/blob/master/DEVELOPMENT.md) need to be made clearer.

MikeRalphson:
Hi @weimengxi, I've tried to change the base branch of this PR to the `v3.0.2-dev` branch, but it created conflicts. Please could you resubmit as an edit to the `versions/3.0.2.md` file on that branch, as per the  [the development guidelines](https://github.com/OAI/OpenAPI-Specification/blob/master/DEVELOPMENT.md)?

PR #1463, opened 2018-01-25
Updated security definitions text link to reference the Security Defi…

Daniel-Houston:
…nitions Object in the 2.0 spec.

I am new to open source pull requests, so if I have done anything incorrectly please let me know and I will fix the problem. 

Thanks!

### DIFF
--- a/versions/2.0.md
+++ b/versions/2.0.md
@@ -2239,13 +2239,13 @@ read:pets: read your pets
 
 Lists the required security schemes to execute this operation. The object can have multiple security schemes declared in it which are all required (that is, there is a logical AND between the schemes).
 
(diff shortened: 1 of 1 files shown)


MikeRalphson:
@Daniel-Houston thanks for the PR! It looks correct, the only issue is that we are treating the 2.0 version of the spec as frozen, warts and all. If you find similar problems in the current version of the spec, please open PRs against the upcoming patch version in the relevant `_dev` branch (currently `versions/3.0.2.md` on the `v3.0.2_dev` branch).

Thanks again.

Daniel-Houston:
Sounds good, thanks for the response!

Issue #1347, opened 2017-09-15
Update license to let Github do its magic.


darrelmiller:

webron:
Implemented through #1541.

PR #1474, opened 2018-02-05
Change to default Apache 2.0 license text

dankohn:
GitHub's license detection isn't smart enough to recognize what is there now


MikeRalphson:
Fixes #1347

webron:
@dankohn - thanks for the PR, we appreciate the effort. We went ahead and merged #1541, really just because we ended up getting official approval over there, no other reason. Thanks for taking the time to suggest the change.

dankohn:
Great!

PR #1485, opened 2018-02-10
Add swagger-node-codegen

richardklose:


webron:
Supports OAS3, merging.

PR #1486, opened 2018-02-11
Revise 405 Response Object description to follow the RFC7231 Specification 's definition

weimengxi:
As the document says, [the key of a responses object's Patterned Fields ](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#patterned-fields-1) is  Http Status Code . 

and the document gives  responses object examples  like this at following part.

```
  "responses": {
    "405": {
      "description": "Invalid input",
      "content": {
        "application/json": {},
        "application/xml": {}
      }
    }
  }
```

As it defined in [ rfc7231 Specification](https://tools.ietf.org/html/rfc7231#section-6.5.5) , 405 indicates `Method Not Allowed` rather than `Invalid Input` .   

>  The 405 (Method Not Allowed) status code indicates that the method
   received in the request-line is known by the origin server but not
   supported by the target resource.  

I think we'd better follow the [ rfc7231 Specification](https://tools.ietf.org/html/rfc7231#section-6.5.5) 's definition

#1449

### DIFF
--- a/versions/3.0.2.md
+++ b/versions/3.0.2.md
@@ -903,7 +903,7 @@ This object MAY be extended with [Specification Extensions](#specificationExtens
       }
     },
     "405": {
-      "description": "Invalid input",
+      "description": "Method Not Allowed",
       "content": {
         "application/json": {},
         "application/xml": {}
@@ -953,7 +953,7 @@ responses:
       'application/json': {}
       'application/xml': {}
   '405':
-    description: Invalid input
+    description: Method Not Allowed
     content: 
       'application/json': {}
       'application/xml': {}


weimengxi:
hi @MikeRalphson,  sorry to disturb you again ~

MikeRalphson:
Not at all, @weimengxi thank you for resubmitting. Looks good now. Just needs another reviewer to sign off.

webron:
@MikeRalphson here you go.

pleothaud:
Hi there, it seems that there is a underlying problem here: as responses ca only be defined for an operation (an accepted method on a specific path) defining a 405 "Method not allowed" response makes no sense, as you're just writing that you accept this method (by exposing it in the OAI Spec file)

IMHO the only place holder to define the 405 response would be at the path level, but that's a huge change.

MikeRalphson:
What if the method is not allowed for some users, but is fir others?

darrelmiller:
What @MikeRalphson  said, and....

The allowed methods are dynamic over the lifetime of a resource.  That's why OPTIONS returns an Allow header.  Something that may be initially updatable, may at some time later be read-only.  Also, paths are templated and therefore some resources under the path may support one set of methods and other resources may support a different set of methods.

pleothaud:
HI Mike, Darrel,

You're damn right! Sorry, working ATM on a product that must filter out what the backend sends back and validate that it cnforms to what's in the OAI Spec file and I was just wondering how you could specify a 405 for a method that is not authorized in any context...

All in all all what I've got to do is define only a 405 response for this method and voila.

Sorry again!

Issue #1507, opened 2018-03-12
Introduce tags hierarchy


dtsaryov:
Hello.

My proposal is to add an ability to create tags hierarchy, for example:

```yaml
tags:
  - name: Entites
    description: CRUD entities operations
      - User
      - Group
  - name: Services
    description: Predefined queries execution
      - FooService:
        - name: ...
          description: ...
      - BarService:
        - name: ...
          description:
```

It'll help to group operations in case of generic REST API.

MikeRalphson:
See also #1367

ymohdriz:
Hi @prozac631,

If I understood correctly, open api provides a way to tag in operation level. Hope we can group operations by using operation level tagging.
Let me know if your requirement is completely different.

Thanks,
Mohammed

prozac631:
Hi @ymohdriz 

Yes, we can use grouping for operations, but it would be nice if we could group operations inside of group of operations. Let me make it more clear.

For example, you have operations and their paths have the following template: `/entities/*` (`entities/User`, `entities/Group`, etc).

And in the same time you have a set operations that can be performed with these entities, for example: `entities/User/create/`, `entities/User/get/`, `entities/User/delete`, etc.

And in this case we can create a tag `Entities` and sub tags `User`, `Group`, etc.

maaland:
I support the request for sub-tags. I want to be able to group (or at least sort) the methods under each of my endpoints, and as far as I can tell, that's not possible right now.

silkentrance:
See also #1642. The path is already a physical group, why do you need an indirection for it?

jukeblimp32:
I really need subtags. My current tags are getting far to cluttered.

farmbotgianpaolo:
I support the request for sub-tags or tag hyerarchy. The reason is simple: if I have a single endpoint where I built public and private APIs I would like to group them into these 2 categories.

storm367:
Tag Hierarchy doesn't seem to be supported yet? I support this request to add sub-tags.

dschmide:
Adding sub-tags is ab obvious advantage. Remember, there is no search and filter option for the apidocs, so this tagging system is all we can provide to devs. 
At this point, my co-workers are telling me to add subgroups, and not believing me when I tell them that "its not possible", but sadly, this is a reality.
I started embedding "tags" right in the paths so the structure within tags is at least somewhat preserved by abusing the alphabetical ordering - but this is clearly a symptom of a lack of feature.

[...]

Issue #1367, opened 2017-10-05
Feature: sub-tags grouping


kamil-kielczewski:
I don't know it is good place to introduce new Feature into swagger specification - if not please give me information where i can put this proposition.

My proposition is to use following convention in tags names (using dots to nesting subtags):

```
users.client     # and use this tag to client GET/POST/...
users.seller     # and use this tag to seller GET/POST/...
users.admin      # and use this tag to admin GET/POST/...
books            # use this to general books GET/POST...
books.sell       # use this to GET/POST... associated with the sell
books.management # use this to GET/POST... associated with the books management
```

And in output generated documentation web-page we will get:

```
user >
       client >
GET
POST
...
       seller >
GET
POST
...
       admin >
GET
POST
...

books >
GET
POST
...
       sell >
GET
POST
...
       management >
GET
POST
...
```

This feature is especially practical for large APIs.

Or may be better way is not change 'tags' but rather add new filed 'groups' which will contain such dot-separated nested groups?

ymohdriz:
Hi @kamil-kielczewski 

Adding the path level tagging solves the requirement, if I understood correctly.
I have raised a PR https://github.com/OAI/OpenAPI-Specification/pull/1509 for the same.

Thanks,
Mohammed

maaland:
I support the request for a sub-tags. I want to be able to group (or at least sort) the methods under each of my endpoints, and as far as I can tell, that's not possible right now.

jukeblimp32:
I support the request for sub-tags. Right now my tags are my collections. But with some collections exposing several different resources that can each support GET, PUT, and PATCH, the tags can become very cluttered. I want to be able have sub-tags for each resource within the collection tag.

farmbotgianpaolo:
I support the request for sub-tags. I have a big API with tens of access points that need categories.

anentropic:
This is one area where apib is better... there is a hierarchy like "Resource Group > Resource > Transition" where a Transition is roughly an OAS `path` and a Resource could be represented with an OAS `tag`.

For large APIs it would be useful to be able to group the endpoints at a level above single tags.

OAS is flexible enough I can represent this in the data with some combination of tags and extension properties, but for it to really be useful you would want a standardised representation that the tools such as Swagger UI then also support.

amandajordaan:
I have a very large API to document, large with a very basic structure:
API {
Service 1 {
Tool 1,
Tool 2,
Tool 3, (etc)
}
[...]

PR #1509, opened 2018-03-15
Adding path level tagging

ymohdriz:
Hi team,

Currently OAS supports tagging in API and operation level. This PR is to add tagging in path level also.
Related feature request related to this:
OAI#1367

Please review the PR and merge the same.

Thanks,
Mohammed

### DIFF
--- a/schemas/v2.0/schema.json
+++ b/schemas/v2.0/schema.json
@@ -338,7 +338,14 @@
         },
         "parameters": {
           "$ref": "#/definitions/parametersList"
-        }
+        },		
+		"tags": {
+		  "type": "array",
+		  "items": {
+			"$ref": "#/definitions/tag"
+		  },
+		  "uniqueItems": true
+		}
       }
     },
     "responses": {


darrelmiller:
Hey @ymohdriz Thanks for the PR.  This is definitely an interesting addition that has a low impact on the spec.  However, we do have to consider the impact on tooling.  We would need to get feedback from folks who are producing documentation based on OpenAPI as to whether they have user requests for these features and whether they are prepared to do the work to incorporate this capability.

Also, this kind of change cannot go into a patch release, it would need to go into a 3.1 branch which I don't believe we have yet created because we are still working on what features will go into 3.1.

We will definitely consider this for inclusion and the more feedback we can get from tooling people on this, the easier it will be to make the call.

ymohdriz:
Hi @darrelmiller,

Thanks for the detailed response.
I would like to contribute to this project. But new PRs may turn into new features. 
Can you please suggest some areas to work on?

Thanks,
Mohammed

emreakca:
Hey @darrelmiller,

What's the state? Are subtags considered to be a feature, because it would be great!
If it's possible, a push would be very helpful!

MikeRalphson:
This PR cannot be accepted as-is, because a) it is targeted at v2.0 of the specification, which is published and will not change, b) it only updates the schema file, which is not the source-of-truth of the specification (the markdown documents are), c) as @darrelmiller says, this change (if accepted) would need to go into a minor release, such as v3.1.0

@emreakca if you need sub-tag grouping / a hierarchy of tags, it would probably be worth commenting on #1367 or #1507 as appropriate.

MikeRalphson:
Closing, for the reasons given in https://github.com/OAI/OpenAPI-Specification/pull/1509#issuecomment-402967954

PR #1521, opened 2018-04-02
PatternProperties are not supported

philsturgeon:
This keyword is supported by [draft-wright-json-schema-validation-00](https://tools.ietf.org/html/draft-wright-json-schema-validation-00#section-5.17) but is not supported by OpenAPI. As such we gotta list it on the spec along with other discrepencies.


MikeRalphson:
@philsturgeon This difference is currently covered by the spec as follows:

> Additional properties defined by the JSON Schema specification that are not mentioned here are strictly unsupported.

It would seem odd to additionally call out `patternProperties` but not `dependencies`, `id` or `$schema` for example.

If you think the differences should be enumerated, could you

* expand this PR
* re-target it to `versions/3.0.2.md` on the `v3.0.2-dev` branch as per the [development guidelines](https://github.com/OAI/OpenAPI-Specification/blob/master/DEVELOPMENT.md)

philsturgeon:
Ah you're right that's already covered. Fair enough.

PR #1522, opened 2018-04-03
Removing type confusion

philsturgeon:
The "Common Name" field was confusing people

So many people were considering dateTime, etc to be an OpenAPI type that tool vendors were implementing functionality to normalize this.

https://github.com/mikunn/openapi-schema-to-json-schema/issues/4

I too in the past have thought I was meant to use dateTime in the format, not date-time. Let's just get rid of this column, it serves no real purpose as far as I can tell.

### DIFF
--- a/versions/3.0.2.md
+++ b/versions/3.0.2.md
@@ -151,19 +151,22 @@ Types that are not accompanied by a `format` property follow the type definition
 
 The formats defined by the OAS are:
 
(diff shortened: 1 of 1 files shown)


MikeRalphson:
@philsturgeon could you target this PR against the `versions/3.0.2.md` file on this branch?

philsturgeon:
@MikeRalphson done! 😅

PR #1524, opened 2018-04-04
Adding marshmallow-code/apispec

earth2marsh:
also cleaning up some inconsistent capitalization and spacing

### DIFF
--- a/IMPLEMENTATIONS.md
+++ b/IMPLEMENTATIONS.md
@@ -2,7 +2,7 @@
 
 Below is a list of known tooling that implements the 3.0.0 specification. While support for the 3.0.0 specification matures,  refer to the details of projects listed below for any notes about stability and roadmap.  The process to improve the 3.x specification includes feedback from end-users and tooling creators. We strongly encourage draft tooling be made available for early users of OAS drafts.
 
-These tools are not necessarily endorsed by the OAI.
+These tools are not endorsed by the OAI.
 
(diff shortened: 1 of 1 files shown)


MikeRalphson:
Apart from the repository description, the README and docs for marshmallowcode/api-spec only refer to OAS 2. See also https://github.com/marshmallow-code/apispec/issues/165 - possibly v3 support is WIP?

Also, if we want people to adhere to our branding guidelines, we should pay it back. GitHub styles themselves 'GitHub' not 'github'. I am not prepared to die on this hill, however.

earth2marsh:
@MikeRalphson thanks, I'd been told it supported v3, but it looks like that's just partial support today. At what point would we deem it fully supported though?

As for the change, my goal was 1) be consistent, and 2) those GitHub references are strikingly path-like, so I decided to write them as GitHub does in paths. I did not force everything to lower case though because orgs and projects can have caps. When GitHub is used to refer to the service and not the resource, I would always put two caps. :)

I like removing `necessarily`, will update. I'll also drop apispec for now until either they think they've hit something that qualifies or we have criteria about what "supported" means. Sound good?

MikeRalphson:
Hard to say, but v2 and v3 should probably at least have parity in the readme.

Thanks for the explanation re: github, makes sense.

Just remembered, we should be getting into the habit of adding a sign-off to our commits as per Jeff-EF.

PR #1531, opened 2018-04-09
Draft features explanation

darrelmiller:
Update to development guidelines to describe the process of creating `draft features`.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -64,6 +64,7 @@ Draft features will be documented as GitHub issues and labeled with the `draft-f
 If during the development of a draft feature, it is determined that the feature needs to change in a way that may break existing draft implementations, the extension name itself may be versioned with a version suffix. e.g. `-v2`
 When a draft feature becomes part of a future update to the specification any version suffix will be removed.
 Draft features that are deemed not appropriate for inclusion MUST be marked with the `abandoned` label.
+Draft-features that are considered suitably specified for implementation will be marked with the `draft-ready` label.
 
 Not all future new features will be introduced in this way.
 Some new features impact the specification in ways that cannot be encapsulated in an extension.
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -47,6 +47,24 @@ Spec changes should be approved by a majority of the committers.  Approval can b
 
 No change should be approved until there is documentation for it, supplied in an accompanying PR.
 
+## Draft Features
+
+Where suitable, features will be introduced as draft but OAI approved extensions.
+By introducing new features this way we enable new features to be designed, documented and then implemented by tools that are interested in the feature, without putting the burden of implementation on all tooling.
+If the feature is successfully implemented and there is demonstrable value added by the feature, it will become a candidate for inclusion in a future release of the specification, at which point all tools will be expected to support the feature.
+
+Draft feature extensions are identified by the `x-oas-draft-` prefix and can only be used where existing extensions are permitted.
+This ensures no exising tooling will affected by the introduction of the draft feature.
+If the feature is deemed appropriate for inclusion in the OAS, the `x-oas-draft-` prefix will be removed.
+Tooling that supports draft features should plan for the future removal of the prefix and accomodate the transition period where descriptions exist with and without the prefix. 
+
+Draft features will be documented as Github issues and labeled with the `draft-feature` label.
+If during the development of a draft feature, it is determined that the feature needs to change in a way that may break existing draft implementations, the extension name itself may be versioned with a version suffix. e.g. `-v2`
+
+Not all future new features will be introduced in this way.
+Some new features impact the specification in ways that cannot be encapsulated in an extension.
+However, where a new feature can be introduced in this way, it should be.
+
 ## Transparency
 
 We should always be as transparent as possible.  Sometimes there will be discussions that use customer names, sensitive use cases, and so on.  These must be anonymized, discussed in a private repository, or conducted offline.
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -54,12 +54,16 @@ By introducing new features this way we enable new features to be designed, docu
 If the feature is successfully implemented and there is demonstrable value added by the feature, it will become a candidate for inclusion in a future release of the specification, at which point all tools will be expected to support the feature.
 
 Draft feature extensions are identified by the `x-oas-draft-` prefix and can only be used where existing extensions are permitted.
-This ensures no exising tooling will affected by the introduction of the draft feature.
+This ensures no existing tooling will affected by the introduction of the draft feature.
 If the feature is deemed appropriate for inclusion in the OAS, the `x-oas-draft-` prefix will be removed.
-Tooling that supports draft features should plan for the future removal of the prefix and accomodate the transition period where descriptions exist with and without the prefix. 
+Tooling that supports draft features should plan for the future removal of the prefix.
+When tooling adds support for a later version of OAS that includes the final implementation of the feature, it MUST not support the use of the draft prefix for that feature.
+Draft features will only be promoted into minor or major releases of the specification and therefore will be transparent to OpenAPI description writers and tooling providers who choose not to use the feature while in its draft state.
 
-Draft features will be documented as Github issues and labeled with the `draft-feature` label.
+Draft features will be documented as GitHub issues and labeled with the `draft-feature` label.
 If during the development of a draft feature, it is determined that the feature needs to change in a way that may break existing draft implementations, the extension name itself may be versioned with a version suffix. e.g. `-v2`
+When a draft feature becomes part of a future update to the specification any version suffix will be removed.
+Draft features that are deemed not appropriate for inclusion MUST be marked with the `abandoned` label.
 
 Not all future new features will be introduced in this way.
 Some new features impact the specification in ways that cannot be encapsulated in an extension.
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -60,11 +60,11 @@ Tooling that supports draft features should plan for the future removal of the p
 When tooling adds support for a later version of OAS that includes the final implementation of the feature, it MUST not support the use of the draft prefix for that feature.
 Draft features will only be promoted into minor or major releases of the specification and therefore will be transparent to OpenAPI description writers and tooling providers who choose not to use the feature while in its draft state.
 
-Draft features will be documented as GitHub issues and labeled with the `draft-feature` label.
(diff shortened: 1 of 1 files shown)


DEVELOPMENT.md
```diff
@@ -47,6 +47,24 @@ Spec changes should be approved by a majority of the committers.  Approval can b
...
 No change should be approved until there is documentation for it, supplied in an accompanying PR.
 
+## Draft Features
+
+Where suitable, features will be introduced as draft but OAI approved extensions.
+By introducing new features this way we enable new features to be designed, documented and then implemented by tools that are interested in the feature, without putting the burden of implementation on all tooling.
+If the feature is successfully implemented and there is demonstrable value added by the feature, it will become a candidate for inclusion in a future release of the specification, at which point all tools will be expected to support the feature.
+
+Draft feature extensions are identified by the `x-oas-draft-` prefix and can only be used where existing extensions are permitted.
+This ensures no exising tooling will affected by the introduction of the draft feature.
```
MikeRalphson: "existing"

DEVELOPMENT.md
```diff
@@ -47,6 +47,24 @@ Spec changes should be approved by a majority of the committers.  Approval can b
...
+If the feature is deemed appropriate for inclusion in the OAS, the `x-oas-draft-` prefix will be removed.
+Tooling that supports draft features should plan for the future removal of the prefix and accomodate the transition period where descriptions exist with and without the prefix. 
+
+Draft features will be documented as Github issues and labeled with the `draft-feature` label.
```
earth2marsh: `GitHub` should have a capital `H`

DEVELOPMENT.md
```diff
@@ -47,6 +47,24 @@ Spec changes should be approved by a majority of the committers.  Approval can b
...
+If during the development of a draft feature, it is determined that the feature needs to change in a way that may break existing draft implementations, the extension name itself may be versioned with a version suffix. e.g. `-v2`
```
darrelmiller: Add note about removing suffix when feature becomes included in the spec.

earth2marsh:
Suggest declaring the purpose of this change along the lines of, "This approach provides an official process by which code and spec can coevolve."

DEVELOPMENT.md
darrelmiller: When a draft feature is deemed not appropriate for inclusion, we should mark the documentation as "obsolete" or "obsoleted by".

DEVELOPMENT.md
MikeRalphson: Is it worth calling out that draft extensions can only be promoted into minor or major (breaking) versions of the spec, not patches? That may be deemed obvious from the rest of the document though.

earth2marsh:
In discussion, we talked about the fact that `x-oas-draft-…` shouldn't be considered in schemas used to validate specs. However, once the feature graduates from draft, should the schema consider draft extensions as invalid OpenAPI?

@usarid also pointed out that overlays may be a way to extend without (or possibly in concert with?) using the spec extension approach.

handrews:
@earth2marsh I'd say that old `x-oas-draft-` keywords are still valid, but as with any other `x-` keyword, support is not reliable.  Rather than forbidding them, perhaps implementations SHOULD warn that the feature has graduated?

DEVELOPMENT.md
```diff
@@ -47,6 +47,29 @@ Spec changes should be approved by a majority of the committers.  Approval can b
...
+Tooling that supports draft features should plan for the future removal of the prefix.
+When tooling adds support for a later version of OAS that includes the final implementation of the feature, it MUST not support the use of the draft prefix for that feature.
+Draft features will only be promoted into minor or major releases of the specification and therefore will be transparent to OpenAPI description writers and tooling providers who choose not to use the feature while in its draft state.
+
+Draft features will be documented as GitHub issues and labeled with the `draft-feature` label.
+If during the development of a draft feature, it is determined that the feature needs to change in a way that may break existing draft implementations, the extension name itself may be versioned with a version suffix. e.g. `-v2`
+When a draft feature becomes part of a future update to the specification any version suffix will be removed.
+Draft features that are deemed not appropriate for inclusion MUST be marked with the `abandoned` label.
+Draft-features that are considered suitably specified for implementation will be marked with the `draft-ready` label.
```
darrelmiller: Alternative labels: "ready", "ready-to-pilot", "ready-for-trial".
"draft:feature", "draft:abandoned", "draft:ready"

earth2marsh: @webron and I caught up on this and have agreed on:
`draft:proposal` - we’re still hashing out the details
`draft:pilot` - go ahead and implement it
`draft:graduated` - we’ve seen the implementations, it will go into the next minor/major release
`draft:abandoned` - didn't succeed, do not use

darrelmiller: Latest status suggestions by @earth2marsh approved by @OAI/tsc

DEVELOPMENT.md
```diff
@@ -47,6 +47,29 @@ Spec changes should be approved by a majority of the committers.  Approval can b
 
 No change should be approved until there is documentation for it, supplied in an accompanying PR.
 
+## Draft Features
+
+Where suitable, features will be introduced as draft but OAI approved extensions.
```
earth2marsh: `but` here is throwing me off... how about `as OAI-approved, draft features`?

DEVELOPMENT.md
```diff
@@ -47,6 +47,29 @@ Spec changes should be approved by a majority of the committers.  Approval can b
...
+
+Draft features will be documented as GitHub issues and labeled with the `draft-feature` label and will be initially labelled as `draft:proposal`. When the proposal is considered sufficiently stable for pilot implementation, it will be labeled `draft:pilot`.
+If during the development of a draft feature, it is determined that the feature needs to change in a way that may break existing draft implementations, the extension name itself may be versioned with a version suffix. e.g. `-v2`
+When a draft feature becomes part of a future update to the specification any version suffix will be removed.
+Draft features that are deemed not appropriate for inclusion MUST be marked with the `draft:abandoned` label.
+Draft-features that are considered suitably specified and have had successful pilot implementations will be marked with the `draft:graduated` label.
+
+Not all future new features will be introduced in this way.
+Some new features impact the specification in ways that cannot be encapsulated in an extension.
+However, where a new feature can be introduced in this way, it should be.
```
earth2marsh: `SHOULD`? :)

DEVELOPMENT.md
```diff
@@ -47,6 +47,29 @@ Spec changes should be approved by a majority of the committers.  Approval can b
 
 No change should be approved until there is documentation for it, supplied in an accompanying PR.
 
+## Draft Features
+
+Where suitable, features will be introduced as draft but OAI approved extensions.
+By introducing new features this way we enable new features to be designed, documented and then implemented by tools that are interested in the feature, without putting the burden of implementation on all tooling.
+If the feature is successfully implemented and there is demonstrable value added by the feature, it will become a candidate for inclusion in a future release of the specification, at which point all tools will be expected to support the feature.
```
earth2marsh: `it will become a candidate` seems like it's a candidate for inclusion in a future release even in draft form. Perhaps this is more like, `it becomes eligible to be included in a future release`?

DEVELOPMENT.md
```diff
@@ -47,6 +47,29 @@ Spec changes should be approved by a majority of the committers.  Approval can b
...
+
+Draft feature extensions are identified by the `x-oas-draft-` prefix and can only be used where existing extensions are permitted.
+This ensures no existing tooling will affected by the introduction of the draft feature.
+If the feature is deemed appropriate for inclusion in the OAS, the `x-oas-draft-` prefix will be removed.
+Tooling that supports draft features should plan for the future removal of the prefix.
+When tooling adds support for a later version of OAS that includes the final implementation of the feature, it MUST not support the use of the draft prefix for that feature.
+Draft features will only be promoted into minor or major releases of the specification and therefore will be transparent to OpenAPI description writers and tooling providers who choose not to use the feature while in its draft state.
+
+Draft features will be documented as GitHub issues and labeled with the `draft-feature` label and will be initially labelled as `draft:proposal`. When the proposal is considered sufficiently stable for pilot implementation, it will be labeled `draft:pilot`.
```
earth2marsh: Is this saying there will be at least two labels, both a `draft-feature` and a `draft:*`? Should it be spelled out that older stage labels are removed?

DEVELOPMENT.md
```diff
@@ -47,6 +47,29 @@ Spec changes should be approved by a majority of the committers.  Approval can b
...
+If during the development of a draft feature, it is determined that the feature needs to change in a way that may break existing draft implementations, the extension name itself may be versioned with a version suffix. e.g. `-v2`
+When a draft feature becomes part of a future update to the specification any version suffix will be removed.
+Draft features that are deemed not appropriate for inclusion MUST be marked with the `draft:abandoned` label.
+Draft-features that are considered suitably specified and have had successful pilot implementations will be marked with the `draft:graduated` label.
```
earth2marsh: Suggest removing `-` from `Draft features` to be consistent with prev line.

DEVELOPMENT.md
```diff
@@ -47,6 +47,29 @@ Spec changes should be approved by a majority of the committers.  Approval can b
...
+## Draft Features
+
+Where suitable, features will be introduced as draft but OAI approved extensions.
+By introducing new features this way we enable new features to be designed, documented and then implemented by tools that are interested in the feature, without putting the burden of implementation on all tooling.
+If the feature is successfully implemented and there is demonstrable value added by the feature, it will become a candidate for inclusion in a future release of the specification, at which point all tools will be expected to support the feature.
+
+Draft feature extensions are identified by the `x-oas-draft-` prefix and can only be used where existing extensions are permitted.
+This ensures no existing tooling will affected by the introduction of the draft feature.
+If the feature is deemed appropriate for inclusion in the OAS, the `x-oas-draft-` prefix will be removed.
+Tooling that supports draft features should plan for the future removal of the prefix.
```
tedepstein: So the final property name will usually (or always?) be the same as the draft property name, minus the `x-oas-draft-` prefix, correct? If tool developers can rely on this, at least as the usual/expected case, it makes it somewhat easier to manage the lifecycle.

DEVELOPMENT.md
```diff
@@ -47,6 +47,29 @@ Spec changes should be approved by a majority of the committers.  Approval can b
...
+When tooling adds support for a later version of OAS that includes the final implementation of the feature, it MUST not support the use of the draft prefix for that feature.
```
tedepstein: @handrews suggested relaxing this:

> I'd say that old x-oas-draft- keywords are still valid, but as with any other x- keyword, support is not reliable. Rather than forbidding them, perhaps implementations SHOULD warn that the feature has graduated?

Was this discussed and rejected for some reason? If not, I think it's a more practical policy, and we should consider adopting it. It will make it easier to migrate OAS documents to new spec versions.

PR #1540, opened 2018-04-12
Fixes #1539. Use the most common X-RateLimit-* instead of X-Rate-Limit-*

### LINKED ISSUE #1539: Use X-RateLimit instead of X-Rate-Limit
## I wish

  - openapi specs to use X-RateLimit instead of X-Rate-Limit in examples ;)

## Because

  - X-RateLimit seems most common in api gateways (kong, wso2) and in services (github, api.data.gov, http://zalando.github.io/restful-api-guidelines/index.html#132)

  - see http://william.holroyd.name/2014/11/02/how-do-most-apis-handle-rate-limiting/ for a comparison

ioggstream:
## This PR

  - just replace X-Rate-Limit-* with X-RateLimit

## Because

  - the X-RateLimit is more common (see #1539) 
  - this avoid confusion when copy-paste part of the specs to create new services.

### DIFF
--- a/versions/3.0.2.md
+++ b/versions/3.0.2.md
@@ -1619,7 +1619,7 @@ requestBody:
           # only accept png/jpeg
           contentType: image/png, image/jpeg
           headers:
-            X-Rate-Limit-Limit:
+            X-RateLimit-Limit:
               description: The number of allowed requests in the current period
               schema:
                 type: integer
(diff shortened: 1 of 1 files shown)


webron:
The @OAI/tsc has performed a search and it looks like you can easily find examples of both forms exist out there in different public popular APIs. Since there's no standard, we don't feel there's a compelling reason to make the change. Thanks for taking the time to bring it up though.

PR #1541, opened 2018-04-12
edit OpenAPI Specification license info so that GitHub recognizes it

eirinikos:
Hello! 👋

(CCing @dankohn who has requested this work so that the license will appear correctly in https://landscape.cncf.io/selected=open-api)

GitHub uses a library called Licensee to identify a project's license
type. It shows this information in the status bar and via the API if it
can unambiguously identify the license.

This commit creates a LICENSE file that stores the full text of the
Apache license, referencing The Linux Foundation as the copyright
holder. It also updates package.json with the correct SPDX license ID
and updates the README "License" header to "Licensing".

Collectively, these changes allow Licensee to successfully identify the
license type of the OpenAPI Specification project as Apache.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,11 +1,201 @@
-Copyright 2017 The Linux Foundation
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
 
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


webron:
Thanks for the PR. We're planning on making this exact change for that exact reason - but we're waiting for a clarification from the legal team on whether we can do it. Granted, it should be simple, but none of us in the team is a lawyer...

eirinikos:
Thanks @webron & @swinslow! Sounds good. I'll be on standby in case you need me to make any changes to this PR.

webron:
Approving and merging after discussing this on the TSC call.

PR #1545, opened 2018-04-16
Fix callback key expression example.

jwalton:


darrelmiller:
@jwalton Thanks for catching this. We would love to merge this, but it needs to be against our v3.0.2-dev branch and against the 3.0.2.md file. Would you create a PR against that branch for us? Thanks from the @OAI/tsc

jwalton:
For sure.

Issue #715, opened 2016-06-13
State whether Open API schema object definitions are open or closed for extension


dehora:
I had a look through the open issues and couldn't find a specific issue for this, apologies if I missed it. It's alluded to in the following but not directly addressed
- https://github.com/OAI/OpenAPI-Specification/issues/333
- https://github.com/OAI/OpenAPI-Specification/issues/668
- https://github.com/swagger-api/swagger-codegen/issues/1318

The clarification I'm looking for is - are object definitions open or closed for extension, and specifically what is the default for handling unexpected JSON properties in the _absence_ of an `additionalProperties` declaration?

If Open API was using vanilla JSON Schema, I would assume the absence of `additionalProperties` means the instances are open for extension, since `additionalProperties` has not been explicitly set to `false` to close the schema. But because Open API redefines parts of JSON Schema, and in particular `additionalProperties` and how it should handle booleans, I'm reluctant to just assume that behaviour is carried forward into Open API. 

At the moment I can't find a definitive statement in existing specifications for this, in particular [http://swagger.io/specification/#schemaObject](http://swagger.io/specification/#schemaObject) (again apologies if I missed it).  I do think it's something that needs to be explicit as it relates to how clients based on Open API handle unexpected data and how Open API tooling should behave by default, especially given `additionalProperties` has been redefined.

darrelmiller:
You might find #568 a relevant discussion. Although there is no definitive answer, I do believe that the general consensus is that OpenAPI is not a closed contract.  I think the only concern is regarding how useful is it to have behavior that is unspecified in the contract.

Relating this back to JSON Schema definitions, it is my understanding that the default behavior of being open for extension is the expected behavior with OpenAPI.  However, I think the spec needs to be much clearer about how it redefines the JSON Schema properties that it claims it does.

webron:
Tackling PR: #741

dehora:
> You might find #568 a relevant discussion.

I did find it very useful in general. This ticket is more specific. 

> Tackling PR: #741

Similar to prior tickets, I see this alluded to but not directly addressed. There's a holding header (`"Clarify additionalProperties default setting'`) but with no text and a [comment](https://github.com/OAI/OpenAPI-Specification/pull/741#issuecomment-236883682) from last August re allOf interacting with additionalProperties' true/false values which misses the key point of what should happen when the additionalProperties field is absent in OAI, and the fact that true/false are not options in OAI. 

[...]

Issue #668, opened 2016-04-22
Make it clear that `additionalProperties` can only have a schema value


ePaul:
JSON Schema allows for `additionalProperties` both a boolean or an object value. `true` is interpreted as "additional properties follow no restrictions", `false` means "no additional restrictions", and an object is interpreted as a JSON schema applied to the property values (the empty object is thus equivalent to `true`).

The [Schema object](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#schemaObject) description says:

> The following properties are taken from the JSON Schema definition but their definitions
> were adjusted to the OpenAPI Specification. Their definition is the same as the one from
> JSON Schema, only where the original definition references the JSON Schema definition,
> the Schema Object definition is used instead.
> - items
> - allOf
> - properties
> - additionalProperties

This would be naively interpreted as `additionalProperties` can have a boolean or a schema value (with a schema being interpreted as an OpenAPI schema, not a JSON schema).

In [Swagger-Codegen #1318](https://github.com/swagger-api/swagger-codegen/issues/1318#issuecomment-146302879), @webron commented that this was actually meant as "the value of additionalProperties can only be a (Swagger) Schema object", not a boolean.

Some more discussion was in #477 here.

(I just hit an API which used booleans in several places, and fails with swagger-codegen.)

To make this clearer in the next version of the spec, I suggest to add the following sentence to the cited paragraph above (before the list):

> Alternative values of other types (boolean) are not allowed.

I'm not totally happy with this wording, feel free to propose a better one. Maybe we should instead simply redefine `additionalProperties` generally?

dehora:
> The following properties are taken from the JSON Schema definition but their definitions
were adjusted to the OpenAPI Specification. Their definition is the same as the one from
JSON Schema, only where the original definition references the JSON Schema definition,
the Schema Object definition is used instead.
 - items
 - allOf
 - properties
 - additionalProperties

This is confusing spec language. In particular the last sentence is too hard to follow - the definitions can't be the same and not be the same. Also there don't seem to be any local definitions defined - the link in the statement "the Schema Object definition is used instead." points back to this section. 

What seems to be happening here operationally, is that four properties from JSON Schema are being changed in Open API while keeping the property names. 

This seems to indicate what's going on, in more direct language

> The following property names are taken from JSON Schema specification but their definitions 
are changed in Open API to disallow boolean values - only an object value representing a JSON schema applied to the property values is allowed. 
[...]

PR #1548, opened 2018-04-17
State explicitly: `additionalProperties : true` is the default

tedepstein:
Address #715 - State (explicitly) whether Open API schema object definitions are open or closed for extension.

This was the topic of many discussions around OpenAPI 2.0 and earlier Swagger specifications.  Here are some of them:
* Make it clear that `additionalProperties` can only have a schema value #668
* JSON Schema Support Proposal #741
* State whether Open API schema object definitions are open or closed for extension #715
* NullPointerException in parser with additionalProperties: false [(swagger-api/swagger-core#1437)](https://github.com/swagger-api/swagger-core/issues/1437)

OpenAPI 2.0 assumed `additionalProperties: false` as the default, and disallowed explicit use of a boolean `additionalProperties` value. This was never fully documented, only stated in out-of-band comments. And those comments were never really reconciled with `allOf` semantics, which magically continued to work as though `additionalProperties: true` were the default. 

I wasn't aware, and I'm guessing that some others also weren't aware, that this changed intentionally in the 3.0 spec. The current 3.0.1 spec explicitly states that boolean values are allowed, but doesn't explicitly state that the _default_ is now `additionalProperties: true`. 

I think it's worth making this explicit now, because anyone who got caught up in the 2.0 `additionalProperties` confusion is likely to need some assurance that this change is intentional, and not subject to informal amendments or clarifications made outside of the spec. 

@handrews and I were both looking for that assurance, and couldn't find it. So I'm suggesting this small change.

### DIFF
--- a/versions/3.0.2.md
+++ b/versions/3.0.2.md
@@ -2301,7 +2301,7 @@ The following properties are taken from the JSON Schema definition but their def
 - not - Inline or referenced schema MUST be of a [Schema Object](#schemaObject) and not a standard JSON Schema.
 - items - Value MUST be an object and not an array. Inline or referenced schema MUST be of a [Schema Object](#schemaObject) and not a standard JSON Schema. `items` MUST be present if the `type` is `array`.
 - properties - Property definitions MUST be a [Schema Object](#schemaObject) and not a standard JSON Schema (inline or referenced).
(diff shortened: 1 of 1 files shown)


tedepstein:
Thanks, @handrews . I made that change.

PR #1559, opened 2018-04-23
Fix callback key expression example.

jwalton:


jwalton:
This is https://github.com/OAI/OpenAPI-Specification/pull/1545, repackaged on the correct branch.  :)

PR #1571, opened 2018-05-04
fix wording on who supplies the default value for a server variable

notEthan:
the server variable default is supplied in the OpenAPI document, not by the consumer.

not sure the best wording to fix it, but the current wording is incorrect, I believe.

### DIFF
--- a/versions/3.0.2.md
+++ b/versions/3.0.2.md
@@ -429,7 +429,7 @@ An object representing a Server Variable for server URL template substitution.
 Field Name | Type | Description
 ---|:---:|---
 <a name="serverVariableEnum"></a>enum | [`string`] | An enumeration of string values to be used if the substitution options are from a limited set.
-<a name="serverVariableDefault"></a>default | `string` |  **REQUIRED**. The default value to use for substitution, and to send, if an alternate value is _not_ supplied. Unlike the [Schema Object's](#schemaObject) `default`, this value MUST be provided by the consumer.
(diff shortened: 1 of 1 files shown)


webron:
It's correct, but perhaps incoherent. The idea is that the consumer of the document must use the `default` value for a server variable, unlike the `default` value of a parameter, for example.

notEthan:
> the consumer of the document must use the `default` value 

sure, the consumer must _use_ it, as you say. but that's completely different than _providing_ the value - that is not done by the consumer but by the document.

I think we're arguing semantics at that point though, and since you describe it as perhaps incoherent, maybe can agree that it should be changed? and if so, what would be a better wording?

darrelmiller:
Text will be reviewed to try and add some more clarity.  @OAI/tsc doesn't feel the current PR accurately conveys the intent.

earth2marsh:
Suggest something like:
```The default value to use for substitution, which will be sent if an alternate value is not supplied by the consumer. Unlike the Schema Object's default, a value MUST be provided by a consumer.```

earth2marsh:
We discussed the language proposed in my last comment on the TSC call today and had verbal approvals for the edited version above. @notEthan would you like to incorporated and re-submit?

webron:
@notEthan I went ahead and modified the PR with the proposed changes as we want to push out a new version. We can revisit it if still needed in 3.0.3.

notEthan:
thanks @webron, looks good to me

Issue #1576, opened 2018-05-08
Support for Streams in Responses


emmanuelproulx:
APIs that download binary data currently must be done by `type: string, format: binary`. This translates to byte arrays (in Java for example, anyway that's what swagger-ui and swagger-codegen do). But _this consumes large amounts of memory_ if the data is very big, and easily gives _out-of-memory errors_.

Typically huge blocks of data are sent via _streams_ instead of byte arrays in most remote APIs. This allows for loading only a block of data at a time in memory.

The workaround used to be the '`file`' format but this was removed in version 3.

What I propose is a **new `type: stream`** (`format: binary`) which would allow this. As an example, this could map to Spring's `ResponseEntity(inputStreamResource,...) `on the server side, and a okhttp's `ResponseBody.byteStream() `on the client side. Lots of HTTP client & server frameworks support streams in API implementations so I'm surprised it doesn't already exist.

handrews:
@emmanuelproulx `type` is from JSON Schema and only refers to the data type, not how it is produced or consumed (because JSON Schema is not aware of those things, although Hyper-Schema is).

If you're talking about HTTP chunked encoding then this should be done through documenting the use of the `Transfer-Encoding` header, which I believe is already possible.

emmanuelproulx:
Right. I understand. But there are times when you don't want to respond with JSON. As an example there's this standard REST API to retrieve images, IIIF ( http://iiif.io/ ) and it returns images not just JSON. I'd like to come up with a Swagger description of it but I don't know what to put in the 200 response.

responses:
    '200':
      content:
        'image/png':
           ???

Sure I could put a schema with type: string + format: binary there but in the end it'll keep the entire image in memory which for sure will blow up during performance testing. Anyway that's not how it's done normally in any API framework; they all use streams.

There should be a way (not sure what the best way is) to specify that the returned entity is not JSON and should be treated as a stream. In OpenAPI 2 there was the "file" type that was used for that purpose but it's gone now.

Would you have a good idea by any chance?

handrews:
If your content is `image/png` then it's not JSON.  Looking at the [Media Type Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#mediaTypeObject) the `schema` field is not required.  So I'm not sure you even need one (although I'd defer to the TSC members on this- my expertise is mostly JSON Schema/Hyper-Schema).

[...]

PR #1579, opened 2018-05-12
Explicitly state that Media Type Object schema field applies to responses as well as requests

jwalton:
Super minor nit-picky sort of change.  :)

### DIFF
--- a/versions/3.0.2.md
+++ b/versions/3.0.2.md
@@ -1363,7 +1363,7 @@ Each Media Type Object provides schema and examples for the media type identifie
 ##### Fixed Fields
 Field Name | Type | Description
 ---|:---:|---
-<a name="mediaTypeSchema"></a>schema | [Schema Object](#schemaObject) \| [Reference Object](#referenceObject) | The schema defining the type used for the request body.
+<a name="mediaTypeSchema"></a>schema | [Schema Object](#schemaObject) \| [Reference Object](#referenceObject) | The schema defining the type used for the request body (or response body, if this is part of a response).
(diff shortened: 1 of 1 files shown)


versions/3.0.2.md
```diff
@@ -1363,7 +1363,7 @@ Each Media Type Object provides schema and examples for the media type identifie
 ##### Fixed Fields
 Field Name | Type | Description
 ---|:---:|---
-<a name="mediaTypeSchema"></a>schema | [Schema Object](#schemaObject) \| [Reference Object](#referenceObject) | The schema defining the type used for the request body.
+<a name="mediaTypeSchema"></a>schema | [Schema Object](#schemaObject) \| [Reference Object](#referenceObject) | The schema defining the type used for the request body (or response body, if this is part of a response).
```
tedepstein: Note that Media Type Object is also used in Parameter Object, in cases where the parameter is specified by the `content` map, as opposed to `schema` and `style`.  The `content` property is a Map[`string`, Media Type Object].

Maybe it would be simpler to say:

 "The schema defining the content of the request, response, or parameter."

jwalton: Good catch, sir.  :)

jwalton: I like that wording a lot better, actually, because the old wording:

> The schema defining the type used

Implies that the schema uniquely identifies the "type" of a request, response, or parameter, but since a schema can validate multiple types, it doesn't really.

handrews:
I think #1576 identifies some confusion in this area as well.  It should perhaps be a separate PR but as I'm confused on the topic myself I thought I'd use this to draw more attention to it.

TL;DR: To me, the schema of the media type object should have nothing to do with whether the media is streamed or not, but apparently that's a thing?

PR #1588, opened 2018-05-22
Add Exegesis to IMPLEMENTATIONS.md

jwalton:
Exegesis is an OpenAPI 3.0.0 server-side implementation on node.js.

It seems like all I ever do is come here and raise issues and complain about things.  But I know exactly how much work goes into a spec like this, so let me take a moment to thank the whole @OAI/tsc for all the hard work they do.  :)


darrelmiller:
@jwalton Having people review our work, point out mistakes, help to clarify confusing words, contribute content, is our lifeblood.  Having folks turn the specs into implementations is what makes the effort all worthwhile.  Don't ever feel bad for the constructive criticism you provide.

jwalton:
> Don't ever feel bad for the constructive criticism you provide.

Oh, I won't.  :P  I just know, when you're running an open source project of any sort, you get a lot of issues pointing out problems, and you never get enough thanks.  :)

PR #1597, opened 2018-06-06
Translate example

felzan:
Translated two words from portuguese to english in two examples.


darrelmiller:
Thanks for the fix, would it be possible to resubmit the PR against the v3.0.2-dev branch.

PR #1598, opened 2018-06-06
Update 3.0.0.md

dpopp07:
I believe the use of the key `representations` here was a typo, and should be `content`. If I am mistaken, please disregard!

### DIFF
--- a/versions/3.0.0.md
+++ b/versions/3.0.0.md
@@ -1737,7 +1737,7 @@ Response with a string type:
 
 ```yaml
 description: A simple string response
-representations:
+content:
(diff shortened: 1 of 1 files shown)


MikeRalphson:
You're not wrong. This oversight was however fixed in https://github.com/OAI/OpenAPI-Specification/commit/d45549e004a54a72d11a9ec9aa676bd2d0748d30#diff-3c215413cf312161e312fcb4327fda00 which was included in v3.0.1 of the specification. (Each version of the specification is fixed once published.)

Thanks again.

Issue #1564, opened 2018-04-25
HTML rendering of the spec (bikeshed/ReSpec)


MikeRalphson:
There has been some discussion, on [twitter](https://twitter.com/philsturgeon/status/933056018011054080) and within the TSC about providing a HTML-rendered version of the specification (in addition to, and **not** replacing the markdown source-of-truth). This would provide an easier-to-use ToC, referenceable section numbers and a bibliography.

This also builds on the earlier discussion in #829 - cc @apowers313 @adrianhopebailie & @philsturgeon

Two examples have been generated, using [Bikeshed](https://github.com/tabatkins/bikeshed) and [ReSpec](https://github.com/w3c/respec).

The following pros/cons/differences are noted, but further feedback is greatly welcomed. Particular areas of interest are accessibility (s.508 and WCAG compliance) - please recommend good tools in this area, browser/mobile compatibility and ToC/section number rendering. Minor formatting tweaks can probably be accommodated, so these would not be a show-stopper, but please do call them out in comments.

I have also added a comment for each option, **purely** to gauge GitHub reactions. Please feel free to use :+1: :-1: :heart: etc. Please avoid one-line comments simply voting for either option - these may be deleted.

| Note | [Bikeshed version](http://mermade.github.io/static/oas_3.0.1.html) | [ReSpec version](http://mermade.github.io/static/oas_3.0.1r.html) |
|---|---|---|
| a11y | pending | pending |
| Size | 342k | 160k + 394k JS |
| w/o javascript | Renders fully | Renders without header/ToC |
| Anchors | Inline | ~~ToC only~~ Inline |
| Version number | Further down | Prominent |
| Logo | Further down | Prominent |
| Copyright | Present | Pending |
| Former Editors | Supported | ~~Not supported~~ Supported in later versions |
| Front matter format | Custom | JSON | 
| Conformance | Repeated | Once |
| GitHub Links | Repo / Issues | Repo / Issues / Commit history |
| Mobile preview | [Bikeshed](http://mobiletest.me/htc_one_emulator/?u=http://mermade.github.io/static/oas_3.0.1.html) | [ReSpec](http://mobiletest.me/htc_one_emulator/?u=http://mermade.github.io/static/oas_3.0.1r.html)
| Implementation | Python 2 | Node.js |

darrelmiller:
[...]

PR #1605, opened 2018-06-12
Remove colons from example headings

MikeRalphson:
Related to #1564 and ToC generation.


earth2marsh:
A colonoscopy, if you will?

PR #1610, opened 2018-06-22
Translate examples

felzan:
Translated two examples from portuguese to english


MikeRalphson:
Thanks for resubmitting. Perhaps mapping `canine` to `Dog` would be a clearer example?

PR #1668, opened 2018-08-17
Added oas-tools and generator to implementations

raffrearaUS:


MikeRalphson:
Thanks for the PR. We are now in the process of moving from `IMPLEMENTATIONS.md` to https://tools.openapis.org - the current implementations list will be removed and a note redirecting readers to the new site added. Instructions on listing your projects are contained in https://github.com/OAI/Tooling#how-can-you-help

PR #1670, opened 2018-08-18
Fix example response reference

darren987469:
`GET /pets/{petId}` returns a pet, not pets.


MikeRalphson:
Also cc @hkosova

Bakerburgh:
@earth2marsh and @MikeRalphson, I would appreciate this getting merged. Is there anything I can do to facilitate that?

PR #1682, opened 2018-09-05
Use semver for API version in example

dragetd:
The specification recommends using semver for version specifications here: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#versions

### DIFF
--- a/examples/v3.0/api-with-examples.yaml
+++ b/examples/v3.0/api-with-examples.yaml
@@ -1,7 +1,7 @@
 openapi: "3.0.0"
 info:
   title: Simple API overview
-  version: v2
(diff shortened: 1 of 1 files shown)


MikeRalphson:
@earth2marsh I'd be happy to see such an informative comment in at least one of the examples.

dragetd:
IMHO there is a spec, which should (and does) specify what is allowed, what not and what is recommended / best pratice. And example are just examples. It is nice if they follow those best practice.

I would also be okay with just reverting the PR than adding comment noise, but that is not my beer.

I like my bike-shed green. :)

PR #1686, opened 2018-09-13
Fix link to #securityDefinitionsObject

Gi60s:
Fix to version 2 documentation link.

Within the section [Security Requirement Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#security-requirement-object) there are two links to `Security Definitions`, the second of which has the wrong URL.

This PR has the fix for that link.

### DIFF
--- a/versions/2.0.md
+++ b/versions/2.0.md
@@ -2245,7 +2245,7 @@ The name used for each property MUST correspond to a security scheme declared in
 
 Field Pattern | Type | Description
 ---|:---:|---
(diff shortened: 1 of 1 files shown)


MikeRalphson:
Many thanks for the PR! Just to note that this fix was included (with others) in #991, but it was decided not to merge it at that time, as v2.0 is a released and set-in-stone specification, and so making changes to it now (even for confirmed errata) might be seen as confusing.

PR #1691, opened 2018-09-20
Re-target two commits from 3.0.1.md to 3.0.2.md

MikeRalphson:
Two commits (3e7d0a8c and 4c79d31f3) were correctly made on the `v3.0.2-dev` branch, but inadvertently against the wrong file (`versions/3.0.1.md` instead of `versions/3.0.2.md`).

This series of commits reverts the two original commits in reverse order, then applies the same diffs to the correct target file.

All metadata from the original commits has been preserved.

### DIFF
--- a/versions/3.0.1.md
+++ b/versions/3.0.1.md
@@ -118,7 +118,6 @@ For example, if a field has an array value, the JSON array representation will b
 }
 ```
 All field names in the specification are **case sensitive**.
-This includes all fields that are used as keys in a map, except where explicitly noted that keys are **case insensitive**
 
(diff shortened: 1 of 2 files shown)


MikeRalphson:
@webron with this PR in place `v3.0.2-dev` will merge into `master` ok (but with a conflict in `README.md` which needs resolving anyway). I tried cherry-picking 6943cc17e into `v3.0.2-dev` but it didn't actually improve the situation with regard to `README.md` needing manual resolution, so I haven't pushed that as a PR.

`.travis.yml` shows up as a modified file, but the relevant changes are already in `master`, so that becomes a nop and therefore does not need to hold up cutting the `v3.0.2` release.

webron:
@MikeRalphson Life saver.

PR #1695, opened 2018-09-26
sync `security` logic with swagger.io

abacaphiliac:
I noticed inconsistent `security` logic in the Swagger 2.0 Markdown (OR vs AND), and found (what I believe to be) the accurate description of `security` logic here: https://swagger.io/docs/specification/2-0/authentication/#multiple

This request synchronizes `security` logic, and copies the description and samples from swagger.io into the `Security Requirement Object` section.


MikeRalphson:
Firstly, thank you for the PR.

There are unfortunately a number of issues with it.

* The PR is against `v2.0` of the OAS which is a published version of the spec. and is thus unchanging
* As per the link you gave, the text/example from `swagger.io` is "Copyright © 2018 SmartBear Software" and cannot be added unless you have permission from the copyright holder to contribute it under this project's license
* The wording "there is a logical OR and AND between the security requirements" is, without reference to an example, very ambiguous

If you find the [wording in the current version of the spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#security-requirement-object) could be clarified further, a PR against the current development branch would be gratefully received.

Thanks again.

PR #1696, opened 2018-09-29
Implementations: add "kin-openapi"

jban332:


MikeRalphson:
Thank you very much for the PR!

Issue #1677, opened 2018-08-27
URL-encoding for operationRef


Alan-Cha:
I was under the impression that *operationRef* does not need URL-encoding. 

For example, in the following snippit taken directly from the [link object documentation](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#link-object)...

```
links:
  UserRepositories:
    # returns array of '#/components/schemas/repository'
    operationRef: '#/paths/~12.0~1repositories~1{username}/get'
    parameters:
      username: $response.body#/username
```

... `{` and `}` are not encoded, i.e. replaced with `%7B` and `%7D`, like in the following:

```
links:
  UserRepositories:
    # returns array of '#/components/schemas/repository'
    operationRef: '#/paths/~12.0~1repositories~1%7Busername%7D/get'
    parameters:
      username: $response.body#/username
```

***

However, after talking to a developer of an OAS validator, I am doubtful... I was told that because *operationRef* is defined as a **reference** in the OAS, it needs URL-encoding.  

Here is the [link](https://github.com/Mermade/oas-kit/issues/85) to the discussion. The last two exchanges are most important to this topic. 

Could you clarify if *operationRef* needs URL-encoding?

darrelmiller:
The operationRef is a fragment identifier and we need to investigate if curly braces require escaping in that portion of a URL.  Some of us think they don't but we will confirm.

Change initial description of OperationRef to "A relative or absolute URI reference to an OAS operation. "

It would be useful to add comments as to how this is different than a regular JSON Reference.

darrelmiller:
According to RFC 3986,

```
fragment    = *( pchar / "/" / "?" )
pchar         = unreserved / pct-encoded / sub-delims / ":" / "@"
unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"
sub-delims  = "!" / "$" / "&" / "'" / "(" / ")"/ "*" / "+" / "," / ";" / "="
```

This means the only way a curly brace can exist in a fragment is if it is percent-encoded.  That makes me sad.

MikeRalphson:
@darrelmiller Thanks for clarifying. Do we also need to update the examples?

jeffory-orrok:
[...]

PR #1713, opened 2018-10-18
Updated text for OperationRef

darrelmiller:
Issue #1677


Prestyn2019 (commented):
<a name="linkOperationRef"></a>operationRef | `string` | A relative or absolute reference to an OAS operation. This field is mutually exclusive of the `operationId` field, and MUST point to an [Operation Object](#operationObject). Relative `operationRef` values MAY be used to locate an existing [Operation Object](#operationObject) in the OpenAPI definit

Prestyn2019 (commented):
The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in BCP 14 RFC2119 RFC8174 when, and only when, they appear in all capitals, as shown here.

PR #1715, opened 2018-10-18
Note that specification extensions beginning x-oas- are reserved

MikeRalphson:


earth2marsh:
Technically makes sense to put this into `3.1.x`, but IMO we should document this in `3.0.2` that this will happen in `3.1.0`

Issue #1687, opened 2018-09-14
Unclear which edition of ECMA-262 is used as regex format


dvh:
While issue #880 links to http://www.ecma-international.org/publications/standards/Ecma-262.htm ([just as JSON Schema does](https://json-schema.org/understanding-json-schema/reference/regular_expressions.html), the OpenAPI Specification [deeplinks to the 5.1 edition](https://www.ecma-international.org/ecma-262/5.1/#sec-7.8.5) in https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#schema-object. Is this a typo or is edition 9.0 not supported (yet)?

MikeRalphson:
Edition 9.0 appears to have been released 11 months **after** OpenAPI 3.0.0, so it should not be a surprise that it is not supported. :slightly_smiling_face: 

It may be that a baseline version was chosen to achieve the broadest tooling compatibility :question:

dvh:
I don't think this is made very clear in the docs. Perhaps adding the specific version to the docs (not only in the deeplink) will help.

MikeRalphson:
A PR for discussion would be warmly welcomed.

handrews:
This got revisited and (assuming the PR is accepted) resolved in issue #1985 and PR #1987.

handrews:
@MikeRalphson perhaps close this in favor of #1985 since it has more details?

MikeRalphson:
Agreed.

PR #1725, opened 2018-10-24
Clarified supported Ecma edition for regex

dvh:
Added supported Ecma edition (5.1) for regular expressions in the link text and used the formal name: [Ecma-262 Edition 5.1](https://www.ecma-international.org/ecma-262/5.1/).

See also: https://github.com/OAI/OpenAPI-Specification/issues/1687

### DIFF
--- a/versions/3.0.3.md
+++ b/versions/3.0.3.md
@@ -2284,7 +2284,7 @@ The following properties are taken directly from the JSON Schema definition and 
 - exclusiveMinimum
 - maxLength
 - minLength
-- pattern (This string SHOULD be a valid regular expression, according to the [ECMA 262 regular expression](https://www.ecma-international.org/ecma-262/5.1/#sec-7.8.5) dialect)
+- pattern (This string SHOULD be a valid regular expression, according to the [Ecma-262 Edition 5.1 regular expression](https://www.ecma-international.org/ecma-262/5.1/#sec-7.8.5) dialect)
 - maxItems
 - minItems
 - uniqueItems


versions/3.0.3.md
```diff
@@ -2284,7 +2284,7 @@ The following properties are taken directly from the JSON Schema definition and
 - exclusiveMinimum
 - maxLength
 - minLength
-- pattern (This string SHOULD be a valid regular expression, according to the [ECMA 262 regular expression](https://www.ecma-international.org/ecma-262/5.1/#sec-7.8.5) dialect)
+- pattern (This string SHOULD be a valid regular expression, according to the [Ecma-262 Edition 5.1 regular expression](https://www.ecma-international.org/ecma-262/5.1/#sec-7.8.5) dialect)
```
earth2marsh: `Ecma-` should be `ECMA-`, no?

dvh: Hm, not sure... https://www.ecma-international.org/ecma-262/5.1/#sec-7.8.5 says:

> This is the HTML rendering of *Ecma-262 Edition 5.1,* The ECMAScript Language Specification.

It looks like ECMA is uppercase mostly when used in 'ECMAScript'.

versions/3.0.3.md
```diff
@@ -2299,7 +2299,7 @@ The following properties are taken directly from the JSON Schema definition and
 - exclusiveMinimum
 - maxLength
 - minLength
-- pattern (This string SHOULD be a valid regular expression, according to the [ECMA 262 regular expression](https://www.ecma-international.org/ecma-262/5.1/#sec-15.10.1) dialect)
+- pattern (This string SHOULD be a valid regular expression, according to the [Ecma-262 Edition 5.1 regular expression](https://www.ecma-international.org/ecma-262/5.1/#sec-15.10.1) dialect)
```
MikeRalphson: Conflict with link fragment change resolved here. Now LGTM.

versions/3.0.3.md
earth2marsh: Wow, what a mess of mixed caps! :)

You're right though, based on their usage in that doc, we can basically do what we want. My own personal sensibilities would argue for capitalization, and the header on that page does use capitalization: 
image

... but then does something else shortly thereafter. My point is definitely a nit, and this shouldn't hold up the acceptance, though if you don't mind using `ECMA` instead, please do, thx.

versions/3.0.3.md
earth2marsh: Thanks Mike. I discussed with @webron , too, and ready to merge.

Issue #1727, opened 2018-10-25
Duplication of tags in operation level, can be restricted as in API Level


r-sreesaran:
Duplication of tags in operation level, can be restricted as in API Level. 

Current Specification 
A list of tags used by the specification with additional metadata. The order of the tags can be used to reflect on their order by the parsing tools. Not all tags that are used by the Operation Object must be declared. The tags that are not declared MAY be organized randomly or based on the tools' logic. **Each tag name in the list MUST be unique.**

why is the uniqueness is not mandated for operation level.

MikeRalphson:
At first glance this appears to be an oversight. Would you like to create a PR to fix it for version 3.1.0 on the `v3.1.0-dev` branch?

r-sreesaran:
Sure I will create a PR to fix it in 3.1.0.

MikeRalphson:
TSC more discussion is required on the current (expected) behaviour before we take the change in #1729 further, which might break existing 3.0.x documents.

r-sreesaran:
@MikeRalphson I have adapted the PR as per suggestion of @earth2marsh.

earth2marsh:
This bubbled back up on today's TSC call. 

Thinking about this again, while I do think that listing the same tag twice may be considered an authoring error, I don't believe it is necessary to specific this behavior in the spec. I would agree that the duplicate tag has no meaning and could reasonably be either ignored or duplicated by any tooling according to the author's preference. I would not consider duplicate tags to cause an invalid spec, though I could imagine that a spec editor tool might helpfully flag the duplicate.

handrews:
Seems like we decided against this in 2020, no further comment so closing.

PR #1729, opened 2018-10-25
Operation Level Tags must be unique, Fix For Issue 1727

r-sreesaran:
Operation Level Tags must be unique, Fix For Issue 1727

### DIFF
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -841,7 +841,7 @@ Describes a single API operation on a path.
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="operationTags"></a>tags | [`string`] | A list of tags for API documentation control. Tags can be used for logical grouping of operations by resources or any other qualifier.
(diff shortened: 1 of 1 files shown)


earth2marsh:
Rather than being so strict as to potentially break existing specs, could we add guidance like, "Duplicate tags MAY be considered an authoring mistake"?

handrews:
@MikeRalphson 

> Rather than being so strict as to potentially break existing specs, could we add guidance like, "Duplicate tags MAY be considered an authoring mistake"?

I'm fond of "The behavior of X is undefined" 😛

r-sreesaran:
@earth2marsh  considering the fact it will break existing code changes, this may be the only way forward.

earth2marsh:
Based on my rationale in #1727 , I'm inclined to close this issue without merging, though I would also like to recognize the author's contribution to the process--thank you!

lornajane:
Closing as I don't see a consensus to accept/merge this. I'll second the acknowledgement of the contribution though! Thanks!

PR #1730, opened 2018-10-26
Registry/Pretty-Spec prep: Travis / package.json simplification

MikeRalphson:
This series of commits:

* Simplifies the Travis CI build configuration using node.js (stable) instead of a hybrid Ruby/Node setup
* Test fixtures for non-current versions of the spec. are removed [1]
* `package.json` is re-worked to prepare for use as a way to publish JSON schemas
* the `schemas/v2.0/README.md` is updated and the old `bower.json` and `package.json` files removed
* Prepares the way for a simpler and easier to maintain setup for publishing the ReSpec render of the specs to the `gh-pages` branch and support for the approved registry implementation [2]

[1] A 'creative' way of interpreting the fact we recommend that tools support the current and previous versions of the spec. allows us to consider v3.1.x and v3.0.0 as current (currently being worked on) and v1.2 and 2.0 as historical. For this reason I've made the decision here to remove all previous test fixtures - as they were previously running against static (non-changing) versions of the spec., added little value, were outdated from a tooling and security point of view and were a maintenance burden.

I envision that we will re-introduce test fixtures at least for the OAS schemas once the v3.x schema(s) is/are merged, and possibly validation of the `examples` too.

[2] I've done the work necessary to allow the registry POC to be hosted on a CNAME'd domain and in a `gh-pages` branch and some minor CSS fixes. It can now be found [here](https://spec.mermade.org.uk/registry/) - still with dummy content.

## TODO

* [ ] check TSC email address in `package.json`


kylef:
@MikeRalphson The new NPM package `oas-schemas` introduced in this PR is still unreleased to the NPM registry, is there anything blocking this? / Is there anything we can do to help get this published?

The existing `swagger-schema-official` NPM package has licensing problems and thus cannot be used while trying to conform to the ambiguous license(s). The SPDX identifier claims it is ISC licensed, the README inside the NPM package tarball states that the package is MIT licensed. Neither licenses are fullfilled in the NPM released package because there is no license file and thus no copyright notice which is a breach of those licenses.

ISC:

> Permission to use, copy, modify, and/or distribute this software for any purpose with or without fee is hereby granted, provided that the above copyright notice and this permission notice appear in all copies. 

MIT:

> Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
>
> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

Neither licenses can be fulfilled as the result of `npm install swagger-schema-official` will result in a local copy of the sources without a copyright notice / permissions notice intact.

MikeRalphson:
@kylef the simplifications to support schema publishing (including rationalising the license situation) were made pending the v3.0 schema work being finished. That was only completed last week. We still need to work on a few details.

kylef:
> We still need to work on a few details.

@MikeRalphson Is there a list of these details / anyway we can help out? Even a pre-release of current state would be great, we're blocked by using the existing `swagger-schema-official` package by our legal team and getting out an update with valid license would allow us to continue.

Issue #1688, opened 2018-09-15
what method to use when link objects use operationRef


dret:
i may just not be able to find it, but it seems to me that when link objects use `operationRef` (instead of `operationId`), it is not specified which method to use for the request? i am looking at the examples ar https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#operationref-examples and wonder how an actual request would look like. i don't know how to figure out which method to use.

MikeRalphson:
The `operationRef` is a reference into an OAS document which points at an `operation` object. The key of that object (the last component of the `operationRef`) is the HTTP method to use.

```yaml
links:
  UserRepositories:
    # returns array of '#/components/schemas/repository'
    operationRef: '#/paths/~12.0~1repositories~1{username}/get'
    parameters:
      username: $response.body#/username
```

here the method is `get`.

MikeRalphson:
@dret does this answer your query?

dret:
not quite, but thanks for looking into this. i am still wondering: clients are expected to parse the URI and extract the last component of the path in the fragment identifier? if that's the logic they're supposed to use, that should be documented. if there's a different way (this way seems sort of brittle and not easy to implement, in particular given that the URI syntax has no concept of paths in fragments identifiers, and thus URI parsers wouldn't help with implementing this), that should be documented instead.

MikeRalphson:
It's not the URI format which dictates the way fragments are interpreted. As I'm lead to understand, it is the mediatype. Here that is (effectively) `application/json` so JSON reference rules are used.

It is the tooling interpreting the OAS document which needs to parse and follow the `operationRef` (not a runtime client interpreting a response), so it is relatively simple for that tool to recognise it has reached an `operation` object where (in this case) `get` is the last component of the ref - and thus the method to use.

dret:
On 2018-10-27 12:40, Mike Ralphson wrote:
> It's not the URI format which dictates the way fragments are 
> interpreted. As I'm lead to understand, it is the mediatype. Here that 
> is (effectively) |application/json| so JSON reference rules are used.

very true. but then you have to parse the fragment identifiers 
themselves. btw, there are no fragment identifiers defined for JSON, so 
what you're referring to here is the way in which OpenAPI defines and 
uses them.

> It is the tooling interpreting the OAS document which needs to parse and 
> follow the |operationRef| (not a runtime client interpreting a 
> response), so it is relatively simple for that tool to recognise it has 
[...]

PR #1733, opened 2018-10-27
minor clarification for operationId usage in link objects

dret:
it's a bit confusing that both the id and the reference are called "operationId", so this tweak makes the text a bit more explicit.

### DIFF
--- a/versions/3.0.3.md
+++ b/versions/3.0.3.md
@@ -2093,7 +2093,7 @@ solely by the existence of a relationship.
 ##### OperationRef Examples
 
 As references to `operationId` MAY NOT be possible (the `operationId` is an optional 
-value), references MAY also be made through a relative `operationRef`:
+value in an [Operation Object](#operationObject)), references MAY also be made through a relative `operationRef`:
 
 ```yaml
 links:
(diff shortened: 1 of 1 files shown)


dret:
this originates from a discussion about #1688

versions/3.0.3.md
```diff
@@ -2093,7 +2093,7 @@ solely by the existence of a relationship.
 ##### OperationRef Examples
 
 As references to `operationId` MAY NOT be possible (the `operationId` is an optional 
-value), references MAY also be made through a relative `operationRef`:
+value in an [Operation Object](#operationObject)), references MAY also be made through a relative `operationRef`:
```
MikeRalphson:
```suggestion
field in an [Operation Object](#operationObject)), references MAY also be made through a relative `operationRef`:
```

webron: Can't we just make `operationId` into a link to `#operationId` without adding the extra text?

MikeRalphson: We could do, but as `operationId` exists in both objects, I think the extra text is worthwhile for those reading the sentence without hovering or clicking on the link.

webron: Duh, missed that 🤦‍♂️

Issue #1726, opened 2018-10-25
Anchor Links in the spec do not work anymore


cebe:
Seems github filters the `<a name="...">` tags...

cebe:
As I am very sure that this has been working fine before it might be worth contacting github support and ask to revert the change or otherwise fix it.

Gi60s:
It's not all links, just any links that have multiple words. For example:

Works: `#schema`
Does not work: `#infoObject`

One way to fix is to update the multi-word links to use dashes instead of camel case. For example, convert `#infoObject` to `#info-object` and links will start working again. 

I'm not sure why Github ever supported camel case links. I've haven't seen another markdown to HTML converter support that.

cebe:
With dashes are the links that are automatically generated by github. camelCase are anchors added in the spec markdown. If you render the spec outside of github links would be working. The filtering done here is specific to github.

MikeRalphson:
Also affects `README`s, reported on #1734

cebe:
I have contacted github support about this.

cebe:
Looks like they fixed it. It's working again for me. Close the issue if you confirm.

MikeRalphson:
Thanks @cebe I'm still seeing the problem, but it may be a cacheing issue.

Gi60s:
It doesn't work for me yet. Tried on two browsers and even disabled cache.

ramsey:
It's still not working for me. Maybe it's cached at an edge location somewhere?

shockey:
Still broken for me as well 😕

webron:
I'm also in touch with GitHub's support about it, will update when I know more.

cebe:
Got a reply from github support today:

> Hey Carsten,
> 
> Our Engineering Team shipped a fix for those anchor links.
> 
> Please let us know if you have any further trouble with it.
> [...]

webron:
Thanks, @cebe - got the same reply. Let the anchor-following resume!

PR #1743, opened 2018-11-08
Changed anchor tags from camelCase to snake-case, Fix for issue #1726

MariusLerf:
Fix for #1726

### DIFF
--- a/versions/3.0.2.md
+++ b/versions/3.0.2.md
@@ -16,64 +16,64 @@ An OpenAPI definition can then be used by documentation generation tools to disp
 <!-- TOC depthFrom:1 depthTo:3 withLinks:1 updateOnSave:1 orderedList:0 -->
 
 - [Definitions](#definitions)
-	- [OpenAPI Document](#oasDocument)
-	- [Path Templating](#pathTemplating)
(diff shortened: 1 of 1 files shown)


ramsey:
I guess the big question is: will GitHub fix the issue, or does this need to be merged to fix the issue? This is a problem on every repository that linked to anchors, so the OAI docs repo is not the only one affected.

cebe:
According to Github support this was an unintentionally made breaking change and they are working on a solution. Given the fact that it is working for me right now, I assume they are rolling out the fix to a limited set of accounts or networks before applying it for all.

cebe:
Btw: While this PR fixes it for master, it breaks external links to the spec, that refer to specific headlines and it does not fix it for links that point to a tag instead of master. So I'd prefer to wait for github fixing it instead of merging this. For future stability it might be wise to not rely on github for hosting it but to put it on the website instead.

MikeRalphson:
Thanks @cebe 

>  For future stability it might be wise to not rely on github for hosting it but to put it on the website instead.

This is under development -- see #1564, though the markdown version will remain the source-of-truth.

PR #1762, opened 2018-11-28
gh-pages; Draft registry content

MikeRalphson:
A deployment of this code can be tested at https://spec.mermade.org.uk/registry/index.html

Before go-live, but not necessarily before merging this PR, probably requires a `CONTRIBUTING.md` or `DEVELOPMENT.md` with some guidelines.


webron:
Approved with a vote by the @OAI/tsc.

earth2marsh:
Thanks for the work, @MikeRalphson, and I approve of the general direction. Would you mind adding for history why the registry is being introduced and what the TODOs are (like contributing.md and development.md) for this to be completed? 

Also, could you remind me why the registry is part of this repo and not standalone? What are the arguments for/against?

darrelmiller:
@earth2marsh We don't have a separate repo because we kept going back and forth on it and we decided not to create a separate repo for anything until it was obvious that it was needed.

Issue #1722, opened 2018-10-23
Yet another Overlay proposal


darrelmiller:
A continuation of the conversation in #1442 but with a new approach.

----

# OpenAPI Overlays

In recent months we have been discussing various use cases for overlays and various solutions.  The following proposal takes a somewhat more radical approach to the problem.  It is a more ambitious proposal than the others we have seen before but the additional complexity does allow for supporting many of the scenarios that have been discussed to date.

#### <a name="overlayDocument"></a>Overlay Document

An overlay document contains a list of [Update Objects](#overlayUpdates) that are to be applied to the target document.  Each [Update Object](#updateObject) has a `target` property and a `value` property.  The `target` property is a JSONPath query that identifies what part of the target document is to be updated and the `value` property contains an object with the properties to be overlayed.

----
Note: 
I have previously argued against using JSONPath due to its lack of formal specification and due to the fact that we already use JSON Pointers in OAS documents, there is risk of confusion.  However, based on further research there do appear to be significant body of implementations and the defacto Javascript implementation uses a BNF Grammar to define the syntax.  This grammar could be used to create a normative description of the JSONPath syntax.   

----

#### <a name="overlayObject"></a>Overlay Object

This is the root object of the [OpenAPI Overlay document](#oasDocument).

##### Fixed Fields

Field Name | Type | Description
---|:---:|---
<a name="overlayDescription"></a>description | `string` | A human readable description of the purpose of the overlay.
<a name="overlayUpdates"></a>updates | [[Update Object](#updateObject)] | A list of update objects to be applied to the target document.

The list of update objects MUST be applied in sequential order to ensure a consistent outcome.  This enables objects to be deleted in one update and then re-created in a subsequent update.

#### <a name="updateObject"></a>Update Object

This object represents one or more changes to be applied to the target document at the location defined by the target JSONPath.

##### Fixed Fields

Field Name | Type | Description
---|:---:|---
<a name="updateTarget"></a>target | `string` | A JSONPath expression referencing the target objects in the target document.
<a name="updateValue"></a>value | [Any](#valueObject) | An object with the properties and values to be updated in the target document.

The properties of the `Value Object` MUST be compatible with the target object referenced by the JSONPath key.  When the Overlay document is applied, the properties in the `Value Object` replace properties in the target object with the same name and new properties are appended to the target object.

##### Structured Overlays Example

[...]

PR #1765, opened 2018-11-28
Define behaviour for specification extensions with null values

MikeRalphson:
Makes the application of the tentative overlay proposal in #1722 easier to reason about, but is a good cleanup regardless.


darrelmiller:
Change rejected as per @OAI/tsc  and dealing with the overlaying null values will be dealt in the overlay spec.

Issue #832, opened 2016-11-08
Summary field in Info Object


rbren:
Several objects have both a `description` and a `summary` field. The distinction seems to be that summaries are plaintext and should be <= 120 chars, while descriptions are markdown and can be arbitrarily long.

I'd like to have an optional `summary` field in the Info Object e.g. for display in API Galleries. Currently to achieve this on [AnyAPI](https://any-api.com) we strip HTML/md out of the `description` field and truncate it, which isn't great.

I'd also like to see a higher level note about `summary` vs. `description` and add the "SHOULD be <= 120 characters" to each summary field (right now it's just mentioned on `operation.summary`)

RobDolinMS:
@bobby-brennan As there have been a bunch of updates to the spec recently, would you want to take a look and see if the latest version addresses some or all of your concerns?

bobby-brennan:
Thanks Rob. Unfortunately I don't think either concern has been addressed:

* The Info Object only has `description`, no `summary` field
* Example Object and Path Item Object have `summary` fields, but no note about recommended max characters

Proposals:
* Add `summary` field to Info Object
* Move "SHOULD be less than 120 characters" note from `Operation.summary` to a higher level, or put that note in each `summary` field.

Thanks!

DavidBiesack:
I also suggested adding summary to Tag Object but that was deferred. I prefer the spec be consistent and allow summary everywhere a description is allowed, and vice versa (unfortunately in JSON schem uses title instead of summary so we can't be fully consistent)

darrelmiller:
Info object has Title.  Do we really need Summary as well?  Consistency is nice, but consistency for consistencies sake is not all that helpful.

darrelmiller:
I'm also tending towards not limiting the size of text fields.  Where do we stop? Should we define limits on names, keys, email, version, urls?  I fear that by fixing the size of every field we are doing busy work that will cause more problems than it will actually solve.

I realize that this may be a controversial position to take, but if we look at other specs like JSON Schema, HTTP, and even JSON itself, the property value sizes are not constrained and people still manage to build tooling that works.

whitlockjc:
I agree with @darrelmiller.

If it ends up becoming the official stance, I would advocate for removing `summary` and using `description` as _the_ place to describe objects.  Of course, tooling will have to handle the fact that this could be plain text or GFM but to me, that's a tooling issue _(and an easy one to solve since many libraries out there handle this)_.

[...]

PR #1779, opened 2018-12-17
832 add info.summary

ioggstream:
## This PR

Adds  `info.summary` field to be a service description (eg. for catalog purposes). The field:

 - is optional
 - it's plain-text (no markdown, which is allowed in `info.description` instead, eg [uspto](https://github.com/OAI/OpenAPI-Specification/blob/master/examples/v3.0/uspto.yaml#L12))
 - there's no enforced limit

## Note

To make an analogy with a product:

```
  title: "Ferrari Testarossa"
  summary: "A 12-cylinder mid-engine sports car manufactured by Ferrari"
  description: |-
    A longer description using markdown, eg. see [uspto]
￼￼
```  

See: https://github.com/OAI/OpenAPI-Specification/issues/832#issuecomment-407108439

### DIFF
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -211,6 +211,7 @@ The metadata MAY be used by the clients if needed, and MAY be presented in editi
 Field Name | Type | Description
 ---|:---:|---
 <a name="infoTitle"></a>title | `string` | **REQUIRED**. The title of the application.
+<a name="infoSummary"></a>summary | `string` | A short, plain-text summary of the application.
 <a name="infoDescription"></a>description | `string` | A short description of the application. [CommonMark syntax](http://spec.commonmark.org/) MAY be used for rich text representation.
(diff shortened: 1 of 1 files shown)


ioggstream:
@bobby-brennan I agree. markdown support and `short` are mutually exclusive imho. 
@MikeRalphson what do you think?

darrelmiller:
@OAI/tsc Can we get a vote on this issue?  I'm for it, we just need to get a majority vote to get it in.  Please weigh in with your opinions.

earth2marsh:
A plain text summary seems reasonable to me. I like that other `summary` fields encourage with "short," so I'd leave it in there, myself. LGTM as is.

darrelmiller:
Update the field descriptions to say:
summary: "A short description of..."  (omitting the plain-text constraint)
description:  "A verbose description..."

ioggstream:
@MikeRalphson can we merge now, or we'll do it at the TSC?

MikeRalphson:
@ioggstream can we leave it to the TSC call? I'm unsure whether the intention was to squash the two commits for clarity or merge as-is.

Issue #1720, opened 2018-10-19
Inconsistent usage of quotes in yaml examples


ndjensen:
This is a minor nitpick but as someone unfamiliar with the OpenAPI specifcation and yaml until now, I found it confusing at first.  There is inconsistent usage of quotes in the yaml examples in the [3.0.2 specification](https://github.com/OAI/OpenAPI-Specification/blob/3.0.2/versions/3.0.2.md) (and presumably earlier).

Specifically, there are places where the yaml examples have content `'application/json'` and other spots where it has `application/json`.  Same confusion applies to other content values.  A notable example is the [file uploads section](https://github.com/OAI/OpenAPI-Specification/blob/3.0.2/versions/3.0.2.md#considerations-for-file-uploads) where it has content `application/octet-stream` without quotes, `'image/jpeg'` and `'image/png'` with single quotes, and then `multipart/form-data` without quotes.

I also noticed inconsistencies with the `$ref: value` where for example it has `$ref: "#/components/schemas/Pet"` with double quotes and `$ref: '#/components/schemas/SomePayload'` with single quotes.

Is there a best practice?  Reading the examples it wasn't immediately clear to me when I should quote vs not quote, and if single quote vs double quote is more standard.

MikeRalphson:
The YAML specification is referenced within the OAS specification, and so we defer the quoting rules to that spec. E.g. a `$ref` value containing a `#` character requires quoting so it isn't seen as beginning a comment.

> Is there a best practice?

I think we should avoid recommending any stylistic choices for the YAML/JSON representations of the OAS document object graph. A simplistic example would be white-space in JSON. There are obvious pros and cons to having formatted versus 'minified' JSON depending on your use-case.

Consistency in examples, i.e. to reduce cognitive load when learning the OAS, is I think a fair point, and a PR for discussion would, I'm sure, be welcomed.

darrelmiller:
I go back and forth between, "we should use different styles to highlight that multiple ways work" and "we should just be consistent".  The challenge of course is deciding what are the "standard" conventions.

I accept that we are making it harder to learn by distracting people with unnecessary variations.  I suppose the rule could be no quoting unless really necessary and default to double quote unless single quote is necessary.

ioggstream:
I agree with @darrelmiller. Where should I PR that?

MikeRalphson:
[...]

PR #1786, opened 2018-12-21
Define a quoting rule for yaml examples.  #1720.

ioggstream:
## This PR

Define a quoting rules for yaml samples.


MikeRalphson:
@darrelmiller did you intend this guidance to make it into the spec? In any case it looks like a breaking change, so would have to wait for v4.0

darrelmiller:
@MikeRalphson My intent with the rule was for it to be guidance for us writing the specification.  I don't think this should be a rule imposed on OpenAPI description writers.  I have no desire to try and invent a new YAML subset.

MikeRalphson:
Just checking!

MikeRalphson:
Closing as not intended to be part of the spec.

PR #1791, opened 2019-01-02
I think these back quotes were forgotten.

seiyab:


MikeRalphson:
Thanks - then I think we can close this one, as all v3.0.3 changes will be forward-ported to v3.1.0. Thanks again.

PR #1792, opened 2019-01-02
explicit expression for '/'

seiyab:
before
```
The field name MUST begin with a slash.
```

after
```
The field name MUST begin with a `/`.
```

### DIFF
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -653,7 +653,7 @@ The path is appended to the URL from the [`Server Object`](#serverObject) in ord
 
 Field Pattern | Type | Description
 ---|:---:|---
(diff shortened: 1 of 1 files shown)


MikeRalphson:
Thanks. Maybe we could say: "with a forward slash (`/`)". Also, changes which don't affect meaning or behaviour of the spec can go into the next patch release.

seiyab:
Thank you for your response.
So should I propose it like this https://github.com/OAI/OpenAPI-Specification/pull/1793 ?

MikeRalphson:
Yes #1793 looks great. Closing this as per your other issue (patch release changes will be forward-ported into v3.1.0 before release).

PR #1802, opened 2019-01-10
fix difference between yaml and json in Response Object Examples

seiyab:
`example` is missed.


MikeRalphson:
Thank you very much for this catch!

PR #1807, opened 2019-01-12
fix typo in Callback Object

seiyab:
It seems to be a typo.


darrelmiller:
I believe you are correct.  Thanks for the PR.

MikeRalphson:
Many thanks for the PR!

Issue #1675, opened 2018-08-24
Questions about Server Variable binding and defaults


tedepstein:
I'm reading about [Server Objects](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#server-object) and [Server Variables](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#server-variable-object), and trying to understand how resolution/expansion is supposed to work. 

The OpenAPI document can specify any number of servers, each with zero or more variables. Each variable can be an arbitrary string, or can be constrained to a string `enum`. Also, each variable MUST specify a `default` value. 

Here are my questions: 
* I guess one of the envisioned scenarios involves a service implementation that binds the server variables at runtime, according to the host, port, and path where the server is running. So it creates a new OpenAPI document with the variables resolved, and passes this to the client. Is that right?
* Another possible scenario is where the API consumer retrieves the original OpenAPI spec, with unbound variables, and dynamically resolves them with some known values. The client may have those server variable values supplied in a configuration file, or they may be provided by a human user at runtime. Correct? 
* I don't see anything in the OpenAPI spec that tells us how variable resolution is expected to happen, or how it should be coordinated between the client and the OpenAPI document provider (which may or may not be the same as the server). I suppose the IANA-registered [media type for OpenAPI documents](https://github.com/OAI/OpenAPI-Specification/issues/110), when we have one, could include a protocol for server variable binding in headers or query parameters, but I don't see any mention of this in the referenced issue. So, is it fair to say that server variable binding is entirely implementation-dependent?  
* The `default` property of a Server Variable Object has this description in the spec:
    > REQUIRED. The default value to use for substitution, and to send, if an alternate value is not supplied. Unlike the Schema Object's default, this value MUST be provided by the consumer.

    I don't understand what "by the consumer" means in this context, or maybe I'm just finding it confusing. 
    * Is this just trying to say that each Server Variable MUST specify a `default` value? In that case it would not be the "consumer" who provides the default; it would be the OpenAPI document author or provider, right? 
    * Or is it trying to say that if a value _other than the default_ needs to be bound to the variable, then the "consumer" must provide that value? In that case it seems odd to say that consumer MUST do this, when there is no explicit mechanism specified for doing so. Maybe it would be a little more natural to say that it is up to the implementation, on the client side or the provider side, to substitute non-default values.

Any clarifications or comments appreciated.

darrelmiller:
[...]

PR #1809, opened 2019-01-14
Server Variable Object clarifications

webron:
Two clarifications:
- The `enum` of the Server Variable Object cannot be empty. This is different than the JSON Schema `enum`. The reason being that Server Variables must always have a value (see `default`), and an empty array suggests a 'no value' is acceptable.
- Since `enum` cannot be empty, and `default` is mandatory, if they both exist, the `default` value has to be one of the `enum` values (otherwise there's a conflict in logic).

If the PR is accepted, I'll submit the same change to 3.1.0.

### DIFF
--- a/versions/3.0.3.md
+++ b/versions/3.0.3.md
@@ -429,8 +429,8 @@ An object representing a Server Variable for server URL template substitution.
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="serverVariableEnum"></a>enum | [`string`] | An enumeration of string values to be used if the substitution options are from a limited set.
-<a name="serverVariableDefault"></a>default | `string` |  **REQUIRED**. The default value to use for substitution, which SHALL be sent if an alternate value is _not_ supplied. Note this behavior is different than the [Schema Object's](#schemaObject) treatment of default values, because in those cases parameter values are optional.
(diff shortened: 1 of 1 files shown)


MikeRalphson:
> If the PR is accepted, I'll submit the same change to 3.1.0.

I do have a script in the works which can create patches from one input file and try to apply them to another based on their last shared git history. It might be cleaner **not** to duplicate commits and to attempt to do them in a regularised manner.

MikeRalphson:
Is adding a `MUST NOT` a bit strong for a patch release, even if it is a clarification of intent?

webron:
Not in my opinion, but I'm biased 😀

MikeRalphson:
My argument has always been that tooling implementations should be compatible no matter which patch release the authors read while implementing. If this change results in different `if` tests being required, and we would actually add non-backwards compatible changes to the schema for it (e.g. `minItems: 1`) then it's actually a breaking change, and would have to wait for v4. :man_facepalming:

darrelmiller:
@webron  to fix this....

webron:
@MikeRalphson please review again

tedepstein:
@webron, @darrelmiller, @MikeRalphson, since you're on the subject, just a quick reminder that some of us out here don't understand server variables at all. Like, how are they supposed to work? 

Please see #1675.

PR #1818, opened 2019-01-29
Fix/spec fixups

BelfordZ:


tedepstein:
@BelfordZ , should we assume this was an accidental upstream push? Looks like a big set of changes, creating a new `open-rpc` specification...

BelfordZ:
My bad @tedepstein ! 

Second time I do that today...

BelfordZ:
@tedepstein also yes, Im working on a simplified spec made specifically for JSON RPC 2.0.

Its in a pretty banged up state right now, but I'd love some feedback when Im ready. Whats a good place to come chat with you folks? gitter?

MikeRalphson:
@BelfordZ I don't think we use gitter, but there is a slack at http://open-api.slack.com/ or you'd be very welcome to join one of our regular Thursday open TSC calls. Details in the pinned issues.

PR #1825, opened 2019-02-04
Fixing response body link comment

AvoH:


AvoH:
Looks like already fixed in 3.0.2.

Issue #1106, opened 2017-05-16
Need clarification on how to handle multiple OAuth scopes in an endpoint


dclucas:
The [specs](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#security-requirement-object) make it very clear that in case of multiple security schemes are present in the spec, that should be read as a logical AND.

On the other hand, when describing the list of scopes within a scheme (see paste below), the text _seems_ to imply a logical AND operation as well, without making it 100% explicit.

>  If the security scheme is of type "oauth2", then the value is a list of scope names required for the execution. For other security scheme types, the array MUST be empty.

So can we make it clear on the text whether those entries should be an "OR", "AND" or whether the specs imply no relationship whatsoever?

webron:
If it's a list of required scopes, it does mean AND. We can certainly improve the wording though.

dclucas:
Cool, tks. I will submit another issue against the new spec, to try to cover for scenarios where an "OR" would be required.
On the current text, more than happy to submit a PR to make this more explicit...

webron:
You're welcome to submit a PR to that effect. As for covering OR, keep in mind we are _very_ close to finalizing this version of the spec, so I suggest you do it fast and we just _might_ be able to squeeze it in. Would suggest that you add a suggestion to the construct itself, otherwise it's going to be a very slim chance it'll get in.

PR #1829, opened 2019-02-06
security; widen use of scopes array to other securityScheme types

MikeRalphson:
Split off from #1764 as discussed. In summary, I believe the outstanding issue is to find a non-contentious term possibly in preference to `scopes` or `roles` and to check this makes sense for all securityScheme types, including `http`'s subtypes.

### DIFF
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -3340,7 +3340,7 @@ When a list of Security Requirement Objects is defined on the [OpenAPI Object](#
 
 Field Pattern | Type | Description
 ---|:---:|---
-<a name="securityRequirementsName"></a>{name} | [`string`] | Each name MUST correspond to a security scheme which is declared in the [Security Schemes](#componentsSecuritySchemes) under the [Components Object](#componentsObject). If the security scheme is of type `"oauth2"` or `"openIdConnect"`, then the value is a list of scope names required for the execution. For other security scheme types, the array MUST be empty.
+<a name="securityRequirementsName"></a>{name} | [`string`] | Each name MUST correspond to a security scheme which is declared in the [Security Schemes](#componentsSecuritySchemes) under the [Components Object](#componentsObject). If the security scheme is of type `"oauth2"` or `"openIdConnect"`, then the value is a list of scope names required for the execution. For other security scheme types, the array MAY contain a list of role names which are required for the execution, but are not otherwise defined or exchanged in-band.
 
 ##### Security Requirement Object Examples
 


ioggstream:
imho this change should include/reference some spec examples.

darrelmiller:
@darrelmiller Review naming of scopes vs roles

darrelmiller:
This PR allows the use of scopes for non OAuth2 security schemes.  It does not provide a way to describe the set of valid scopes for those scenarios.  We do believe that it would be valuable to have a place to define the allowed scopes but we need to spend some more time discussing how this can be incorporated into the SecurityScheme object.  We will do this in the rc1 timeframe.

ioggstream:
>  we need to spend some more time discussing how this can be incorporated into the SecurityScheme object. We will do this in the rc1 timeframe.

WDYT to add some example to this PR as a comment?

glroman:
As mentioned in #1106 and #1903, it would be really useful to support the ORing of scopes.  For example:

  /pet:
    post:
      security:
        - oauthScopes: [read, write] | [admin]

Will this be in the PR?  Thanks.

MikeRalphson:
@glroman can you not do:

```yaml
  /pet:
    post:
      security:
        - oauthScopes: [read, write]
        - oauthScopes: [admin]
```

> Security Requirement Objects that contain multiple schemes require that all schemes MUST be satisfied for a request to be authorized. This enables support for scenarios where multiple query parameters or HTTP headers are required to convey security information.

This is ANDing. Here you can't have two references to `oauthScopes` because that would be a duplicate YAML/JSON key, but you could merge the scopes arrays.

> When a list of Security Requirement Objects is defined on the OpenAPI Object or Operation Object, only one of the Security Requirement Objects in the list needs to be satisfied to authorize the request.

This is ORing. Here you **can** have multiple references to `oauthScopes` because they are in separate elements of the `security` array.

glroman:
When I connect with a client that has "admin", this works:

    /pet:  # Scenario 1
       post:
         security:
           - oauthScopes: [read, write]
           - oauthScopes: [admin]

but this doesn't:

    /pet:  # Scenario 2
       post:
         security:
           - oauthScopes: [admin]
           - oauthScopes: [read, write]

I get:

    {
      "detail": "Provided token doesn't have the required scope",
      "status": 403,
      "title": "Forbidden",
      "type": "about:blank"
    }


Is this expected behavior?  My env:

    python                  3.7.1
    connexion               2.6.0
    openapi-spec-validator  0.2.8

MikeRalphson:
I think that may be an issue with `connexion` then.

darrelmiller:
@ioggstream  We have the opportunity to add examples in the examples not included in the spec.  I don't think we need to include an example in the specification because I do not think there is widespread usage of scopes in non-OAuth2 scenarios.

@MikeRalphson  I have come to terms with the use of the term "role".  I don't feel the need to change it here.

Defining scopes/roles at the security scheme level is not going to make it into 3.1.

Issue #1830, opened 2019-02-06
Path Templating Clarification: Is a variable without a corresponding path parameter valid?


tedepstein:
The specification [makes it clear](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#fixed-fields-10) that each defined path parameter must correspond to a same-named variable placeholder in the path:

> If [`in`](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#parameterIn) is `"path"`, the `name` field MUST correspond to the associated path segment from the [path](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#pathsPath) field in the [Paths Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#pathsObject). See [Path Templating](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#pathTemplating) for further information.

But I cannot find anything in the spec that states the converse rule: 

**Each variable expression in the path MUST correspond to a path parameter that is included in the [Path Item](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#path-item-object) itself and/or in each of the Path Item's [Operations](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#operation-object).**

My guess is that this rule should apply, and that this was the intent. But I don't want to make assumptions, given the long history of the spec and the abundance of edge cases where something needs to be allowed for non-obvious reasons. 

This is important for OpenAPI document validation.
* Is there ever a reason to have a variable placeholder in the path without a corresponding path parameter?
* If so, I think we should add some language to the spec clarifying that this is allowed, hopefully providing some rationale for this, and a clear description the API semantics in such cases. 
* Otherwise, would we consider adding the rule as stated above to be a breaking change? 
    * If it's a breaking change, I'd propose adding the above rule with SHOULD instead of MUST, and leaving this issue (or a follow-up isse) open for the next major spec revision so we can change it to MUST.
    * Otherwise, I'd propose adding the above rule to [Path Templating](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#pathTemplating) or to the `/{path}` field pattern in [Paths Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#user-content-pathspath).

webron:
Closed as PR is merged.

PR #1831, opened 2019-02-06
Path Templating Clarification - proposed fix for #1830. 

tedepstein:
Proposed fix for #1830. Each variable expression in a path must have a corresponding path parameter.

See the associated issue for further details and questions. 

Notes:
* The proposed spec change uses the term _variable expression_ to refer to a curly brace-delimited parameter name used within the path. Borrowed this terminology from [RFC 6570](https://tools.ietf.org/html/rfc6570).
* For consistency, I changed the language in the [Parameter Object `name` field](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#fixed-fields-10). 
    * The previous text referred to an "**associated path segment** from the path field in the Paths Object." 
    * I changed this to refer instead to refer to a "**variable expression** occurring at least once within the path field in the Paths Object."
    * I wanted to flag this here because I see that there are still open issues about variable expressions and path segments, asking for clarification or additional flexibility in the use of variable expressions as partial path segments and/or spanning path segments. Maybe the "associated path segment" language is meaningful, in which case we should discuss further.
* This is branched from `v3.0.3-dev`, optimistically hoping it will be deemed a clarification, not a breaking  change. The issue asks that question, and proposes SHOULD instead of MUST if we need to make it non-breaking. Also, if it matters, we can target this change to 3.1 instead of 3.0.3.


MikeRalphson:
@tedepstein generally LGTM. Your language "appearing at least once", is that meant to cater for the ambiguous case here https://github.com/microsoftgraph/microsoft-graph-openapi/issues/9#issuecomment-439895379 and can we instead look at outlawing it (i.e. each variable-expression / parameter MUST be named differently)?

tedepstein:
@MikeRalphson , it's not intended to allow duplicate parameter definitions, but rather >1 occurrence of a single parameter within a path, like this:

```yaml
paths:
  /events/{eventID}/comments/{eventID}/{commentID}: 
```

the `{eventID}` path parameter occurs in two different segments. Offhand, I can't think of a good reason to include a path parameter more than once in a path. But I imagine there must be some APIs in the wild that do this. And AFAICT, there's nothing in the OpenAPI spec that disallows it.

If this is surfacing a new question that we'd rather not address right now, I can remove "at least once."

MikeRalphson:
But is `eventID` here *meant* to be defined identically, or is there an unintended ambiguity?

tedepstein:
> But is `eventID` here _meant_ to be defined identically, or is there an unintended ambiguity?

The `eventID` parameter can only be defined once, so repeated references to that parameter should always resolve to identical values.  I don't think that's ambiguous.

tedepstein:
For reference, the specification addresses parameter uniqueness [here](https://github.com/OAI/OpenAPI-Specification/blob/v3.0.3-dev/versions/3.0.3.md#user-content-pathitemparameters), [here](https://github.com/OAI/OpenAPI-Specification/blob/v3.0.3-dev/versions/3.0.3.md#user-content-operationparameters), and [here](https://github.com/OAI/OpenAPI-Specification/blob/v3.0.3-dev/versions/3.0.3.md#user-content-parameterobject).

MikeRalphson:
Yes. I'm concerned with variable-expression uniqueness as they relate to parameters.

tedepstein:
@MikeRalphson, 

> Yes. I'm concerned with variable-expression uniqueness as they relate to parameters.

OK, I'm pretty sure we don't need to address that concern in the scope of this PR, so I'll remove "at least once."

versions/3.0.3.md
```diff
@@ -71,7 +71,9 @@ An OpenAPI definition can then be used by documentation generation tools to disp
 A document (or set of documents) that defines or describes an API. An OpenAPI definition uses and conforms to the OpenAPI Specification.
 
 ##### <a name="pathTemplating"></a>Path Templating
-Path templating refers to the usage of curly braces ({}) to mark a section of a URL path as replaceable using path parameters.
+Path templating refers to the usage of variable expressions, delimited by curly braces ({}), to mark a section of a URL path as replaceable using path parameters.
```
darrelmiller: It would be good if we could align with the RFC6570 terminology and call these things template expressions.
```suggestion
Path templating refers to the usage of template expressions, delimited by curly braces ({}), to mark a section of a URL path as replaceable using path parameters.
```

versions/3.0.3.md
```diff
@@ -1010,7 +1012,7 @@ There are four possible parameter locations specified by the `in` field:
...
+<a name="parameterName"></a>name | `string` | **REQUIRED**. The name of the parameter. Parameter names are *case sensitive*. <ul><li>If [`in`](#parameterIn) is `"path"`, the `name` field MUST correspond to a variable expression occurring within the [path](#pathsPath) field in the [Paths Object](#pathsObject). See [Path Templating](#pathTemplating) for further information.<li>If [`in`](#parameterIn) is `"header"` and the `name` field is `"Accept"`, `"Content-Type"` or `"Authorization"`, the parameter definition SHALL be ignored.<li>For all other cases, the `name` corresponds to the parameter name used by the [`in`](#parameterIn) property.</ul>
```
darrelmiller:
```suggestion
<a name="parameterName"></a>name | `string` | **REQUIRED**. The name of the parameter. Parameter names are *case sensitive*. <ul><li>If [`in`](#parameterIn) is `"path"`, the `name` field MUST correspond to a template expression occurring within the [path](#pathsPath) field in the [Paths Object](#pathsObject). See [Path Templating](#pathTemplating) for further information.<li>If [`in`](#parameterIn) is `"header"` and the `name` field is `"Accept"`, `"Content-Type"` or `"Authorization"`, the parameter definition SHALL be ignored.<li>For all other cases, the `name` corresponds to the parameter name used by the [`in`](#parameterIn) property.</ul>
```

tedepstein:
@darrelmiller, thanks, that's fine with me. Your suggestions missed one occurrence of "template variable," so I just made the change in a new commit #6929fee75b86f31303d97e70760798819ff3977b. 

Diffs on this PR should now show "template expression."

PR #1857, opened 2019-03-05
Fix link to RFC 2119

Glenn-Aaron-Fisher-BP:
The link to http://www.ietf.org/rfc/rfc2119.txt is dead. Updated the link within the incorporation phrase to match the current standard—https://tools.ietf.org/html/rfc2119.

### DIFF
--- a/versions/2.0.md
+++ b/versions/2.0.md
@@ -3,7 +3,7 @@
 
 #### Version 2.0
 
-The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](http://www.ietf.org/rfc/rfc2119.txt).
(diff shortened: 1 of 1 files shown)


darrelmiller:
The .txt link works fine for me. Although,  I don't have any objections to updating it to the prettier version.

MikeRalphson:
@darrelmiller this change is targeted at version 2.0 of the spec, is that an issue?

darrelmiller:
@MikeRalphson Ugh. Sorry didn't notice that.  I don't think we should update 2.0 for any reason.

ioggstream:
In that case we I'd:

 - close the PR 
 - eventually address the improvement in the newer specs.

my 2¢, R.

Glenn-Aaron-Fisher-BP:
@darrelmiller When I double-checked the .txt link, it was still down. But on further investigation, it turned out to be a bizarre corporate proxy issue that affects www.ietf.org but not tools.ietf.org.

I have no idea why BP's proxy has this issue, or even if it's intentional. But either way it seems to be entirely on my end.

Closing this PR, as it's not necessary.

Issue #645, opened 2016-04-11
Spec translations


webron:
We're getting approached from time to time by users who'd like to translate the spec into other languages (French, Chinese, ...).

Suggest we define a (simple) procedure to propose such translations and hopefully validate them. May also want to consider asking for translation of supporting documents as well.

Parent: #589

mailtokun:
I am going to translate it to Chinese.  You can get it on http://ibpm.org.cn/swagger_cn.htm page.

mailtokun:
I hope I can finish the translation work in 2 months.

epastorino:
Hi, @webron , I'd like to collaborate with Spanish translations.
Do you have any translation guidelines?
If not, I was thinking in just forking and create new documents with an +es suffix. E.g: /versions/3.0.1+es.md
I'm using + instead of - so the version number is [semver compliant](https://semver.org/#spec-item-10).

seiyab:
I also want to collaborate with Japanese translations.
If there are no guidelines other than @epastorino 's one, I will follow @epastorino and use /versions/3.0.2+ja.md

↓current Japanese translations.
https://github.com/seiyab/OpenAPI-Specification/blob/v3.0-ja-dev/versions/3.0.2-ja.md

handrews:
@swaldron58 should this be moved over to the Outreach repo?

PR #1861, opened 2019-03-06
want to add Japanese translation

seiyab:
I want to add Japanese OAS.

If supporting various languages is troublesome, feel free to reject it.


seiyab:
Related issue https://github.com/OAI/OpenAPI-Specification/issues/645

earth2marsh:
Thanks for offering a translation. I've invited a colleague to review it, and hopefully give feedback. Two quick thoughts:

1. It will be hard to review this given the lack of Japanese language expertise in the technical steering committee. 
1. Because the written explanation takes precedent over anything else, it would have to defer to the English version as the source of truth. I think we'd need to declare that in any translation? 

The second point makes me wonder if the right approach is to list unofficial translations that are maintained as forks maintained outside of the OAI organization? How would you feel about a list of known community translations of the OpenAPI Specification?

seiyab:
@earth2marsh 
Thank you for your response.
> the right approach is to list unofficial translations that are maintained as forks maintained outside of the OAI organization

It seems to be good.
Should the translation be located in an organization repository ?

PR #1864, opened 2019-03-14
Overlay proposal

darrelmiller:
To improve the process around working on the overlay feature, the contents of the Overlay issue have been moved into a markdown file that lives in the proposals folder in the master branch.  The goal is to bring this document to the point where it can be copied and pasted into the official spec when finalized.  For overlays, this may be a different spec document than the OpenApi spec document.

### DIFF
--- a/proposals/004_Overlays.md
+++ b/proposals/004_Overlays.md
@@ -14,6 +14,7 @@
 |Date |Responsible Party |Description |
 |---- | ---------------- | ---------- |
 | 24th December 2019 | Darrel Miller | Initial draft |
+| 2nd January 2019 | Darrel Miller | Update to wording around removing items from arrays.  Added section on backward compatibility. Clarified process around applying a set of updates. Started to add supported scenarios.|
 
 ## Introduction
 
@@ -22,7 +23,7 @@ In recent months we have been discussing various use cases for overlays and vari
 
 #### <a name="overlayDocument"></a>Overlay Document
 
-An overlay document contains a list of [Update Objects](#overlayUpdates) that are to be applied to the target document.  Each [Update Object](#updateObject) has a `target` property and a `value` property.  The `target` property is a [JMESPath](http://jmespath.org/specification.html) query that identifies what part of the target document is to be updated and the `value` property contains an object with the properties to be overlayed.
+An overlay document contains a list of [Update Objects](#overlayUpdates) that are to be applied to the target document.  Each [Update Object](#updateObject) has a `target` property and a `value` property.  The `target` property is a [JMESPath](http://jmespath.org/specification.html) query that identifies what part of the target document is to be updated and the `value` property contains an object with the properties to be overlaid.
 
 
 #### <a name="overlayObject"></a>Overlay Object
@@ -38,7 +39,9 @@ Field Name | Type | Description
 <a name="overlayExtends"></a>extends | `url` | URL to an OpenAPI document this overlay applies to. 
 <a name="overlayUpdates"></a>updates | [[Update Object](#updateObject)] | A list of update objects to be applied to the target document.
 
-The list of update objects MUST be applied in sequential order to ensure a consistent outcome.  This enables objects to be deleted in one update and then re-created in a subsequent update.
+The list of update objects MUST be applied in sequential order to ensure a consistent outcome.  Updates are applied to the result of the previous updates. This enables objects to be deleted in one update and then re-created in a subsequent update.
+
+The `extends` property can be used to indicate that the Overlay was designed to update a specific OpenAPI description.  This is an optional property.  Where no `extends` is provided it is the responsibility of tooling to apply the Overlay documents to the appropriate OpenAPI description.
 
 #### <a name="overlayInfoObject"></a>Info Object
 
@@ -64,7 +67,6 @@ Field Name | Type | Description
 <a name="updateRemove"></a>remove | `boolean` | A boolean value that indicates that the target object is to be removed from the the map or array it is contained in. The default value is false.  
 
 The properties of the `Value Object` MUST be compatible with the target object referenced by the JMESPath key.  When the Overlay document is applied, the properties in the `Value Object` replace properties in the target object with the same name and new properties are appended to the target object.
-
 
 ##### Structured Overlays Example
 
@@ -138,7 +140,7 @@ updates:
 
 ##### Array Modification Examples
 
-Due to the fact that we can now reference specific elements of the parameter array, it does open the possibilty to being able to add parameters and potentially remove them using a `null` value.
+Due to the fact that we can now reference specific elements of the parameter array, it allows adding parameters. Parameters can be deleted using the `remove` property.  Use of indexes to remove array items should be avoided where possible as indexes will change when items are removed.
 
 ```yaml
 overlay: 1.0.0
@@ -162,7 +164,6 @@ updates:
     remove: true
 ```
 
-
 ## Proposal Summary
 
 ### Benefits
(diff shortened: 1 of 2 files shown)


MikeRalphson:
Sorry to bikeshed, but I'm not really a fan of the name `proposals` for this directory. For one thing the contents are not named for the version of the spec. they apply to and for another I'm unclear as to the lifetime of the contents. Is it intended to retain proposals indefinitely in this directory even once they have progressed to the spec. proper, to avoid breaking people's links?

darrelmiller:
@MikeRalphson Your bikeshedding is welcome and you raise valid questions.  We should address them and change the name appropriately.  I just didn't want to block the other work while we debated the name.

namdeirf:
Sorry I have been out of pocket for a while. I am looking into the specifics of the Overlay proposal and can't find it here https://github.com/OAI/OpenAPI-Specification/tree/master/proposals . The alternative schema proposal is there, and the template for future proposals, but I don't see overlays. Sorry if I missed it.

darrelmiller:
@namdeirf The proposal is now there.

@MikeRalphson The proposals folder is intended to be long lived.  It serves as a place to keep a record of why the change happened and the scenarios that it intends to address.   Your point is valid that we should have metadata that indicates what version of the spec the proposal is targeted.  At least the major version, maybe the minor but not the patch.  In the future we could decide to create subfolders for spec versions to partition the proposals.  However, I'm not sure that is desirable.   Someone creating a proposal may not be aware of the future version timeline and the proposal may take a number of released versions before it is integrated.

I apologize for not addressing this issue earlier, I had seen your comment and then lost where it was until I cam to merge overlays.

proposals/004_Overlays.md
```diff
@@ -0,0 +1,178 @@
...
+      x-safe: true
+- target: paths.*.get.parameters[?name=='filter' && in=='query']
+    value:
+      schema:
+        $ref: "/components/schemas/filterSchema"
+```
+
+##### Array Modification Examples
+
+Due to the fact that we can now reference specific elements of the parameter array, it does open the possibilty to being able to add parameters and potentially remove them using a `null` value.
```
MikeRalphson: Is "using a null value" a hangover from a previous version without the `remove` property?

proposals/004_Overlays.md
```diff
@@ -0,0 +1,178 @@
...
+
+### Benefits
+
+- This approach addresses the two distinct approaches of structured overlay vs targeted overlay which suits distinct but equally valid scenarios.
+- Addresses the problem of modifying the parameters array and removes the need to replace the entire array when a small change is required.
+- Allows sets of related overlays to be stored in a same file.
+- Enables updating a set of objects based on a pattern. This might be an effective way of apply common behaviour across many operations in an API.
+
+### Challenges
+- Tooling will need a JMESPath implementation.
```
MikeRalphson: Suggest adding a "Alternatives considered" paragraph as per the template, which explains why JSONPath was considered and JMESPath preferred (no formal published specification, and security / compatibility concerns over evaluation mechanism).

proposals/004_Overlays.md
MikeRalphson: We should probably also document what happens to array indices when an item is removed.

Consider the array: `[0, 1, 2, 3, 4]`

If we `remove` item `1`, and then remove item `2`, are we left with `[0, 3, 4]` or `[0, 2, 4]`?

darrelmiller: Yes. The `null` language is left over.  I will remove it.

Yes, it would be good to clarify the behavior around remove.  My understanding is the second case is true.  All Updates are applied sequentially and use the result of the previous update as the baseline.  If we don't do that then I think we could get into some funky merge situations even within a single document.

darrelmiller: Thanks for the feedback

proposals/004_Overlays.md
```diff
@@ -0,0 +1,194 @@
...
+When updating properties throughout the target document it may be more efficient to create a single `Update Object` that mirrors the structure of the target document. e.g.
+
+```yaml
+overlay: 1.0.0
+info:
+  title: Structured Overlay
+  version: 1.0.0
+updates:
+- target: "@"
+    value:
```
fmvilas: Bad indentation (it's repeated on every example).

darrelmiller: Sequences (aka arrays) don't need to be indented https://yaml.org/spec/1.2/spec.html#id2802662

> Since people perceive the “-” indicator as indentation, nested block sequences may be indented by one less space to compensate
https://yaml.org/spec/1.2/spec.html#id2799181

fmvilas: Mmm, I didn't mean the `- target` but the indentation on `value`.

darrelmiller: Dammit. Sorry about that.  I'll fix it.

proposals/004_Overlays.md
```diff
@@ -0,0 +1,194 @@
...
+| 2nd January 2019 | Darrel Miller | Update to wording around removing items from arrays.  Added section on backward compatibility. Clarified process around applying a set of updates. Started to add supported scenarios.|
+
+## Introduction
+
+In recent months we have been discussing various use cases for overlays and various solutions.  The following proposal takes a somewhat more radical approach to the problem.  It is a more ambitious proposal than the others we have seen before but the additional complexity does allow for supporting many of the scenarios that have been discussed to date.
+
+
+#### <a name="overlayDocument"></a>Overlay Document
+
+An overlay document contains a list of [Update Objects](#overlayUpdates) that are to be applied to the target document.  Each [Update Object](#updateObject) has a `target` property and a `value` property.  The `target` property is a [JMESPath](http://jmespath.org/specification.html) query that identifies what part of the target document is to be updated and the `value` property contains an object with the properties to be overlaid.
```
mattbishop: Nice, I love JMESPath!

proposals/004_Overlays.md
```diff
@@ -0,0 +1,194 @@
...
+- Large overlays may be slow to process.
+- Multiple complex pattern based overlays may cause overlapping updates causing confusing outcomes.
+
+## Alternatives considered
+
+JMESPath was chosen over JSONPath due to the fact that JMESPath has a [specification](http://jmespath.org/specification.html) and a set of test cases.  This will help to ensure compatibility between implementations.
+
+## Backwards compatibility
+
+Overlays will be described in a new specification that can be used alongside an OpenAPI Description, therefore there will be no compatibility issues for the initial release. Overlay documents can be used against OpenAPI v2 and v3 descriptions.
```
mattbishop: I'm glad to see this, though it might be good to only support v3 to help give people another reason to move to v3.

proposals/004_Overlays.md
```diff
@@ -0,0 +1,194 @@
...
+- This approach addresses the two distinct approaches of structured overlay vs targeted overlay which suits distinct but equally valid scenarios.
+- Addresses the problem of modifying the parameters array and removes the need to replace the entire array when a small change is required.
+- Allows sets of related overlays to be stored in a same file.
+- Enables updating a set of objects based on a pattern. This might be an effective way of apply common behaviour across many operations in an API.
+
+### Challenges
+
+- Tooling will need a JMESPath implementation.
+- Large overlays may be slow to process.
+- Multiple complex pattern based overlays may cause overlapping updates causing confusing outcomes.
```
mattbishop: The overlay order needs to be unambiguous. Moreover, previous overlays in a chain must be immutable. A change to an earlier overlay will likely ruin subsequent overlays. Perhaps an overlay can declare a hash of the document it is updating? This would probably need to be coupled with some kind of ordering convention in the overlay files.

proposals/004_Overlays.md
darrelmiller: I considered it.  I just didn't like the idea of creating an arbitrary constraint that I couldn't really justify.  I'm on the fence.  I suspect if we tell people they can't people will do it anyway.

proposals/004_Overlays.md
darrelmiller: That last comment of mine is poorly worded.   An Overlay document has an array of update objects.  The order of applying those updates is defined based on the order of the array.  Therefore within an Overlay document the order is well defined.  However, there still could be overlapping updates.  That could make the Overlay document tricky to understand.  Best practice should be to try and avoid the situation.

Having multiple Overlay documents apply to a single OpenAPI description is a different challenge.  Currently this is up to tooling to manage the order of application.  I don't like the idea of linking or ordering Overlay documents explicitly in the document, but I could imagine the use of priority value that provides an ordering hint to tooling.

Issue #1442, opened 2017-12-15
RAML Overlay equivalent?


darrelmiller:
In a separate document.  
Has a separate lifecycle.

How does the overlay reference the source document? - By document structure
Is OperationId a valid way of referencing core documents?

earth2marsh:
Could the URL for `extends` be relative? That way I could work on a local copy.

Can multiple `extends` be chained together so several documents that extend each other?

I can see a couple places where overlays would be useful to me:

1. as a way to keep any implementation details for a spec separate from the consumer-oriented doc, such as API management concerns from the API provider's perspective, for example attaching a policy to an operation.
2. to allow a writer to create better descriptions/summaries/titles for attributes like a parameter or an operation without having to edit the spec itself. But since parameters are an array, in order to edit a single description, I would have to reproduce the entire array, no? Is there a way to more precisely target?

jstoiko:
Those are both good use cases, 2. particularly for the localization of API docs .

> Could the URL for extends be relative?

Yes

> Can multiple extends be chained together so several documents that extend each other?

Yes and that's an important point. I updated my comment accordingly.

> I would have to reproduce the entire array, no? Is there a way to more precisely target?

That's right. Arrays are hard to work with when it comes to merging. Instead of replacing arrays, we could decide to concat or union the overlaying array with the target array but this would introduce another (worse) problem by making it impossible to rewrite arrays from scratch. We could also decide to have a more imperative syntax which would include some sort of merge rule to apply to the target document, like "add" or "replace", but this would be to the detriment of readability/usability (IMHO).

MikeRalphson:
> But since parameters are an array, in order to edit a single description, I would have to reproduce the entire array, no? Is there a way to more precisely target?

In the case of `parameters` etc, the issue with targeting one element of the array can be worked around by using `$ref`s to named parameters under the `components` object in the base OAS document.

earth2marsh:
Thanks for clarifying, @jstoiko. What are some other use cases for merging that are more along the lines of signature changes? For example, a proposed change to an API design? 

The crux of this for me comes down to identifying the key use cases. IMO, it is critical to make it easy to invite collaboration (e.g., the second case above, including your localization point), and it would help to keep the syntax simple. 

As a thought experiment, say I wanted to update the `limit` parameter's description in your [sample](https://raw.githubusercontent.com/OAI/OpenAPI-Specification/master/examples/v3.0/petstore.yaml), the easiest I could imagine is:

[...]

PR #1865, opened 2019-03-14
Traits proposal

darrelmiller:
To improve the process around working on the traits feature, the contents of the Overlay issue have been moved into a markdown file that lives in the proposals folder in the master branch. The goal is to bring this document to the point where it can be copied and pasted into the official spec when finalized.

### DIFF
--- a/proposals/Traits.md
+++ b/proposals/Traits.md
@@ -1,0 +1,117 @@
+[Source](https://github.com/OAI/OpenAPI-Specification/issues/1843)
+
+
+I'm opening this issue simply as a place to collect some ideas about how the concepts of [Overlays](#1722) and [Traits](#613) might be brought together.
+
+In both proposals, I think the key notion is a "fragment", which I would describe as: a "sparse" sub-object of an OpenAPI definition.  In the Overlay proposal, a fragment is the `value` of an "Update Object" and has a type of `any`.
+
+I think fragments -- which I would like to call "mixins" -- can have a more well-defined structure than just `any`.  If we use the `discriminator` approach already present in OpenAPI for "mixins", we can require (and validate) conformance to a particular structure.  In particular, we can require a mixin to be a "sparse" form of any well-defined OpenAPI object, e.g. Operation, Response, Parameters, or even the whole OpenAPI definition.
+
+Mixins could be defined as just another flavor of "component".  So
+```
+components:
+  mixins:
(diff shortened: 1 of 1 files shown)


darrelmiller:
- Decide on the name.
- Decided : Mixin references should be array of JSON references
- Merging rules as defined here https://github.com/OAI/OpenAPI-Specification/issues/1442#issuecomment-418549259  by @jstoiko

proposals/Traits.md
```diff
@@ -0,0 +1,117 @@
...
+```
+
+and an operation would "apply" the "pageable" mixin with a $mixin property, as follows:
+```
+paths:
+  /foo:
+    get:
+      description: search for foo resources
+      $mixin: 
+        - pagable
```
darrelmiller: This should be a JSON ref, not just a simple string.

darrelmiller:
We need to collect a set of use cases for internal vs external "overlay/trait/mixin/whatever".  @tedepstein  volunteered to help.

cmheazel:
Yes, we will do the same thing with Alternative Schema.

cmheazel:
@darrelmiller where will the use cases live?  Should we have discrete directories for each proposal?  That way we don't have to include use cases and other supporting material in the markup targeted for the OAS spec.

earth2marsh:
Regarding use cases, there were a few from #1442 I'll bring over here:
as a way to keep any implementation details for a spec se

1. parate from the consumer-oriented doc, such as API management concerns from the API provider's perspective, for example attaching a policy to an operation.
2. to allow a writer to create better descriptions/summaries/titles for attributes like a parameter or an operation without having to edit the spec itself. But since parameters are an array, in order to edit a single description, I would have to reproduce the entire array, no? Is there a way to more precisely target?
3. To identify an SLA. You might have a valid spec that doesn't have that information, but you could ask for the SLA variant for that spec.
4. Codegen is another, as we're polluting specs with metadata for codegen. This gives us a way to separate that out.

darrelmiller:
@cmheazel I was considering the use cases as living in the comments here, but I see your point that it is going to get just as messy.  Perhaps each proposal should have a folder and the spec wording and all related documents, e.g. examples, use cases, spec wording should all live in that folder.

Having a permenant record of why we created a feature is definitely not a bad thing.

MikeRalphson:

PR #1880, opened 2019-03-28
Clarify constraints on Security Scheme Object Scheme Property

darrelmiller:
When declaring the use of a 'http' SecurityScheme Type, the Scheme property is a pointer to the appropriate specification for describing the authentication protocol.  Ideally this scheme should be listed in the IANA Authentication Scheme Registry.  However, a number of popular schemes are not, and some APIs use their own custom scheme.

This wording change clarifies that the scheme SHOULD be listed in the IANA registry, but it is not a MUST.  This leaves the door open for specifications to describe and tooling to support a wider set of authentication schemes.

There should be no expectation for tooling to support unregistered schemes.


darrelmiller:
Add note that "bearer" in applies to is a case insensitive rule.

darrelmiller:
Consider adding wording on expectation of tooling support for external registries.

darrelmiller:
Decided not to add not about case sensitivity of bearer because a) there is no clean way to add that annotation and b) the auth scheme is defined as case insensitive, so it should be expected that any form of the bearer identifier should be sufficient.

darrelmiller:
Chose not to add wording around tooling support.  The use of a registry for defining a set of values implicitly indicates that that tooling may not support a particular value.  We cannot impose the requirement to support all registry values on tooling as any new addition to the registry would then make tooling not compliant with the spec.

Issue #513, opened 2015-11-21
Scopes Object should be optional?


dwhieb:
The [Security Scheme Object](http://swagger.io/specification/#securitySchemeObject) lists the Scopes Object as required. However, the [OAuth 2.0 specification for the Access Token Scope](http://tools.ietf.org/html/rfc6749#section-3.3) allows the client to omit the scope parameter, as long as a default behavior is specified:

> If the client omits the scope parameter when requesting authorization, the authorization server MUST either process the request using a pre-defined default value or fail the request indicating an invalid scope. The authorization server SHOULD document its scope requirements and default value (if defined).

In [other places in the spec](http://tools.ietf.org/html/rfc6749#section-4.2.1), the scope parameter is explicitly stated as optional.

Does this mean that the Scopes Object should be optional too? Or allowed to be empty?

I'm wondering because I'm implementing an API with the Implicit grant flow where there's only one scope. Since there's just the one, I didn't want to have to declare it explicitly.

webron:
You can leave it empty, but that means none is required.

dwhieb:
Great - that's exactly what I was hoping for.

Follow-up question though: For the Security Requirement Object, the spec says:
> If the security scheme is of type "oauth2", then the value is a list of scope names required for the execution.

I assume the array can just be left empty too?

webron:
Yup.

dwhieb:
Awesome - thanks @webron!

tedepstein:
@webron , would you mind reopening this issue? I'm submitting a PR to clarify that this usage is allowed. The spec is ambiguous or maybe misleading IMO, and we tripped over this in our implementation.

The [OAuth Flow Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#oauth-flow-object) says that the `scopes` map is REQUIRED, and doesn't specify any meaning for an empty object value. We took this to mean that `scopes` is required to be present AND non-empty, otherwise an empty value would seem to be circumventing the intention of the spec.

Likewise, [Security Requirement Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#securityRequirementObject) says:

> If the security scheme is of type "oauth2" or "openIdConnect", then the value is a list of scope names required for the execution. For other security scheme types, the array MUST be empty.

[...]

PR #1888, opened 2019-04-05
Clarify the spec to allow optional or unspecified OAuth scopes

tedepstein:
Referencing issue #513. Clarify the spec to accommodate OAuth schemes where scope is optional in the authorization request, or where scope is not used at all.

The [OAuth Flow Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#oauth-flow-object) says that the scopes map is REQUIRED, and doesn't specify any meaning for an empty object value. We took this to mean that the `scopes` property is required to be present AND non-empty, otherwise an empty value would seem to be circumventing the intention of the spec.

This PR makes it explicit that the scopes map MAY be empty.

Likewise, [Security Requirement Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#securityRequirementObject) says:

> If the security scheme is of type "oauth2" or "openIdConnect", then the value is a list of scope names required for the execution. For other security scheme types, the array MUST be empty.

It doesn't explicitly disallow an empty scope list for an OAuth security requirement. But taken in context, it's easy to assume, mistakenly, that the scopes array must be non-empty, so this PR explicitly allows an empty scopes list on oauth2 security requirements.

### DIFF
--- a/versions/3.0.3.md
+++ b/versions/3.0.3.md
@@ -3286,13 +3286,13 @@ Field Name | Type | Applies To | Description
 <a name="oauthFlowAuthorizationUrl"></a>authorizationUrl | `string` | `oauth2` (`"implicit"`, `"authorizationCode"`) | **REQUIRED**. The authorization URL to be used for this flow. This MUST be in the form of a URL.
 <a name="oauthFlowTokenUrl"></a>tokenUrl | `string` | `oauth2` (`"password"`, `"clientCredentials"`, `"authorizationCode"`) | **REQUIRED**. The token URL to be used for this flow. This MUST be in the form of a URL.
 <a name="oauthFlowRefreshUrl"></a>refreshUrl | `string` | `oauth2` | The URL to be used for obtaining refresh tokens. This MUST be in the form of a URL.
-<a name="oauthFlowScopes"></a>scopes | Map[`string`, `string`] | `oauth2` | **REQUIRED**. The available scopes for the OAuth2 security scheme. A map between the scope name and a short description for it. If scope is optional, the map MAY include an entry with an empty string as its key to represent the default scope. If scope is not used in the authorization scheme, the map MAY be empty.
+<a name="oauthFlowScopes"></a>scopes | Map[`string`, `string`] | `oauth2` | **REQUIRED**. The available scopes for the OAuth2 security scheme. A map between the scope name and a short description for it. If scope is not required or not specified in the authorization scheme, the map MAY be empty.
 
 This object MAY be extended with [Specification Extensions](#specificationExtensions).
 
 ##### OAuth Flow Object Examples
 
-###### OAuth Flows with Required Scope
+###### OAuth Flows with Defined Scopes
 
 ```JSON
 {
@@ -3333,7 +3333,7 @@ flows:
       read:pets: read your pets 
 ```
 
-###### OAuth Flows with Unspecified and Optional Scope
+###### OAuth Flows with Optional or Unspecified Scope
 
 ```JSON
 {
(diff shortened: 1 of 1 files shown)


tedepstein:
**Note:** #513 suggests that there are OAuth implementations where scope is _optional_ in the authorization request. In cases where the authorization request omits the scope parameter, the authorization server uses a pre-defined _default value_ for the scope. 

I tried to address this in the [first commit](https://github.com/OAI/OpenAPI-Specification/pull/1888/commits/79470fe5a1398a0f53d79cf863087055993ccdc6) by allowing a scope with an empty string as its name (i.e.  `""` as its map key) to represent the default scope.  Other named scopes could also be defined alongside this one.

But I removed the empty-string provision in the [second commit](https://github.com/OAI/OpenAPI-Specification/pull/1888/commits/0e189448a593baa7850118173d213c8c298393c2) because it seemed to create ambiguities in the Security Requirement Object. For example: 
* If empty string is a defined scope, can Security Requirement include the empty string in its scope list?
* In that case, what's the difference between a scopes array with empty string vs. an empty scopes array?  
    * Maybe this:
        * `"security": [ { "oauth_scheme": [""] } ]` (scopes array with empty string) means "The caller must be authorized in the default scope."
        * `"security": [ { "oauth_scheme": [] } ]`  (empty scopes array) means "The caller must be authorized in some scope(s) that are not specified here, therefore the scope requirements are implementation-defined."
    * Or maybe they just mean the same thing. 
* Should we disallow scopes array with empty string and a non-empty string?
    * `"security": [ { "oauth_scheme": ["", "read:pets"] } ]` would seem to say, "the caller must be authorized in the default scope AND in the 'read:pets' scope."
    * This doesn't make much sense to me, but I guess this would be just one of many nonsensical things that designers can express in their API specs. So maybe we just leave it to API designers to specify a sane authorization scheme. 

If anyone has good answers to these questions and/or feels strongly that this needs to be addressed in this PR, let's discuss. 

Otherwise, I'm happy to defer these gnarly questions and just address the case of empty `scopes` in this PR.

versions/3.0.3.md
```diff
@@ -3286,12 +3286,14 @@ Field Name | Type | Applies To | Description
 <a name="oauthFlowAuthorizationUrl"></a>authorizationUrl | `string` | `oauth2` (`"implicit"`, `"authorizationCode"`) | **REQUIRED**. The authorization URL to be used for this flow. This MUST be in the form of a URL.
 <a name="oauthFlowTokenUrl"></a>tokenUrl | `string` | `oauth2` (`"password"`, `"clientCredentials"`, `"authorizationCode"`) | **REQUIRED**. The token URL to be used for this flow. This MUST be in the form of a URL.
 <a name="oauthFlowRefreshUrl"></a>refreshUrl | `string` | `oauth2` | The URL to be used for obtaining refresh tokens. This MUST be in the form of a URL.
-<a name="oauthFlowScopes"></a>scopes | Map[`string`, `string`] | `oauth2` | **REQUIRED**. The available scopes for the OAuth2 security scheme. A map between the scope name and a short description for it.
+<a name="oauthFlowScopes"></a>scopes | Map[`string`, `string`] | `oauth2` | **REQUIRED**. The available scopes for the OAuth2 security scheme. A map between the scope name and a short description for it. If scope is not required or not specified in the authorization scheme, the map MAY be empty.
```
webron: Why not just keep it simple and say that the array may be empty? There's no need to describe why it might be empty. Users who need it to be empty, know why the need it empty.

versions/3.0.3.md
```diff
@@ -3331,6 +3333,36 @@ flows:
       read:pets: read your pets 
 ```
 
+###### OAuth Flows with Optional or Unspecified Scope
```
webron: Let's remove this, please. Adding more examples for edge cases just makes the spec longer and harder to read. This belongs in a tutorial.

tedepstein:
@webron, done.

PR #1899, opened 2019-04-19
Callbacks key definition - fixes #1856

### LINKED ISSUE #1856: Bug in definition of callbacks
I'm talking about the [callbacks field](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#operation-object) `callbacks :== Map[string, Callback Object | Reference Object]` 

> The key value used to identify the **callback object** is an expression, evaluated at runtime, that identifies a URL to use for the callback operation.

both defined in
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#operationObject
and
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#callbackObject

This means that this is a valid `callbacks` map:
```yml
callbacks:
  "$request.query.queryUrl": # key value used to identify the callback object
    <CallbackObject>
  "$url":  # key value used to identify the callback object
    <CallbackObject>
```

Now, by definition of a Callback Object:
> The key that identifies the **Path Item Object** is a runtime expression that can be evaluated in the context of a runtime HTTP request/response to identify the URL to be used for the callback request.

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#key-expression

so we end up with some mess like this:

```yml
callbacks:
  "$request.query.queryUrl": # key value used to identify the callback object
    "$request.query.queryUrl": # key that identifies the Path Item Object
      <PathItemObject>
    "$url": # key that identifies the Path Item Object
      <PathItemObject>
  "$url":  # key value used to identify the callback object
    "$request.query.queryUrl":  # key that identifies the Path Item Object
      <PathItemObject>
    "$url": # key that identifies the Path Item Object
      <PathItemObject>
```

This makes no sense to me and I'm assuming it's a bug. Reading through #1141, I think that the intention was to make this example valid
```yml
callbacks:
  "foo":
    <CallbackObject>
  "bar":
    <CallbackObject>
```
but it's not, because the literals `"foo"`, `"bar"` are not [runtime expressions](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#runtime-expressions).

ahl:
Fixes #1856 

The current text of the `callbacks` field is this:

```
A map of possible out-of band callbacks related to the parent operation. The key is a unique identifier for the Callback Object. Each value in the map is a Callback Object that describes a request that may be initiated by the API provider and the expected responses. The key value used to identify the callback object is an expression, evaluated at runtime, that identifies a URL to use for the callback operation.
```

The key is multiply defined; I believe there's leftovers from previous iterations as the Callback Object definition includes this:

```
The key value used to identify the callback object is an expression, evaluated at runtime, that identifies a URL to use for the callback operation.
```

I deleted that from the former as it seems to apply only to the latter.


ahl:
@webron are you the right person to review this? Thanks!

darrelmiller:
@ahl  Thanks for this PR.  You are correct, that looks like a failed copy/paste effort.

ahl:
Apologies for the delay. Did I understand correctly that @webron wanted to preserve the typo in 3.0.2? Accordingly I've only applied the correction to 3.0.3.

cc @darrelmiller

darrelmiller:
@ahl  We don't change published versions of the spec for any reason.  That's what patch versions are for.

ahl:
@darrelmiller yes, I was just confirming my understanding. Do you require any further changes or is this sufficient as-is? Thanks!

ahl:
@oai/tsc can someone take a look?

MikeRalphson:
Do we need to specify the scope within which the keys must be unique (i.e. within the `callbacks` property / the `operation` object, not the whole document)?

ahl:
@MikeRalphson that would seem to be implied by "The key is a unique identifier for the Callback Object" and by the structure definition itself: Map[string, Callback Object | Reference Object].

Do you suspect that's a legitimate source of confusion? If so perhaps simply omitting the word "unique" would suffice to bring it in line with other Map descriptions. My suggestion though is that if it hasn't confused anyone, it's probably not worth clarifying.

ahl:
@darrelmiller or @MikeRalphson is this ready to be merged or are there changes you'd like to see? Thanks!

PR #1908, opened 2019-04-28
add Scorpio HTTP client library to IMPLEMENTATIONS.md

notEthan:
Scorpio is a Ruby library that offers an HTTP client built around OpenAPI.
I added a section for this as it does not seem to fit the existing sections.


notEthan:
thanks!

PR #1917, opened 2019-05-03
oauth2 rfc6749 requires TLS for authorization endpoint

AxelNennker:
Signed-off-by: Axel Nennker <axel.nennker@telekom.de>

Referencing [OAuth2 Authorization Endpoint](https://tools.ietf.org/html/rfc6749#section-3.1)
```Since requests to the authorization endpoint result in user authentication and the transmission of clear-text credentials (in the HTTP response), the authorization server MUST require the use of TLS as described in Section 1.6 when sending requests to the authorization endpoint.```

Changed the other http URLs to https as well because 'confidentiality' e.g. user information in some examples

### DIFF
--- a/examples/v2.0/yaml/petstore-expanded.yaml
+++ b/examples/v2.0/yaml/petstore-expanded.yaml
@@ -113,12 +113,14 @@ definitions:
       - $ref: '#/definitions/NewPet'
       - required:
         - id
+        type: "object"
         properties:
           id:
             type: integer
             format: int64
 
   NewPet:
+    type: "object"
     required:
       - name  
     properties:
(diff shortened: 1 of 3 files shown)


MikeRalphson:
Thanks for the PR. As v3.0.2 is released, changes should go against the most recent patch version on the associated `-dev` branch.

You will see that a number of these changes have already been made in v3.0.3.

As to the meat of this change, I think we should note what users of the specification MUST do, and not what components of the oAuth dance MUST do, to make it clear we are not intending to make changes to the referenced RFC behaviour. I.e. something like "The authorization URL must specify a secure scheme" rather than "The authorization server MUST require the use of TLS."

AxelNennker:
Closing this as it is not based on the latest -dev branch.
Created new PR based on v3.1.0-dev
https://github.com/OAI/OpenAPI-Specification/pull/1959

PR #1922, opened 2019-05-09
TOB members 2019

ncaidin:
To reflect the latest TOB elections.

### DIFF
--- a/TOB.md
+++ b/TOB.md
@@ -4,7 +4,7 @@
 > The TOB is responsible for managing conflicts, violations of procedures or guidelines or other issues that cannot be resolved in the TSC for the OAS. For further details please consult the OpenAPI Project Charter.
 
 ## TSC Elected - terms through May 2021
-Isabelle Mauny - 42Crunch
+Isabelle Mauny
 
 Uri Sarid @usarid
 
--- a/TOB.md
+++ b/TOB.md
@@ -1,0 +1,21 @@
+#Technical Oversight Board ("TOB")
+
+## Description: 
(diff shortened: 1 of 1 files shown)


ncaidin:
@OAI/tsc  could you please review my pull request?

usarid:
@ncaidin I may well be too pedantic here, but when I read

> and any cross-project or high-level issues that cannot be resolved in the TSC for the OAS

I worry that someone might interpret this as being the governing body, vs the BGB. Changing the charter, for example, is a cross-project high-level issue, but should not be done by the TOB. So should we somehow constrain this language?

ncaidin:
Thanks @usarid .

I simply did a copy/paste from the Charter to make sure it was consistent. Please bear in mind that the TOB has been invoked zero times since OAI has been formed. I think the part where it says "in the TSC"  makes it clear. The TOB is intended to resolve conflicts that the TSC can not in order to avoid having to go to the BGB. 

What if I change it to this:
"The TOB is responsible for managing conflicts, violations of procedures or guidelines or other issues that cannot be resolved in the TSC for the OAS. For further details please consult the OpenAPI Project Charter."

Personally I would be okay keeping the language as is or changing to this. Also open to other suggestions, but suggest we KIS (keep it simple). 

What do you think?

usarid:
The new language sounds great to me. Thanks!


________________________________

ncaidin:
@usarid  text updated. Thanks for your review.

ncaidin:
@usarid , is there anything else I need to do to get this PR approved and merged?

ncaidin:
@OAI/tsc  is there anything more that needs to be done with the request to get it approved and merged? 

Thanks.

PR #1923, opened 2019-05-10
Fix missing schema type in YAML example spec

sjaensch:
These were fixed for the JSON version of the examples back in [mid 2015](https://github.com/OAI/OpenAPI-Specification/commit/94924733c072fb4784a35be529c37947ea99ed83#diff-c3f2d48f1c40a088bbac6fdff2f9895a), but some are still missing for the YAML versions. This is the root cause for Yelp/bravado#416.

### DIFF
--- a/examples/v2.0/yaml/petstore-expanded.yaml
+++ b/examples/v2.0/yaml/petstore-expanded.yaml
@@ -109,6 +109,7 @@ paths:
             $ref: '#/definitions/Error'
 definitions:
   Pet:
+    type: "object"
     allOf:
       - $ref: '#/definitions/NewPet'
       - required:
@@ -119,6 +120,7 @@ definitions:
             format: int64
 
   NewPet:
+    type: "object"
     required:
       - name  
     properties:
@@ -128,6 +130,7 @@ definitions:
         type: string    
 
   Error:
+    type: "object"
     required:
       - code
       - message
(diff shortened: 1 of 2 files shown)


darrelmiller:
The problem is that the absence of the type is not an error.  Type is not a required property in a schema.  By fixing the examples we lead people down the path of assuming type will always be there, when it will not.

sjaensch:
> The problem is that the absence of the type is not an error. Type is not a required property in a schema. By fixing the examples we lead people down the path of assuming type will always be there, when it will not.

I think that's a bit besides the point - the YAML spec is different in content _and meaning_ from the JSON spec now (example: the JSON version doesn't allow non-`object` types, the YAML spec does). They should be identical in content and meaning.

There might be a separate issue here in how bravado recognizes its models.

webron:
@sjaensch we just talked about on the call and explained @darrelmiller why the `type` should completely be there. We did make one minor change to your PR to make it a bit more consistent, hopefully you won't object to it, but if you do, please let us know.

jstoiko:
Also, if you don't mind @sjaensch, could you make those same changes in the OAS 3.0 version of those examples (i.e. in `examples/v3.0/yaml/`)? I could have made a separate PR like I had volunteered to do during the tsc call but I now feel too lazy to do so (and it really belongs in the same PR anyway :) ).

sjaensch:
@jstoiko sure thing! Since this PR has been merged I opened #1931.

PR #1931, opened 2019-05-17
Add missing object type to 3.0 examples

sjaensch:
As discussed in #1923 I'm adding the `object` type to the 3.0 example specs where it's missing.


MikeRalphson:
Many thanks!

PR #1937, opened 2019-05-30
allow, but discourage, requestBody for GET, HEAD, DELETE

### LINKED ISSUE #1801: Allow requestBody for the DELETE method.
I would like the specification to allow a requestBody in the DELETE method and other methods without explicitly defined semantics.

One of the answers in this [StackOverflow ](https://stackoverflow.com/questions/299628/is-an-entity-body-allowed-for-an-http-delete-request) post states: "The spec does not explicitly forbid or discourage it, so I would tend to say it is allowed."
I would agree with that statement.

Currently the OpenAPI spec says "The requestBody is only supported in HTTP methods where the HTTP 1.1 specification RFC7231 has explicitly defined semantics for request bodies."

I think this should raise a notice, but not be unsupported and raise an error.

Addendum:

As I mentioned in a comment below, if you do choose to create a batch delete method on your API, whether or not anyone recommends it, make sure to check your cache settings and think about how these settings will interact with DELETE requests. Calling `DELETE` on `/users`  with data `[3,4]`  will not invalidate the cache for `/users/3` or `/users/4`, so a GET request to either of them may return invalid data depending on cache settings. My comment below describes methods of mitigating this.

n2ygk:
Fixes #1801 per TSC discussion of 2019-05-30 which requested a PR for review.


philsturgeon:
This is a good change.

n2ygk:
@philsturgeon Thanks. I suppose I need to forward-port this against `3.1.0-dev` as well?

philsturgeon:
@n2ygk I expect there are periodic merges from master to 3.1.0-dev.

n2ygk:
> @n2ygk I expect there are periodic merges from master to 3.1.0-dev.

@philsturgeon Yeah I was thinking that, except the filename is `3.1.0.md` vs. `3.0.3.md` so a merge  won't "just do it" as it's a different file. I'll wait for direction if and when this is approved and merged. Thanks.

MikeRalphson:
1. We have a script which will forward port changes from patch releases (e.g. 3.0.3) to minor releases (e.g. 3.1.0) across files, obviating the need for any merges from `master`.
2. As this change calls for different behaviour for tools written to this spec as opposed to versions 3.0.0, 3.0.0 and 3.0.2 I would *personally* prefer to see this targetted against v3.1.0.
3. Typo: "symantics" for "semantics".

ioggstream:
I'd remember the following from RFC7230 too, as HTTP intermediaries could rejejct the request. 

>   sending a payload body on a GET request might cause some existing
>   implementations to reject the request.

n2ygk:
@MikeRalphson:
> 1. We have a script which will forward port changes from patch releases (e.g. 3.0.3) to minor releases (e.g. 3.1.0) across files, obviating the need for any merges from `master`.
> 2. As this change calls for different behaviour for tools written to this spec as opposed to versions 3.0.0, 3.0.0 and 3.0.2 I would _personally_ prefer to see this targetted against v3.1.0.
OK, so I'll await for direction from the TSC on whether I should submit this against 3.1.0 instead.
> 3. Typo: "symantics" for "semantics".
D'oh! Thanks. commit coming.

n2ygk:
> I'd remember the following from RFC7230 too, as HTTP intermediaries could rejejct the request.
> 
> > sending a payload body on a GET request might cause some existing
> > implementations to reject the request.

@ioggstream: I believe that's in [RFC 7231 4.3.1](https://tools.ietf.org/html/rfc7231#section-4.3.1), not 7230. I prefer not to quote too much from other references rather than just citing them as is already done in the prior sentence.

ioggstream:
@n2ygk 

> I believe that's in RFC 7231 4.3.1, not 7230.
True ;)

> I prefer not to quote too much from other references.
Agree. I think the caveat I cited is easier to understand respect
to the current proposal.

I generally  agree on the content.

philsturgeon:
👍  latest wording looks good. Again we don't want to ban this, but we can suggest folks avoid it because spooky confusion my occur. SHOULD is the way to go.

n2ygk:
@darrelmiller is there a plan to merge this at some point? Just trying to clean up my various open PRs....

rcgroup-areuin:
Hello,
When this change will be release (if not already done)?
The version **_3.0.3 is the one?_**
We are looking up to upgrade from Swagger 2 to **_OpenAPI 3_** but this issue had us blocked.
Is this finally **_approved?_** (if so, we can begin in parallel and wait for release) 
Thanks.

n2ygk:
> I agree with @MikeRalphson that as this is actually changing allowed behaviour we should defer this change to 3.1

@darrelmiller Do you want me to rebase this PR against 3.1?

BTW, I would make the counter-argument that it fixes a mistake in 3.0 vs. adding new functionality,  as 3.0 breaks functionality from Swagger 2.  So it could go in 3.0.3, thereby unblocking @rcgroup-areuin and others who are documenting pre-existing APIs -- not inventing new ones. Anyone "doing it right" would never run across this bug/feature in the spec anyway.

Is there a need for another TSC meeting agenda item to move this forward to a conclusion? I see a note in the June TSC #1941 needing to follow up on this. I imagine the September TSC is coming up shortly. Also, are there meeting minutes posted somewhere?

tomoya-kinsho:
vd

handrews:
@darrelmiller @MikeRalphson should this be added to #2025 for 3.1 tracking?

earth2marsh:
As discussed in today's TSC call, this relaxes our stance on an issue that, as Ron observed, is something v2 was more relaxed about. General consensus was to accept this PR, thank you.

MikeRalphson:
Bah! If I had remembered which branch this was targeted against, I would have rattled my semver stick.

webron:
@MikeRalphson - you're right. This should be reverted and reapplied to 3.1 - it can't be changed in 3.0.

webron:
@n2ygk - unfortunately we had to revert it (see #2114) but we'd still like to see the change in 3.1.0. Would you like to submit a similar PR against that version? I'd rather you get the credit as you put in the work.

n2ygk:
> @n2ygk - unfortunately we had to revert it (see #2114) but we'd still like to see the change in 3.1.0. Would you like to submit a similar PR against that version? I'd rather you get the credit as you put in the work.

will do

n2ygk:
@webron see #2117

PR #1944, opened 2019-06-11
Update 3.0.0.md

jhonyourangel:
this looks like a typo to me.

should be `content:` ?
not sure `representations:` is correct

all others are `content`
also `representations` is not part of the openapi specification

### DIFF
--- a/versions/3.0.0.md
+++ b/versions/3.0.0.md
@@ -1737,7 +1737,7 @@ Response with a string type:
 
 ```yaml
 description: A simple string response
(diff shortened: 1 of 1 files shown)


MikeRalphson:
This was fixed in https://github.com/OAI/OpenAPI-Specification/commit/d45549e004a54a72d11a9ec9aa676bd2d0748d30 in v3.0.1 - (each release of the specification is unchanging once released). Thanks for the PR.

Issue #801, opened 2016-09-29
Proposal: Expand OpenAPI to include RPC APIs


timburks:
The OpenAPI 2.0 Specification states that its goal is [“to define a standard, language-agnostic interface to REST APIs”](https://github.com/OAI/OpenAPI-Specification). REST is an extremely popular style for implementing APIs that run over HTTP and related protocols (HTTPS, HTTP/2, QUIC). But other transport mechanisms are common and APIs are often defined in ways that correspond to procedure calls. In some ways, procedure calls are more fundamental, as REST APIs are often implemented by systems that convert REST requests into procedure calls.

At times it is desirable to use remote procedure calls (RPCs) as the primary form of an API. RPC systems allow expression of semantics that go outside the range of the HTTP verbs used by REST. RPC systems map naturally to the function calls that are used to implement most software systems, and RPC provides a design pattern that can be use to guide system implementations. The RPC abstraction is also much older than REST, and new systems such as [gRPC](http://www.grpc.io/) and [Thrift](https://thrift.apache.org/) show its enduring usefulness.

With this in mind, we think it would be beneficial to expand the scope of the OpenAPI specification to include remote procedure call semantics. Here we propose an expanded specification that builds on our experience using RPCs to connect very large distributed systems while being general enough to apply to a broad range of variations.
### Protocol Buffers

We give special consideration to the [Protocol Buffer](https://developers.google.com/protocol-buffers/) message representation. At Google, Protocol Buffers are used to send and receive tens of billions of messages each second. Protocol Buffers are also used for messages in the open-source [gRPC](http://www.grpc.io/) framework and other more specialized RPC frameworks [[1]](http://pages.cs.wisc.edu/~nater/trpc/index.html) [[2]](http://blogs.mulesoft.com/biz/mule/lightweight-rpc-with-%C3%B8mq-and-protocol-buffers/) [[3]](http://hstore.cs.brown.edu/documentation/development/protorpc/).

Protocol Buffers are typically represented in a binary form, and the latest version ([proto3](https://developers.google.com/protocol-buffers/docs/proto3)) also allows mapping to representations in JSON and YAML formats. Messages can be generally treated as collections of typed fields, just as they are in other serialization formats including [Thrift](https://thrift.apache.org/), [Avro](https://avro.apache.org/), and [Cap’n Proto](https://capnproto.org/). Thus message description can be treated similarly for all of these representations, including the JSON and YAML that OpenAPI uses to describe request and response objects.

[...]

PR #1946, opened 2019-06-12
OpenAPI supports any type of plain HTTP API

philsturgeon:
Twice in the last few days I have had people ask if its ok to use OpenAPI for RPC, and I would 
say its better at describing RPC than REST currently. OpenAPI functionality is mostly aiming at APIs at Level 1 or Level 2 of the Richardson Maturity Model, with a quick nudge towards hypermedia control concepts via links, but definitely nothing around Level 3.

image

You have to get preeeeetty creative to get a REST (Level 3) API described with OpenAPI, and even then it mostly involves [adding your own hypermedia format](https://apisyouwonthate.com/blog/json-api-openapi-and-json-schema-working-in-harmony) and kinda mentioning it exists in OpenAPI.

Seeing as RMM Level 1 and Level 2 is what OpenAPI is helping with, it seems a little off that the OpenAPI is always talking about REST APIs. I'm not sure if that was for marketing, or born from the old "REST is HTTP and JSON" confusion, but... it seems rather incorrect. 

Lets remove the limitation by fixing this wording, which would unblock larger talks about things like gRPC support for v4, and maybe even other level 0 implementations like GraphQL.

Relates to #801.


darrelmiller:
The definition of RPC that OpenAPI cannot support is where the identification of the "operation" is contained inside the request payload.  That's not going to be possible while we limit operation to only be discriminated by path.

philsturgeon:
That’s fine. Level 1 = URIs, Level 0 is method=sendMail.

So when I say we’re allowing ROC, we’re allowing some types of HTTP RPC API, and anything RESTish.

darrelmiller:
We are waiting on BGB approval to change wording of the charter.  Then we should be good to go ahead with these changes.

Issue #1410, opened 2017-11-10
Create issue template and PR template for the project


darrelmiller:

lornajane:
Closing, we've handled this in other issues since this one.

PR #1951, opened 2019-06-14
Github Issue Template with "Spec Issues Only" Advisory

tedepstein:
See issue #1410

This PR creates a single GitHub issue template with boilerplate text advising that this repo is only for the written OpenAPI Specification, not tooling.  

This is meant to reduce the number of tooling-related issues that have to be swatted away. Later, we can add other helpful header info, template placeholders, and we can create multiple issue templates for different issue types.

Here's what workflow looks like with this change: 

image

image

image

image

I wasn't sure about including the links, given that users will see the raw markdown in the editor. But a lot of issue templates include links, and they are easily visible and usable in preview mode. I made them relative links to reduce clutter in the editor view.


.github/ISSUE_TEMPLATE/specification-issue.md
```diff
@@ -0,0 +1,12 @@
+---
+name: Specification Issue
+about: Create a new issue related to the OpenAPI Specification
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**PLEASE NOTE:** This project only maintains the written [OpenAPI Specification](../blob/master/versions/3.0.2.md) and a few other supporting artifacts. No OpenAPI-related tooling lives here. If your issue is with a documentation format, UI, editor, programming framework, or other [OpenAPI implementation](../blob/master/IMPLEMENTATIONS.md), please post your issue in the appropriate project repository or contact the provider.
```
MikeRalphson: Could we reword "documentation format", as this might dissuade people from submitting issues against the ReSpec (pretty) version of the spec?

.github/ISSUE_TEMPLATE/specification-issue.md
```diff
@@ -0,0 +1,12 @@
...
+
+_Please delete the above notice before submitting._
```
MikeRalphson: As an alternative, the text above could be wrapped in an HTML comment, so would be visible during issue creation, but not when the submitted issue was viewed. Thus saving the submitter from having to delete the advice paragraph.

tedepstein: @MikeRalphson, you mean a `<!-- comment like this -->`, right? 

The tradeoff is that we won't be able to create links that work in preview mode. The entire notice will be hidden in the preview and in the final issue description, only visible when editing. In that case, maybe I'd take the relative links out of the text, and add absolute links to the end so the user can copy and paste to get to the referenced pages.

> \

How does that look?

MikeRalphson: Apart from the links to the current version of the specification, which would create a bit more work for the release manager to keep up to date, this LGTM. Maybe we could just link to the `versions` directory, though it's not ideal?

BelfordZ:
Just a suggestion: Make two issue templates, the first one titled "Tooling Bug", second "Specification Bug". 

In the "Tooling Bug" template, put `**STOP** this is not the place for this!!` or w.e.

This way, you can still have a template that is tailored towards people leaving issues for the spec.

lornajane:
Closing in favour of newer pull request #3211

Issue #1389, opened 2017-10-25
nullable vs type arrays (JSON Schema compatibility)


handrews:
I understand that fully supporting `type` arrays is a challenge for a variety of reasons.  I would like to suggest that a compromise that would be functionally identical to `nullable` but compatible with JSON Schema:

* `type` MUST be a string or a two-element array with unique items
* if an array, one of the unique items MUST be `"null"`

The only way in which this is inherently more difficult to implement than `nullable` is that parsing `type` becomes slightly harder.  The functionality, validation outcomes, code generation, etc. should be identical.

philsturgeon:
I know this has been discussed greatly in the past, and as a brief recap:

I love OpenAPI, have been pushing it hard at work, and I also love contract testing. JSON Schema is fantastic at contract testing, doing this sort of thing:

<img width="512" alt="image" src="https://user-images.githubusercontent.com/67381/33454894-4b94fc8a-d5e8-11e7-9d5e-7c95839368d6.png">


Do we special case that particular use of the array? Or do we give the middle finger to people who want to generate models?

handrews:
> We'd still be left with an item on the "JSON Schema vs OAI discrepancies" list

I think the key thing is that the remaining item would be a "feature of JSON Schema that is not supported", while the current situation is "feature of JSON Schema that is functionally supported but requires a conflicting syntax".

I'm not that bothered by a system saying "hey, we don't quite want to use all of JSON Schema as defined for validation".  In fact, I want to spend effort in draft-08 making it *easier* to do that in general, so that projects like OpenAPI don't have to awkwardly document the deltas and either write their own tooling or try to wrap up existing tooling some way to forbid bits and pieces.

With the two-element array option, and some effort in draft-08 around being able to declare support for things like vocabulary subsets, we could get OAI and JSON Schema fully compatible with each other, without requiring OAI to support everything JSON Schema dreams up for validation.  After all, validation and code generation are really not the same things, and we should not pretend that they are.

RobJDean185:
Another reason for supporting @handrews suggestion is that despite commentary/opinion otherwise on whether the v2.0 spec formally permits this or not, the real world continues to intrude.

Companies like SAP (see http://api.sap.com/) produce exactly that construct in their v2.0 output. They expect tool providers to consume it because:

- It is formally valid from the official spec, 
- The Swagger 2 on-line editor is happy with it, 
- A null value is essential in business APIs, 
- And they are big enough to insist

[...]

PR #1957, opened 2019-06-21
OpenAPI v3.1 as a JSON Schema "Vocabulary"

philsturgeon:
As a follow-up to the June 13th TSC call, I am working on showing you what OpenAPI v3.1 on "JSON Schema Draft 8" could look like. The WIP JSON Schema draft added a fantastic new feature called vocabularies, which allow for extension languages like OpenAPI to add their own keywords. 

Basically, instead of the OpenAPI spec listing out every single similarity and difference in its subset superset sideset confusing situation, OpenAPI just extends the JSON Schema core and valaidation vocabs and adds a few keywords of its own. 

This solves the OpenAPI <≠> JSON Schema divergence nicely. The biggest thing here is explaining how exclusiveMinimum and exclusiveMaximum are different from latest JSON Schema, which was done to support BC between v3.0 and v3.1. 

The most controversial part of this is my choice to mark things like `nullable: true` and `example` as deprecated. This can certainly be removed, but I think it would be a good idea to start pushing folks towards one way of doing things instead of two. 

Timeline for removal would essentially be _maybe_ v4.0, or v5.0, or never, but deprecation itself is a good way to move folks in the right direction, and allow tooling to start giving warnings.


versions/3.1.0.md
```diff
@@ -91,7 +91,7 @@ Some examples of possible media type definitions:
   application/vnd.github.v3.patch
 ```
-The HTTP Status Codes are used to indicate the status of the executed operation. 
+The HTTP Status Codes are used to indicate the status of the executed operation.
```
MikeRalphson: @philsturgeon could you change your editor settings to preserve whitespace at the end of lines in markdown files? (In some markdown variants two trailing spaces equals a hard line break, though I don't know if we depend on that anywhere). It makes it difficult to see what you've actually changed. Also affects #1946

handrews: @MikeRalphson in the meantime, on the diff page under the settings menu (the gear thing) there is an "Ignore Whitespace" option that you can turn on which takes out the visual noise.

versions/3.1.0.md
```diff
@@ -2276,64 +2276,64 @@ This object is an extended subset of the [JSON Schema Specification Wright Draft
...
 <a name="schemaXml"></a>xml | [XML Object](#xmlObject) | This MAY be used only on properties schemas. It has no effect on root schemas. Adds additional metadata to describe the XML representation of this property.
-<a name="schemaExternalDocs"></a>externalDocs | [External Documentation Object](#externalDocumentationObject) | Additional external documentation for this schema. 
+<a name="schemaExternalDocs"></a>externalDocs | [External Documentation Object](#externalDocumentationObject) | Additional external documentation for this schema.
 <a name="schemaExample"></a>example | Any | A free-form property to include an example of an instance for this schema. To represent examples that cannot be naturally represented in JSON or YAML, a string value can be used to contain the example with escaping where necessary.
-<a name="schemaDeprecated"></a> deprecated | `boolean` | Specifies that a schema is deprecated and SHOULD be transitioned out of usage. Default value is `false`.
 
 This object MAY be extended with [Specification Extensions](#specificationExtensions).
```
handrews: Are we still restricting unrecognized keywords to the Specification Extensions syntax?  While this doesn't make sense in terms of vocabularies, it is totally valid for the OAS meta-schema to continue to use `patternProperties` to enforce this.  But it will break anyone who is trying to share schemas that also use other vocabularies e.g. for home-grown tooling that do not begin with `x-`.

schemas/v3.1/openapi-vocabulary.json
```diff
@@ -0,0 +1,1409 @@
...
+  "$vocabulary": {
+    "https://json-schema.org/draft/2019-WIP/vocab/core": true,
+    "https://json-schema.org/draft/2019-WIP/vocab/applicator": true,
+    "https://json-schema.org/draft/2019-WIP/vocab/validation": true,
+    "https://json-schema.org/draft/2019-WIP/vocab/meta-data": true,
+    "https://json-schema.org/draft/2019-WIP/vocab/format": true,
+    "https://json-schema.org/draft/2019-WIP/vocab/content": true,
+    "https://specs.openapis.org/oas/3.1/vocab/openapi": true
+  },
+  "$recursiveAnchor": true,
```
handrews: This is not doing what you want it to do- the OAS file schema is not a recursive extension of a regular meta-schema.  It is a schema for a totally separate document format that embeds extended schemas.  Take a look at http://json-schema.org/work-in-progress/WIP-links.json for an example of how this sort of thing should work.

As noted in the comment on the vocabulary meta-schema you need three documents:

* A schema for the vocabulary (which can work as a meta-schema on its own, but is not very useful that way)
* A meta-schema for the OAS Schema Object (essentially the same as the standard JSON Schema `schema`, but with the OAS vocabulary added as the 7th entry in `$vocabularies`, and the OAS vocabulary schema as the 7th entry in the `allOf`)
* A schema for the OAS file format, which references the meta-schema for the OAS Schema Object

The OAS Schema Object meta-schema needs to be a standalone document because of how `$recursiveAnchor` and `$recursiveRef` work.  `$recursiveRef` can only target root schemas (entire schema documents) because otherwise it would be incredibly difficult to implement.

handrews: Also, if we're moving to the new draft for the OAS meta-schema, I would strongly recommend starting from https://gist.github.com/handrews/6dfebd56ef97328f9e4dc7a47a1e8bc7 which is more concise, although not tested and not necessarily correct.

schemas/v3.1/meta-openapi.json
```diff
@@ -0,0 +1,113 @@
...
+          "$ref": "#/$defs/XML"
+        }
+      }
+    }
+  ],
+  "patternProperties": {
+    "^x-": {
+    }
+  },
+  "additionalProperties": false,
```
handrews: Needs to be `unevaluatedProperties` or it won't work.

schemas/v3.1/meta-openapi.json
```diff
@@ -0,0 +1,113 @@
...
+  "$recursiveAnchor": true,
+  "title": "OpenAPI Vocabulary Schema",
+  "type": ["object", "boolean"],
+  "allOf": [
+    {
+      "$ref": "http://json-schema.org/draft/2019-WIP/schema"
+    },
+    {
+      "properties": {
+        "type": {
```
handrews: You don't need to redefine this, as far as I can tell.

schemas/v3.1/meta-openapi.json
```diff
@@ -0,0 +1,113 @@
...
+  "$id": "https://spec.openapis.org/oas/3.1/meta",
+  "$vocabulary": {
+    "https://specs.openapis.org/oas/3.1/vocab/openapi": true
+  },
+  "$recursiveAnchor": true,
+  "title": "OpenAPI Vocabulary Schema",
+  "type": ["object", "boolean"],
+  "allOf": [
+    {
+      "$ref": "http://json-schema.org/draft/2019-WIP/schema"
```
handrews: You don't actually want this here- a vocabulary meta-schema should only describe itself.  You still want `recursiveAnchor` the way you have it a few lines up because that ensures that this vocabulary meta-schema can be combined and extended properly, but you don't want the `allOf`.  Note that there is no `allOf` in http://json-schema.org/work-in-progress/meta/WIP-validation.json

You will need a separate meta-schema that combines this vocabulary with the standard ones and serves as the meta-schema for OAS Schema Objects.  Then you will reference that meta-schema wherever you want to allow schemas that use all of the combined vocabularies.  That separate meta-schema will look almost identical to http://json-schema.org/work-in-progress/WIP-schema.json except with the OAS vocabulary added to `$vocabularies`.  So you'll have seven entries in `$vocabularies` instead of six.

versions/3.1.0.md
```diff
@@ -2276,64 +2276,64 @@ This object is an extended subset of the [JSON Schema Specification Wright Draft
...
-Additional properties defined by the JSON Schema specification that are not mentioned here are strictly unsupported.
-
-Other than the JSON Schema subset fields, the following fields MAY be used for further schema documentation:
+In addition to the JSON Schema fields, the following OpenAPI vocabulary fields MAY be used for further schema documentation:
 
 ##### Fixed Fields
 Field Name | Type | Description
 ---|:---:|---
-<a name="schemaNullable"></a>nullable | `boolean` | Allows sending a `null` value for the defined schema. Default value is `false`.
+<a name="schemaNullable"></a>nullable | `boolean` | Allows sending a `null` value for the defined schema. Default value is `false`. Deprecated. Move towards using `"type": "null"` & type arrays , e.g: `"type": ["string", "null"]`.
```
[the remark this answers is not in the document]
tedepstein: +1 for deprecating `nullable`!

Also +1 for changing its default as suggested.

But I don't think we can deprecate it until we allow at least the simple form of type arrays that you proposed in #1389:
> * `type` MUST be a string or a two-element array with unique items
> * if an array, one of the unique items MUST be `"null"`

... or just add support for type arrays, without restricting it to the simplified two-element form. I'm not sure there's really any benefit to that restriction. 

If we can make these changes in 3.1, then the proposed JSON Schema vocabulary can assume that change will already be in effect.

versions/3.1.0.md
```diff
@@ -2276,64 +2276,64 @@ This object is an extended subset of the [JSON Schema Specification Wright Draft
...
+- required
+- allOf / oneOf / andOf
 - multipleOf
 - maximum
 - exclusiveMaximum
 - minimum
 - exclusiveMinimum
-- maxLength
-- minLength
+- maxLength / minLength
```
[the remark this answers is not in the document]
philsturgeon: This list got big and I saved some lines. 🤷‍♂

versions/3.1.0.md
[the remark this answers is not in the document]
philsturgeon: @tedepstein "or just add support for type arrays" this PR does that. :)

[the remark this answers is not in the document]
philsturgeon: @handrews yeah I need to write a thing in the spec which says "nullable: true" should be treated the same as a null appearing in the array, instead of leaving it as default false. they're certainly not meant to contradict each other.

schemas/v3.1/meta-openapi.json
philsturgeon: Greaat! off of memory I thought openapi was more restrictive but the list matches JSON Schema, so away it goes!

schemas/v3.1/meta/oas.json
```diff
@@ -0,0 +1,1424 @@
...
+        "type": "string"
+      }
+    },
+    "patternProperties": {
+      "^x-": {
+      }
+    },
+    "additionalProperties": false
+  },
+  "Contact": {
```
philsturgeon: These definitions are nothing to do with the schema object, which is what this (meta?)-schema file is describing in "properties", should I make another file which is just defs?

versions/3.1.0.md
philsturgeon: I think in the interests of minimal change, right now we just make oas a vocab and don't worry about supporting other custom vocabs and lots of other exciting things the OAS community has never heard of, or asked for. This is a 3.1 remember so better not to change too much. 

With all the chats about traits and overlays and other forms of extensibility, im sure we can make a case for proper vocab support in the future. :)

versions/3.1.0.md
```diff
@@ -2276,64 +2276,64 @@ This object is an extended subset of the [JSON Schema Specification Wright Draft
...
+
+- exclsuiveMinimum - Value can be a numeric value just like JSON Schema, or for backwards compatibility with older drafts it can accept a boolean to act as a modifier to the `minimum` keyword. The numeric value usage is recommended.
+- exclsuiveMaximum - Value can be a numeric value just like JSON Schema, or for backwards compatibility with older drafts it can accept a boolean to act as a modifier to the `minimum` keyword. The numeric value usage is recommended.
 - description - [CommonMark syntax](http://spec.commonmark.org/) MAY be used for rich text representation.
 - format - See [Data Type Formats](#dataTypeFormat) for further details. While relying on JSON Schema's defined formats, the OAS offers a few additional predefined formats.
-- default - The default value represents what would be assumed by the consumer of the input as the value of the schema if one is not provided. Unlike JSON Schema, the value MUST conform to the defined type for the Schema Object defined at the same level. For example, if `type` is `string`, then `default` can be `"foo"` but cannot be `1`.
```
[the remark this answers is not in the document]
philsturgeon: I'm not sure I follow your suggestion here. I know you don't love it, but could you throw a few more words at what you're trying to say and I can turn that into specific words for the spec once I understand the intention? 

The original sentence makes more sense to me than "documents the required behavior" because its basically "The default value represents something a user might provide as input for this value" which is one way to talk about a default (in OAS land anyway).

philsturgeon:
@MikeRalphson fixed the whitespace! :)

versions/3.1.0.md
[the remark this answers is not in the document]
philsturgeon: Reverted

schemas/v3.1/meta/oas.json
```diff
@@ -0,0 +1,1424 @@
...
+    "xml": {
+      "$ref": "#/$defs/XML"
+    }
+  },
+  "patternProperties": {
+    "^x-": {
+    }
+  },
+  "unevaluatedProperties": false,
+  "$defs": {
```
handrews: While it make sense to have `Discriminator` and `XML`, which are used in this file, defined here, I'm a little unclear on why you put all of the other definitions here.  They are describing other aspects of the OAS format, not parts of the schema, so they don't make sense to me for storing in the vocabulary meta-schema.  I'd put them in `schemas/v3.1/schema.json`

Since OAS is a document format that embeds schemas, the thing that describes the schema format is a meta-schema.  The thing that describes the overall format is not a meta-schema, it's just a schema.  Schemas have a root schema at the.. um... root of the document, which is not how the root object of OAS behaves.

versions/3.1.0.md
```diff
@@ -2276,11 +2276,16 @@ This object is an extended subset of the [JSON Schema Specification Wright Draft
 For more information about the properties, see [JSON Schema Core](https://tools.ietf.org/html/draft-wright-json-schema-00) and [JSON Schema Validation](https://tools.ietf.org/html/draft-wright-json-schema-validation-00).
 Unless stated otherwise, the property definitions follow the JSON Schema.
 
-##### Properties 
+##### Properties
 
-The following properties are taken directly from the JSON Schema definition and follow the same specifications:
+OpenAPI v3.1 is a JSON Schema vocabulary which eetends JSON Schema Core and Validation vocabularies. As such any keyword available for those vocabularies is by definition available in OpenAPI, and will work the exact same way, including but not limited to:
```
cebe:
```suggestion
OpenAPI v3.1 is a JSON Schema vocabulary which extends JSON Schema Core and Validation vocabularies. As such any keyword available for those vocabularies is by definition available in OpenAPI, and will work the exact same way, including but not limited to:
```

versions/3.1.0.md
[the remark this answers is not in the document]
Relequestual: The greater issue here is that the default, as in "if omitted", has a constricting behaviour.
Omitting keywords should not cause additional validation constraints to be imposed, according to JSON Schema principals.

schemas/v3.1/schema.json
```diff
@@ -0,0 +1,60 @@
+{
+  "$schema": "http://json-schema.org/draft/2019-WIP/schema#",
+  "$id": "https://spec.openapis.org/oas/3.1/schema",
+  "$vocabulary": {
```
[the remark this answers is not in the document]
Relequestual: @philsturgeon I have to admit, this was part of why I found reading your PR really confusing.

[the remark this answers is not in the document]
philsturgeon: Fixed!

philsturgeon:
This one got too confusing because I took on too much at once. Instead of worrying about updating the spec AND updating the schema, let's just talk about the spec. This can be revisited later.

PR #1958, opened 2019-06-21
Update v3.1.0-dev from master

philsturgeon:
master and v3.1.0-dev are diverging and its getting a bit out of hand. Can we update this branch, or delete it, or something?


philsturgeon:
I have to keep flipping between branches to see the current state of things in other files and its weird. Why not just.. have branches be up to date? 😕

PR #1959, opened 2019-06-21
fix OIDC url and OAuth2 requirements

AxelNennker:
Signed-off-by: Axel Nennker <axel.nennker@telekom.de>

OAuth2 requires the use of TLS
OpenID Connect requires the use of TLS
Fixed the url to the OpenID Connect standard
Changed all links to external references from http to https
Changed all links to example endpoints from http to https


MikeRalphson:
Many thanks!

Issue #1038, opened 2017-04-12
Clarify or remove language about "conflicts" in Path Item Object $ref property


tedepstein:
The [Path Item Object](https://github.com/OAI/OpenAPI-Specification/blob/3.0.0-rc0/versions/3.0.md#user-content-pathItemObject) documentation for the `$ref` property says:
 
> Allows for an external definition of this path item. The referenced structure MUST be in the format of a Path Item Object. If there are conflicts between the referenced definition and this Path Item's definition, the behavior is undefined.

I don't understand what conflict could arise. I assume that the value of this `$ref` property is a JSON Reference, like all of the other `$ref` properties defined in the spec. In that case, `$ref` should not have any sibling properties, and if it did, those would be ignored according to the JSON Reference spec. 

So, for consistency and clarity, can we remove this language?

darrelmiller:
That $ref is not a `Reference Object` like the other ones.  It is special.  There are historical reasons why this happened.  @webron  explained it to me, but my memory is a sieve, so I forget the details.  Yes it sucks that this is different.  I'm not sure it can be fixed without breaking lots of existing specs.

tedepstein:
Thanks @darrelmiller. 

To clarify: I understand that this is not a Reference Object, as defined in the OpenAPI spec.  But isn't it still a JSON Reference? Or, trying to be more precise:  Shouldn't a Path Item Object having a valid `$ref` property (and no other properties) be treated as a JSON Reference Object, as described in the [draft spec](https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03)?

If so, JSON Reference specifies:

> Any members other than "$ref" in a JSON Reference object SHALL be ignored.

That would seem to preclude any possibility of the Path Item Object having a conflict between an inline definition and a referenced definition. 

I don't see anything in the Path Item Object's `$ref` description saying that it should conform to JSON Reference.  In fact, it doesn't say much at all about the expected value and semantics of the `$ref` property.  It also doesn't assert the requirement that if `$ref` is present, all other Path Item Object properties are disallowed. I just assumed, as I think most users will, that this `$ref` is supposed to _work_ like the other `$ref` properties, even if it's not _defined_ in the same way, via Reference Object.

@webron, maybe you can help clear this up:

1. Would it really break something to add this constraint, and to say that the value of the `$ref` conforms to JSON Reference and JSON Pointer, same as it is in the Reference Object?
[...]

Issue #1575, opened 2018-05-07
Certification program for OAS consuming implementations


tedepstein:
Some organizations considering adoption of OpenAPI have asked what assurances we can offer with regard to interoperability.  
* They want to see evidence that the OpenAPI specification itself is mature enough, and well-enough specified, that tools and libraries can be tested for conformance.  
* They want to see a commitment to interop and specification conformance on the part of OAI as an organization. 
* When considering adoption of specific tools, they want some objective evidence that those tools conform to the OpenAPI specification.

This issue builds on some early discussions in #1520, which was originally scoped to defining a _format_ for the rules and test cases that comprise a test suite. That is one aspect of the program.

This issue is about the first phase of the overall certification program. The TSC has agreed that this first phase should focus on OpenAPI _consumers_, i.e. implementations that take an OpenAPI document as an input.

tedepstein:
Here's a strawman proposal for a milestone plan: 

Open Issue:
https://github.com/OAI/OpenAPI-Specification/issues/1520

**M0. Suggested Next steps:**

- Example YAML file and schema for validation. 
- Provision some resources in the current OAI project, or create a new one. 
    - YAML data document with initial subset of rules and test cases
    - JSON Schema describing the format of the YAML doc.
        - Readme markdown document describing: 
        - Purpose of the conformance test Suite
        - Scope (consumers only for now) 
        - Current status (experimental, draft, comments welcome) 
        - Available resources (YAML data document with rules and test cases)
        - Plans

**M1. Prototype Test Automation Tool**

- Complete a subset of rules and test cases
- Prototype test automation tool to inject each test input & inspect response. 
- Prototype test adapter for one tool, same platform as test driver (e.g. Java) 
- Review with TSC

**M2. Test Automation Tool RFC**

- Evolve subset of rules & test cases
- Develop 1 or 2 additional test adapters for other tools, other platforms. 
- Add to the readme: overview of the certification process:
    - How to start, by self-verifying conformance. 
    - How to raise questions & issues. 
    - How to request review & certification from the TSC (or certifying authority, whoever that is.)
- Package and polish the the readme file and related artifacts (github project under OAI recommended), with a CTA to providers of OpenAPI consuming tools & libraries: 
    - Please look at the conformance test suite and the automated testing tool. 
    - Please try building your own adapter & running the test suite. 
    - Please give feedback!
- Limited publicity and/or targeted outreach to OAS consuming tool providers. 

[...]

Issue #1520, opened 2018-04-02
Formalize conformance test description format


tedepstein:
Discussed on TSC Meeting: April 2, 2018:

Each conformance tests in a suite might have a descriptive structure, something like this: 
* Rule description (human-readable)
* Reference to the relevant part of the spec. 
* Example input
* Expected result (valid, invalid) 
* Notes (optional) 

OpenAPI implementations, including editors, generators, documentation formats, etc., are expected to validate the content, correctly detect errors (without false-positives on valid content), expose and handle those errors in whatever way is appropriate for that implementation. 

Some goals, and possibly non-goals, pending discussion:
* Create a definitive conformance test suite, with all of these test cases. 
* Use a uniform structure for the test cases in the suite, something like the preceding bullet list. 
* Make the rule structure machine-_readable_. 
* Make the rules machine-_executable_ by adding executable representations of the rule, which could take different forms: 
    * Where the rule can be expressed in JSON Schema, include it in a schema, which can be designed to be the standard schema for OpenAPI, or could be given a more limited scope of use, as an informative addition to the conformance test suite. 
    * In other cases, or possibly in all cases, state the rule using an expression language, preferably one that can be embedded or interpreted in different programming languages.

darrelmiller:
@tedepstein Looks great to me.  I'm going to add a note to tomorrow's agenda to discuss what next steps we can take.  If you are able to make the meeting then it would be great to hear your thoughts and how we can progress.

cmheazel:
The Open Geospatial Consortium (OGC) runs a compliance test program at http://cite.opengeospatial.org/.  This program is supported by the Team Engine test server at https://github.com/opengeospatial/teamengine.  Team Engine executes test scripts written in TestNG (http://testng.org/doc/index.html).  As part of their Testbed 14 initiative, the OGC will develop compliance tests for their OpenAPI based WFS 3.0 spec.  This work may provide us with most of what we need for a general purpose OpenAPI compliance test.  Any interest?

tedepstein:
@darrelmiller  and @cmheazel, sorry to have missed your earlier comments.  My Gmail filter was a bit overzealous, and I just corrected that.  

I'll be on Monday's call to discuss next steps, and I'll take a look at what the OGC has done.  Thanks!

PR #1964, opened 2019-06-28
Explain unclear semantics of property `$ref` in Path Item Object

cebe:
Currently, as explained in https://github.com/OAI/OpenAPI-Specification/issues/1038#issuecomment-295594246 the description of `$ref` in [Path Item Object](https://github.com/OAI/OpenAPI-Specification/blob/3.0.2/versions/3.0.2.md#pathItemObject) is unclear about the semantics behind it. I took the explaination from issue #1038 to make it more clear.

Hope I got the branching right, if not let me know which branch to send the PR to, I'll fix it then.

### DIFF
--- a/versions/3.0.3.md
+++ b/versions/3.0.3.md
@@ -732,7 +732,7 @@ The path itself is still exposed to the documentation viewer but they will not k
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="pathItemRef"></a>$ref | `string` | Allows for an external definition of this path item. The referenced structure MUST be in the format of a [Path Item Object](#pathItemObject). A Path Item Object containing a `$ref` property is **not** a [Reference Object](#referenceObject). The properties of the referenced structure are merged with the local Path Item Object. If the same property exists in both, the referenced structure and the local one, this is a conflict. In this case the behavior is *undefined*.
+<a name="pathItemRef"></a>$ref | `string` | Allows for an external definition of this path item. The referenced structure MUST be in the format of a [Path Item Object](#pathItemObject).  In case a Path Item Object field appears both in the defined object and the referenced object, the behavior is undefined.
 <a name="pathItemSummary"></a>summary| `string` | An optional, string summary, intended to apply to all operations in this path.
 <a name="pathItemDescription"></a>description | `string` | An optional, string description, intended to apply to all operations in this path. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation.
 <a name="pathItemGet"></a>get | [Operation Object](#operationObject) | A definition of a GET operation on this path.
--- a/versions/3.0.3.md
+++ b/versions/3.0.3.md
@@ -732,7 +732,7 @@ The path itself is still exposed to the documentation viewer but they will not k
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="pathItemRef"></a>$ref | `string` | Allows for an external definition of this path item. The referenced structure MUST be in the format of a [Path Item Object](#pathItemObject). If there are conflicts between the referenced definition and this Path Item's definition, the behavior is *undefined*.
+<a name="pathItemRef"></a>$ref | `string` | Allows for an external definition of this path item. The referenced structure MUST be in the format of a [Path Item Object](#pathItemObject). A Path Item Object containing a `$ref` property is **not** a [Reference Object](#referenceObject). The properties of the referenced structure are merged with the local Path Item Object. If the same property exists in both, the referenced structure and the local one, this is a conflict. In this case the behavior is *undefined*.
 <a name="pathItemSummary"></a>summary| `string` | An optional, string summary, intended to apply to all operations in this path.
 <a name="pathItemDescription"></a>description | `string` | An optional, string description, intended to apply to all operations in this path. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation.
 <a name="pathItemGet"></a>get | [Operation Object](#operationObject) | A definition of a GET operation on this path.


versions/3.0.3.md
```diff
@@ -732,7 +732,7 @@ The path itself is still exposed to the documentation viewer but they will not k
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="pathItemRef"></a>$ref | `string` | Allows for an external definition of this path item. The referenced structure MUST be in the format of a [Path Item Object](#pathItemObject). If there are conflicts between the referenced definition and this Path Item's definition, the behavior is *undefined*.
+<a name="pathItemRef"></a>$ref | `string` | Allows for an external definition of this path item. The referenced structure MUST be in the format of a [Path Item Object](#pathItemObject). A Path Item Object containing a `$ref` property is **not** a [Reference Object](#referenceObject). The properties of the referenced structure are merged with the local Path Item Object. If the same property exists in both, the referenced structure and the local one, this is a conflict. In this case the behavior is *undefined*.
```
cebe:
```suggestion
<a name="pathItemRef"></a>$ref | `string` | Allows for an external definition of this path item. The referenced structure MUST be in the format of a [Path Item Object](#pathItemObject). A Path Item Object containing a `$ref` property is **not** a [Reference Object](#referenceObject) but the semantics of the `$ref` property are the same. The properties of the referenced structure are merged with the local Path Item Object. If the same property exists in both, the referenced structure and the local one, this is a conflict. In this case the behavior is *undefined*.
```

cebe: Hm... GitHub is unable to highlight the change in my suggestion above, here is a better diff:

```diff
-A Path Item Object containing a `$ref` property is **not** a [Reference Object](#referenceObject).
+A Path Item Object containing a `$ref` property is **not** a [Reference Object](#referenceObject) but the semantics of the `$ref` property are the same.
```

tedepstein:
@cebe, thanks for picking this up where the issue left off. :-)

I have a few suggested changes for precision and alignment with standard terminology:

----
Allows for an external definition of this path item. A Path Item Object containing a `$ref` property is not a [Reference Object](#referenceObject), but the semantics of the `$ref` property value are the same. The `$ref` value MUST be a URI that resolves to a [Path Item Object](#pathItemObject). The properties of the referenced object are merged into the referring Path Item Object. If the same property exists in both the referenced object and the referring object, this is a conflict. In this case, the behavior is *undefined*. Applications that do not define their own resolution behavior SHOULD report such conflicts as a warning or error condition.

----

Notes: 

* I had some hesitation about saying that the referenced object MUST be a Path Item Object, because it could be a partial object, meaning it doesn't have all of the required properties of a valid Path Item Object. But in fact, Path Item Object doesn't have any required values. So we can say that the `$ref` URI MUST resolve to a Path Item Object, meaning an object that _does not_ contain any set of properties allowed in a Path Item Object. This seemed to be the cleanest way to state it. 
* I added the last sentence as an incremental step towards better defining validation behavior. This is a core part of conformance testing, discussed in #1520 and #1575. The OpenAPI spec uses the term "undefined" in a few places, but this is the only instance of a property definition with an explicitly undefined value space. I assume that "undefined" allows applications to define their own behavior, whereas a pure error condition would not. So I'm trying to clarify how consuming applications should behave in these cases:
* Either define your own interpretation of the undefined value space, or report it as an error or warning. 
* The choice of error or warning severity depends on the consuming application. If the consumer is a validator or linter, a warning might be the most appropriate default severity, because you don't want to block a downstream code generator or documentation format that actually defines useful behavior for these cases. If the consuming application is a code generator or documentation format that _does not_ provide its own interpretation of the undefined value space, error severity might be more appropriate.

cebe:
Thanks for the detailed feedback!

I agree with the change about referencing a Path Item Object, but:

> Either define your own interpretation of the undefined value space

I'd like to avoid undefined behavior and would prefer to mark the situation as an error condition explicitly. Allowing different tools to handle the situation differently will definitively cause some trouble. Making it easier to implement the spec without ambiguation will prevent tooling implementors and also users from getting into weird situations.

ArtonaDigital:
What do you offer I’m pretty straight forward with what I’m looking for
Do you want to be part of it so let’s put numbers and start

Get Outlook for iOS<https://aka.ms/o0ukef>

________________________________

tedepstein:
>> Either define your own interpretation of the undefined value space
>
> I'd like to avoid undefined behavior and would prefer to mark the situation as an error condition explicitly. Allowing different tools to handle the situation differently will definitively cause some trouble. Making it easier to implement the spec without ambiguation will prevent tooling implementors and also users from getting into weird situations.

I agree, but I assume there is some historical reason why the spec says "undefined" instead of making it an error condition. I think we need some clarification from the TSC.

philsturgeon:
If something changes from "undefined" to "definite nope" then it should probably go in 3.1.0, because its kinda maybe sorta breaking but only if you're relying on undefined behavior which is kinda your own fault? Seems like the wiggle room is there to make it happen in v3.1.

cebe:
Note that my adjustment does not change the definition, so the adjustment made here can go into 3.0, we could then discuss a change of defined behavior for 3.1.

tedepstein:
> Note that my adjustment does not change the definition, so the adjustment made here can go into 3.0, we could then discuss a change of defined behavior for 3.1.

Right. What we've tried to do here is retain "undefined" so as not to introduce a breaking change, but add a bit of guidance as to how implementations should handle this undefined case. I think that could go into a 3.0.x patch release without any issues. 

Disallowing duplicate properties (or conflicting, or overlapping properties) would be a breaking change, so should probably go into 4.0 rather than 3.1. And before we suggest this breaking change, I'm hoping the TSC, or someone, can provide the history of this, explaining why the behavior was left undefined. Maybe there's a good reason for it, and in that case, we might want to define the resolution rules rather than disallowing the duplicate properties.

tedepstein:
@cebe , there might be some confusion because your PR doesn't incorporate my [suggested change here](#issuecomment-506972221). If you like the suggested revisions, you might want to commit those to your `cebe:patch-1` branch so we're all talking about the same thing. :-)

MikeRalphson:
@tedepstein I would prefer to drop your final proposed sentence:

> Applications that do not define their own resolution behavior SHOULD report such conflicts as a warning or error condition.

As declaring the behaviour as undefined and the defining what applications (implementations) **SHOULD** do appears to be in conflict. From a purist point of view, I would like the specification to declare its own requirements, but never impose behaviour (such as reporting errors or warnings) on implementations.

tedepstein:
@MikeRalphson , I'm OK with removing that last sentence. 

Please tell me if "undefined" means the same thing to you as it does to me: 

* We're acknowledging that the OpenAPI spec has consciously declined to specify a meaning or prescribe behavior in this case.
* We _could_ choose to disallow these conflicts, or specify semantics (e.g. properties defined in the referring context take precedence) in a later release. And we could even argue that the newly defined semantics are an addition, not a breaking change, because we've already blocked off this space as undefined.

Common-sense guidance, implied but not stated:
* It's safest to avoid this undefined value space in your OpenAPI documents. 
* Applications _may_ define behaviors, and users _may_ rely on those behaviors, at their own risk. We won't stop you from doing your own thing in an undefined value space, as long as you understand that your special semantics might be clobbered by a subsequent spec release. 

Is that right?

MikeRalphson:
> Please tell me if "undefined" means the same thing to you as it does to me:

@tedepstein I couldn't swear that your definition covers every case of undefined behaviour, but I don't see anything to disagree with in what you have there.

handrews:
@tedepstein @earth2marsh this gets into the objections @whitlockjc has around three different `$ref` behaviors in OAS 3.1, which @webron and I were going to take a look at.

In order to make 3.0.3 compatible with the most likely approach in 3.1, I would recommend slightly different wording.  I find "merge" misleading here, because it only works one level down and become undefined if there's s conflict.  Which I guess is technically a merge but one that only works in a small set of cases.

As I think @MikeRalphson noted on the call, we should look at this a bit more like JSON schema's annotations in terms of defining behavior.  Possible wording for that might replace 

> The properties of the referenced structure are merged with the local Path Item Object.
>  If the same property exists in both, the referenced structure and the local one, this is a conflict. In this case the behavior is *undefined*.

to (changed words in bold italics):

> The properties of the referenced structure are **_examined alongside those in_** the local Path Item Object.
>  If the same property exists in both, the referenced structure and the local one, this is a conflict. In this case the behavior is *undefined*.

If there is a concern over "examined" not being clear enough, we could add a third sentence:

> **_The referenced structure's properties that do not exist in the local structure are used as if they were present in the local structure._**

or something like that.

This avoids terminology that sounds like it might combine properties in some deeper way, which is what I think of with the word "merge" (and is what the discarded `$merge` JSON Schema proposal did, so anyone who is familiar with that will probably think of it similarly).

webron:
Working on this as part of the 3.0.3 release, so we're a bit more constrained in time.

Some history - in 2.0 we added the ability to $ref a Path Item Object within a Path Item Object to allow reusability (putting aside whether it makes sense or not). Two issues were involved here. The first, we did not allow users to define reusable PIOs within an API definition (should be solved in 3.1.0). The second, is that we used the construct of $ref without fully understanding the restrictions of JSON References.

Since we did not realize back then that JSON References do not allow sibling properties, we had no idea that it would be technically not possible to get to the point where there's a conflict. And so, we decided to allow a 'merge-like' behavior, and since we didn't want to deal with the case of duplication, we just said that it's undefined (instead of saying one takes precedence over the other, for example). Little did we know what this would mean for standard JSON Reference implementations out there.

While we did realize the problem before working on 3.0, actually dealing with the issue slipped through the cracks and the same definition remained. These things happen, and it means the behavior stays the same as in 2.0.

As for the change in 3.03 - I'm not a fan of the rewording. I find that specifying it is not the Reference Object is not needed. Nowhere does it say that it is. Claiming that because the field is called `$ref` people should assume it's the Reference Object is false, because we also have a field `type` in Security Scheme Object, and nobody confuses it with the Schema Object's `type`. The spec, in this case, is actually quite concise in the fact that it is _not_ a JSON Reference Object, because the type of the field is _not_ a JSON Reference Object.

I agree there's room to clarify the behavior of the $ref. I would suggest we change
``
 If there are conflicts between the referenced definition and this Path Item's definition, the behavior is undefined.
``
to
``
In case a Path Item Object field appears both in the defined object and the referenced object, the. behavior is undefined.
``

@handrews - as for rewording/resolving things in 3.1, that should come in 3.1 and not here. As long as we don't use the latest JSON Schema draft in 3.0.3, it's better we steer clear of using similar terminology.

handrews:
@webron my concern is to make sure we don't make things harder to resolve in 3.1.

Currently, the word "merge" does not appear in 3.0.2.  If it did, I would not argue to remove it.  But I am going to push very hard to avoid adding it.  There is no single word involved in JSON Schema or JSON Reference that has been the subject of more fights and assumptions than "merge."

I will be happy with literally any other word.  "examine", "combine", "zipper", "splat", seriously ***anything*** but "merge."  The way people make assumptions around the word "merge" and connect it to past proposals elsewhere will make the change from 3.0.3 to 3.1.0 in this specific area much, much harder.

webron:
Sure, but the change I recommend making doesn't have the term "merge" in it either. :)

handrews:
@webron oh I assumed you meant changing

> The properties of the referenced structure are merged with the local Path Item Object.
If the same property exists in both, the referenced structure and the local one, this is a conflict. In this case the behavior is undefined.

to

> The properties of the referenced structure are merged with the local Path Item Object.  In case a Path Item Object field appears both in the defined object and the referenced object, the. behavior is undefined.

leaving the sentence with "merge" intact.

handrews:
If you want to cut that first sentence and just have

> In case a Path Item Object field appears both in the defined object and the referenced object, the. behavior is undefined.

then I am totally fine with that 😁

webron:
To clarify, I think what I suggested was the only change we might want to do. I find the rest not needed.

MikeRalphson:
> To clarify, I think what I suggested was the only change we might want to do. I find the rest not needed.

:+1: I think this is both a positive change and all we need to do for v3.0.3. Hopefully all the `$ref` language will be brought as far into line as possible in v3.1

versions/3.0.3.md
webron:
```suggestion
<a name="pathItemRef"></a>$ref | `string` | Allows for an external definition of this path item. The referenced structure MUST be in the format of a [Path Item Object](#pathItemObject).  In case a Path Item Object field appears both in the defined object and the referenced object, the behavior is undefined.
```

Adding the suggestion based on the discussion so we can merge the PR and release 3.0.3.

PR #1974, opened 2019-07-17
Proposal: Support Webhooks as well as Callbacks

lornajane:
I work at Nexmo and we have a bunch of HTTP APIs, some of them offer webhooks where the server will send payloads to the client as well as the client making API requests to the server. We're looking for a solution that allows us to document and use other tools with our APIs, covering both outgoing and incoming requests.

The OpenAPI `callbacks` functionality _nearly_ supports what we need, but these APIs have the subscribe step completely out of band (you set it up which URL to send a webhook to when a particular event occurs either on a different API or via a web interface). I also speak about OpenAPI at conferences and I've had this question a few times. So I offer this proposal to build on the callbacks feature and add support for webhooks (with no parent path) as well.

### DIFF
--- a/proposals/002_Webhooks.md
+++ b/proposals/002_Webhooks.md
@@ -1,0 +1,192 @@
+# Webhooks
+
+
+## Metadata
+
+|Tag |Value |
+|---- | ---------------- |
+|Proposal |[002_Webhooks](https://github.com/OAI/OpenAPI-Specification/tree/master/proposals/002_webhooks.md)|
+|Authors|[Lorna Mitchell](https://github.com/lornajane)|
+|Review Manager |TBD |
+|Status |Proposal|
+|Issues |[#1968](https://github.com/OAI/OpenAPI-Specification/issues/1968)|
+
+## Change Log
+
+|Date |Responsible Party |Description |
+|---- | ---------------- | ---------- |
+| 17th July 2019 | Lorna Mitchell | Initial draft |
+
+## Introduction
+
+Modern APIs often consist of two-way API traffic, but only outgoing API calls or incoming callbacks directly related to an outgoing API call are currently supported by OpenAPI. Incoming HTTP reqests ("webhooks") cannot be described in the current version of OpenAPI.
+
+## Motivation
+
+OpenAPI supports a `callback` element, where the result of an API call is delivered at some later time as an incoming HTTP request to a nominated URL. However it does not support webhooks, where events arrive as an incoming HTTP request but the configuration of these requests was arranged outside of the scope of the API, e.g. on a website.
+
+For example: at Nexmo we have an SMS API (the docs are here: <https://developer.nexmo.com/api/sms> and the source spec here: <https://github.com/Nexmo/api-specification/blob/master/definitions/sms.yml>). It supports:
+
+* sending an SMS (an outgoing API call, currently supported)
+* receiving a delivery receipt when you just sent an SMS (callback, currently supported)
+* receiving an incoming SMS (webhook, not currently supported)
+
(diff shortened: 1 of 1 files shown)


pvcarrera:
Thanks for opening this issue @lornajane 👍 

I'm also finding it hard to describe webhooks with the OpenAPI specification. I'm currently trying out two workarounds:

- Defining the webhooks endpoints as normal paths and then adding tags to indicate/document the request's direction (outgoing | incoming)
- Using variables for the webhooks path names:
```
paths:
  /{$yourResource}:
    post:
    .
    .
```
For now, it looks like the first approach is easier to undertand for our clients and it also looks nicer on the documentation tools we use (Stoplight)

proposals/002_Webhooks.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Change Log
+
+|Date |Responsible Party |Description |
+|---- | ---------------- | ---------- |
+| 17th July 2019 | Lorna Mitchell | Initial draft |
+
+## Introduction
+
+Modern APIs often consist of two-way API traffic, but only outgoing API calls or incoming callbacks directly related to an outgoing API call are currently supported by OpenAPI. Incoming HTTP reqests ("webhooks") cannot be described in the current version of OpenAPI.
```
earth2marsh: I stumbled over `incoming` and `outgoing`. Consider `push` or another word to describe the relationship?

proposals/002_Webhooks.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Motivation
+
+OpenAPI supports a `callback` element, where the result of an API call is delivered at some later time as an incoming HTTP request to a nominated URL. However it does not support webhooks, where events arrive as an incoming HTTP request but the configuration of these requests was arranged outside of the scope of the API, e.g. on a website.
+
+For example: at Nexmo we have an SMS API (the docs are here: <https://developer.nexmo.com/api/sms> and the source spec here: <https://github.com/Nexmo/api-specification/blob/master/definitions/sms.yml>). It supports:
+
+* sending an SMS (an outgoing API call, currently supported)
+* receiving a delivery receipt when you just sent an SMS (callback, currently supported)
+* receiving an incoming SMS (webhook, not currently supported)
```
earth2marsh: It may just be that I'm unfamiliar with the specific API here, but I'm missing why the webhook cannot be described by the `callback` object? IIRC, we intentionally declared the subscription mechanism out of band, so an expression substitutes for the unknown path that is necessarily dynamic.

proposals/002_Webhooks.md
```diff
@@ -0,0 +1,192 @@
...
+
+```
+
+## Backwards compatibility
+
+Adding a new top-level entry is not something to take lightly, however hopefully most tools will simply ignore what they weren't expecting and continue to operate on the parts of the spec they do understand until their functionality catches up with the spec change.
+
+## Alternatives considered
+
+Another option is to add a special `path` that could contain the various webhooks using the exisiting `callback` syntax but existing tools which aren't expecting this special value may not handle it well, so this option was discounted.
```
earth2marsh: Agree that `paths` is fraught, since the paths for a webhook handler are necessarily unknown and will vary. I agree that it's worth considering a new element, effectively a peer of `paths` that describe emitting (webhooks) rather than listening (paths). Thanks for the proposal.

proposals/002_Webhooks.md
nogates: `inbound` & `outbound` ?

lornajane: Thanks! I ended up expanding this a bit to try to be more accurate and clear.

PR #1981, opened 2019-07-20
Update petstore-expanded.yaml

Grinseteddy:
Only a small change in the description getPetById

### DIFF
--- a/examples/v3.0/petstore-expanded.yaml
+++ b/examples/v3.0/petstore-expanded.yaml
@@ -79,7 +79,8 @@ paths:
                 $ref: '#/components/schemas/Error'
   /pets/{id}:
     get:
-      description: Returns a user based on a single ID, if the user does not have access to the pet
+      description: Returns a pet based on a single ID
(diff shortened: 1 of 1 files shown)


examples/v3.0/petstore-expanded.yaml
```diff
@@ -79,7 +79,8 @@ paths:
                 $ref: '#/components/schemas/Error'
   /pets/{id}:
     get:
-      description: Returns a user based on a single ID, if the user does not have access to the pet
+      description: Returns a pet based on a single ID
+      #Description seems quite right before
```
MikeRalphson: I think it would be confusing to leave a comment such as this here.

Grinseteddy: OK Thank you very much.  Obviously I missed something, Anyhow thank you very much for the really helpful samples.

Grinseteddy:
Misunderstanding of the meaning of the interface by myself.

PR #1988, opened 2019-08-18
fix: rename wrong operationId find pet by id to findPetById

exfly:


MikeRalphson:
Please could you add some context as to why you believe this `operationId` is wrong?

exfly:
other codebase operationID is something like addPet, this wrong operationID is `find pet by id`, so I think this is wrong, isn't it?

MikeRalphson:
> other codebase operationID is something like addPet, this wrong operationID is find pet by id, so I think this is wrong, isn't it?

No, the only restriction on an `operationId` is that it must be unique, there are no constraints on the format of the string.

Tooling should take appropriate steps to be able to use `operationId`s in various casings and spacings.

MikeRalphson:
@OAI/tsc Motion to close?

MikeRalphson:
Closing as not an error.

PR #2014, opened 2019-09-23
Merge OAS 3.0 schema to OAS 3.1

handrews:
Since we have the earlier historical OAS schemas present in this branch, should we have the 3.0 schema as well?

I just noticed this as I was about to try writing up some 3.1 schemas to show how the new vocabulary stuff fits in.  Having the 3.0 schemas present is not a prerequisite for that, it just seemed like an odd omission to me.

Feel free to decline if the omission is intentional, of course!


handrews:
Hm.... not sure what to make of those CI failures.  Maybe that's why this hasn't been done 😛

PR #2021, opened 2019-10-02
Update IMPLEMENTATIONS.md

sugimomoto:
Add CData API Server at Server Implementations.
CData API Server is a commercial API generation application.
An API corresponding to OpenAPI 2.0 and 3.0 is automatically generated. So I added it to the list.

https://www.cdata.com/apiserver/#overview


MikeRalphson:
Thanks for the PR. We are now in the process of moving from `IMPLEMENTATIONS.md` to https://tools.openapis.org - the current implementations list will be removed and a note redirecting readers to the new site added. Instructions on listing your projects are contained in https://github.com/OAI/Tooling#how-can-you-help

PR #2022, opened 2019-10-03
Paraphrasing SemVer Caused Confusion

philsturgeon:
TSC 2019-10-03 Feedback

This sentance was meant to paraphrase semver for clarifty, but ended up suggesting a confusion situation where new things essentially cannot really happen.

@webron has context on this.

### DIFF
--- a/versions/3.0.3.md
+++ b/versions/3.0.3.md
@@ -102,9 +102,9 @@ The OpenAPI Specification is versioned using [Semantic Versioning 2.0.0](https:/
 
 The `major`.`minor` portion of the semver (for example `3.0`) SHALL designate the OAS feature set. Typically, *`.patch`* versions address errors in this document, not the feature set. Tooling which supports OAS 3.0 SHOULD be compatible with all OAS 3.0.\* versions. The patch version SHOULD NOT be considered by tooling, making no distinction between `3.0.0` and `3.0.1` for example.
 
-Each new minor version of the OpenAPI Specification SHALL allow any OpenAPI document that is valid against any previous minor version of the Specification, within the same major version, to be updated to the new Specification version with equivalent semantics. Such an update MUST only require changing the openapi property to the new minor version.
+Each new minor version of the OpenAPI Specification SHALL allow any OpenAPI document that is valid against any previous minor version of the Specification, within the same major version, to be updated to the new Specification version with equivalent semantics. Such an update MUST only require changing the `openapi` property to the new minor version.
 
-For example, a valid OpenAPI 3.0.2 document, upon changing its openapi property to 3.1.0, SHALL be a valid OpenAPI 3.1.0 document, semantically equivalent to the original OpenAPI 3.0.2 document. New minor versions of the OpenAPI Specification MUST be written to ensure this form of backward compatibility.
+For example, a valid OpenAPI 3.0.2 document, upon changing its `openapi` property to `3.1.0`, SHALL be a valid OpenAPI 3.1.0 document, semantically equivalent to the original OpenAPI 3.0.2 document. New minor versions of the OpenAPI Specification MUST be written to ensure this form of backward compatibility.
 
 An OpenAPI document compatible with OAS 3.\*.\* contains a required [`openapi`](#oasVersion) field which designates the semantic version of the OAS that it uses. (OAS 2.0 documents contain a top-level version field named [`swagger`](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#swaggerObject) and value `"2.0"`.)
 
--- a/versions/3.0.3.md
+++ b/versions/3.0.3.md
@@ -102,8 +102,6 @@ The OpenAPI Specification is versioned using [Semantic Versioning 2.0.0](https:/
 
 The `major`.`minor` portion of the semver (for example `3.0`) SHALL designate the OAS feature set. Typically, *`.patch`* versions address errors in this document, not the feature set. Tooling which supports OAS 3.0 SHOULD be compatible with all OAS 3.0.\* versions. The patch version SHOULD NOT be considered by tooling, making no distinction between `3.0.0` and `3.0.1` for example.
 
-Subsequent minor version releases of the OpenAPI Specification (incrementing the `minor` version number) SHOULD NOT interfere with tooling developed to a lower minor version and same major version.  Thus a hypothetical `3.1.0` specification SHOULD be usable with tooling designed for `3.0.0`.
-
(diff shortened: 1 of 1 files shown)


darrelmiller:
Alternate proposal for clarifying semver confusion:

If an OAS definition has the minor version incremented, it MUST still be a valid definition for the new minor version.

e.g. If we have a document that is 3.0 and we change the version to 3.1, then it MUST a valid 3.1 document.

versions/3.0.3.md
```diff
@@ -102,8 +102,6 @@ The OpenAPI Specification is versioned using [Semantic Versioning 2.0.0](https:/
 
 The `major`.`minor` portion of the semver (for example `3.0`) SHALL designate the OAS feature set. Typically, *`.patch`* versions address errors in this document, not the feature set. Tooling which supports OAS 3.0 SHOULD be compatible with all OAS 3.0.\* versions. The patch version SHOULD NOT be considered by tooling, making no distinction between `3.0.0` and `3.0.1` for example.
 
-Subsequent minor version releases of the OpenAPI Specification (incrementing the `minor` version number) SHOULD NOT interfere with tooling developed to a lower minor version and same major version.  Thus a hypothetical `3.1.0` specification SHOULD be usable with tooling designed for `3.0.0`.
```
earth2marsh: If an OAS definition has its minor version incremented, it MUST still be a valid definition for the new minor version.

E.g., if we have a document that is `3.0.2` and we change the version to `3.1.0`, then it MUST be a valid 3.1.0 document.

earth2marsh: (updated @darrelmiller 's text with "minor" changes)

darrelmiller:
@philsturgeon I wish I could agree. Perhaps semver could speak for itself if we were deploying a versioned software component, but all semver says about minor versions is the following:

> Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backwards compatible functionality is introduced to the public API. 

It however doesn't define what "backwards compatible" means.  The statement we are adding here is a concrete test to determine backwards compatibility.

tedepstein:
@darrelmiller wrote: 

> [SemVer] however doesn't define what "backwards compatible" means. The statement we are adding here is a concrete test to determine backwards compatibility.

I agree, we need to clarify what we mean by backward-compatible because a textbook definition of backward compatibility (if there is such a thing) might imply a kind of compatibility that we do not and cannot support. Here's [Wikipedia](https://en.wikipedia.org/wiki/Backward_compatibility):

> Backward compatibility is a property of a system, product, or technology that allows for interoperability with an older legacy system, or with input designed for such a system. 

The language in the current 3.0 spec, deleted in this PR, seems aligned with that definition: 

> Subsequent minor version releases of the OpenAPI Specification (incrementing the minor version number) SHOULD NOT interfere with tooling developed to a lower minor version and same major version. Thus a hypothetical 3.1.0 specification SHOULD be usable with tooling designed for 3.0.0.

The problem is, this isn't true!

Some 3.0 tools, e.g. code generators, are strictly OpenAPI _consumers_, and they _could_ process a 3.1 document as a 3.0 document, ignoring unknown keywords. Whether it's advisable to do so depends partly on what guarantees OpenAPI makes w.r.t. "backward compatibility" as we define it. 

But 3.0 editors, validators, and other OpenAPI _producers_ really can't support OpenAPI 3.1 documents, and probably shouldn't try. 

Tooling for 3.0 _won't_ work with 3.1 because there are new keywords that 3.0 tooling would consider to be errors; and because an `openapi: 3.1` property value would probably cause a well-behaved 3.0 tool to exit with an error. 

> If an OAS definition has its minor version incremented, it MUST still be a valid definition for the new minor version.
> 
> E.g., if we have a document that is 3.0.2 and we change the version to 3.1.0, then it MUST be a valid 3.1.0 document.

This is exactly what I think we mean, intuitively, by backward compatibility. Maybe it's better to state it as a guarantee made by the OpenAPI specification, rather than a requirement on the document: 

> Each new minor version of the OpenAPI Specification SHALL allow any OpenAPI document that is valid against any previous minor version of the Specification, within the same major version, to be updated to the new specification by setting its `openapi` property to the new version.
> 
> E.g., a valid OpenAPI 3.0.2 document, upon changing its `openapi` property to `3.1`, SHALL be a valid OpenAPI 3.1 document. New minor versions of the OpenAPI specification MUST be written to ensure this form of backward compatibility.

   valid under the new Specification upon updating   

If an OAS definition has its minor version incremented, it MUST still be a valid definition for the new minor version.
> 
> E.g., if we have a document that is 3.0.2 and we change the version to 3.1.0, then it MUST be a valid 3.1.0 document.

versions/3.0.3.md
philsturgeon: Done

philsturgeon:
@darrelmiller you win! 🚢 🤡

versions/3.0.3.md
```diff
@@ -102,7 +102,7 @@ The OpenAPI Specification is versioned using [Semantic Versioning 2.0.0](https:/
...
+If an OAS definition has its minor version incremented, it MUST still be a valid definition for the new minor version. E.g., if we have a document that is 3.0.2 and we change the version to 3.1.0, then it MUST be a valid 3.1.0 document.
```
philsturgeon: Is "OAS definition" here consistent wording? Not to be confused with `definitions`? Do we mean `a document` or something?

mkistler: I think we need to make a stronger statement than just
> it MUST be a valid 3.1.0 document.

I think we should also require that it have the same interpretation as as the 3.0.x document.

To illustrate: consider the "required" attribute in a parameter object.  It currently defaults to `false`.  I think that changing this default to `true` would be considered a "breaking change", but it would not affect the "validity" of any 3.0.x API doc that was updated to 3.1.0.

I'm not sure how best to word this in the spec, but perhaps something like:
>  If an OAS definition has its minor version incremented, it MUST still be a valid definition for the new minor version with an equivalent interpretation to the original definition. E.g., if we have a document that is 3.0.2 and we change the version to 3.1.0, then it MUST be a valid 3.1.0 document with an equivalent interpretation to the 3.0.2 definition.

philsturgeon: @mkistler fair enough! @darrelmiller @webron what do you think? Can somebody hit that sweet sweet Suggest button and try to solve this feedback without removing your own meaning?

tedepstein:
Updated language, discussed on the call:

> Each new minor version of the OpenAPI Specification SHALL allow any OpenAPI document that is valid against any previous minor version of the Specification, within the same major version, to be updated to the new Specification version with equivalent semantics. Such an update MUST only require changing the `openapi` property to the new minor version. 
> 
> For example, a valid OpenAPI 3.0.2 document, upon changing its `openapi` property to `3.1.0`, SHALL be a valid OpenAPI 3.1.0 document, semantically equivalent to the original OpenAPI 3.0.2 document. New minor versions of the OpenAPI Specification MUST be written to ensure this form of backward compatibility.

PR #2028, opened 2019-10-11
Add spring-openapi to low-level tooling in IMPLEMENTATIONS.md

jrcodeza:
There is new OpenAPI 3 spec generator which can generate OpenAPI 3 spec from java code (using spring annotations). It is also able to generate java client from OpenAPI 3 spec (with usage of Jackson annotation for inheritance implementation).


MikeRalphson:
Thanks - in the description can we avoid saying it generates OpenAPI specifications or specs (this repo is the one and only OpenAPI Specification), OpenAPI "document" or just OpenAPI is fine.

jrcodeza:
@MikeRalphson  thanks for reviewing. You're right, I adjusted description based on your proposal

PR #2039, opened 2019-10-23
Support for tagging a persona to an API operation.

chainhead:
# Introduction

This feature request (#2040) proposes to assign authorized personae with every API definition. Such a feature will serve as a clean bridge between a business process owner and a developer.

- [Introduction](#introduction)
  - [Development Process](#development-process)
    - [Specification Change Criteria evaluation](#specification-change-criteria-evaluation)
    - [Specification Change Process evaluation](#specification-change-process-evaluation)
  - [Proposal](#proposal)
    - [Personae](#personae)
    - [Specification change](#specification-change)
  - [Files changed](#files-changed)

## Development Process

The [Development Process document](../OpenAPI-Specification/DEVELOPMENT.md) calls out the following evaluations.

### Specification Change Criteria evaluation

* Clarity. _Not applicable_
* Consistency. _Not applicable_
* Necessary functionality. **Yes.**
* Forward-looking designs. _Not applicable_

### Specification Change Process evaluation

* Migration. **Easy.** Ideally, this feature should appear, at least, as an optional item with default value corresponding to the semantics of `any`. This means, APIs in older documents will be treated as if they can be invoked by `any`.
* Tooling. **Known.** Code generation may use popular authentication and authorization schemes such as JSON Web Token (JWT). For example, the `sub` and `aud` keys in JWT could be used to generate a snippet of JWT. See [JWT RFC](https://tools.ietf.org/html/rfc7519). The selection of the persona itself maybe via an organizational directory or an implementation of various personae (e.g. employee, administrator, public, etc.) schemas.
* Visualization. **Not sure.**

## Proposal

This proposal covers the following:

- the scope of personae
- its usage in the specification

### Personae

1. The personae can be either **registered users** or **unregistered users**. For example, registered users could be bank customers and unregistered users could be an entity - human or computer - running a web search.
   1. Multiple registered users may map to a single individual. For example, a bank employee may be a bank customer too.
   2. An API author MIGHT want to discriminate between human and system users.

2. A persona of the type, **registered user** MAY be qualified with a `location` attribute. For example, an employee may have limited access to functions if he/she is on a GPRS network through a corporate owned handheld device, BYOD, laptop, etc. Similarly, the same employee may have different level of access to functions if he/she connects to corporate LAN or WLAN via corporate owned handheld device, laptop, etc.
   1. It is possible that the device (corporate or personal) is configured with different persona in which case the `location` attribute maybe redundant.
   2. Where the `location` attribute is preferred to be used, it will be an array of identifiers. For example, an API maybe accessible from home and office for a particular persona in which case, the values of `location` attribute will be an array of strings e.g. `home` and `office`. For variations of identifiers e.g. location hierarchy, specification extensions may be used.

3. A persona of the type, **unregistered user** MAY be qualified with a `location` attribute to indicate all of internet or regions of internet, etc.

4. The details of the persona attribute will be implementation dependent. This feature request does not propose any particular schema. However, at a minimum, the attribute should carry the name of a persona (e.g. employee, department, contractor, developer, administrator, www) and location (e.g. office, home, field, corporate HQ, users from India, IP CIDR, etc.)

### Specification change

The change to the specification is proposed in two ways:

1. A new `personae` object
   1. Defined as `#/definitions/Personae`
   2. Referred to at `#/definitions/Operation`
2. Vendor extensions through `x-personae-` object.

## Files changed

| File | Remark |
| ---- | ------ |
| `schemas/v3.0/schemas.yaml` | Copied from the `master` branch and edited. |
| `examples/v3.0/personae.yaml` | New file added for an example with personae. |
| `versions/3.1.0.md` | Added reference to `personae` object. |

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ target
 atlassian-ide-plugin.xml
 node_modules/
 Gemfile.lock
+pr.md
--- a/examples/v3.0/personae.yaml
+++ b/examples/v3.0/personae.yaml
@@ -1,0 +1,167 @@
+openapi: "3.0.0"
+info:
+  title: Simple API overview
+  version: v2
+paths:
+  /:
+    get:
+      operationId: listVersionsv2
+      summary: List API versions
+      responses:
+        '200':
+          description: |-
+            200 response
+          content:
+            application/json:
+              examples: 
+                foo:
+                  value: {
+                    "versions": [
+                        {
+                            "status": "CURRENT",
+                            "updated": "2011-01-21T11:33:21Z",
+                            "id": "v2.0",
+                            "links": [
+                                {
+                                    "href": "http://127.0.0.1:8774/v2/",
+                                    "rel": "self"
+                                }
+                            ]
+                        },
+                        {
+                            "status": "EXPERIMENTAL",
+                            "updated": "2013-07-23T11:33:21Z",
+                            "id": "v3.0",
+                            "links": [
+                                {
+                                    "href": "http://127.0.0.1:8774/v3/",
+                                    "rel": "self"
+                                }
+                            ]
+                        }
+                    ]
+                 }
+        '300':
+          description: |-
+            300 response
+          content:
+            application/json: 
+              examples: 
+                foo:
+                  value: |
+                   {
+                    "versions": [
+                          {
+                            "status": "CURRENT",
+                            "updated": "2011-01-21T11:33:21Z",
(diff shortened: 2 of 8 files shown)


darrelmiller:
I would suggest getting some buy-in from the larger community on this feature before attempting to make a spec change.  This feels like something that should start out life as an x- extension to determine viability and adoption before we make the request for tooling vendors to support this.

PR #2042, opened 2019-10-24
The examples keyword is not supported inside schema

philsturgeon:
Relates to https://github.com/stoplightio/prism/issues/722

This example of examples is a mistake.


philsturgeon:
@darrelmiller can we get this up on the agenda for next week or do it kinda sooner before it gets lost in the deluge of other PRs.

earth2marsh:
Good catch, @philsturgeon , and your edits to remove the faulty examples LGTM. 

I have one concern though, which is that while an examples object is inappropriate to appear within a schema, it is appropriate within a parameter object. By removing the erroneous example there is now a void left of having no valid parameter example of the examples object. 

If there's a good reason for why we shouldn't have one of those, then let's keep moving forward with a merge, but otherwise, this would be improved by adding that IMO.

philsturgeon:
@earth2marsh every day is a school day!

image

What does this mean. In the Property Object it says "example of the media type", but... it's a parameter? Is this actually a named objects array like in the Media Type Object? If so, AGH! If not, AGH! ?

earth2marsh:
Huh, that strikes me as odd, for both `example` and `examples`. The feeling of, "WTH were we thinking when we wrote that?" encroaches...

darrelmiller:
@philsturgeon Even parameters can be described using Content objects and therefore they can have media types.  It's a bit weird, but then people insisted on being able to shove objects in the URL.
`text/plain` is a valid media type.  For simpler parameters that just use style and schema, there is wording that says what example and examples should do:

> For more complex scenarios, the content property can define the media type and schema of the parameter. A parameter MUST contain either a schema property, or a content property, but not both. When example or examples are provided in conjunction with the schema object, the example MUST follow the prescribed serialization strategy for the parameter.

versions/3.0.3.md
```diff
@@ -1968,20 +1956,6 @@ requestBody:
...
-In a parameter:
-
-```yaml
-parameters:
-  - name: 'zipCode'
-    in: 'query'
-    schema:
-      type: 'string'
-      format: 'zip-code'
-      examples:
```
darrelmiller: Rather than deleting this, it should be moved up to be under parameter

philsturgeon:
@darrelmiller hey I’m not sure about the words here, it doesn’t make any sense. What do examples have to do with media type? Where is text/plain coming from? Multiple examples are given arbitrary names, nothing to do with media types? What is happening I’m scared.

darrelmiller:
@philsturgeon Media types are a description of how to convey meaning in a series of bytes.  Parameter values in a URL are a series of bytes that follow some kind of format.  The only difference between values in the URL and values in the HTTP payload is the payload is usually bigger....  but not always.  When we moved the location of examples in V3, we changed the key from being a media type identifier to just be any old identifier that can be used to identify a scenario.

versions/3.0.3.md
philsturgeon: Done!

philsturgeon:
Ok so this explains my confusion perfectly, thank you @darrelmiller.

> When we moved the location of examples in V3, we changed the key from being a media type identifier to just be any old identifier that can be used to identify a scenario.

Yeah so examples contain arbitrary values keyed by arbitrary strings, and there is no content negotiation. Great, The only connection between Media Types and Parameters is... examples used to be tied to media types as keys but _they aren't anymore_. That makes me think we should change the wording.

I switched the wording, lmk what you think.

philsturgeon:
@darrelmiller @MikeRalphson can somebody merge this one now it's approved?

philsturgeon:
@MikeRalphson thanks for merging #2054, this is a very similar one: just updating an example.

@OAI/tsc any concerns with this one?

mindstretch23:
@darrelmiller  Sorry guys for this dumb question...I read above very carefully to try and answer but need a little help...

Does this PR mean that we can now use `examples` (plural) within the `components` section (previously unsupported)?

PR #2043, opened 2019-10-25
Host the JSON schema at a URL corresponding to its ID

wbazant:
I think this PR will result in the URL

https://spec.openapis.org/oas/3.0/schema/2019-04-02

containing the document currently hosted under https://raw.githubusercontent.com/OAI/OpenAPI-Specification/master/schemas/v3.0/schema.json.

I think this was the case once already, and I'm not sure why it's not like that currently.

I would like to use a Perl package Mojolicious::Plugin::OpenAPI with the v3 schema. I see it was configured to co-op with v3, based on https://github.com/jhthorsen/mojolicious-plugin-openapi/commit/9acade417f1b284929d4b206c756e664cf542b06 - in May the author of that package seems to have found the schema under the URL corresponding to the schema's ID, because he lets the user use the package with merely "v3" referring to the document.

### DIFF
--- a/oas/3.0/schema/2019-04-02
+++ b/oas/3.0/schema/2019-04-02
@@ -1,0 +1,1654 @@
+{
+  "id": "https://spec.openapis.org/oas/3.0/schema/2019-04-02",
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Validation schema for OpenAPI Specification 3.0.X.",
+  "type": "object",
+  "required": [
+    "openapi",
+    "info",
+    "paths"
+  ],
+  "properties": {
+    "openapi": {
+      "type": "string",
+      "pattern": "^3\\.0\\.\\d(-.+)?$"
+    },
+    "info": {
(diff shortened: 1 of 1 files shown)


handrews:
Note that the JSON Schema draft spec advises against automatically downloading schemas in order to avoid heavy load on sites hosting popular schemas.  This is one reason that the draft spec also notes that schemas are not required to be downloadable from their `$id`.

In general, libraries should package appropriate schemas and use a local copy, unless the schema publisher specifically indicates that downloading is OK.  If published, the cache expiration should be set to something rather high.

wbazant:
Thank you! The logic for not hosting the schema under its ID makes sense, I was expecting the wrong thing.

For my actual problem, I've just found where the schema for Mojolicious::Plugin::OpenAPI actually is: https://github.com/mojolicious/json-validator/blob/master/lib/JSON/Validator/cache/a516498b60c53096b2ce2cd83ebe0abc - bundled into JSON::Validator.

PR #1977, opened 2019-07-18
Update Schema Objects to JSON Schema Draft 2019-09

philsturgeon:
Fixes #333
Fixes #397
Fixes #470
Fixes #1026
Fixes #1238
Fixes #1313
Fixes #1368
Fixes #1389
Fixes #1494
Fixes #1523
Fixes #1666
Fixes #1719
Fixes #1766
Fixes #1869
Fixes #1985

Checklist items from #1766 so we can close.
* [x] @handrews "We should probably make sure none of the formats added in the newer draft conflict with the ones OpenAPI defines."
* [x] @handrews "At most you might want to check if we have any slight incompatibilities for readOnly and writeOnly which were moved/added to the validation spec in [Draft 07]"

Picking up from where https://github.com/OAI/OpenAPI-Specification/pull/1766  left off, this PR is helping OpenAPI v3.1 make a non-breaking change, to catch up with JSON Schema's latest draft. 

Instead of OpenAPI continuing to confuse everyone by being a JSON Schema draft 05 sub-set, and a super-set, with some notable differences to watch out for, it will now be a JSON Schema draft 2019-09 superset. [More info on that here](https://apisyouwonthate.com/blog/openapi-and-json-schema-divergence-part-1).

The goal here is to help people write pure JSON Schema for their data model if they want to, or write OpenAPI-flavoured JSON Schema if they want to, but we are closing the gaps on what these two things mean. The list of differences is now tiny, only 4 extra keywords, with 2 notes about how `exlusiveMinimum` and `exclusiveMaximum` can be used in two ways (in order to maintain BC with OpenAPI v3.0).

## Alternative Solutions

Alternative Schemas are a big topic which needs [more work to be done](https://github.com/OAI/OpenAPI-Specification/issues/1943). We should address the JSON Schema divergence issue to make the v3.x branch a happy place to be, and follow up with Alternative Schemas for v3.2 or v4.0, so our XML using friends can play with XML Schema, and all the other good things.

## Risks

There are concerns that tooling vendors who work with static languages will have a hard time adding some of JSON Schemas newer more dynamic features. I understand this in theory, but most of JSON Schema's new features like if/then/else, and the old "type can be an array" issue, are just [sugar for allOf/anyOf/oneOf](https://github.com/OAI/OpenAPI-Specification/pull/1766#issuecomment-442652805), as can be seen in [wework/json-schema-to-openapi-schema](https://github.com/wework/json-schema-to-openapi-schema). 

[...]

PR #2046, opened 2019-10-28
Resolved undocumented nullable behavior

philsturgeon:
OpenAPI Schema Objects and JSON Schema have a few fundamental differences, and this clears up a bit of confusion about one of them.

Basically, in JSON Schema, a schema of `{}` is totally valid, and when type is missing it essentially means "type can be anything", which would allow `null`. 

With OpenAPI Schema Objects, `{}` should be allowed, but nullable confuses that with its default of false. You'd need to write `{ nullable: true }` at the very least to let a `null` squeak through.

This seems like its breaking, but I dug around and most tooling will fall over if type is missing. 

image

Seeing as some tooling prevents `{}` and some tooling supports it, we have a bit of a nasty situation that needs resolving. This change basically says "nullable adds null to the type that exist and missing type essentially means `type: string|integer|array|etc`... so adding null to that list is fine.


versions/3.0.3.md
```diff
@@ -2326,7 +2326,7 @@ Other than the JSON Schema subset fields, the following fields MAY be used for f
 ##### Fixed Fields
 Field Name | Type | Description
 ---|:---:|---
-<a name="schemaNullable"></a>nullable | `boolean` | Allows sending a `null` value for the defined schema. Default value is `false`.
+<a name="schemaNullable"></a>nullable | `boolean` | Allows sending a `null` value for the defined schema. Default value is `false`.<br><br>A `nullable` value of `true` allows `null` in addition to other specified `type` values only if `nullable` and `type` are both specified within the same schema object. The effect of `"nullable" : true` is limited to its expansion of the set of allowed types, within the scope of its containing schema. An `enum`, `const`, `allOf`, or other keyword can independently prohibit `null` values, effectively overriding `"nullable" : true`.<br><br>A `nullable` value of `false` leaves the specified or default `type` unmodified. It has no effect on the schema object.
```
tedepstein: This is the revised language we agreed to on the TSC call: 

```suggestion
<a name="schemaNullable"></a>nullable | `boolean` | A `true` value expands the allowed type defined by the `type` keyword, only if `type` is explicitly defined within the same Schema Object. Other Schema Object constraints retain their defined behavior, and therefore may disallow the use of `null` as a value. A `false` value leaves the specified or default `type` unmodified. The default value is `false`.
```

tedepstein: @darrelmiller , @webron , @handrews, @philsturgeon , please note the following edits to our edit:

A `true` value ~expands~ **adds `"null"` to** the allowed type ~defined~ **specified** by the `type` keyword, only if `type` is explicitly defined within the same Schema Object. Other Schema Object constraints retain their defined behavior, and therefore may disallow the use of `null` as a value. A `false` value leaves the specified or default `type` unmodified. The default value is `false`.

Somehow in our collaborative real-time editing, we neglected to state that `nullable` adds `"null"` to the set of allowed types. We just said that it "expands" the `type` value, but didn't state the obvious (to us) nature of said expansion.

I have corrected that here and in the proposal in PR #2050.

versions/3.0.3.md
```diff
@@ -2326,7 +2326,7 @@ Other than the JSON Schema subset fields, the following fields MAY be used for f
 ##### Fixed Fields
 Field Name | Type | Description
 ---|:---:|---
-<a name="schemaNullable"></a>nullable | `boolean` | Allows sending a `null` value for the defined schema. Default value is `false`.
+<a name="schemaNullable"></a>nullable | `boolean` | A `true` value adds `"null"` to the allowed type specified by the `type` keyword, only if `type` is explicitly defined within the same Schema Object. Other Schema Object constraints retain their defined behavior, and therefore may disallow the use of `null` as a value. A `false` value leaves the specified or default `type` unmodified. The default value is `false`.
```
philsturgeon: Do we also need to mention and clear up that type being missing is string, integer, null, array, etc?

tedepstein: > Do we also need to mention and clear up that type being missing is string, integer, null, array, etc?

I think that's covered by JSON Schema. `type` is a constraint, and in the absence of a `type` constraint, values of any type are allowed. 

I would not mind stating it explicitly, but TSC folks on the call were pushing to keep the spec minimal and concise. Paraphrasing what was said on the call, people will inevitably get confused and ask questions, and no amount of explanatory verbiage in the spec will prevent that from happening.

handrews: @philsturgeon we covered this in the call.  We want to keep the text minimal, and defer to JSON Schema as much as possible.

tedepstein:
@darrelmiller , this is the PR that we would like to merge for the 3.0.3 patch release. 

I believe we have discussed this exhaustively on TSC calls. The new `nullable` definition language is what we composed in a real-time editing session with @webron, @handrews, and others on a TSC call a few weeks ago. The complete background, motivation, and analysis are documented in the [proposal here](https://github.com/OAI/OpenAPI-Specification/blob/master/proposals/003_Clarify-Nullable.md).

As discussed, please herd the TSC cats to approve and merge.  Thank you!  :-)

handrews:
@darrelmiller my recollection matches @tedepstein's.

I also put wording as close to exactly matching this as possible given `type` arrays in #1977, in case anyone is worried about wether/how this would fit with possible OAS 3.1 changes.

philsturgeon:
giphy

darrelmiller:
I think it is fair to say we have done our due diligence on this particular issue.   Hey @OAI/tsc please review and and sign off so we can get this merged into 3.0.3 so we can clear the path to updating JSON Schema in 3.1.  Ted's proposal provides all the background to this conversation and the reasoning as to why this solution was chosen.

philsturgeon:
Yaaaaay, @darrelmiller could you click that button?

darrelmiller:
I was going to say that we a majority TSC vote, but I re-read the governance rules and we only need that for merging the patch branch to master.  Sooooooo.  Time to click the big green button.

philsturgeon:
Thank you Darrel!!

darrelmiller:
@philsturgeon No, thank you and all the other community folks who put the effort into solving this tricky problem  /cc @tedepstein @handrews

webron:
@philsturgeon, @tedepstein, @handrews - indeed, thank you all for pushing this forward. I love this came to be an elegant solution!

PR #2048, opened 2019-10-30
Improved callback examples

philsturgeon:
Looking at this before I had no idea how callbacks were meant to work, but this example makes it a biiiit clearer. 

Still, could do with some tutorial content here, but this is better (and still compatible).

### DIFF
--- a/versions/3.0.3.md
+++ b/versions/3.0.3.md
@@ -1893,7 +1893,7 @@ $response.header.Location | http://example.org/subscription/1
 
 ##### Callback Object Examples
 
-The following example uses the user provided `queryUrl` query string parameter to define the callback URL.
+The following example uses the user provided `queryUrl` query string parameter to define the callback URL.  This is an example of how to use a callback object to describe a WebHook callback that goes with the subscription operation to enable registering for the WebHook.
 
 ```yaml
 myCallback:
--- a/versions/3.0.3.md
+++ b/versions/3.0.3.md
@@ -1891,12 +1891,29 @@ $request.body#/successUrls/2 | http://clientdomain.com/medium
 $response.header.Location | http://example.org/subscription/1
 
 
-##### Callback Object Example
+##### Callback Object Examples
 
-The following example shows a callback to the URL specified by the `id` and `email` property in the request body.
+The following example uses the user provided `queryUrl` query string parameter to define the callback URL.
 
 ```yaml
-myWebhook:
+myCallback:
+  '{$request.query.queryUrl}':
+    post:
+      requestBody:
+        description: Callback payload
+        content: 
+          'application/json':
+            schema:
+              $ref: '#/components/schemas/SomePayload'
+      responses:
+        '200':
+          description: callback successfully processed
+```
+
+The following example shows a callback where the server is hard-coded, but the query string parameters are populated from the `id` and `email` property in the request body.
+
+```yaml
+transactionCallback:
   'http://notificationServer.com?transactionId={$request.body#/id}&email={$request.body#/email}':
     post:
       requestBody:
@@ -1907,9 +1924,8 @@ myWebhook:
               $ref: '#/components/schemas/SomePayload'
       responses:
         '200':
-          description: webhook successfully processed and no retries will be performed
+          description: callback successfully processed
 ```
-
 
 #### <a name="exampleObject"></a>Example Object
 


versions/3.0.3.md
```diff
@@ -1891,13 +1891,13 @@ $request.body#/successUrls/2 | http://clientdomain.com/medium
...
 
 
-##### Callback Object Example
+##### Callback Object Examples
 
-The following example shows a callback to the URL specified by the `id` and `email` property in the request body.
+The following example uses the user provided `queryUrl` query string parameter to define the callback URL.
 
 ```yaml
-myWebhook:
```
philsturgeon: here we were conflating callbacks and webhooks, which could be confusing with Webhooks getting their own proposal: #1974.

versions/3.0.3.md
```diff
@@ -1891,13 +1891,13 @@ $request.body#/successUrls/2 | http://clientdomain.com/medium
...
-  'http://notificationServer.com?transactionId={$request.body#/id}&email={$request.body#/email}':
+myCallback:
+  '{$request.query.queryUrl}':
```
philsturgeon: Common use case, in fact this one was used above but the example was dropped on the floor for another less common example. Let's have both!

versions/3.0.3.md
darrelmiller: Hence, why I am not a fan of calling the new thing WebHooks.  Callbacks were designed to support Webhooks.

versions/3.0.3.md
darrelmiller: Are you suggesting that for something to be called a webhook the consumer needs to provide the entire URL?

versions/3.0.3.md
philsturgeon: Let's save discussions for naming of other things in the PRs for other things. This is callbacks and we should talk about callbacks and whether callbacks support webhooks nicely or not, we can be a bit less confusing here by sticking to the term callback consistently.

versions/3.0.3.md
philsturgeon: This PR exists solely because the Callbacks section is so confusing to me and the team that we struggled to build tooling around the callback feature. We got there in the end https://github.com/stoplightio/prism/pull/716 but it could have been easier. This PR makes it easier.

I am not talking about webhooks here at all. Note that the only mention of webhook in the PR is in red, not in green. This PR is about examples, and one example is using an entire URL, another example is using parameters to construct a URL.

This example was elluded to earlier in the doc then just dropped on the floor. I'm bringing it back here to visualize what that earlier example might look like.

darrelmiller: Our original intent was that Callback is a generic term that includes WebHooks as a type of CallBack.  I am concerned about trying to make a distinction between Callbacks and Webhooks.  Having an API to subscribe to a webhook should be a normal scenario.

philsturgeon: I hear you but this PR does not make a distinction between callbacks and webooks. This seems like a conversation that should be had with the webook proposal, not on this PR, which is just improving examples.

philsturgeon: I am not disagreeing with your opinion just saying you are putting it in the wrong place. 😅

darrelmiller: We should probably have this conversation in a different medium.  However, I don't think this is the wrong place to have the conversation.  Having an example of how to use a Callback Object to describe a webhook that has a subscription API seems like something value to me.

philsturgeon: Possibly! I dunno if you're saying the two words are interchangeable then who cares if they're interchangeable? It seemed weird to suddenly throw the word Webhook in there, so as part of my general cleanup I fixed that weird inconsistency. I dont like arbitrarily introducing words unexplained.

With this change, we are not randomly throwing a word in unexplained, and if the proposal we have already merged to master takes off then we don't have to worry about a collision. If that proposal does not take off, then no harm no foul, because the terms are interchangeable and so it doesn't matter at all?

lornajane:
Personally (sorry this was on my list but I have been travelling and slow to review!) I think since the incoming-request-after-previous-API-call pattern is called "callback", it is nice to use the same word throughout the spec. The additional example is useful too, so IMO this change is a net gain.

versions/3.0.3.md
darrelmiller: I don't see them as interchangeable.  In my head, callback is a more generic term.  Webhook is a special case of a callback.  Callbacks include scenarios where I call operation A and at some point later I get a HTTP call back with the response.   A webhook is a pattern where you register to notify when something happens.  The registration and the callback are not correlated.  
From the Webhooks.org site:

     A WebHook is an HTTP callback: an HTTP POST that occurs when something happens; a simple event-notification via HTTP POST.

I understand that I introduced a new term in the example, but we also introduced cat and dog when defining the pets API.  Having said, that I don't mind adding some text to the callback object description to mention the fact that callback object can be used to describe WebHooks and the corresponding registration API.

What I don't want to have happen, is introduce the new top level concept for callbacks that are registered out of band, and have people think that you can only describe  WebHooks using that mechanism and that callback objects don't work for Webhooks.  Hence why I want to call the new thing Events because there is no expectation that you have to register something before events start happening.  They just happen.

philsturgeon: It's a safe assumption that most readers of the spec speak English, and most English speakers know what a cat and dog are without extra clarifications, so I think that's probably not a relevant comparison. 😝

> I am concerned about trying to make a distinction between Callbacks and Webhooks.

> I don't see them as interchangeable. 

AGH which is it. they're either the same, or they're different, but you gotta pick one fella!

> I don't mind adding some text to the callback object description to mention the fact that callback object can be used to describe WebHooks and the corresponding registration API.

That sounds fine to me. Want to come up with something and I copy and paste it in here, or another PR, or what are we doing? Merge this one then we sit down and smash one out with a pint in Paris?

darrelmiller: I don't want the specification to have two distinct concepts of Callbacks and Webhooks because Webhooks are Callbacks but Callbacks are not necessarily WebHooks.   
It all makes sense in my addled brain  :-)
I'll add some words.

philsturgeon: > I don't want the specification to have two distinct concepts of Callbacks and Webhooks

That is why I removed the awkward mention of Webhooks that was just slid into an example without any explanation. I am improving the situation you are describing in this comment!

philsturgeon: I am not introducing any new meaning for Webhooks. That is being done in an already merged-to-master proposal which has nothing to do with this change.

darrelmiller: The use of WebHooks in the callback example was intentional as a way of showing that you can model WebHooks with callback objects.   I would do the same in examples of the Event Object to show you can model WebHooks that are registered out of band with an Event Object.

If I add text to explain that Callback Objects can model WebHooks, are you ok putting the WebHook example back in?

darrelmiller: And you don't need to remind me that Lorna's proposal is "merged-to-master" as some badge of approval.  I'm not falling for that :-P   I'm still going to make the case for the rename to Events.

philsturgeon: That's not the intention of the reference, I am explaining that I am not giving Webhooks any new meaning, only that proposal is giving Webhooks new meaning. I am removing confusion not adding it, but you keep saying I'm adding confusion, and that is confusing _me_. 

If you add a thing introducing that you can use callbacks for webooks then great, I'll take that, even if it needs to be resolved later on through some possible future movement on the webhooks proposal. That's not relevant to this PR so we can cross that bridge if we get to it I guess.

versions/3.0.3.md
```diff
@@ -1891,13 +1891,13 @@ $request.body#/successUrls/2 | http://clientdomain.com/medium
 $response.header.Location | http://example.org/subscription/1
 
 
-##### Callback Object Example
+##### Callback Object Examples
 
-The following example shows a callback to the URL specified by the `id` and `email` property in the request body.
+The following example uses the user provided `queryUrl` query string parameter to define the callback URL.
```
darrelmiller:
```suggestion
The following example uses the user provided `queryUrl` query string parameter to define the callback URL.  This is an example of how to use callback objects to describe a WebHook callback that goes with the subscription operation to enable registering for the callback.
```
@philsturgeon How do you feel about adding this?  I'm ok with taking the webhook word out of the actual examples.

philsturgeon: Ship it!

PR #2054, opened 2019-11-12
Fix #2053 `style` keyword is not supported inside Schema object

### LINKED ISSUE #2053: `style` keyword is not supported inside Schema object
According to the documentation of Schema object, a keyword `style` is not supported inside.
However, in the example of PathItem object, `style` keyword is used inside Schema.

``` yaml
get:
  description: Returns pets based on ID
  summary: Find pets by ID
  operationId: getPetsById
  responses:
    '200':
      description: pet response
      content:
        '*/*' :
          schema:
            type: array
            items:
              $ref: '#/components/schemas/Pet'
    default:
      description: error payload
      content:
        'text/html':
          schema:
            $ref: '#/components/schemas/ErrorModel'
parameters:
- name: id
  in: path
  description: ID of pet to use
  required: true
  schema:
    type: array
    style: simple
    items:
      type: string  
```

The `style` is not inside Schema in JSON format example so I think this is just a mistake.
Maybe like:

``` yaml
get:
  description: Returns pets based on ID
  summary: Find pets by ID
  operationId: getPetsById
  responses:
    '200':
      description: pet response
      content:
        '*/*' :
          schema:
            type: array
            items:
              $ref: '#/components/schemas/Pet'
    default:
      description: error payload
      content:
        'text/html':
          schema:
            $ref: '#/components/schemas/ErrorModel'
parameters:
- name: id
  in: path
  description: ID of pet to use
  required: true
  schema:
    type: array
    items:
      type: string
  style: simple
```

This mistake is not fiexed `v3.0.3-dev`, too.

nasa9084:
fix #2053

### DIFF
--- a/versions/3.0.3.md
+++ b/versions/3.0.3.md
@@ -830,9 +830,9 @@ parameters:
   required: true
   schema:
     type: array
-    style: simple
     items:
       type: string  
+  style: simple
 ```
 
 #### <a name="operationObject"></a>Operation Object


darrelmiller:
@OAI/tsc  This seems like a easy fix to a broken example.  Anyone object to this being merged?

PR #2066, opened 2019-11-25
Support style, explode, allowReserved encoding for multipart/form-data

### LINKED ISSUE #2018: OpenAPI 3.0 does not support csv-serialized form-data arrays
## The OpenAPI 2.0 behavior

In OpenAPI 2.0, a parameter defined as `in: form` with `type: array` defaulted to `collectionFormat: csv` — meaning that the values of the array should be joined into a comma-separated string and passed in a single form data element [ref][oas2].

[oas2]: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields-7

## The OpenAPI 3.0 behavior

In OpenAPI 3.0, the default for arrays in a request body appears to be the equivalent of `collectionFormat: multi` — pass each element of the array in its own form data element.
This was hard to track down but here's my step-by-step reading of the spec:

### Details -- feel free to skip

In OpenAPI 3.0, form parameters are now described as properties in a request body.

An [encoding attribute](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#encoding-object) can be used to specify the serialization of parts of multipart request bodies.

Within the encoding object is a `style` property which
>  Describes how a specific property value will be serialized depending on its type. See Parameter Object for details on the style property. The behavior follows the same values as query parameters, including default values. This property SHALL be ignored if the request body media type is not `application/x-www-form-urlencoded`.

So when the media type is other than `application/x-www-form-urlencoded`, the serialization will be handled using the default `style` for query parameters, which is 'form' [ref][style].

[style]: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#fixed-fields-10

In [Style Values](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#style-values), `style` of `form` means:
> Form style parameters defined by RFC6570. This option replaces collectionFormat with a csv (when explode is false) or multi (when explode is true) value from OpenAPI 2.0.

So what is `explode`? Back to the encoding object: 
- This property SHALL be ignored if the request body media type is not `application/x-www-form-urlencoded`.
- When `style` is `form`, the default value is `true`.  

So when media type is other than `application/x-www-form-urlencoded`, all properties of the request body are serialized as `style: form, explode: true` -- the equivalent of `collectionFormat: multi` in OpenAPI 2.0.

## Conclusion

So the default value changed.  That's not great but would probably be acceptable in a new major release of the spec, except that there appears to be _no way_ to specify the OpenAPI 2.0 default behavior (`collectionFormat: csv` ) for properties in`multipart` request bodies in OpenAPI 3.0. Both the `style` and `explode` properties in `encoding` are explicitly restricted to media type `application/x-www-form-urlencoded`, so only the default serialization -- `style: form, explode: true` is possible.

mkistler:
This PR extends the `style`, `explode`, and `allowReserved` of the `Encoding Object` to also apply to request bodies with a `content-type` of `multipart/form-data`.

As discussed in the OpenAPI TSC meeting on 11/21/19, `multipart/form-data` is just a different serialization of url-encoded form and therefore should support the same set of behaviors.

Fixes #2018.


darrelmiller:
@OAI/tsc  Could you please review, comment and approve.  This fix simply corrects the fact that we didn't explicitly call out that both url-encoded forms and multipart/form-data can have parameter serialization customized using style, explode and allowReserved.

versions/3.1.0.md
```diff
@@ -1533,6 +1533,8 @@ When passing complex objects in the `application/x-www-form-urlencoded` content
 
 It is common to use `multipart/form-data` as a `Content-Type` when transferring request bodies to operations.  In contrast to 2.0, a `schema` is REQUIRED to define the input parameters to the operation when using `multipart` content.  This supports complex structures as well as supporting mechanisms for multiple file uploads.
 
+In a `multipart/form-data` request body, each property, or each element of an array property, takes a section in the payload with an internal header. For example, for the header `Content-Disposition: form-data; name="submit-name"`, the name of the property is `submit-name`. This type of form property is more commonly used for file transfers. The serialization strategy for each property of a `multipart/form-data` request body can be specified in an associated [`Encoding Object`](#encodingObject).
```
MikeRalphson: Apart from the last sentence, this new paragraph seems to be summarising the defined behaviour of the `multipart/form-data` spec, not defining how the OAS works in conjunction with it. Is it needed?

darrelmiller: I agree.  I don't think it is needed.

mkistler: I am open to modifying or even removing this paragraph, but let me make a few points:

- I mostly borrowed the wording for this paragraph from the OpenAPI 2.0 spec [here](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameter-object). So maybe this was also not needed in the 2.0 spec, but I personally found it helpful and thought we should move it forward to 3.0.

- In particular, I think there is value in describing how schema property values map to form data / payload sections. You may consider this "obvious", but I think it is better for a specification to state the obvious than to leave things undefined.

- In multiple conversations I’ve had on this issue, including in the last TSC, I think there has been agreement that the current spec needs to be clarified on how multipart/form-data is handled.  This was my attempt to address that.

- By "the multipart/form-data spec", do you mean RFC 7578?  If so, then at the very least I would like to replace this with a reference to that RFC (currently not referenced anywhere in the spec).

So should I take a stab at revising this paragraph or are folks still opposed to providing any explanation/clarification on this?

darrelmiller: I missed the fact that you were explaining how the Schema property maps to a parameter in multipart/form-data.   I think that aspect is worth clarifying.  I think it is important to be explicit about the form-data parameter mapping to the corresponding *schema* property.

versions/3.1.0.md
```diff
@@ -1583,9 +1585,9 @@ Field Name | Type | Description
...
-<a name="encodingAllowReserved"></a>allowReserved | `boolean` | Determines whether the parameter value SHOULD allow reserved characters, as defined by [RFC3986](https://tools.ietf.org/html/rfc3986#section-2.2) `:/?#[]@!$&'()*+,;=` to be included without percent-encoding. The default value is `false`. This property SHALL be ignored if the request body media type is not `application/x-www-form-urlencoded`.
+<a name="encodingStyle"></a>style | `string` | Describes how a specific property value will be serialized depending on its type.  See [Parameter Object](#parameterObject) for details on the [`style`](#parameterStyle) property. The behavior follows the same values as `query` parameters, including default values. This property SHALL be ignored if the request body media type is not `application/x-www-form-urlencoded` or `multipart/form-data`.
```
webron:
```suggestion
<a name="encodingStyle"></a>style | `string` | Describes how a specific property value will be serialized depending on its type.  See [Parameter Object](#parameterObject) for details on the [`style`](#parameterStyle) property. The behavior follows the same values as `query` parameters, including default values. This property SHALL be ignored if the request body media type is not `application/x-www-form-urlencoded` or `multipart/form-data`. If a value is explicitly defined, then the value of [`contentType`](#encodingContentType) (implicit or explicit) SHALL be ignored.
```

versions/3.1.0.md
```diff
@@ -1583,9 +1585,9 @@ Field Name | Type | Description
...
+<a name="encodingExplode"></a>explode | `boolean` | When this is true, property values of type `array` or `object` generate separate parameters for each value of the array, or key-value-pair of the map.  For other types of properties this property has no effect. When [`style`](#encodingStyle) is `form`, the default value is `true`. For all other styles, the default value is `false`. This property SHALL be ignored if the request body media type is not `application/x-www-form-urlencoded` or `multipart/form-data`.
```
webron:
```suggestion
<a name="encodingExplode"></a>explode | `boolean` | When this is true, property values of type `array` or `object` generate separate parameters for each value of the array, or key-value-pair of the map.  For other types of properties this property has no effect. When [`style`](#encodingStyle) is `form`, the default value is `true`. For all other styles, the default value is `false`. This property SHALL be ignored if the request body media type is not `application/x-www-form-urlencoded` or `multipart/form-data`. If a value is explicitly defined, then the value of [`contentType`](#encodingContentType) (implicit or explicit) SHALL be ignored.
```

versions/3.1.0.md
```diff
@@ -1583,9 +1585,9 @@ Field Name | Type | Description
...
+<a name="encodingAllowReserved"></a>allowReserved | `boolean` | Determines whether the parameter value SHOULD allow reserved characters, as defined by [RFC3986](https://tools.ietf.org/html/rfc3986#section-2.2) `:/?#[]@!$&'()*+,;=` to be included without percent-encoding. The default value is `false`. This property SHALL be ignored if the request body media type is not `application/x-www-form-urlencoded` or `multipart/form-data`.
```
webron:
```suggestion
<a name="encodingAllowReserved"></a>allowReserved | `boolean` | Determines whether the parameter value SHOULD allow reserved characters, as defined by [RFC3986](https://tools.ietf.org/html/rfc3986#section-2.2) `:/?#[]@!$&'()*+,;=` to be included without percent-encoding. The default value is `false`. This property SHALL be ignored if the request body media type is not `application/x-www-form-urlencoded` or `multipart/form-data`. If a value is explicitly defined, then the value of [`contentType`](#encodingContentType) (implicit or explicit) SHALL be ignored.
```

mkistler:
I've merged @webron 's suggested changes and replaced the details of multipart/form-data with a reference to RFC 7578.  I think this is ready for re-review.

versions/3.1.0.md
```diff
@@ -1533,6 +1533,8 @@ When passing complex objects in the `application/x-www-form-urlencoded` content
 
 It is common to use `multipart/form-data` as a `Content-Type` when transferring request bodies to operations.  In contrast to 2.0, a `schema` is REQUIRED to define the input parameters to the operation when using `multipart` content.  This supports complex structures as well as supporting mechanisms for multiple file uploads.
 
+In a `multipart/form-data` request body, each property, or each element of an array property, takes a section in the payload with an internal header as defined by [RFC 7578](https://tools.ietf.org/html/rfc7578). The serialization strategy for each property of a `multipart/form-data` request body can be specified in an associated [`Encoding Object`](#encodingObject).
```
darrelmiller:
```suggestion
In a `multipart/form-data` request body, each schema property, or each element of a schema array property, takes a section in the payload with an internal header as defined by [RFC 7578](https://tools.ietf.org/html/rfc7578). The serialization strategy for each property of a `multipart/form-data` request body can be specified in an associated [`Encoding Object`](#encodingObject).
```
I feel the need to qualify `property` here.

Issue #1385, opened 2017-10-20
Generate 3.0 JSON examples from 3.0 YAML examples in json/yaml subfolders


darrelmiller:

MikeRalphson:
Refs #1357

darrelmiller:
Yes, I should have mentioned that the PR referenced inspired this decision.  We don't want to add maintenance burden by having both formats, but we can generate them.

cebe:
What kind of tooling would you accept into the repo? Is there any preference for language/toolchain to use?

With PHP installed, the following `Makefile` would generate JSON files from all YAML files in a directory:

```make


YAMLS := $(wildcard ./*.yaml)
JSONS := $(patsubst ./%.yaml, ./%.json, $(YAMLS))

all: vendor/bin/php-openapi $(JSONS)

vendor/bin/php-openapi:
	echo '{}' > composer.json
	composer require cebe/php-openapi 'symfony/yaml:~4.2.0'

%.json: %.yaml
	vendor/bin/php-openapi convert --read-yaml $< --write-json $@

.PHONY: all

```

If PHP is an acceptable tooling, I can send a PR with setting it up.

SephReed:
I'd put in a request for this too.  It's not like it's hard to convert on ones own, but I was just looking for some quick test data.  Even just one json example would have been enough to save me the trouble of converting it myself.

https://onlineyamltools.com/convert-yaml-to-json

MikeRalphson:
Closing as #2199 was merged. Thanks all.

PR #2036, opened 2019-10-21
Fixed syntax errors in api-with-examples.yaml

mabar:
With original, php implementation of yaml symfony/yaml I got some syntax errors. This PR fixes them.

Online parsers which consider current file valid:
- https://www.json2yaml.com
- http://www.yamllint.com
- https://yaml-online-parser.appspot.com

Online parsers which reports error:
- https://codebeautify.org/yaml-validator
- https://jsonformatter.org/yaml-validator
- https://onlineyamltools.com/validate-yaml
- https://onlineyamltools.com/edit-yaml

With proposed changes is file valid for all of them

PR #2068, opened 2019-11-27
Implement GitHub Actions workflow for automatically updating the example JSON files

cebe:
As discussed in #1385 the example `.json` files for version 3 should be automatically generated from the `.yaml` example files.

I thought this was a good case for trying out Github Actions and the result is the following:

- When a push to `master` (or any other branch we wish to include) is made, a Github Actions Workflow is triggered ([example](https://github.com/cebe/OpenAPI-Specification/commit/9c98e819ae876af92c2a9112dcfa6dfcb929e7dc/checks?check_suite_id=331067708))
- it will generate new JSON files from the YAML examples
- if there are any changes, it will create a pull request to the branch that triggered the Workflow ([example](https://github.com/cebe/OpenAPI-Specification/pull/3))

### DIFF
--- a/.github/workflows/yaml2json.yml
+++ b/.github/workflows/yaml2json.yml
@@ -35,11 +35,9 @@ jobs:
         for file in *.yaml ; do
             target=`basename -s .yaml $file`.json
             echo "converting  $file  to  $target  ..."
-            if [ "$file" != "api-with-examples.yaml" ] ; then
-                vendor/bin/php-openapi convert --read-yaml $file --write-json $target
-                vendor/bin/indent --tabs --tabstop=4 $target
-                vendor/bin/indent --spaces --tabstop=2 $target
-            fi
+            vendor/bin/php-openapi convert --read-yaml $file --write-json $target
+            vendor/bin/indent --tabs --tabstop=4 $target
+            vendor/bin/indent --spaces --tabstop=2 $target
         done
         
     - name: cleanup
@@ -51,7 +49,7 @@ jobs:
       run: |
         cd examples/v3.0
         git add *.json
-        git diff --staged
+        git --no-pager -c color.diff=always diff --staged
 
     - name: Create Pull Request
       uses: peter-evans/create-pull-request@v1
--- a/.github/workflows/yaml2json.yml
+++ b/.github/workflows/yaml2json.yml
@@ -12,7 +12,6 @@ on:
   push:
     branches:    
       - master
-      - try-github-actions
 
 jobs:
   yaml2json:
--- a/.github/workflows/yaml2json.yml
+++ b/.github/workflows/yaml2json.yml
@@ -1,0 +1,70 @@
+name: YAML2JSON
+
+#
+# This workflow updates the *.json files in the examples/v3.0 directory,
+# when the corresponding *.yaml files change.
+# JSON example files are automatically generated from the YAML example files.
+# Only the YAML files should be adjusted manually.
+#
+
+# run this on push to master
+on:
+  push:
+    branches:    
+      - master
+      - try-github-actions
+
+jobs:
+  yaml2json:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1 # checkout repo content
+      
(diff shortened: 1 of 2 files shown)


.github/workflows/yaml2json.yml
```diff
@@ -0,0 +1,51 @@
+name: YAML2JSON
+
+# run this on push to master
+on:
+  push:
+    branches:    
+      - master
+      - try-github-actions
```
webron: Is that still needed? :)

cebe: probably not, going to remove that before merging this PR. But it will run the action if I make adjustments to this branch so I'd keep it for testing until it is ready to merge.

cebe:
> Would it make sense to change the Action to run on `*.yaml` in that directory instead?

what do you mean? As far as I am aware there is no way to run an action based on whether some files changed or not.

cebe:
@webron is there anything to change or to do from my side to get this merged? (except from removing the `try-github-actions` branch)

webron:
Sorry, I didn't mean run it based on whether a file changed, but rather here:
```
        vendor/bin/php-openapi convert --read-yaml callback-example.yaml  --write-json callback-example.json
        vendor/bin/php-openapi convert --read-yaml link-example.yaml      --write-json link-example.json
        vendor/bin/php-openapi convert --read-yaml petstore.yaml          --write-json petstore.json
        vendor/bin/php-openapi convert --read-yaml petstore-expanded.yaml --write-json petstore-expanded.json
        vendor/bin/php-openapi convert --read-yaml uspto.yaml             --write-json uspto.json
```

instead of running it on each file individually, it'd be more robust if it ran on all yaml file in the directory, in case additional files are added.

MikeRalphson:
Also, what indentation is used in the output json? I guess it should match the v2.0 examples for consistency.

cebe:
Indentation is currently 4 spaces, I'll change that to 2.

MikeRalphson:
@cebe is the exclusion due to #2036 ?

Just a thought for after this is merged, would you have time to create a similar workflow to keep the json versions of the yaml json-schemas in step? It would need to run on versions (directories) 3.0 and above.

cebe:
> @cebe is the exclusion due to #2036 ?

yes, symfony/yaml has some problems with these files, did not see that PR, will try to merge the commit and see if it fixes the problem.

> Just a thought for after this is merged, would you have time to create a similar workflow to keep the json versions of the yaml json-schemas in step? It would need to run on versions (directories) 3.0 and above.

sure I can do that.

cebe:
Merged the changes from #2036, they fix the issue with api-with-examples.yaml.

> instead of running it on each file individually, it'd be more robust if it ran on all yaml file in the directory, in case additional files are added.

done.

> Also, what indentation is used in the output json? I guess it should match the v2.0 examples for consistency.

indentation is 2 now.

Ready to merge from my side.

MikeRalphson:
My only question is why we indent the target with tabs and then (presumably) overwrite it with a space indented version. Otherwise LGTM.

cebe:
> My only question is why we indent the target with tabs and then (presumably) overwrite it with a space indented version.

The tool I use for indentation can not convert 4 space to 2 space indentation. it can only convert tabs to spaces or spaces to tabs. So I do the conversion in two steps, 4 spaces to tabs, then tabs to 2 spaces.

cebe:
> we need to remove the test branch from the script

done.

MikeRalphson:
@webron are you ok to create the `GITHUB_TOKEN` before merging?

webron:
@MikeRalphson If I understand correctly, `GITHUB_TOKEN` is created automatically when GitHub Actions are enabled (which they are). See https://help.github.com/en/actions/automating-your-workflow-with-github-actions/authenticating-with-the-github_token.

MikeRalphson:
👍

cebe:
there is no need to add a github token manually, it is provided by Github actions. Anything stopping this from being merged?

cebe:
Closing this, as #2199 is replacing it.

PR #2077, opened 2019-12-18
Added support for multiple externalDocs (#1034)

### LINKED ISSUE #1034: Permit multiple externalDocs
Documenting APIs often depend on multiple external factors - functional and non-functional aspects, change notes, governance, strategy, many more - and a single externalDocs reference is often insufficient, especially in complex scenarios where multiple parties are responsible for design/deployment/delivery of these aspects (not everyone is DevOps)

Would we want to move the single object to an array/collection of objects:
```
externalDocs:
- description: link 1
  url: http://link/1
- description: link
  url: http://link/2
```
or
```
externalDocs:
  link1:
    description: link 1
    url: http://link/1
  link2
    description: link
    url: http://link/2
```
 Gong further, form 2 may also benefit from becoming a reference:
```
externalDocs:
  link1:
    url: http://link/1

paths:
  /path1:
    get:
      externalDocs:
      - $ref: '#/externalDocs/link1'
```

While some of this may overlap with (the single) externalDocs avaulable in a tag, it is not entirely overlapping, and the link should ideally reside close to where it is needed instead of tracking it back through a tag that may not be immediately apparent will contain the required link.

nirnir93:
This allows for multiple external docs without breaking current API.
I will add another pull request containing breaking changes for future releases

Fixes #1034

### DIFF
--- a/schemas/v2.0/schema.json
+++ b/schemas/v2.0/schema.json
@@ -80,19 +80,8 @@
       "uniqueItems": true
     },
     "externalDocs": {
-      "oneOf": [
-        {
-          "$ref": "#/definitions/externalDocs"
-        },
-        {
-          "type": "array",
-          "items": {
-            "$ref": "#/definitions/externalDocs"
-          },
-          "uniqueItems": true
-        }
-      ]
-}
+      "$ref": "#/definitions/externalDocs"
+    }
   },
   "definitions": {
     "info": {
@@ -274,18 +263,7 @@
           "description": "A longer description of the operation, GitHub Flavored Markdown is allowed."
         },
         "externalDocs": {
-          "oneOf": [
-            {
-              "$ref": "#/definitions/externalDocs"
-            },
-            {
-              "type": "array",
-              "items": {
-                "$ref": "#/definitions/externalDocs"
-              },
-              "uniqueItems": true
-            }
-          ]
+          "$ref": "#/definitions/externalDocs"
         },
         "operationId": {
           "type": "string",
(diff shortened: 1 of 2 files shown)


nirnir93:
I's just noticing that I worked on the V2.0 schema (because there is no V3.0 schema).
Is there a branch which containg a V3.1 schema that I can rebase onto?

MikeRalphson:
@nirnir93 the 3.0 schema document wasn't finalised when we branched `3.1.0-dev` so that's the historical reason this branch doesn't contain that file. For now please just ignore changes to the schema file, we can pick them up later.

It would also be helpful if you could cross-reference pre-existing requests for this change in the PR. Thanks.

darrelmiller:
How would this be processed by tooling?  What is the difference between the first doc link and the second one?  Are they additive? Are they alternative docs in different languages?  

I don't see value that outweighs the additional complexity.

nirnir93:
@darrelmiller that's up to the `name`. i.e. I can specify each doc's language, version, function and the functionality it refers. UI tools should just show all the docs, one by one.

Also <span>#</span>1: After the community starts using it, it may suggest changing the `External Documentation Object` itself (e.g. adding a `labels`/`tags` property to control visibility by some kind of logic such as language).

Also <span>#</span>2: The workaround suggested now (creating an external website for docs) will still work. This will only add power to it (e.g. I can refer multiple endpoints in the site, and handle logic in the site itself. I can also ignore this and just use a single `externalDoc`).

darrelmiller:
@nirnir93 When we add a new feature like this to the specification, we are asking many of the tools listed here https://openapi.tools/ to add support for that feature.  While it is only changing a few lines of text to the specification it represents hundreds of hours of work in total for tooling folk.  

I would suggest reaching out to some of the people building the tools to ask them if they would be prepared to prioritize adding support for this feature.  When you get their buy in, it will be much easier to make a case for adding it to the specification.  

In the meanwhile, I would encourage you to create an `x-` extension to track your additional documents.

PR #2080, opened 2019-12-20
OpenAPI not Open API

philsturgeon:
image


webron:
We can't change the spec files (we can (and should) 3.0.3 and 3.1.0) - at the time they were published, the space existed.

philsturgeon:
Alright, removed those changes and followed up with #2081 and #2082. Let's merge all three, I've got a glossary to make correct. https://github.com/openapi-contrib/glossary

PR #2083, opened 2019-12-23
OpenAPI not Open API (3.1.0 Take II)

philsturgeon:


MikeRalphson:
Just a reminder that PRs against patch releases don't need to be additionally submitted against minor dev versions as we have a script that will handle this.

philsturgeon:
Ooo ok so #2081 will take care of things and I can close this one? Great. Thanks Mike.

PR #2102, opened 2020-01-14
introduce schema definition SchemaReference

notEthan:
I'm working on tooling which relies on the openapi schema for implementing openapi functionality. schemas are handled, of course, quite differently than other structures which may be pointed to by a Reference such as a Parameter or Header.

in order to identify what is a schema and what is not, using this openapi schema, having a separate Reference particular to Schema references would be very helpful.

I can do a workaround where I examine the target of a reference and see if its schema is `#/definitions/Schema`, but it's much simpler, from my point of view, to just have schema references be their own thing.

### DIFF
--- a/schemas/v3.0/schema.yaml
+++ b/schemas/v3.0/schema.yaml
@@ -43,6 +43,14 @@ definitions:
       '^\$ref$':
         type: string
         format: uri-reference
+  SchemaReference:
+    type: object
+    required:
+      - $ref
+    patternProperties:
+      '^\$ref$':
+        type: string
+        format: uri-reference
   Info:
     type: object
     required:
@@ -139,7 +147,7 @@ definitions:
           '^[a-zA-Z0-9\.\-_]+$':
             oneOf:
               - $ref: '#/definitions/Schema'
-              - $ref: '#/definitions/Reference'
+              - $ref: '#/definitions/SchemaReference'
       responses:
         type: object
         patternProperties:
@@ -269,39 +277,39 @@ definitions:
       not:
         oneOf:
           - $ref: '#/definitions/Schema'
-          - $ref: '#/definitions/Reference'
+          - $ref: '#/definitions/SchemaReference'
       allOf:
         type: array
         items:
           oneOf:
             - $ref: '#/definitions/Schema'
-            - $ref: '#/definitions/Reference'
+            - $ref: '#/definitions/SchemaReference'
       oneOf:
         type: array
         items:
           oneOf:
             - $ref: '#/definitions/Schema'
-            - $ref: '#/definitions/Reference'
+            - $ref: '#/definitions/SchemaReference'
       anyOf:
         type: array
         items:
           oneOf:
             - $ref: '#/definitions/Schema'
-            - $ref: '#/definitions/Reference'
+            - $ref: '#/definitions/SchemaReference'
       items:
         oneOf:
           - $ref: '#/definitions/Schema'
-          - $ref: '#/definitions/Reference'
+          - $ref: '#/definitions/SchemaReference'
       properties:
         type: object
         additionalProperties:
           oneOf:
             - $ref: '#/definitions/Schema'
-            - $ref: '#/definitions/Reference'
+            - $ref: '#/definitions/SchemaReference'
       additionalProperties:
         oneOf:
           - $ref: '#/definitions/Schema'
-          - $ref: '#/definitions/Reference'
+          - $ref: '#/definitions/SchemaReference'
           - type: boolean
         default: true
       description:
@@ -397,7 +405,7 @@ definitions:
       schema:
         oneOf:
           - $ref: '#/definitions/Schema'
-          - $ref: '#/definitions/Reference'
+          - $ref: '#/definitions/SchemaReference'
       example: {}
       examples:
         type: object
@@ -457,7 +465,7 @@ definitions:
       schema:
         oneOf:
           - $ref: '#/definitions/Schema'
-          - $ref: '#/definitions/Reference'
+          - $ref: '#/definitions/SchemaReference'
       content:
         type: object
         additionalProperties:
@@ -669,7 +677,7 @@ definitions:
       schema:
         oneOf:
           - $ref: '#/definitions/Schema'
-          - $ref: '#/definitions/Reference'
+          - $ref: '#/definitions/SchemaReference'
       content:
         type: object
         additionalProperties:


schemas/v3.0/schema.yaml
```diff
@@ -43,6 +43,14 @@ definitions:
...
         type: string
         format: uri-reference
+  SchemaReference:
+    type: object
+    required:
+      - $ref
+    patternProperties:
+      '^\$ref$':
+        type: string
+        format: uri-reference
```
notEthan: this could also just be

```yaml
SchemaReference:
  $ref: '#/definitions/Reference'
```

to avoid duplication. just having a separate schema to differentiate a SchemaReference vs a plain Reference is all that matters to me.

handrews:
@notEthan you should be checking the source of the reference, because references to schemas are only legal where Schema Objects are legal (which includes inside of Schema Objects).  So you should just be looking for what things allow Schema Objects.  Any references in the same location are required to reference schemas.  (that requirement is not in the JSON Schema here because that sort of thing is outside of the scope of JSON Schema).

notEthan:
> you should just be looking for what things allow Schema Objects

that's the idea. I do that by seeing if the schema for an object resolves to `#/definitions/Schema` in this openapi schema. that part is easy enough. I can easily indicate to my schema tooling, "this schema describes other schemas".

but when the schema resolves to `#/definitions/Reference`, I have to do more to determine if it points at a schema or not, and it is not so easy to say "the Reference schema sometimes describes a schema, sometimes not". there are a few ways I could work around this:

- deref the Reference and see if it ends up pointing at a thing whose schema is openapi's `#/definitions/Schema`
- check if its schema contains a oneOf pointing at both Schema and Reference, like 
  ```json
  {
    "oneOf": [
      {"$ref": "#/definitions/Schema"},
      {"$ref": "#/definitions/Reference"}
    ]
  }
  ```

  then conditionally treat such a Reference like an openapi schema. I don't much care for that option, it feels like more brittle logic than I want, when this should be expressible with schemas alone.
- the approach I've PR'd here where a SchemaReference is a different schema than a Reference to things that are not schemas.

I feel like what you're suggesting is closest to the second option - checking to see whether an object which is a Reference could have been a Schema. I think the first option is preferable to that, if I have to dynamically determine what's a schema or not. however, with the third option enabled by this PR, I'm hoping to distinguish schema-ness using the schema definitions alone, by telling my tooling that both `#/definitions/Schema` and `#/definitions/SchemaReference` indicate schemas.

handrews:
@notEthan This all seems rather complicated- with OpenAPI, the places where schemas are legal are well-known, and do not change frequently (like... maybe once every couple of years).  It's not supposed to be this hard to parse OpenAPI, and I don't think a complex runtime tool should be needed.

Also, the schema for OpenAPI is non-normative.  It's there to help, and it's believed to _only_ produce errors for invalid OAS documents.  But it doesn't catch every possible OAS document error.

MikeRalphson:
@notEthan you say

> schemas are handled, of course, quite differently than other structures which may be pointed to by a Reference such as a Parameter or Header.

But I don't think you've made your case for why this is necessary for your tool. The only `$ref` which is not treated as a referenceObject is that within the pathItem object.

notEthan:
true, I have not very well explained the goal I'm working toward with this. I'll try to explain briefly.

at a high level, the implementation that I'm working on uses schemas (json schema or openapi-flavored schema) to dynamically generate object-oriented classes in order to instantiate schema instances. so, for one example, when you call a petstore's getPetById operation, the resulting response would give you a pet whose class is defined by the petstore's openapi schema, and whose instance comes from the response body.

for another example, the petstore openapi document is instantiated using a dynamically generated class constructed from this schema. this works extremely well for almost everything, except for this Reference issue.

> why, Ethan, would you want to do that?

it's pretty neat to be able to extend schemas with class methods. you can put business logic relating to a Pet on the class, and it's implicitly there whether you get back one pet, an array of pets, or a pet in any other nested structure.

similarly, an openapi-flavored schema is implicitly instantiated as a schema wherever it lives in an openapi document. unfortunately this is made more difficult by a Reference schema being too polymorphous. I need some workaround to identify a Schema in any of the 11 places a Pointer may be a Schema.

with this change, the intent of a schema pointer is clear and no workaround would be needed.

handrews:
@notEthan every object that allows a `$ref` requires that the target be of the correct structure matching the context of the `$ref`.  It's not that there is one for schemas, and one for everything else.  There is one for each object type that can be referenced.

Honestly, I don't understand the problem here.  If you're in a schema, or looking directly a field with a schema value (and I'm pretty sure they're all called `schema` so that's easy to figure out), the only thing you can `$ref` is a schema.  All other possible places a `$ref` might be used are ***not*** schemas.  Looking for fields named `schema` seems more robust, as that's actually part of the specification.  Definition names in the schema describing OAS documents are non-normative, and in JSON Schema are usually assumed to not have any semantic behavior.

handrews:
I'm still opposed to this, as it is intended to cause tooling to rely on definition names, which is an anti-pattern in JSON Schema.

@OAI/tsc can we get a decision on this and (preferably) close it out?

notEthan:
I know this is worthwhile, if I can explain it sensibly.

I'm looking at the pet store, and I want to know if the object at [`#/paths/~1pets/get/responses/200/content/application~1json/schema`](https://github.com/OAI/OpenAPI-Specification/blob/3.0.3/examples/v3.0/petstore.yaml#L35) is a schema.

how do I determine if this is a schema?

I want to look at the schema of that object - defined by this openapi document schema - and see if the schema for that object itself describes a schema.

so I end up at [MediaType's schema property](https://github.com/OAI/OpenAPI-Specification/blob/3.0.3/schemas/v3.0/schema.yaml#L397), and from there I match the oneOf option going to `#/definitions/Reference`.

`#/definitions/Reference` does not describe a schema; it can refer to any number of non-schema things. I have to do something else to determine that this reference itself represents a schema.

I talk about possible "something else"s above, but they are all ugly and overly complex compared to just having a 'reference object pointing at a schema' be a thing that describes a schema, separate from the generic Reference object.

> It's not that there is one for schemas, and one for everything else. There is one for each object type that can be referenced.

there is only one Reference schema, I don't know what you're talking about with "one for each object type that can be referenced". there _could_ be that much distinction among Reference objects, but I only care about this distinction for schemas at the moment.

> This all seems rather complicated- with OpenAPI, the places where schemas are legal are well-known, and do not change frequently

I'm not sure how you are using 'place' here. the paths to schemas can be arbitrarily nested, so I'll assume (please do correct me) you mean 'place' as in a property of a known type of object (e.g. the 'schema' property of definitions/MediaType instance, or the 'items' property of a definitions/Schema instance).

it is possible that I can go to each of the 11 object properties where I know that a Reference must refer to a schema (the same ones I changed in this PR) and indicate that their instances are schemas. (I think this would work anyway - I haven't coded this out because this change that I am trying to share made it so much easier.)

now I have 11 schemas under definitions that I am keeping track of in order to know what in the document is a schema.

> I'm still opposed to this, as it is intended to cause tooling to rely on definition names, which is an anti-pattern in JSON Schema

I'm not sure how you're supposed to refer to schemas except by the names they are given, but taking as read this being an anti-pattern: this change makes it so I can go from relying on paths to 11 different schemas nested at various depths under definitions, down to 1.

I just want to be able to tell what the heck is a schema in an openapi document, using this schema. describing this with SchemaReference makes it so nice and simple.

Issue #1968, opened 2019-07-05
Callbacks without a subscription API call


lornajane:
I have an API that sends webhooks/callbacks to a client, but not as a response to a previous API call, such as requesting async data or making a subscription. The "subscribe" step happens elsewhere, through a web frontend, so there is no `path` entry in the API spec for me to put my `callback` entries under.

I would like to have `callback` items in my spec but they need to be independent of any previous API call since in this case I don't have one. Is there a recognised workaround for this? Or would we need a formal spec change to accommodate this? I'd really appreciate any advice!

My use case: the Nexmo SMS API allows users to make an API call to send a message. When they sign up for an API key and secret, they configure a URL to receive webhooks when an SMS arrives at this number. So the incoming webhook is not really a callback, and doesn't belong to a previous API call. I have also been asked how this works by a few other organisations looking to move to OpenAPI (I have given a few conference talks about OpenAPI) and I'd love to have a good answer myself and to offer to otthers.

lornajane:
I got some positive noise in response to this on slack so I created a format proposal for adding webhooks https://github.com/OAI/OpenAPI-Specification/pull/1974. I'd still love to hear any advice or comments you have either on this issue or over on the proposal

darrelmiller:
@lornajane How about putting the callbacks in /components/callbacks?  This allows you to describe the webhook without attaching it to a path.  We could consider adding an optional property to a callback to indicate that it could be registered out of band.

Regarding the proposal, I can see the value of adding a new key to describe between callbacks that are registered out of band, however, I wouldn't want to use the term webhook to make that distinction as most implementations of webhooks are registered by subscribing to an API.

lornajane:
Thanks @darrelmiller ! I am not sure I agree that "most implementations of webhooks are registered by subscribing to an API". The ones I've used have almost all been a website setup step to configure the destination URL rather than an API call. Maybe I'm just using the wrong APIs ? :)

I don't feel strongly about naming or structure, I really just need the feature of being able to describe the incoming requests from server to API client. To me it seems like these are equivalent to (but distinct from) the entries in `paths` that describe the requests which is why I proposed the sibling entry to that entry. If some other name or structure would help this to be acceptable so that I could continue to use OpenAPI, that would also be fine!

darrelmiller:
@lornajane What about describing the webhook callbacks in /components/callbacks?  That way you don't have to have a path.

MikeRalphson:
@darrelmiller 

[...]

PR #2103, opened 2020-01-14
Add support for webhooks as a top-level element

lornajane:
After some discussion (#1968 ), an accepted proposal (#1974 ), and some more discussion on the TSC calls, we're at a a point where we can add support for webhooks to OpenAPI and this PR does that.

Of particular note is that after TSC discussion, it became clear that what we have here is a map of PathItems, rather than a new object type so the spec changes reflect this.

Without a specific Webhook Object, the documentation is technically complete but felt a bit ... thin. Therefore an example has also been added and linked to help users understand how this feature can be used.


lornajane:
Worth a mention: I manually repeated the changes from #2048 as well as they seemed really relevant and like we'd want to see the two together but they weren't already in this branch (they're in the 3.0.3 one) and it wasn't clear to me how to merge that into this file.

versions/3.1.0.md
```diff
@@ -20,7 +20,7 @@ An OpenAPI definition can then be used by documentation generation tools to disp
 	- [Path Templating](#pathTemplating)
 	- [Media Types](#mediaTypes)
 	- [HTTP Status Codes](#httpCodes)
-- [Specification](#specification)
+    - [Specification](#specification)
```
webron: This seems to have snuck in, should be removed.

versions/3.1.0.md
```diff
@@ -197,6 +197,8 @@ Field Name | Type | Description
...
 <a name="oasTags"></a>tags | [[Tag Object](#tagObject)] | A list of tags used by the specification with additional metadata. The order of the tags can be used to reflect on their order by the parsing tools. Not all tags that are used by the [Operation Object](#operationObject) must be declared. The tags that are not declared MAY be organized randomly or based on the tools' logic. Each tag name in the list MUST be unique.
 <a name="oasExternalDocs"></a>externalDocs | [External Documentation Object](#externalDocumentationObject) | Additional external documentation.
+<a name="oasWebhooks"></a>webhooks | Map[`string`, [Path Item Object](#pathItemObject)] | The incoming webhooks that may be received as part of this API. The key name is a unique string to refer to each webhook, while the Path Item Object describes a request that may be initiated by the API provider and the expected responses. An [example](../examples/v3.1/webhook-example.yaml) is available.
```
webron: - I'd like to see this entry right after the `paths`. 
- `MAY`
- Not a huge fan of the link to the external example. I realize we also have it in the callbacks, not sure we should have it there either.
- It might also prove useful to further explain that this can be used to describe out of bound webhook registration, the relationship with callbacks and that these are meant to be implemented by the API consumer. It might be clear to us as we've all discussed it in the past, not sure a first time reader will get it as well.

versions/3.1.0.md
lornajane: good spot, removed.

versions/3.1.0.md
lornajane: I've made the changes you suggest and expanded more on the wording, would you take another look? I have left the external example in since I think it is useful to have one somewhere but I'm open to suggestions about where!

examples/v3.1/webhook-example.yaml
```diff
@@ -0,0 +1,61 @@
+openapi: 3.1.0
+info:
+  title: Webhook Example
+  version: 1.0.0
+paths:
+  # OpenAPI specs all need a paths element
```
MikeRalphson: documents for specs, but we can fix this later.

examples/v3.1/webhook-example.yaml
```diff
@@ -0,0 +1,61 @@
...
+    # This is a Path Item Object, the only difference is that the request is initiated by the API provider
+    post:
+      requestBody:
+        description: Information about a new pet in the system
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/Pet"
+      responses:
+        200:
```
MikeRalphson: `200` needs quoting. Again, a nit.

versions/3.1.0.md
```diff
@@ -193,11 +193,13 @@ Field Name | Type | Description
...
 <a name="oasServers"></a>servers | [[Server Object](#serverObject)] | An array of Server Objects, which provide connectivity information to a target server. If the `servers` property is not provided, or is an empty array, the default value would be a [Server Object](#serverObject) with a [url](#serverUrl) value of `/`.
 <a name="oasPaths"></a>paths | [Paths Object](#pathsObject) | **REQUIRED**. The available paths and operations for the API.
+<a name="oasWebhooks"></a>webhooks | Map[`string`, [Path Item Object](#pathItemObject)] | The incoming webhooks that MAY be received as part of this API and that the API consumer is expected to implement. Closely related to the `callbacks` feature, this section describes requests initiated other than by an API call, for example by an out of band registration. The key name is a unique string to refer to each webhook, while the Path Item Object describes a request that may be initiated by the API provider and the expected responses. An [example](../examples/v3.1/webhook-example.yaml) is available.
```
MikeRalphson: Just a note that if we are linking examples from the spec, we should ensure links are not broken from the "pretty" ReSpec version on openapis.org.

lornajane: Ah, what's the correct way to refer to the examples? I copied the way the Callback Object did it but @MikeRalphson is correct (of course!) - that link doesn't work in the nicely rendered spec version http://spec.openapis.org/oas/v3.0.2.html#callback-object

MikeRalphson: I'd be happy with either an absolute URL or I could fix relative links in the spec rendering pre-processor. What do people think? It's good that we've spotted that the problem was there already,

examples/v3.1/webhook-example.yaml
lornajane: Fixed!

examples/v3.1/webhook-example.yaml
lornajane: Good catch, fixed now

examples/v3.1/webhook-example.yaml
```diff
@@ -0,0 +1,61 @@
...
+          description: A paged array of pets
+          content:
+            application/json:    
+              schema:
+                $ref: "#/components/schemas/Pets"
+
+webhooks:
+  # Each webhook needs a name
+  newPet:
+    # This is a Path Item Object, the only difference is that the request is initiated by the API provider
```
fmvilas: If the request is initiated by the API provider, where is the server defined? Where is this request going?

(Sorry for the late review comment)

Issue #2092, opened 2020-01-09
Define the scope of $id resolution 


darrelmiller:

philsturgeon:
Let's make sure this work solves https://github.com/OAI/OpenAPI-Specification/issues/2038 too.

handrews:
@philsturgeon please keep these issues separate.  It's not clear that action should be taken on #2038, while the basic work here is straightforward.

handrews:
I think part of what needs to happen here is to point to the [soon-to-be-consolidated](https://github.com/json-schema-org/json-schema-spec/issues/849) (in the bugfix/clarification draft that will come out in coordination with OAS 3.1) guidance on loading JSON Schemas, and explain how that fits into processing OpenAPI files.

@darrelmiller @webron I seem to recall a reluctance to get too much into OAS file processing, but since JSON Schema processing **is** defined and has a lot to do with ensuring that `$id`, `$anchor`, and `$ref` function correctly, making sure that it's clear how and when to hand off OAS processing to JSON Schema processing would be helpful.

The main issue is determining the base URI.  The relevant part of RFC 3986 and the JSON Schema spec is that in the absence of an `%id` in the root of the OAS Schema Object, its retrieval URI is used to determine the base.

Since the retrieval URI is `<oas-document-base-URI>#/path/to/schemaobject`, we probably want to just make clear that that does in fact mean that `<oas-document-base-uri>` is the base URI in the absence of `$id`.  This is the thing that's already clear if you can recite the RFC3986 base URI determination process in your sleep, but that probably should be called out for everyone else :-)

handrews:
We might also want to make sure that it's clear that if your Schema Object does have an `$id`, relative URI-references in `$ref` in child schemas **cannot** do things like `"#/components/schemas/foo"` because that would go relative to the `$id` and not to the OAS document as a whole.

If you are using a Schema Object with a `$id`, then you need to write you relative references the same as if the Schema Object were a separate file/resource at that URI.  So you'd have to write out `<oas-document-base-uri>#/components/schemas/foo` in full.  Or possibly some path-relative thing, but definitely not a fragment-only thing.

handrews:
@webron @darrelmiller if I put a non-Schema Object (like a Response Object) in its own file, what is the base URI for Reference Objects in that file?  In particular, is `#/components/whatever` expected to work (because it is resolved relative to the primary OAS file) or fail (because Response Objects don't have a `components` property)?

[...]

PR #2107, opened 2020-01-16
Reference Object and Schema Object use of $ref updates for 2019-09 / OAS 3.1

### LINKED ISSUE #1887: $ref value must be accessible URL for external reference?
Based on JSON reference spec, the $ref value is an URI, which by definition is a resource identifier only, and is independent of access.
It seems it is valid to use a well formed URI, but not an accessible URL (assuming http).
However, if it is not a valid URL for external reference, how would different tools resolve the references?

handrews:
Partially handles #2092 
Fixes #1887

### DIFF
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -173,7 +173,7 @@ Where OpenAPI tooling renders rich text it MUST support, at a minimum, markdown 
 ### <a name="relativeReferences"></a>Relative References in URIs
 
 Unless specified otherwise, all properties that are URIs MAY be relative references as defined by [RFC3986](https://tools.ietf.org/html/rfc3986#section-4.2).
-Relative references are resolved using the URLs defined in the [`Server Object`](#serverObject) as a Base URI.  This includes relative references in [`Reference Objects`](#referenceObjects), outside of any [`Schema Object`](#schemaObject)l
+Relative references are resolved using the URLs defined in the [`Server Object`](#serverObject) as a Base URI.  This includes relative references in [`Reference Objects`](#referenceObjects), outside of any [`Schema Object`](#schemaObject).
 
 Relative references in [`Schema Objects`](#schemaObject), including any that appear as `$id` values, use the nearest parent `$id` as a Base URI, as described by [JSON Schema Specification Draft 2019-09](https://json-schema.org/draft/2019-09/json-schema-core.html).  If no parent schema contains an `$id`, then the Base URI is determined as in the previous paragraph.
 
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -170,12 +170,12 @@ The formats defined by the OAS are:
 Throughout the specification `description` fields are noted as supporting CommonMark markdown formatting.
 Where OpenAPI tooling renders rich text it MUST support, at a minimum, markdown syntax as described by [CommonMark 0.27](https://spec.commonmark.org/0.27/). Tooling MAY choose to ignore some CommonMark features to address security concerns. 
 
-### <a name="relativeReferences"></a>Relative References in URLs
+### <a name="relativeReferences"></a>Relative References in URIs
 
-Unless specified otherwise, all properties that are URLs MAY be relative references as defined by [RFC3986](https://tools.ietf.org/html/rfc3986#section-4.2).
-Relative references are resolved using the URLs defined in the [`Server Object`](#serverObject) as a Base URI.
+Unless specified otherwise, all properties that are URIs MAY be relative references as defined by [RFC3986](https://tools.ietf.org/html/rfc3986#section-4.2).
+Relative references are resolved using the URLs defined in the [`Server Object`](#serverObject) as a Base URI.  This includes relative references in [`Reference Objects`](#referenceObjects), outside of any [`Schema Object`](#schemaObject)l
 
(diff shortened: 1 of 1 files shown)


webron:
\*ahem\* https://github.blog/2019-02-14-introducing-draft-pull-requests/

handrews:
😮 
Since no one has commented on this anyway I'll close it and open one of those- I had no idea draft PRs were a thing!

handrews:
Except this repo doesn't seem to offer the draft option?  I'm confused so I'm re-opening this- I just went through all of the ways to set up a PR I could think of and none of them gave me that drop down on the button.

versions/3.1.0.md
```diff
@@ -2290,7 +2288,7 @@ $ref: definitions.yaml#/Pet
 
 #### <a name="schemaObject"></a>Schema Object
 
-The Schema Object allows the definition of input and output data types. These types can be objects, but also primitives and arrays. This object is a superset of the [JSON Schema Draft 2019-09](http://json-schema.org/) specification.
+The Schema Object allows the definition of input and output data types. These types can be objects, but also primitives and arrays. This object is a superset of the [JSON Schema Specification Draft 2019-09](http://json-schema.org/) specification.
```
MikeRalphson: Can we lose the final "specification"?

philsturgeon: I think it's correct the way it is. We don't shove Specification into the name of our project we just accept that its a project which happens to have a specification. :)

MikeRalphson: Do you mean this is ok

> This object is a superset of the JSON Schema **Specification** Draft 2019-09 **specification**.

Or that the link should not be changed?

handrews: @MikeRalphson @philsturgeon I changed it to make it consistent with all of the other references in the OAS spec.  I don't care whether the word is there or not, but it should either always be there or never be there.

versions/3.1.0.md
```diff
@@ -170,12 +170,12 @@ The formats defined by the OAS are:
 Throughout the specification `description` fields are noted as supporting CommonMark markdown formatting.
 Where OpenAPI tooling renders rich text it MUST support, at a minimum, markdown syntax as described by [CommonMark 0.27](https://spec.commonmark.org/0.27/). Tooling MAY choose to ignore some CommonMark features to address security concerns. 
 
-### <a name="relativeReferences"></a>Relative References in URLs
+### <a name="relativeReferences"></a>Relative References in URIs
 
-Unless specified otherwise, all properties that are URLs MAY be relative references as defined by [RFC3986](https://tools.ietf.org/html/rfc3986#section-4.2).
-Relative references are resolved using the URLs defined in the [`Server Object`](#serverObject) as a Base URI.
+Unless specified otherwise, all properties that are URIs MAY be relative references as defined by [RFC3986](https://tools.ietf.org/html/rfc3986#section-4.2).
+Relative references are resolved using the URLs defined in the [`Server Object`](#serverObject) as a Base URI.  This includes relative references in [`Reference Objects`](#referenceObjects), outside of any [`Schema Object`](#schemaObject)l
```
darrelmiller:
```suggestion
Relative references are resolved using the URLs defined in the [`Server Object`](#serverObject) as a Base URI.  This includes relative references in [`Reference Objects`](#referenceObjects), outside of any [`Schema Object`](#schemaObject).
```

versions/3.1.0.md
```diff
@@ -170,12 +170,12 @@ The formats defined by the OAS are:
 Throughout the specification `description` fields are noted as supporting CommonMark markdown formatting.
 Where OpenAPI tooling renders rich text it MUST support, at a minimum, markdown syntax as described by [CommonMark 0.27](https://spec.commonmark.org/0.27/). Tooling MAY choose to ignore some CommonMark features to address security concerns. 
 
-### <a name="relativeReferences"></a>Relative References in URLs
+### <a name="relativeReferences"></a>Relative References in URIs
 
-Unless specified otherwise, all properties that are URLs MAY be relative references as defined by [RFC3986](https://tools.ietf.org/html/rfc3986#section-4.2).
-Relative references are resolved using the URLs defined in the [`Server Object`](#serverObject) as a Base URI.
+Unless specified otherwise, all properties that are URIs MAY be relative references as defined by [RFC3986](https://tools.ietf.org/html/rfc3986#section-4.2).
+Relative references are resolved using the URLs defined in the [`Server Object`](#serverObject) as a Base URI.  This includes relative references in [`Reference Objects`](#referenceObjects), outside of any [`Schema Object`](#schemaObject).
```
MikeRalphson: The build failure here is related to the `#referenceObjects` plural, which I think should be singular.
```suggestion
Relative references are resolved using the URLs defined in the [`Server Object`](#serverObject) as a Base URI.  This includes relative references in [`Reference Objects`](#referenceObject), outside of any [`Schema Object`](#schemaObject).
```

MikeRalphson: @handrews suggestion for ya.

versions/3.1.0.md
philsturgeon: I misread the diff @MikeRalphson, what you said is correct. 👇 


```suggestion
The Schema Object allows the definition of input and output data types. These types can be objects, but also primitives and arrays. This object is a superset of the [JSON Schema Specification Draft 2019-09](http://json-schema.org/).
```

PR #2115, opened 2020-01-23
Align the OAS 3.1 nullable language with the 3.0.3

handrews:
This adapts the language from PR #2046, with minimal wording tweaks
to account for type now being able to have multiple values (type arrays).

This is exactly the change from the suggestion at https://github.com/OAI/OpenAPI-Specification/pull/1977/files#diff-b83206d953b4be5d312b068f82ba0217R2295 which @tedepstein approved but I didn't realize needed to be actually merged in.  Because I can't GitHub, apparently.


tedepstein:
@handrews , thanks for taking this up. The final language we landed on in 3.0.3 is in commit 7fd23cd743ca596c9c6faaaf571a17d55d36c80a:

> A `true` value adds `"null"` to the allowed type specified by the `type` keyword, only if `type` is explicitly defined within the same Schema Object. Other Schema Object constraints retain their defined behavior, and therefore may disallow the use of `null` as a value. A `false` value leaves the specified or default `type` unmodified. The default value is `false`.

It needs to be adapted a bit for 3.1, because we now allow type arrays. I think we also agreed to deprecate `nullable` in 3.1.

> A `true` value adds `"null"` to the allowed types specified by the `type` keyword, only if `type` is explicitly defined within the same Schema Object. Other Schema Object constraints retain their defined behavior, and therefore may disallow the use of `null` as a value. A `false` value leaves the specified or default `type` unmodified. The default value is `false`.  The `nullable` keyword is **deprecated** and may be removed in a future revision of the OpenAPI specification, as JSON Schema type arrays are now supported.

handrews:
@tedepstein I think I have your language in there now.  I kept the first line as you didn't seem to be replacing it.  The deprecation text was identical.

tedepstein:
@handrews , I think the first line shouldn't be there. 

> Allows sending a `null` value for the defined schema. Default value is `false`.

We intentionally replaced that first sentence with the more nuanced language, because "Allows sending a `null` value" sounds like an expanding assertion, which is the perception we are trying to avoid. 

Also "Default value is 'false'" duplicates the last sentence of the new language (before the deprecation notice).

handrews:
@tedepstein ok fixed.

tedepstein:
@handrews , thanks! 

BTW, how are these changes happening without creating new commits? Some kind of exotic Git-Fu, like squashing, or rebasing? It's messing with my git-tuition!

handrews:
@tedepstein oh sorry about that.  I tend to essentially squash and force-push if there doesn't seem to be any point in keeping the history (such as for this very small change).  I usually do it because I never want to automatically squash commit histories (I often construct them very carefully for large changes), and because of that I never remember to manually squash-on-merge.

I can leave the commit trail next time.

handrews:
It does mess folks up if they pull the branch down, I've just usually worked on projects where no one seems to do that.

tedepstein:
@handrews, no worries, I was just interested to know how and why the commit history was missing from this task branch. I know different projects employ a variety of different Git usage patterns. I know what to expect in my own projects, but it's always informative to learn how other folks are doing it.

philsturgeon:
@darrelmiller @OAI/tsc can we get this one merged? It uses the language from 3.0.3 so is not controversial.

Issue #1635, opened 2018-07-17
Proposal: support fragment identifier in path object


link89:
OpenAPI spec is design for modeling REST APIs, but in real world we may find some APIs are mixture of restful style and RPC style.
Those RPC APIs may share same URI, but their schema or content of their request body or query parameters are different.
It's common to meet this problem when you try to model APIs that have been in your company for a long time.
As far as I know OpenAPI spec cannot model those kind of APIs properly because the combination of verb+path uniquely identify an operation.
I found that there are other people have similar problem with us:

https://github.com/swagger-api/swagger-core/issues/935
https://github.com/swagger-api/swagger-editor/issues/854

But we found that this problem can be solved by simply adding [fragment identifier](https://en.wikipedia.org/wiki/Fragment_identifier) in path objects.

Here is the example. There are some RPC style APIs sharing same path /api/team/{team_id} and verb "POST". Their request body may look like
```javascript
// Add members
{"operation": "add members", "data": {"members": [1, 2]}}

// Remove members
{"operation": "remove members", "data": {"members": [3, 4]}}
```
(I know they are bad designed, but they do exist and we have to model them.) By default, tools like Swagger-UI/Editor or swagger code generators cannot handle this case properly because their path and verb are same. In order to model those APIs, we add a fragment identifier after path for each RPC APIs
```yaml
/api/team/{team_id}#AddMember:
    post:
        ...
/api/team/{team_id}#RemoveMember:
    post:
        ...
```
And Swagger-UI/Editor will treat them as different paths after we add this fragment, so that we can now model these RPC APIs like other restful ones in swagger editor.
Because standard HTTP clients (browsers, ajax, urllib/requests in python, etc) will automatically omit fragment in url when making reqeusts, even we add fragment in path object, tools like Swagger-UI/Editor or clients created from swagger code genenerators can still work properly.
And that's all we need to do.

Accoding to wikiepdia
```
In computer hypertext, a fragment identifier is a short string of characters that refers to a resource that is subordinate to another, primary resource. The primary resource is identified by a Uniform Resource Identifier (URI), and the fragment identifier points to the subordinate resources
```

I think it is reasonable to use fragment identifier this way in OpenAPI. 
And I hope this proposal can become parts of OpenAPI spec to help other people who have similar problems.

silkentrance:
The fragment identifier, correct me if I am wrong, is something that gets evaluated on the client side and not on the server.

[...]

PR #2123, opened 2020-02-02
Paths MAY contain fragments

link89:
This PR is based on discussion in https://github.com/OAI/OpenAPI-Specification/issues/1635

Paths MAY contain fragments as defined in [RFC3986](https://tools.ietf.org/html/rfc3986#section-3.5). Paths with different fragments are considered different.

This feature can be used to further divide resources under same endpoint.

### DIFF
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -657,7 +657,7 @@ The path is appended to the URL from the [`Server Object`](#serverObject) in ord
 
 Field Pattern | Type | Description
 ---|:---:|---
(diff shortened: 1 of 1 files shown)


handrews:
This doesn't seem to address @darrelmiller's [comment](https://github.com/OAI/OpenAPI-Specification/issues/1635#issuecomment-441103603) about how a server is supposed to understand this.  One thing paths are for is server-side routing, and servers don't see fragments and therefore would be unable to distinguish your two examples.

link89:
@handrews I think [Specification Extensions](https://github.com/OAI/OpenAPI-Specification/blob/v3.1.0-dev/versions/3.1.0.md#specificationExtensions) can be used to solve this.

handrews:
@link89 I don't see how.  The problem is that the fragment doesn't get sent to the server.  I mean, I guess if the point is to completely ignore the OAS spec on the server and only use it for documentation, then I suppose this is useful, but it definitely seems odd.  And makes it even more annoying that query params can't be managed in the path, although that's not your/this PR's problem.

handrews:
I have elaborated on my concerns in the issue.  I do not think this topic is ready for a PR.

ioggstream:
@link89 Can we close? I think the use-case is real, though, and it deserves a more engineered solution.

Issue #2126, opened 2020-02-05
PathItem definition misuses patternProperties


axelcostaspena:
IMHO the PathItem definition abuses of the patternProperties feature. Since the list of valid Operation properties is kown and finite, I think that we should use specific properties for each one.

Advantages:

* Defined properties help code tools to be smarter and to assist better in coding. If we define those properties specifically, editors which use an improved language completion based on schema will offer a better completion experience - IntelliJ IDEA only offers validation for this, but after improving the schema as per the PR I will attach, it does offer completion for get|put|post... - and this is the reason nº1 for this improvement.
* Offers a clearer undesrtanding of the allowed properties when reading the schema, with plain full-matching literals instead of a regex.
* Avoids misuse of the patternProperties feature, which IMHO is for the need to allow an open set of constrained properties matching a pattern, or for a defined huge set of properties which can be expressed in a form of a pattern.
* Schema is ready for future diferent Operation specifications from different request methods - maybe?

Disadvantages:

* Longer schema.
* Somewhat repeating code.

Attaching PR immediately.

PR #2127, opened 2020-02-05
Make PathItem operation properties specific

axelcostaspena:
PR for #2126.

Move the Operations properties definitions from `patternProperties` to the `properties` set, explicitly defining all of them. The goal is to get better coding tools completion - IntelliJ IDEA does - and to improve the spec quality with a purer spec schema usage.

Changes are 100% backwards compatible, all validating schemas will validate and all non validating schemas won't validate after merge. After merge, coding tools will get more information because of replacing a pattern with a defined literal set.

### DIFF
--- a/schemas/v3.0/schema.yaml
+++ b/schemas/v3.0/schema.yaml
@@ -495,6 +495,22 @@ definitions:
         type: string
       description:
         type: string
+      get:
+        $ref: '#/definitions/Operation'
+      put:
+        $ref: '#/definitions/Operation'
+      post:
+        $ref: '#/definitions/Operation'
+      delete:
+        $ref: '#/definitions/Operation'
+      options:
+        $ref: '#/definitions/Operation'
+      head:
+        $ref: '#/definitions/Operation'
+      patch:
+        $ref: '#/definitions/Operation'
+      trace:
+        $ref: '#/definitions/Operation'
       servers:
         type: array
         items:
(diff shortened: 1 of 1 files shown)


handrews:
As the primary author of recent drafts of the JSON Schema spec, this is not an abuse of `patternProperties`.  It's certainly valid to talk about what is most useful for tools, but it's a valid and useful technique as written.

axelcostaspena:
@handrews your opinion is welcome :)

Never mind, #2126 specifically states that its nº 1 goal is better tool support. Also, commit message first line is about specificity.

axelcostaspena:
@MikeRalphson hmm sorry! Issue sems worse since the modification comes **FROM** OpenApi.next, which is unmerged into master. So let me solve it, I will create proper PR. But which work branch to choose? DEVELOPMENT.md:47 and README.md:24 seem to tell me to use 3.1.0-dev - non breaking change - instead of 3.0.3-dev - typos or clarifications - but none of that branch contain the v3 schema ¿?

MikeRalphson:
That guidance is for changes to the specification. The schema is non-normative, so I think `master ` should be OK.

axelcostaspena:
@MikeRalphson PR source force updated
PR target updated.
Thank you for support on this.

karenetheridge:
Should this also be merged into v3.1/schema*, as the same keyword usage exists there?

webron:
Wouldn't hurt...

PR #2147, opened 2020-02-19
Do not force "" for http status code

pkuczynski:
In JSON the key which is a number must be wrapped into "". This is not the case of yaml format. Different formats, different requirements. It does not make sense to force json requirements into yaml.

Please also note that https://editor.swagger.io does not wrap status codes into "", which is correct, but contradictory to this spec:


webron:
As a general rule, we're looking for compatibility between JSON and YAML representations of API definitions, and have added some restrictions in the spec to support it. Since in YAML keys are *typed*, the following is legal:
```yaml
responses:
  "400": {...}
  400: {...}
```
and doesn't really have a JSON equivalent. Even in YAML, that can cause parsing issues and we don't want to complicate parsing. 

The fact that Swagger Editor does not complain about it is an issue with the tool and should be filed there.

pkuczynski:
This makes it really annoying to have to use `"400": ...` in YAML :(

webron:
I feel your pain (seriously). Regardless, this is definitely not a change we can do in 3.0.x because it changes the behavior of the spec.

pkuczynski:
Well, I didnt know how to propose it for the next version?

PR #2148, opened 2020-02-20
OAS v3.0.3 Release

webron:
As a patch release, the following changes were made to improve the specification in terms of readability and accuracy. None of these modifications change the behavior of the spec.

- Clarified how Path Templating works.
- Clarified the meaning of Semantic Versioning as it applies to the OpenAPI Specification (note, this is the `openapi` field, not the `version` field).
- Changed some hyperlinks from `http` to `https`.
- Clarified add the notion of optional security on operations.
- Added an explanation that the `Server Variable Object`'s `enum` should not be empty. This is not a breaking change but should be considered as guidance for a more explicit restriction in the next major version.
- Clarified paths under the `Paths Object` should start with a forward slash.
- Clarified `Path Item Object`'s `$ref` behavior with sibling fields.
- Fixed a few examples.
- Clarified the map structure of `callbacks` under the `Operation Object`.
- Clarified how path parameters are being matched.
- Clarified `example`/`examples` value(s) in the `Parameter Object`.
- Fixed example for `pipeDelimited` `object` value.
- Fixed `Callback Object` description.
- Clarified `Link Object`'s `operationDef`'s description.
- Improved ABNF for `Runtime Expressions`.
- Clarified valid regex for `pattern` under `Schema Object`.
- Clarified the behavior of `nullable` under `Schema Object`.
- Fixed names of OAuth2 flows in the description of `Security Scheme Object`.
- Improved description of `Security Filtering` section.


Blessedlybroken:
Debug

PR #2181, opened 2020-03-19
Allow summary and description as $ref siblings outside schema objects

MikeRalphson:
As discussed re: 3.1-RC0

* I'm not wedded to either the "SHOULD by default" override wording or the "has no effect" behaviour (i.e. we could outlaw it instead) but at least they are placeholder answers to questions we should address.
* By "object-type" I was trying to clearly indicate the target OAS object, and whether it defines a summary/description field as distinct from whether the referenced component has a summary/description property in the OAS document instance.

### DIFF
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -2251,6 +2251,8 @@ A simple object to allow referencing other components in the OpenAPI document, i
 Field Name | Type | Description
 ---|:---:|---
 <a name="referenceRef"></a>$ref | `string` | **REQUIRED**. The reference string.
+<a name="referenceSummary"></a>summary | `string` | A short summary which by default SHOULD override that of the referenced component. If the referenced object-type does not define a `summary` field, then this field has no effect.
+<a name="referenceDescription"></a>description | `string` | A description which by default SHOULD override that of the referenced component. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation. If the referenced object-type does not define a `description` field, then this field has no effect.
 
 This object cannot be extended with additional properties and any properties added SHALL be ignored.
 


handrews:
I'm fine with the SHOULD, although may I suggest leaving `summary` as override and considering the option of concatenation in `description`?  I do not feel at all strongly about this and am fine with override.  I've just previously done things where for `description` instead of overriding, it appends a paragraph or paragraphs to further explain.

I am puzzled by the "has no effect", though.  Adding a `summary` or `description` at point of reference to a structure that is too generic to have a sensible default for those keywords is an important use case for me.  Although maybe I'm not thinking the non-schema cases through properly?

MikeRalphson:
> I am puzzled by the "has no effect", though. Adding a summary or description at point of reference to a structure that is too generic to have a sensible default for those keywords is an important use case for me. Although maybe I'm not thinking the non-schema cases through properly?

This is not about the case where (say) a referenced `parameter` component does not have a `description` **property**, but the case where a ref to a `parameter` component has a `summary` property, but the `parameter` object itself (structurally) does not have such a **field**. I guess I was right that it needs more clarity :smile:

handrews:
Oh now I see how "object-type" fits in.  But yeah, my brain glossed over that even after reading your PR description where you called it out.

tedepstein:
Any thoughts about allowing other annotations, like `x-` specification extensions and `externalDocs`?

MikeRalphson:
@tedepstein see https://gist.github.com/MikeRalphson/d3b6bdbc3c4a9f2e69310a356824e174 for the current state of what we're thinking re; RC0 and possible subsequent RCs.

What would be the use-case for `externalDocs` (i.e. which component would be being referenced, a `pathItem`)?

Issue #1547, opened 2018-04-17
[oas3] base64 vs byte


jwalton:
The spec makes numerous references to `{format: base64}` ([here](https://github.com/OAI/OpenAPI-Specification/blob/fbe62006211838a8bb7bf2433a1d15f1a5838a03/versions/3.0.1.md#considerations-for-file-uploads) and [here](https://github.com/OAI/OpenAPI-Specification/blob/fbe62006211838a8bb7bf2433a1d15f1a5838a03/versions/3.0.1.md#special-considerations-for-multipart-content) for example).  However this format is not defined in either [the wright-00 JSON-schema spec](https://tools.ietf.org/html/draft-wright-json-schema-validation-00#section-7.3), nor in [the OAS3 defined data types](https://github.com/OAI/OpenAPI-Specification/blob/fbe62006211838a8bb7bf2433a1d15f1a5838a03/versions/3.0.1.md#data-types).

handrews:
@jwalton `format` is extensible and OpenAPI defines numerous extensions including `base64`, although per #1310 and #1544, `base64` should be handled differently in the long run with newer versions of JSON Schema.

But as far as the JSON Schema spec goes, what OpenAPI is doing with formats is valid.

jwalton:
Oh, it's valid, it's just that [the defined data types](https://github.com/OAI/OpenAPI-Specification/blob/fbe62006211838a8bb7bf2433a1d15f1a5838a03/versions/3.0.1.md#data-types) section defines `{format: 'byte'}` as "base64 encoded characters".  It seems weird to define our own special name for base64 and then not use it.  :P

darrelmiller:
As per @tsc meeting we are going make a decision between have a `byte` format or a `base64` but not both.  Some historical spelunking will be done by @webron before making the call.

jwalton:
FWIW, my vote is for `base64`.  It much more clearly conveys the intent.

ralfhandl:
I also prefer `base64`

handrews:
In OAS 3.1, this can now be handled properly from the JSON Schema perspective.  Specifically:

```YAML
type: string
contentEncoding: base64
```
is a base64-endcode string (where base64 is per RFC 4648).  You can use any RFC 4648 or RFC 2045 encoding, so `base64url` or `quoted-printable` are also valid.

If you are encoding a specific media type, such as a PNG file:
```YAML
type: string
contentEncoding: base64
contentMediaType: image/png
```

[...]

PR #2191, opened 2020-04-02
Change base64 to byte in examples

smhc:
format is 'byte' not 'base64'.

### DIFF
--- a/versions/3.0.3.md
+++ b/versions/3.0.3.md
@@ -1446,7 +1446,7 @@ In contrast with the 2.0 specification, `file` input/output content in OpenAPI i
 # content transferred with base64 encoding
 schema:
   type: string
-  format: base64
+  format: byte
 ```
 
 ```yaml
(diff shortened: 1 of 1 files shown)


hkosova:
`base64` isn't necessarily a mistake. There's an open issue #1547 asking to clarify the difference between `byte` and `base64`.

smhc:
Comments in that referenced github issue state:

"OpenAPI defines numerous extensions including base64"

However, in the specification it has a section which lists the formats defined by OAS:


----

The formats defined by the OAS are:

[`type`](#dataTypes) | [`format`](#dataTypeFormat) | Comments
------ | -------- | --------
`integer` | `int32` | signed 32 bits
`integer` | `int64` | signed 64 bits (a.k.a long)
`number` | `float` | |
`number` | `double` | |
`string` | | |
`string` | `byte` | base64 encoded characters
`string` | `binary` | any sequence of octets
`boolean` | | |
`string` | `date` | As defined by `full-date` - [RFC3339](https://xml2rfc.ietf.org/public/rfc/html/rfc3339.html#anchor14)
`string` | `date-time` | As defined by `date-time` - [RFC3339](https://xml2rfc.ietf.org/public/rfc/html/rfc3339.html#anchor14)
`string` | `password` | A hint to UIs to obscure input.

----

in this table, 'base64' is not mentioned. So it there an inconsistency there.
If 'base64' is predefined by OAS it should be listed in this table.

handrews:
This is being addressed as part of PR #2200, which drops `"format": "byte"` entirely in favor of `"contentEncoding": "base64"`.  `contentEncoding` is more flexible, and compatible with the latest JSON Schema (which will be used in OAS 3.1).

If there will be an OAS 3.0.4, then a decision on `byte` vs `base64` is needed, as tracked by #1547, followed by a PR against the appropriate branch.  That issue still has not resolved which of `byte` or `base64` was intended.  Since this is not against the right branch for 3.0.4 and it's not clear which value is correct, I'm closing this.  We'll need a PR against the right branch depending on the outcome of #1547.

PR #2192, opened 2020-04-02
Fixed format of 'base64' to correct 'byte'

smhc:


hkosova:
@handrews
> I'm not sure if we have an issue tracking that, if not I'll file one today.

#1547?

handrews:
@hkosova that's at least related.  I didn't manage to file an issue but I discovered a 2/3rds done PR in my local repo.  Wonder when I did that... I'll try to post that soonish.

handrews:
@smhc all of this stuff is in the middle of being changed around for OAS 3.1, that's why it doesn't make any sense right now.

handrews:
Let me finish updating this section with my PR that aligns it with previous work done to update OAS 3.1 to use the latest JSON Schema.  If it still doesn't make sense, let's work through it then.

handrews:
This is being addressed as part of PR #2200, which drops `"format": "byte"` entirely in favor of `"contentEncoding": "base64"`.  `contentEncoding` is more flexible, and compatible with the latest JSON Schema (which will be used in OAS 3.1).

If there will be an OAS 3.0.4, then a decision on `byte` vs `base64` is needed, as tracked by #1547, followed by a PR against the appropriate branch.  That issue still has not resolved which of `byte` or `base64` was intended.  Since this is not against the right branch for 3.0.4 and it's not clear which value is correct, I'm closing this.  We'll need a PR against the right branch depending on the outcome of #1547.

PR #2200, opened 2020-04-08
Update "format" and "content*" for new JSON Schema

### LINKED ISSUE #1547: [oas3] base64 vs byte

### LINKED ISSUE #2093: Clarify text relating to formats due to dependency on JSON Schema 2019-09
Review if it is necessary to point out that JSON Schema 2019-09 doesn't use formats for validation by default.
Clarify that format is an annotation not a constraint

handrews:
Closes #2093.  Probably also closes #1547, #2191, #2192.

This removes OAS formats and examples that are now superfluous
as they are part of the 2019-09 JSON Schema draft, and notes
that "format" is not validated by default.

Similarly it deprecates the "byte" and "binary" formats in favor
of JSON Schema's "contentEncoding" and "contentMediaType" keywords,
and updates various related examples and other guidance.

Note that I had to get a bit creative interpreting how some of the
"content*" keywords might change things, and may have gone
in the wrong direction, so please review those parts carefully.

It also removes confusingly blank rows in the OAS format table.


versions/3.1.0.md
```diff
@@ -147,12 +147,14 @@ Primitive data types in the OAS are based on the types supported by the [JSON Sc
...
+OAS uses several known additional formats to define in fine detail the data type being used.
 However, to support documentation needs, the `format` property is an open `string`-valued property, and can have any value.
-Formats such as `"email"`, `"uuid"`, and so on, MAY be used even though undefined by this specification.
+Additional formats MAY be used even though undefined by either JSON Schema or this specification.
 Types that are not accompanied by a `format` property follow the type definition in the JSON Schema. Tools that do not recognize a specific `format` MAY default back to the `type` alone, as if the `format` is not specified.
 
+Note that by default, JSON Schema validators MUST NOT attempt to validate the `format` keyword.  It is primarily treated as an annotation that informs applications of the intended nature of the data.  For historical reasons, it is possible to configure JSON Schema validators to validate "format", but this is not consistently implemented and SHOULD NOT be relied upon in any environment where an unknown validator or unknown validator configuration might be used.
```
jcarres-mdsol, line 156: In the company I work for we use UUIDs to identify resources so `type: string  format: uuid` is super common in our schemas. And we do leverage that information for a number of things. 
It seems that this note is motivated by the current implementations in the wild, which does not seem like a good reason to include it to a standard. If these validators start improving would this note go away? At what point? How many good validators out there would it make it happen?

MikeRalphson, line 156: Is there a reason for the double-quotes as opposed to backticks on the last mention of `"format"`?

versions/3.1.0.md
```diff
@@ -1453,14 +1455,14 @@ In contrast with the 2.0 specification, `file` input/output content in OpenAPI i
...
-  format: base64
+  contentEncoding: base64
 ```
 
 ```yaml
 # content transferred in binary (octet-stream):
 schema:
   type: string
-  format: binary
+  contentMediaType: application/octet-stream
```
MikeRalphson, line 1465: Should be include a `mediaTypeObject` in one of these examples? Is that the preferred OAS way to indicate media-type? What if there is a conflict between the `mediaTypeObject` key and a `contentMediaType` keyword?

versions/3.1.0.md
handrews: @jcarres-mdsol `uuid` and all other formats that are no longer explicitly listed are now part of JSON Schema proper.  See [Section 7.3.5 "Resource Identifiers"](https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.7.3.5) under the `format` keyword's specification for `uuid` and also how it relates to URNs of the form `urn:uuid:`.

handrews: @jcarres-mdsol As for validators actually validating `format`, no, that's not going to happen.  I don't mean that in a "we (the JSON Schema project) decided it shouldn't happen", I mean it in "this has been a problem for years, and after talking with numerous maintainers of major implementations, it is 100% clear that `format` is a confusing mess that will not be validated consistently no matter how much the spec says it should be."  At some point you have to bow to reality and figure out a way forward, and saying that by default implementations MUST NOT attempt to validate using the `format` keyword was the only possible way to produce consistent default behavior.  This is [explained in detail](https://json-schema.org/draft/2019-09/release-notes.html#format-vocabulary) in the JSON Schema draft's release notes.

The problem with `format` is that it tries to do a huge number of things, *and* be open-ended to any extension anyone wants to throw in there.  There's not even agreement on what fully validating `format` would involve.  My hope is that with re-usable extension vocabularies being a thing, people will design vocabularies with keywords that validate specific things, instead of shoving them into this all-purpose-but-never-reliable keyword.

So an `isUuid` keyword would be straightforward to implement b/c UUIDs are well-specified.  For something like `email`, where there's no consensus on what "valid email" means, people could create keywords with specific criteria.  `isProbablyEmail` might just check for an `@` in the string (this is what several validators actually do right now).  `isEmailByRegex` might specify validation against a specific regex (instead of what happens now, which is that each validator comes up with its own incredibly complex regex that they probably got from a web site somewhere).  `isEmailBySomeLibraryName` would define what library to use to validate, etc.

These keywords would then be supported or not, which could be detected/enforced using the `$vocabulary` mechanism, and would reliably provide whatever level of validation they specify.  You could require support for `isUuid` but not require any sort of email validator, for example- that would be likely to be supported.  As would keywords for RFC 3339 date and time values, which are likewise well-specified.

Ultimately, I hope to deprecate `format` because it's always been a compatibility nightmare.  But that will only happen if enough replacement vocabularies are developed and adopted widely enough to be a viable migration path.

handrews: Nope, I'll fix that, thanks!

versions/3.1.0.md
handrews: > What if there is a conflict between the `mediaTypeObject` key and a `contentMediaType` keyword?

So.... technically I would *not* use the schema in this example in the first place for a purely binary resource.  Because what this means is that there is a JSON string (double-quoted) with binary content, which doesn't actually make any sense.  You would have to encode the binary content into a legal JSON string (as indicated by `contentEncoding`).

For pure binary content, I would just us a Media Type Object without a schema.

Does that make sense?  I didn't want to start revamping larger things without first trying the minimal changes, as I assume this approach (`{"type": "string", "format": "binary"}`) has more or less worked for OAS so far.  But I wouldn't actually advise it, personally.

versions/3.1.0.md
jcarres-mdsol: Thanks for the detailed response. Hopefully the plan will go as expected. I'm worried about the complexity of all this, I wonder if people will shy away and just keep shoveling everything into format anyways. But time will tell. 
Thanks for your efforts!

handrews: @jcarres-mdsol we're definitely starting to see people understand the utility of the vocabulary concept on our slack.  But really, if enough people accept vocabularies as a solution here, we'll deprecate/remove `format`.  If they don't, we won't.

versions/3.1.0.md
handrews: @MikeRalphson @webron should I make the larger change here (to not use a Schema Object for binary content, and just use a Media Type Object)?

versions/3.1.0.md
```diff
@@ -147,12 +147,14 @@ Primitive data types in the OAS are based on the types supported by the [JSON Sc
...
+OAS uses several known additional formats to define in fine detail the data type being used.
 However, to support documentation needs, the `format` property is an open `string`-valued property, and can have any value.
-Formats such as `"email"`, `"uuid"`, and so on, MAY be used even though undefined by this specification.
+Additional formats MAY be used even though undefined by either JSON Schema or this specification.
 Types that are not accompanied by a `format` property follow the type definition in the JSON Schema. Tools that do not recognize a specific `format` MAY default back to the `type` alone, as if the `format` is not specified.
 
+Note that by default, JSON Schema validators MUST NOT attempt to validate the `format` keyword.  It is primarily treated as an annotation that informs applications of the intended nature of the data.  For historical reasons, it is possible to configure JSON Schema validators to validate `format`, but this is not consistently implemented and SHOULD NOT be relied upon in any environment where an unknown validator or unknown validator configuration might be used.
```
darrelmiller, line 156:
```suggestion
Note that by default, JSON Schema validators will not attempt to validate the `format` keyword.  https://json-schema.org/draft/2019-09/release-notes.html#format-vocabulary
```

versions/3.1.0.md
```diff
@@ -161,14 +163,14 @@ The formats defined by the OAS are:
...
 `number` | `double` | |
-`string` | | |
-`string` | `byte` | base64 encoded characters
-`string` | `binary` | any sequence of octets
-`boolean` | | |
-`string` | `date` | As defined by `full-date` - [RFC3339](https://xml2rfc.ietf.org/public/rfc/html/rfc3339.html#anchor14)
-`string` | `date-time` | As defined by `date-time` - [RFC3339](https://xml2rfc.ietf.org/public/rfc/html/rfc3339.html#anchor14)
 `string` | `password` | A hint to UIs to obscure input.
+`string` | `byte` | base64 encoded characters **Deprecated**
+`string` | `binary` | any sequence of octets **Deprecated**
```
darrelmiller, line 168: Should these simply be removed?  What would the impact be on other content-types?  If someone is using JSON Schema to infer schema on other non-JSON formats, what does format even mean?

versions/3.1.0.md
```diff
@@ -1474,7 +1476,7 @@ requestBody:
       schema:
```
darrelmiller, line 1476: Should this schema be removed?

versions/3.1.0.md
```diff
@@ -1487,11 +1489,11 @@ requestBody:
     'image/jpeg':
       schema:
```
darrelmiller, line 1490: Schemas should be removed.

versions/3.1.0.md
darrelmiller: Personally, I think we can remove these.

handrews:
I've updated the first part of the file (and rebased the whole change- you can see the update on its own in the last commit).

I have not yet updated the part around the Encoding Object and the examples because I'm having some trouble figuring out how to get it to work consistently in both the Media Type Object and the Encoding Object scenarios, and with the complex defaulting of the Encoding Object content-type based on various schema characteristics.

I've made my brain hurt enough for today, I'll try to sort it out or at least explain the issues tomorrow.

PR #2216, opened 2020-04-26
Improve "Discriminator object" section of the OAS specification.

sebastien-rosset:
This PR attempts to clarify the **Discriminator Object** section of OAS 3.1. Discriminators seem to cause a number of questions, confusion and interoperability issues. I think some of these questions could be addressed by formulating the normative statements more precisely. I am aware of #2143 but a goal of deprecating the discriminator object is not in conflict with improving the existing specification.

### DIFF
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -2698,6 +2698,7 @@ A payload MAY be described to be exactly one of any number of types:
 ```yaml
 MyType:
   oneOf:
+  - type: 'null'
   - $ref: '#/components/schemas/Cat'
   - $ref: '#/components/schemas/Dog'
   - $ref: '#/components/schemas/Lizard'
@@ -2709,6 +2710,7 @@ which means the payload _MUST_, by validation, match exactly one of the schemas 
 ```yaml
 MyType:
   oneOf:
+  - type: 'null'
   - $ref: '#/components/schemas/Cat'
   - $ref: '#/components/schemas/Dog'
   - $ref: '#/components/schemas/Lizard'
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -2677,7 +2677,9 @@ components:
 When request bodies or response payloads may be one of a number of different schemas, a `discriminator` object can be used to aid in serialization, deserialization, and validation.  The discriminator is a specific object in a schema which is used to inform the consumer of the document of an alternative schema based on the value associated with it.
 
 The discriminator object is legal only when using one of the composite keywords `oneOf`, `anyOf`, `allOf`. Both _inline_ and references are valid child schemas.
-When a child schema is a reference, the discriminator property _MUST_ be present in the payload. A discriminator MAY act as a "hint" to shortcut validation and selection of the matching schema which may be a costly operation, depending on the complexity of the schema. Even when the discriminator is used as shortcut hint, the payload MUST still be validated against the JSON schema.
+When a child schema is a reference, the discriminator property _MUST_ be present in the payload.
+A discriminator MAY act as a "hint" to shortcut validation and selection of the matching schema which may be a costly operation, depending on the complexity of the schema.
+Even when the discriminator is used as shortcut hint, the payload MUST still be validated against the JSON schema.
 When a child schema is specified _inline_, the discriminator property is not required in the payload. Validation is applied to determine if the payload matches the child JSON schema.
 
 The discriminator `propertyName` field MUST be a required field in the schema object. However, as shown below, the discriminator is not required to be present in the payload for _inline_ schemas.
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -2674,7 +2674,7 @@ components:
 
 #### <a name="discriminatorObject"></a>Discriminator Object
 
-When request bodies or response payloads may be one of a number of different schemas, a `discriminator` object can be used to aid in serialization, deserialization, and validation. The discriminator is a specific object in a schema which is used to inform the consumer of the document of an alternative schema based on the value associated with it.
+When request bodies or response payloads may be one of a number of different schemas, a `discriminator` object can be used to aid in serialization, deserialization, and validation.  The discriminator is a specific object in a schema which is used to inform the consumer of the document of an alternative schema based on the value associated with it.
 
 A discriminator MAY act as a "hint" to shortcut validation and selection of the matching schema which may be a costly operation, depending on the complexity of the schema.
 The discriminator `propertyName` field MUST be a required field in the schema object. However, as shown below, the discriminator is not required to be present in the payload for _inline_ schemas.
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -2674,9 +2674,14 @@ components:
 
 #### <a name="discriminatorObject"></a>Discriminator Object
 
-When request bodies or response payloads may be one of a number of different schemas, a `discriminator` object can be used to aid in serialization, deserialization, and validation.  The discriminator is a specific object in a schema which is used to inform the consumer of the document of an alternative schema based on the value associated with it.
+When request bodies or response payloads may be one of a number of different schemas, a `discriminator` object can be used to aid in serialization, deserialization, and validation. The discriminator is a specific object in a schema which is used to inform the consumer of the document of an alternative schema based on the value associated with it.
 
-When using the discriminator, _inline_ schemas will not be considered.
+A discriminator MAY act as a "hint" to shortcut validation and selection of the matching schema which may be a costly operation, depending on the complexity of the schema.
+The discriminator `propertyName` field MUST be a required field in the schema object. However, as shown below, the discriminator is not required to be present in the payload for _inline_ schemas.
+
+The discriminator object is legal only when using one of the composite keywords `oneOf`, `anyOf`, `allOf`. The child schemas MAY be a mix of _inline_ or references schemas.
+When a child schema is specified as a reference, the discriminator property _MUST_ be present in the payload.
+When a child schema is specified _inline_, the discriminator property is not required in the payload. Validation is applied to determine if the payload matches the child JSON schema.
 
 ##### Fixed Fields
 Field Name | Type | Description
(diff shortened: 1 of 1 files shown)


versions/3.1.0.md
```diff
@@ -2689,10 +2689,10 @@ This object MAY be extended with [Specification Extensions](#specificationExtens
 
 The discriminator object is legal only when using one of the composite keywords `oneOf`, `anyOf`, `allOf`.
 
-In OAS 3.0, a response payload MAY be described to be exactly one of any number of types:
+A payload MAY be described to be exactly one of any number of types:
```
sebastien-rosset, line 2692: Remove "In OAS 3.0":
1) It is the 3.1 spec, not 3.0
2) There is no need to start every sentence with "In OAS X.Y". The entire document is about a specific version of the OAS spec (in this case 3.1).

sebastien-rosset, line 2692: I suggest removing the word "response", because the discriminator can be used in the request and response.

versions/3.1.0.md
```diff
@@ -2675,9 +2674,14 @@ components:
 
 #### <a name="discriminatorObject"></a>Discriminator Object
 
-When request bodies or response payloads may be one of a number of different schemas, a `discriminator` object can be used to aid in serialization, deserialization, and validation.  The discriminator is a specific object in a schema which is used to inform the consumer of the document of an alternative schema based on the value associated with it.
+When request bodies or response payloads may be one of a number of different schemas, a `discriminator` object can be used to aid in serialization, deserialization, and validation. The discriminator is a specific object in a schema which is used to inform the consumer of the document of an alternative schema based on the value associated with it.
 
-When using the discriminator, _inline_ schemas will not be considered.
```
sebastien-rosset, line 2680: Problem 1: the verb "consider" is not conducive to precision: "to think carefully about (something), typically before making a decision." "take (something) into account when making an assessment or judgment."...

Problem 2: it’s not clear how this sentence interacts with other sentences in the spec. In particular, does the above sentence nullify the requirement for the discriminator property in the payload: 

> The expectation now is that a property with name `petType` MUST be present in the response payload.

versions/3.1.0.md
```diff
@@ -2675,9 +2674,14 @@ components:
...
+A discriminator MAY act as a "hint" to shortcut validation and selection of the matching schema which may be a costly operation, depending on the complexity of the schema.
```
sebastien-rosset, line 2679: I moved this sentence up in the section because it was in the middle of an example, but this sentence is always true regardless of the specific example that was provided (?).

What does the word MAY mean for validation exactly? Who decides whether the discriminator is used as a hint or not? It would be good to be completely unambiguous such that there is no discrepancy between OAS validators.

1. Loop through all _inline_ child schemas and apply JSON schema validation (because discriminators are not "considered" for inline schemas?)
1. Loop through all _referenced_ child schema and use discriminator as a hint to determine if payload matches child schema or not. Then is the validator supposed to validate the matching payload against the JSON schema. I would think the JSON schema should still be applied after the discriminator match, but the sentence seem to imply validation may be skipped.

versions/3.1.0.md
```diff
@@ -2675,9 +2674,14 @@ components:
...
+The discriminator `propertyName` field MUST be a required field in the schema object. However, as shown below, the discriminator is not required to be present in the payload for _inline_ schemas.
+
+The discriminator object is legal only when using one of the composite keywords `oneOf`, `anyOf`, `allOf`. The child schemas MAY be a mix of _inline_ or references schemas.
+When a child schema is specified as a reference, the discriminator property _MUST_ be present in the payload.
+When a child schema is specified _inline_, the discriminator property is not required in the payload. Validation is applied to determine if the payload matches the child JSON schema.
```
sebastien-rosset, line 2684: This is my attempt to clarify **When using the discriminator, _inline_ schemas will not be considered.** Since that sentence is ambiguous, I'm not sure what it really means. I wrote something in the hope that it can incrementally lead to a better wording of the spec. I am not claiming my proposed change is necessarily correct AS IS.

sebastien-rosset, line 2684: One special case is the "null" type.

```yaml
oneOf:
  - type: 'null'
  - $ref: ''

File: versions/3.1.0.md [hunk 5]
```diff
@@ -2334,7 +2334,6 @@ The OpenAPI Specification allows combining and extending model definitions using
 While composition offers model extensibility, it does not imply a hierarchy between the models.
 To support polymorphism, the OpenAPI Specification adds the `discriminator` field.
 When used, the `discriminator` will be the name of the property that decides which schema definition validates the structure of the model.
-As such, the `discriminator` field MUST be a required field.
```
sebastien-rosset, line 2337: I have move this statement in the "Discriminator Object" section. Grouping all the normative statements in one place helps to understand the specification and how multiple normative statements interact with each other.

versions/3.1.0.md
```diff
@@ -2687,23 +2691,23 @@ Field Name | Type | Description
 
 This object MAY be extended with [Specification Extensions](#specificationExtensions).
 
-The discriminator object is legal only when using one of the composite keywords `oneOf`, `anyOf`, `allOf`.
+##### Discriminator Object Example
```
sebastien-rosset, line 2694: Added this section title to clearly differentiate between the normative statements versus the example section.

versions/3.1.0.md
```diff
@@ -2677,7 +2676,14 @@ components:
 
 When request bodies or response payloads may be one of a number of different schemas, a `discriminator` object can be used to aid in serialization, deserialization, and validation.  The discriminator is a specific object in a schema which is used to inform the consumer of the document of an alternative schema based on the value associated with it.
 
-When using the discriminator, _inline_ schemas will not be considered.
+The discriminator object is legal only when using one of the composite keywords `oneOf`, `anyOf`, `allOf`. Both _inline_ and references are valid child schemas.
+When a child schema is a reference, the discriminator property _MUST_ be present in the payload.
+A discriminator MAY act as a "hint" to shortcut validation and selection of the matching schema which may be a costly operation, depending on the complexity of the schema.
```
sebastien-rosset, line 2681: I moved this sentence up in the section because it was in the middle of an example. Isn't this sentence always true regardless of the specific example that was provided? If so, I think it would make sense to put it outside the example section.

What does the word MAY mean for validation exactly? Who decides whether the discriminator is used as a hint or not? It would be good to be completely unambiguous such that there is no discrepancy between OAS validators.

What are the OAS validators supposed to do?

1. For _inline_ child schemas, do not use the discriminator as a hint and validate the payload using JSON schema validation. 
1. For _referenced_ child schemas, the discriminator MAY be used as a hint to quickly determine if the payload matches the child schema or not. If the payload does match the discriminator value, the payload MUST be validated against the JSON schema.

versions/3.1.0.md
```diff
@@ -2677,7 +2676,14 @@ components:
 
 When request bodies or response payloads may be one of a number of different schemas, a `discriminator` object can be used to aid in serialization, deserialization, and validation.  The discriminator is a specific object in a schema which is used to inform the consumer of the document of an alternative schema based on the value associated with it.
 
-When using the discriminator, _inline_ schemas will not be considered.
+The discriminator object is legal only when using one of the composite keywords `oneOf`, `anyOf`, `allOf`. Both _inline_ and references are valid child schemas.
```
sebastien-rosset, line 2679: I am proposing to clarify **When using the discriminator, _inline_ schemas will not be considered.** I am not claiming my proposed change is necessarily correct AS IS. The problem is that we find the spec ambiguous. I wrote something in the hope that it can incrementally lead to a better wording of the spec.

versions/3.1.0.md
```diff
@@ -2677,7 +2676,14 @@ components:
...
+When a child schema is a reference, the discriminator property _MUST_ be present in the payload.
+A discriminator MAY act as a "hint" to shortcut validation and selection of the matching schema which may be a costly operation, depending on the complexity of the schema.
+Even when the discriminator is used as shortcut hint, the payload MUST still be validated against the JSON schema.
+When a child schema is specified _inline_, the discriminator property is not required to be present in the payload. Validation is applied to determine if the payload matches the child JSON schema.
```
sebastien-rosset, line 2683: As a special case, what if all child schemas are specified inline? Does this mean the discriminator is completely ignored and has no effect?

sebastien-rosset, line 2683: This is a proposal to clarify **When using the discriminator, _inline_ schemas will not be considered.** I am not claiming my proposed change is necessarily correct AS IS. The problem is that we find that sentence ambiguous. I wrote something in the hope that it can incrementally lead to a better wording of the spec.

versions/3.1.0.md
```diff
@@ -2687,32 +2693,34 @@ Field Name | Type | Description
...
+##### Discriminator Object Example
 
-In OAS 3.0, a response payload MAY be described to be exactly one of any number of types:
+A payload MAY be described to be exactly one of any number of types:
 
 ```yaml
-MyResponseType:
+MyType:
   oneOf:
+  - type: 'null'
```
sebastien-rosset, line 2703: One special case is the "null" type, which is introduced in 3.1. I am suggesting to provide one "null" type example to present the implications of the "null" type when it is used along with a discriminator object and oneOf/anyOf. There is only one sentence in this 3.1 draft spec that mentions the "null" type. 

My understanding is this schema is supported, and it is the new recommended approach in lieu of the deprecated "nullable". But the "null" type does not have a discriminator field in its payload, so adding it as an example would help to clarify that it's ok to use.

On one hand, the "null" type does not have a discriminator value which breaks the normative statement about the discriminator being a required field. But on the other hand, it seems to be valid because of the following sentence: **When using the discriminator, inline schemas will not be considered.**

versions/3.1.0.md
sebastien-rosset, line 2679: The normative statements and examples in this section assume the child schemas are **object**. What if they are string, integer, boolean, array, number or "null" type? On one hand, these types are not allowed because **the `discriminator` field MUST be a required field**. On the other hand, it seems to be possible as _inline_ schemas, because **When using the discriminator, _inline_ schemas will not be considered**

versions/3.1.0.md
spacether: How about `one or more of the composite keywords`
You can have a schema that includes allOf and oneOf where the payload conforms to both definitions, the allOf validation passes and those properties are loaded in as additional properties in the chosen oneOf schema.

sebastien-rosset: > How about `using one of the composite keywords` -> `using one or more of the composite keywords`
> You can have a schema that includes allOf and oneOf where the payload conforms to both definitions, the allOf validation passes and those properties are loaded in as additional properties in the chosen oneOf schema.

Thank you I have updated the sentence accordingly.

PR #2243, opened 2020-05-27
[3.1.0-dev] drop OAS semver requirement

MikeRalphson:
First pass at removing semver requirement for OAS document versioning, whilst still leaving something in its place.

### DIFF
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -100,15 +100,11 @@ The available status codes are defined by [RFC7231](https://tools.ietf.org/html/
 
 ### Versions
 
-The OpenAPI Specification is versioned using [Semantic Versioning 2.0.0](https://semver.org/spec/v2.0.0.html) (semver) and follows the semver specification.
+The OpenAPI Specification is versioned using a `major`.`minor`.`patch` versioning scheme. The `major`.`minor` portion of the version string (for example `3.1`) SHALL designate the OAS feature set. *`.patch`* versions address errors in, or provide clarifications to, this document, not the feature set. Tooling which supports OAS 3.1 SHOULD be compatible with all OAS 3.1.\* versions. The patch version SHOULD NOT be considered by tooling, making no distinction between `3.1.0` and `3.1.1` for example.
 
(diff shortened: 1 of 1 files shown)


versions/3.1.0.md
```diff
@@ -100,15 +100,11 @@ The available status codes are defined by [RFC7231](https://tools.ietf.org/html/
 
 ### Versions
 
-The OpenAPI Specification is versioned using [Semantic Versioning 2.0.0](https://semver.org/spec/v2.0.0.html) (semver) and follows the semver specification.
+The OpenAPI Specification is versioned using a `major`.`minor`.`patch` versioning scheme. The `major`.`minor` portion of the version string (for example `3.1`) SHALL designate the OAS feature set. *`.patch`* versions address errors in, or provide clarifications to, this document, not the feature set. Tooling which supports OAS 3.1 SHOULD be compatible with all OAS 3.1.\* versions. The patch version SHOULD NOT be considered by tooling, making no distinction between `3.1.0` and `3.1.1` for example.
 
-The `major`.`minor` portion of the semver (for example `3.0`) SHALL designate the OAS feature set. Typically, *`.patch`* versions address errors in this document, not the feature set. Tooling which supports OAS 3.0 SHOULD be compatible with all OAS 3.0.\* versions. The patch version SHOULD NOT be considered by tooling, making no distinction between `3.0.0` and `3.0.1` for example.
+Occasionally, non-backwards compatible changes may be made in `minor` versions of the OAS where impact is believed to be low.
```
handrews, line 105: Would it make sense to say "impact is believed to be low relative to the benefit provided."?

That would provide some reassurance of not just making random small breaking changes for the lulz.

Or another option: "impact is believed to be low and can be clearly and concisely communicated."?

I'm just brainstorming wording that might help make it clear that minor is not the new major, without constraining the process through hard-and-fast rules.

darrelmiller, line 105:
```suggestion
Occasionally, non-backwards compatible changes may be made in `minor` versions of the OAS where impact is believed to be low relative to the benefit provided.
```

PR #2255, opened 2020-06-04
[master] x-oas- to x-oai-

MikeRalphson:
Updates `DEVELOPMENT.md` and the `proposals` folder to reflect the decision to change the reserved `x-oas-` prefix to `x-oai-`.


MikeRalphson:
Needs updating to reserve both.

PR #2256, opened 2020-06-04
[v3.1.0-dev] reserve both x-oas- and x-oai- prefixes

MikeRalphson:
Update to reflect the decision to change the reserved specification extension prefix from `x-oas-` to `x-oai-` and removes a redundant sentence.


MikeRalphson:
Needs updating to reserve both.

webron (approved):

PR #2266, opened 2020-06-15
$schema Guidance

### LINKED ISSUE #2017: "$schema" guidance for OAS 3.1
#1977 includes support for `$schema` along with all of the other JSON Schema keywords.  This is really important for being able to experiment with things like new code gen vocabularies or using Hyper-Schema and OAS together, but we'll want to put some boundaries on it.  We don't want people thinking they can just shove anything in there and expect tooling to support it.

I opened PR #2016 primarily to have something concrete to reference for this.

***TL;DR:***
* The default meta-schema for OAS Schema objects should describe the JSON Schema Core and Validation keywords, plus OAS's extensions and compatibility adjustments (e.g. `exclusive*`).
* Will OAS require that all meta-schemas forbid unknown keywords that do not begin with `x-`?  Requiring this would mean that the standard JSON Schema meta-schemas _cannot_ be used, even though they are compatible, which would seem to defeat the purpose of JSON Schema compatibility
* Will OAS allow using additional optional vocabularies (meaning that tooling that doesn't care about them can safely ignore them) that use keywords without `x-` prefixes?  Examples would be hyper-schema and some future hypothetical code generation vocabulary that might get designed outside of OAS.

## Default `$schema` value

I used `https://spec.openapis.org/oas/3.1/meta/validation/2019-10` as a placeholder URI for what I think the default Schema Object meta-schema should be.  This meta-schema declares that all of the JSON Schema Core and Validation semantics are present, with the same level of requirement as the standard `https://json-schema.org/draft/2019-09/schema` meta-schema (everything except the Format vocabulary is required, while support for `format` is [somewhat best effort](http://json-schema.org/draft/2019-09/release-notes.html#format-vocabulary)- as far as I can tell, that matches how OpenAPI manages `format`).

The meta-schema also requires support for OpenAPI extension keywords, and compatibility syntax for `exclusiveMinimum` and `exclusiveMaximum`.  It uses a somewhat tweaked version of the standard validation meta-schema plus a meta-schema for the extensions in order for this to work.  It also enforces that the only extension keywords allowed must start with `x-`.

OpenAPI should define this as its default for `$schema` (as a special-purpose implementation, it can [determine its own default assumptions](http://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.8.1.2.1).

### The standard core/validation meta-schema

To facilitate interoperability with regular JSON Schema implementations, we should allow the standard `https://json-schema.org/draft/2019-09/schema` meta-schema.  As it declares strictly a subset of the keyword semantics and syntax declared by the OAS Schema object meta-schema, this should work just fine.

The only exception to compatibility is that it does not forbid unknown keywords that start with something other than `x-`.  I think it is fine for us to document that just because a meta-schema does not enforce that restriction, that does not mean that OpenAPI implementations must suddenly support non-`x-`-prefixed extension keywords.  

### The standard hyper-schema meta-schema

If we want to encourage experimentation with Hyper-Schema, we should also allow `https://json-schema.org/draft/2019-09/hyper-schema`.  Of course, that introduces two additional non-`x-`-prefixed keywords, `links` and `base`.  This is where things get interesting.

If we want to have a strict adherence to the `x-` policy, then that would mean that in order to use Hyper-Schema, its vocabulary should be added to the default Schema Object meta-schema (probably with value `false` indicating that tools can decide to process the schema even if they do not understand Hyper-Schema- Hyper-Schema is safe for tools that don't care about it to ignore).

_OR_, we could say that only keywords that are defined outside of the JSON Schema vocabulary concept need to be prefixed with `x-`.  So any keyword with semantics defined by a vocabulary referenced in the meta-schema's `$vocabulary` object would be allowable.  Only keywords that do not have formally defined semantics need an `x-` prefix.  I assume the `x-` prefix was designed to avoid future name clashes.  The `$vocabulary` system reduces the need for that somewhat, I think.

### Future code-gen or other vocabularies

The main reason I wanted to bring up this idea of allowing keywords from defined vocabularies to not be prefixed with `x-` is that one of the goals of updating to `2019-09` is to allow using the vocabulary concept to facilitate use cases like code generation.

This is something for which there is substantial demand both inside and outside of the OpenAPI ecosystem.  If a new vocabulary is developed for this outside of OpenAPI, then it will not have `x-`-prefixed keywords.  It would seem to me to be important to allow tools to add support for such a vocabulary without forcing the existence of a parallel vocabulary with all of the keywords renamed with `x-` prefixes.

## Vocabulary-based compatibility

One of the key reasons for separating vocabularies and meta-schemas is that new/unknown meta-schemas are expected to be far more common than new/unknown vocabularies.

So, defining what is allowable in `$schema` based on the vocabulary set is the most flexible option.  In this approach, any meta-schema declaring a subset of the vocabularies in `https://spec.openapis.org/oas/3.1/meta/validation/2019-10` would be allowed (which would include the standard core/validation meta-schema).

The question would again be whether additional vocabularies that _do not contradict_ those declared in `https://spec.openapis.org/oas/3.1/meta/validation/2019-10` would also be allowed, even though their keywords do not start with `x-` (for example, Hyper-Schema, or a code gen vocabulary).  A good option might be to say that they are, but only if they are declared with `false` (meaning that tooling can safely ignore them if it is not designed to work with them).

This exact scenario is why the `true`/`false` vocabulary declarations exist: we did not want to force all tooling to understand irrelevant vocabularies that do not impact validation outcomes.

philsturgeon:
Took a swing at this one but I'll absolutely need some guidance myself. 

Closes #2017.

### DIFF
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -2342,7 +2342,7 @@ OpenAPI documents are made up of multiple resources. Sometimes resources are all
 
 It is important for tooling to be able to detect what meta-schema any given resource wishes to be processed with: JSON Schema Core, JSON Schema Validation, OpenAPI Schema Object, or some custom meta schema.
 
-The `$schema` keyword can be used to define which meta-schema should be used when interacting with any schema resource.
+The `$schema` keyword can be used to identify which meta-schema should be used when interacting with any schema resource.
 
 Inline schema resources will default to using `$schema: "https://spec.openapis.org/oas/3.1/schema-object"`.
 
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -2338,15 +2338,17 @@ The [XML Object](#xmlObject) contains additional information about the available
 
 ###### Picking Schema Vocabularies
 
-OpenAPI documents are made up of multiple resources, sometimes resources are all in the same root OpenAPI document (inline resources), or resources are contain their own documents with `$ref` (separate resources). It is important for tooling to be able to detect what meta-schema any given resource wishes to be processed with: JSON Schema Core, JSON Schema Validation, OpenAPI Schema Object, or some custom meta schema.
+OpenAPI documents are made up of multiple resources, sometimes resources are all in the same root OpenAPI document (inline resources), or resources are contain their own documents with `$ref` (separate resources). A schema object could exist as an inline resource, or a separate resource. 
 
-The `$schema` keyword can be used to define which meta-schema should be used when interacting with any resource.
+It is important for tooling to be able to detect what meta-schema any given resource wishes to be processed with: JSON Schema Core, JSON Schema Validation, OpenAPI Schema Object, or some custom meta schema.
 
-Inline resources will default to using `$schema: "https://json-schema.org/draft/2019-09/schema"`.
+The `$schema` keyword can be used to define which meta-schema should be used when interacting with any schema resource.
 
-Separate resources will default to using `$schema: "https://spec.openapis.org/oas/3.1/schema-object"`.
+Inline schema resources will default to using `$schema: "https://spec.openapis.org/oas/3.1/schema-object"`.
 
-Any resource can override this default by setting the `$schema` property to something else. 
+Separate schema resources will default to using `$schema: "https://json-schema.org/draft/2019-09/schema"`.
+
+Any schema resource can override this default by setting the `$schema` property to something else. 
 
 ##### Schema Object Examples
 
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -2338,17 +2338,11 @@ The [XML Object](#xmlObject) contains additional information about the available
 
 ###### Picking Schema Vocabularies
 
-OpenAPI documents are made up of multiple resources. Sometimes resources are all in the same root OpenAPI document (inline resources); also, resources can contain their own documents with `$ref` (separate resources). A schema object could exist as an inline resource, or a separate resource. 
-
 It is important for tooling to be able to detect what meta-schema any given resource wishes to be processed with: JSON Schema Core, JSON Schema Validation, OpenAPI Schema Object, or some custom meta schema.
 
-The `$schema` keyword can be used to identify which meta-schema should be used when interacting with any schema resource.
+`$schema` MAY be present in any Schema Object, and if present MUST be used to determine which dialect should be used when processing the schema.
 
-Inline schema resources will default to using `$schema: "https://spec.openapis.org/oas/3.1/schema-object"`.
-
-Separate schema resources will default to using `$schema: "https://json-schema.org/draft/2019-09/schema"`.
-
-Any schema resource can override this default by setting the `$schema` property to something else. 
+When `$schema` is not present, the default the following dialect MUST be assumed: `$schema: "https://spec.openapis.org/oas/3.1/schema-object"`.
 
 ##### Schema Object Examples
 
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -2336,6 +2336,18 @@ As such, inline schema definitions, which do not have a given id, *cannot* be us
 The [xml](#schemaXml) property allows extra definitions when translating the JSON definition to XML.
 The [XML Object](#xmlObject) contains additional information about the available options.
 
+###### Picking Schema Vocabularies
+
+OpenAPI documents are made up of multiple resources, sometimes resources are all in the same root OpenAPI document (inline resources), or resources are contain their own documents with `$ref` (separate resources). It is important for tooling to be able to detect what meta-schema any given resource wishes to be processed with: JSON Schema Core, JSON Schema Validation, OpenAPI Schema Object, or some custom meta schema.
+
+The `$schema` keyword can be used to define which meta-schema should be used when interacting with any resource.
+
+Inline resources will default to using `$schema: "https://json-schema.org/draft/2019-09/schema"`.
+
+Separate resources will default to using `$schema: "https://spec.openapis.org/oas/3.1/schema-object"`.
+
+Any resource can override this default by setting the `$schema` property to something else. 
+
 ##### Schema Object Examples
 
 ###### Primitive Sample
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -2338,7 +2338,7 @@ The [XML Object](#xmlObject) contains additional information about the available
 
 ###### Picking Schema Vocabularies
 
-OpenAPI documents are made up of multiple resources, sometimes resources are all in the same root OpenAPI document (inline resources), or resources are contain their own documents with `$ref` (separate resources). A schema object could exist as an inline resource, or a separate resource. 
+OpenAPI documents are made up of multiple resources. Sometimes resources are all in the same root OpenAPI document (inline resources); also, resources can contain their own documents with `$ref` (separate resources). A schema object could exist as an inline resource, or a separate resource. 
 
 It is important for tooling to be able to detect what meta-schema any given resource wishes to be processed with: JSON Schema Core, JSON Schema Validation, OpenAPI Schema Object, or some custom meta schema.
 


versions/3.1.0.md
```diff
@@ -2336,6 +2336,18 @@ As such, inline schema definitions, which do not have a given id, *cannot* be us
 The [xml](#schemaXml) property allows extra definitions when translating the JSON definition to XML.
 The [XML Object](#xmlObject) contains additional information about the available options.
 
+###### Picking Schema Vocabularies
+
+OpenAPI documents are made up of multiple resources, sometimes resources are all in the same root OpenAPI document (inline resources), or resources are contain their own documents with `$ref` (separate resources). It is important for tooling to be able to detect what meta-schema any given resource wishes to be processed with: JSON Schema Core, JSON Schema Validation, OpenAPI Schema Object, or some custom meta schema.
```
MikeRalphson, line 2341: Can we reserve the use of the word "resource" to mean a separate file, HTTP-style resource or (:copyright: @handrews an octet-stream), rather than overloading it by using the term "inline resources". Are these better just called "components"?

I think we also need to address that not all resources are JSON Schema or even collections of JSON Schema. Resources can be pretty much arbitrary JSON/YAML as long as they comply with the object structure of the location they're being `$ref`d from.

`s/are contain/MAY be contained/`

I'm not clear on the use-case for setting the meta-schema to JSON Schema Core in an OAS document would be.

We should be consistent on whether it's `meta-schema` or `meta schema`.

versions/3.1.0.md
```diff
@@ -2336,6 +2336,18 @@ As such, inline schema definitions, which do not have a given id, *cannot* be us
...
+
+The `$schema` keyword can be used to define which meta-schema should be used when interacting with any resource.
```
MikeRalphson, line 2343: Again, only valid within a valid JSON Schema resource or a schema-object resource or within schema-objects within an OAS sub-document containing a `components` object.

versions/3.1.0.md
```diff
@@ -2336,6 +2336,18 @@ As such, inline schema definitions, which do not have a given id, *cannot* be us
...
+
+Inline resources will default to using `$schema: "https://json-schema.org/draft/2019-09/schema"`.
+
+Separate resources will default to using `$schema: "https://spec.openapis.org/oas/3.1/schema-object"`.
```
MikeRalphson, line 2347: I'm confused by this distinction. It could be it's just the wrong way around?

philsturgeon: Totally backwards! 🙃

versions/3.1.0.md
Relequestual: In JSON Schema world, we have started to refer to any schema which identifies itself as a "resource", which is probably where Phil is using the termonology from in this case. Specifically how schema resources can be embedded in other documents. If a JSON Schema has an `$id`, it's a "schema resource".

versions/3.1.0.md
philsturgeon: Would you like to take a swing at improving my nonsense with the Suggestion feature? I would appreciate the help.

versions/3.1.0.md
philsturgeon: I've changed this to be schema resource in a bunch of places. If the new version is no good lets start with new feedback and try to use the suggestion feature wherever possible to take a swing at making it better.

versions/3.1.0.md
philsturgeon: I think I might have got this sorted.

philsturgeon:
New attempt! LMK what you think.

Relequestual:
The `$schema` URI is only an identifier. It's usually network addressable returning the associated meta schema, but you should not assume that to be the case.

philsturgeon:
@Relequestual what is that feedback aimed at? Do I need to add that in or did I say something which suggests otherwise?

versions/3.1.0.md
```diff
@@ -2336,6 +2336,20 @@ As such, inline schema definitions, which do not have a given id, *cannot* be us
 The [xml](#schemaXml) property allows extra definitions when translating the JSON definition to XML.
 The [XML Object](#xmlObject) contains additional information about the available options.
 
+###### Picking Schema Vocabularies
+
+OpenAPI documents are made up of multiple resources, sometimes resources are all in the same root OpenAPI document (inline resources), or resources are contain their own documents with `$ref` (separate resources). A schema object could exist as an inline resource, or a separate resource.
```
karenetheridge, line 2341:
```suggestion
OpenAPI documents are made up of multiple resources. Sometimes resources are all in the same root OpenAPI document (inline resources); also, resources cam contain their own documents with `$ref` (separate resources). A schema object could exist as an inline resource, or a separate resource. 
```

handrews: @karenetheridge `cam` => `can`

karenetheridge: doh, fixed

MikeRalphson: Where words like "could" and "can" take the meanings of words in the [RFC2119](https://tools.ietf.org/html/rfc2119) list like "MAY", we usually prefer those (and they should be capitalised).

Relequestual:
> @Relequestual what is that feedback aimed at? Do I need to add that in or did I say something which suggests otherwise?

I think I read it slightly differently the first time. I think I have a one word change I'll suggest.

versions/3.1.0.md
```diff
@@ -2336,6 +2336,20 @@ As such, inline schema definitions, which do not have a given id, *cannot* be us
 The [xml](#schemaXml) property allows extra definitions when translating the JSON definition to XML.
 The [XML Object](#xmlObject) contains additional information about the available options.
 
+###### Picking Schema Vocabularies
+
+OpenAPI documents are made up of multiple resources, sometimes resources are all in the same root OpenAPI document (inline resources), or resources are contain their own documents with `$ref` (separate resources). A schema object could exist as an inline resource, or a separate resource. 
+
+It is important for tooling to be able to detect what meta-schema any given resource wishes to be processed with: JSON Schema Core, JSON Schema Validation, OpenAPI Schema Object, or some custom meta schema.
+
+The `$schema` keyword can be used to define which meta-schema should be used when interacting with any schema resource.
```
Relequestual, line 2345:
```suggestion
The `$schema` keyword can be used to identify which meta-schema should be used when interacting with any schema resource.
```

My feeling is "define" suggests the URI should be network acessable, where as it may not be. "Identify" hints at more work must be done in order to actually GET the meta-schema.

philsturgeon:
@Relequestual @MikeRalphson hey friends! Think I got all your feedback in in time for tomorrows call. @handrews I'd appreciate a check from you too. This is based off of some of your words from a while back, but I tried to make it more OpenAPIish.

versions/3.1.0.md
```diff
@@ -2336,6 +2336,20 @@ As such, inline schema definitions, which do not have a given id, *cannot* be us
 The [xml](#schemaXml) property allows extra definitions when translating the JSON definition to XML.
 The [XML Object](#xmlObject) contains additional information about the available options.
 
+###### Picking Schema Vocabularies
+
+OpenAPI documents are made up of multiple resources. Sometimes resources are all in the same root OpenAPI document (inline resources); also, resources can contain their own documents with `$ref` (separate resources). A schema object could exist as an inline resource, or a separate resource.
```
MikeRalphson, line 2341: The opening sentence sounds like a re-visit of the `oasDocument` definition further up:

> A self-contained or composite resource which defines or describes an API or elements of an API.

We probably don't want two places saying nearly the same thing in different terms.

MikeRalphson: The use of "inline" to mean within the same root OAS document conflicts with our use elsewhere (e.g. in the `discriminator` section) where "inline" means "not accessed via a `$ref` (internal or external)".

MikeRalphson: Rather than say "sometimes this can happen" just say "this MAY happen".

MikeRalphson: Do you actually mean "resources can **contain** their own **documents** with `$ref` (separate resources)"?

MikeRalphson: Please can we avoid "could" and "can" along with "sometimes". Stick to "MAY" and other RFC2119 terms.

versions/3.1.0.md
```diff
@@ -2336,6 +2336,20 @@ As such, inline schema definitions, which do not have a given id, *cannot* be us
 The [xml](#schemaXml) property allows extra definitions when translating the JSON definition to XML.
 The [XML Object](#xmlObject) contains additional information about the available options.
 
+###### Picking Schema Vocabularies
+
+OpenAPI documents are made up of multiple resources. Sometimes resources are all in the same root OpenAPI document (inline resources); also, resources can contain their own documents with `$ref` (separate resources). A schema object could exist as an inline resource, or a separate resource. 
+
+It is important for tooling to be able to detect what meta-schema any given resource wishes to be processed with: JSON Schema Core, JSON Schema Validation, OpenAPI Schema Object, or some custom meta schema.
```
MikeRalphson, line 2343: I appreciate we're importing the JSON Schema terminology here, but are "resources" only `schemaObjects` or is any component a resource? Only `schemaObjects` can indicate their `$schema` so it is only "important for tooling" when dealing with `schemaObjects`, not for example `parameters` or `responses`.

MikeRalphson: What would be the use-case for specifying JSON Schema Core as the meta-schema for a `schemaObject` / resource?

> Keywords declared in this section, which all begin with "$", make up the JSON Schema Core vocabulary.

A `schemaObject` or resource which only uses `$`-prefixed keywords doesn't sound very useful in an OAS context.

versions/3.1.0.md
```diff
@@ -2336,6 +2336,20 @@ As such, inline schema definitions, which do not have a given id, *cannot* be us
...
+
+The `$schema` keyword can be used to identify which meta-schema should be used when interacting with any schema resource.
```
MikeRalphson, line 2345:
```suggestion
The `$schema` keyword MAY be used to identify which meta-schema should be used when processing a given schema resource.
```

versions/3.1.0.md
```diff
@@ -2336,6 +2336,20 @@ As such, inline schema definitions, which do not have a given id, *cannot* be us
...
+
+Inline schema resources will default to using `$schema: "https://spec.openapis.org/oas/3.1/schema-object"`.
```
MikeRalphson, line 2347:
```suggestion
Inline schema resources MUST default to using `$schema: "https://spec.openapis.org/oas/3.1/schema-object"`.
```

versions/3.1.0.md
```diff
@@ -2336,6 +2336,20 @@ As such, inline schema definitions, which do not have a given id, *cannot* be us
...
+
+Separate schema resources will default to using `$schema: "https://json-schema.org/draft/2019-09/schema"`.
```
MikeRalphson, line 2349:
```suggestion
Separate schema resources MUST default to using `$schema: "https://json-schema.org/draft/2019-09/schema"`.
```

versions/3.1.0.md
```diff
@@ -2336,6 +2336,20 @@ As such, inline schema definitions, which do not have a given id, *cannot* be us
...
+
+Any schema resource can override this default by setting the `$schema` property to something else.
```
MikeRalphson, line 2351:
```suggestion
Any schema resource MAY override this default by setting the `$schema` property to another value. 
```

versions/3.1.0.md
MikeRalphson: Pending resolution of the outstanding question of whether these defaults are desirable (allowing for future versions of JSON Schema) or implicit/explicitly stated in some new top-level property/object.

versions/3.1.0.md
handrews: This is where it would be good to use the _dialect_ vs _vocabulary_ terminology: https://github.com/json-schema-org/json-schema-spec/issues/902

> JSON Schema Core, JSON Schema Validation

When written out and capitalized like this, these are usually the name of the specification documents, which don't quite map to meta-schemas.  The meta-schema that most people used is referred to as the ["default meta-schema"](https://tools.ietf.org/html/draft-handrews-json-schema-validation-02#section-5) in 2019-09.  We might say "default dialect meta-schema" as we add the "dialect" term in the next draft.

It would be very strange to use only the core vocabulary as all it does is bootstrap the JSON Schema system.  There is no standard meta-schema for the combination of vocabularies in the JSON Schema Core specification document (== core and applicator vocabularies).

versions/3.1.0.md
handrews: `$schema` MAY be present, and if present MUST be used to determine which dialect should be used when processing the schema.  So for the schema author it's a MAY, for tooling consuming the schema, if present, it is a MUST.

philsturgeon:
Two weeks ago it was discussed on the TSC that due to the nature of 3.1 now removing the awkward exclusiveMinimum/exclusiveMaximum and readOnly/writeOnly differences, we can just assume everything is `https://spec.openapis.org/oas/3.1/schema-object`. 

If somebody uses discriminator in JSON Schema tooling which doesnt understand it then meh whatever.

versions/3.1.0.md
philsturgeon: This paragraph was setting up some stuff to make the following clear, but now that the following is simplified this can go away.

philsturgeon:
Click it 😈

darrelmiller:
@philsturgeon Consider it clicked.

PR #2291, opened 2020-07-20
Update TOB.md

ncaidin:
Darrel, Jerome, and Jeremy have been re-elected for another 2 year term. Updating the BGB elected term through 2022.


ncaidin:
Thanks @MikeRalphson and @webron !

PR #2300, opened 2020-07-30
Updated Overlay proposal

darrelmiller:
Updated with "add" and "merge" fields instead of just "value".


MikeRalphson (approved):

FWiesner:
@darrelmiller what is the right ticket to comment on Overlays as shown on the TSC call today?

PR #2302, opened 2020-08-01
v3.1.0: rephrase data-type section because `format` keyword can be used for any data type.

### LINKED ISSUE #2301: OAS 3.1: Suggest rephrasing data-types section for `format` keyword and primitive types.
The OAS 3.1 [data types section](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.1.0.md#data-types) states: 

> Primitives have an optional modifier property: format, which is defined by JSON Schema.

Implied in that sentence is that the `format` keyword does not apply to non-primitive types. However, In the [JSON schema specification section 7](http://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.7), it is stated the `format` keyword applies to all model instance types:

> All format attributes defined in this section apply to strings, but a format attribute can be specified to apply to any instance types defined in the data model defined in the core JSON Schema.

Based on the above the `format` keyword can apply to `null`, `boolean`, `object`, `array`, `number` and `string`.

One of the goal of OAS 3.1 is to remove some of the constraints that existed in 3.0.3 and support full JSON schema specification. Therefore, shouldn't https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.1.0.md#data-types be rephrased to to avoid conveying the `format` keyword applies to primitive types? Even if in practice most people use the  `format` keyword for primitive types, the spec doesn't have such restriction.

sebastien-rosset:
Fix #2301

### DIFF
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -139,11 +139,11 @@ It is RECOMMENDED that the root OpenAPI document be named: `openapi.json` or `op
 
 ### <a name="dataTypes"></a>Data Types
 
-Primitive data types in the OAS are based on the types supported by the [JSON Schema Specification Draft 2019-09](http://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.4.2). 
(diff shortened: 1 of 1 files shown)


darrelmiller:
Thanks for the PR @sebastien-rosset

PR #2329, opened 2020-08-28
Removed required constraint on responses object

darrelmiller:
This change would allow the following to be a valid OpenAPI description

```
openapi: 3.1.0
info:
   title: An api with currently undefined responses
   version: 1.0.0
paths:
   /: 
     get: {}
```

This is useful for doing initial design work on an API without having to define what request and response payloads should contain.  Currently to be a valid document it is required to include a responses object, which must contain at least one response code which then must contain a description.


MikeRalphson (approved):

darrelmiller:
No objections yet.  Leaving this open for folks to stew on it for a few days.

PR #2335, opened 2020-09-02
Make Server Variable Object's properties more strict

webron:
Followup to #1809, now that we allow breaking changes.


MikeRalphson (approved):
LGTM.

darrelmiller (approved):

PR #2351, opened 2020-09-17
Fix, clarify, and simplify content type schemas

### LINKED ISSUE #2349: Ambiguity between JSON string and plain string
The following is in the 3.1 dev spec:
```json
{
  "description": "A simple string response",
  "content": {
    "text/plain": {
      "schema": {
        "type": "string"
      }
    }
  }

}
```

The intention is that the response body contains an arbitrary string, e.g.:
```
Hello world!
```

However, this would be invalid given the above JSON schema. A correct response would be:
```json
"Hello world!"
```

It seems like the spec ignores this inconsistency and somehow implicitly turns JSON schema off in cases where the type is primitive (or just for strings?). The same is true here:

```yaml
content:
    image/png:
        schema:
            type: string
            contentEncoding: base64
```

A valid body would be:
```json
"SGVsbG8gd29ybGQh"
```

But that is again not the intention here.

In summary, I'm missing some language in the spec that describes this behavior explicitly.

Slightly unrelated, there's another example which confuses me:
```json
{
  "description": "user to add to the system",
  "content": {
    "text/plain": {
      "schema": {
        "type": "array",
        "items": {
          "type": "string"
        }
      }
    }
  }
}
```
Is this supposed to be for example `["Hello", "world!"]`? If so, why is the content type `text/plain` instead of `application/json`?

handrews:
This fixes #2349, which caught that an encoded PNG image
is encoded into a text media type.

In the process I realized some other errors, and simplified things.

* HTTP `Content-Type` is always handled by OAS
    * Media Type Object key in most cases
    * Encoding object (possibly inferred from schema) in `multipart/form-data`
* HTTP-level `Content-Encoding` is always handled by the OAS Header Object
* JSON Schema "content*" is used for embedding one media type into another
    * the encoded resource is of media type `text/plain`
    * `"contentMediaType"` is the embedded media type after decoding
    * `"contentEncoding"` is how to encode/decode binary to/from text

This removes any chance of `"contentMediaType"` conflicting with
the Media Type Object key or with `contentType` in the Encoding Object,
as they now always do different things.

Likewise, the HTTP `Content-Encoding` header (with values like
gzip, deflate, etc.) does different things than `"contentEncoding"`
(which has values like base64, base64url, quoted-printable, etc.).

The deprecated part header `Content-Transfer-Encoding` is likewise
handled in the Encoding Object, but is probably never used.


versions/3.1.0.md
```diff
@@ -1458,12 +1455,15 @@ Binary content transferred with base64 encoding:
 
 ```yaml
 content:
-    image/png:
+    text/plain:
```
darrelmiller, line 1458: I believe this should stay as image/png to convey the semantics of the HTTP payload.

versions/3.1.0.md
```diff
@@ -1458,12 +1455,15 @@ Binary content transferred with base64 encoding:
...
         schema:
             type: string
+            contentMediaType: image/png
             contentEncoding: base64
 ```
 
+Note that the `Content-Type` is `text/plain`, since from an HTTP perspective, base64-encoded data is simply text.  The `contentMediaType` and `contentEncoding` fields inform the application how to encode and decode the resource from and to its actual media type.
```
darrelmiller, line 1465: Believe this text should be removed.

versions/3.1.0.md
handrews: @darrelmiller I don't remember why I did that, but I'll take your word for it.  I am having a difficult time finding a reference (which I was hoping would remind me why I thought this was correct).  But yeah, the JSON Schema provides all of the processing information and the logical output is `image/png` so that makes sense.  And of course the text that asserted this should indeed be removed.  Lemme do that right now...

versions/3.1.0.md
handrews: @darrelmiller I have changed this paragraph to explain why there is a redundancy, but if you feel the new text is more confusing than helpful, I'm happy to remove it.

versions/3.1.0.md
```diff
@@ -1535,10 +1536,9 @@ When passing in `multipart` types, boundaries MAY be used to separate sections o
 
 * If the property is a primitive, or an array of primitive values, the default Content-Type is `text/plain`
 * If the property is complex, or an array of complex values, the default Content-Type is `application/json`
-* If the property is a `type: string` with a `contentEncoding`, the default Content-Type is `application/octet-stream`
-* If the JSON Schema keyword `contentMediaType` is used and no Encoding Object is present, then the Content-Type is that which is specified by `contentMediaType`, however if an Encoding Object is present, then `contentMediaType` SHALL be ignored
+* If the property is a `type: string` with a `contentEncoding`, the default Content-Type is `text/plain`, and the media type of the embedded resource is specified in `contentMediaType`
```
darrelmiller, line 1539:
```suggestion
* If the property is a `type: string` with a `contentEncoding`, the default Content-Type is `application/octet-stream`
```
Changed this back to match wording here: https://tools.ietf.org/html/draft-bhutton-json-schema-validation-00#section-8.3

versions/3.1.0.md
```diff
@@ -1581,7 +1583,7 @@ A single encoding definition applied to a single schema property.
...
+<a name="encodingContentType"></a>contentType | `string` | The Content-Type for encoding a specific property. Default value depends on the property type: when `type` is absent - `application/octet-stream`; for primitive types - `text/plain`; for `object` - `application/json`; when `type` is `string` and `contentEncoding` is present, the default Content-Type is `text/plain`, and the media type of the encoded resource is specified in `contentMediaType`; for `array` – the default is defined based on the inner type. The value can be a specific media type (e.g. `application/json`), a wildcard media type (e.g. `image/*`), or a comma-separated list of the two types.
```
mkistler, line 1586: I thought that we agreed in the 1/28 TSC meeting that when `type` is `string` the default Content-Type should be `application/octet-stream`.

mkistler:
I've read through this PR and it left some questions unanswered in my mind.

I think the basic issue to be addressed is: Now that full JSON Schema is allowed in OpenAPI 3.1, what does it mean for the JSON Schema `contentMediaType` and `contentEncoding` keywords to be present in a request body schema?

I think the 3.1.0 spec already covers `contentMediaType` with this statement:

> JSON Schema also offers a contentMediaType keyword. However, when the media type is already specified by the Media Type Object's key, or by the contentType field of an Encoding Object, the contentMediaType keyword SHALL be ignored if present.

Though that statement is buried in the section titled "Considerations for File Uploads" and should be moved to the general section on the Media type object.

Regarding `contentEncoding`, I don't think this is properly addressed for the "root" schema of a request body content entry.  In particular, what I think should be said is:

- If the “root” schema of a request body `content` entry contains a `contentEncoding`, then:
  - If no `Content-encoding` header parameter is defined for the operation, the allowed encodings for the content entry are defined by the `contentEncoding` in the schema
  - If a `Content-encoding` header parameter is defined, the allowed encodings for the content entry are defined by the intersection of the allowable values for the `Content-encoding` header and the `contentEncoding` in the schema

This is just my take but I suppose other approaches could be taken, such as simply ignoring `Content-encoding` for a "root" schema of a request body `content` entry, but whatever is decided, I think it should be clearly stated in the spec.

versions/3.1.0.md
darrelmiller, line 1586:
```suggestion
<a name="encodingContentType"></a>contentType | `string` | The Content-Type for encoding a specific property. Default value depends on the property type: for `object` - `application/json`;  for `array` – the default is defined based on the inner type; for all other cases the default is `application/octet-stream`. The value can be a specific media type (e.g. `application/json`), a wildcard media type (e.g. `image/*`), or a comma-separated list of the two types. 
```

darrelmiller:
Content-Encoding header overrides the JSON schema contentEncoding.

versions/3.1.0.md
```diff
@@ -1433,14 +1433,11 @@ application/json:
...
-
-JSON Schema also offers a `contentMediaType` keyword.  However, when the media type is already specified by the
-Media Type Object's key, or by the `contentType` field of an [Encoding Object](#encodingObject), the `contentMediaType` keyword SHALL be ignored if present.
+In contrast with the 2.0 specification, `file` input/output content in OpenAPI is described with the same semantics as any other schema type.  In contrast with the 3.0 specification, such schemas either omit the `type` (in place of `format: binary`), or use `contentMediaType` and `contentEncoding` with `type: string`. The `contentEncoding` keyword supports all encodings defined in [RFC4648](https://tools.ietf.org/html/rfc4648), including "base64" (which replaces `format: byte`) and "base64url", as well as "quoted-printable" from [RFC2045](https://tools.ietf.org/html/rfc2045#section-6.7).
```
mkistler, line 1436: Here is my proposed rewording to clarify the priority of `Content-Encoding` headers (in various places) and the JSON Schema `contentEncoding`:
```suggestion
In contrast with the 2.0 specification, `file` input/output content in OpenAPI is described with the same semantics as any other schema type. 

In contrast with the 3.0 specification, the `format` keyword has no effect on the content-encoding of the schema. The Content-Encoding for the schema can be specified explicitly with the `Content-Encoding` parameter of the operation, for schemas in request bodies, or in a `Content-Encoding` header in the `headers` of the response, for response schemas, or in a `Content-Encoding` header in the `headers` of an Encoding object associated with a request or response body property. JSON Schema also offers a `contentEncoding` keyword, which may be used to specify the `Content-Encoding` for the schema, but `contentEncoding` SHALL be ignored if a `Content-Encoding` header is defined. The `contentEncoding` keyword supports all encodings defined in [RFC4648](https://tools.ietf.org/html/rfc4648), including "base64" (which replaces `format: byte`) and "base64url", as well as "quoted-printable" from [RFC2045](https://tools.ietf.org/html/rfc2045#section-6.7). When not explicitly specified, the content-encoding

JSON Schema also offers a `contentMediaType` keyword.  However, when the media type is already specified by the Media Type Object's key, or by the `contentType` field of an [Encoding Object](#encodingObject), the `contentMediaType` keyword SHALL be ignored if present.
```

versions/3.1.0.md
handrews: @darrelmiller I'm fairly sure we went over the wording and cases for this in a call and there's a reason it was done this way.  I confess I can't remember what that was.

versions/3.1.0.md
handrews: @mkistler the `Content-Encoding` HTTP header and `contentEncoding` JSON Schema keyword are unrelated.  The HTTP header is for things like gzip that the HTTP client and server will encode/decode automatically.  The JSON Schema keyword is for when both HTTP and the media type think that something is text, but the application needs to be informed that some specific piece of text needs to be encoded/decoded in a particular way.

handrews: You could in fact have a gzipped payload that includes a base64-encoded PNG file.  That would use both the header and the keyword.

mkistler: @handrews I struggling to understand this all.  Do you mean to say that `contentEncoding` keyword has no effect whatsoever on the value that should/will be passed in the `Content-Encoding` header? If that is the case then it certainly will confuse some readers (like me :smile:).

In the example you give, how would this look in the API doc?  Maybe:
```
  /v1/photos:
    post:
      parameters:
        - name: Content-Encoding
           in: header
           schema:
             type: string
             enum:
               - gzip
        requestBody:
          content:
            image/png:
              schema:
                contentEncoding: base64
```

And the http request would specify:
- `Content-type: image/png`
- `Content-encoding: gzip`
- a gzipped, base64-encoded image/png

Is that right?

If I've got this right, then I think you are saying that _both_ the encoding specified in the `Content-Encoding` header _and_ the encoding specified in the `contentEncoding` keyword are applied to the request body (or whichever element they pertain to).

If that's right, I'll make a stab at rewording my suggestion to describe this.

handrews: @mkistler yeah this is the challenge of combining multiple standards that were made in different contexts at different times.

If you go back to MIME, there is a `Content-Transfer-Encoding` header, which is where values such as `base64` from RFC 4648 (or its predecessors) are used.  Perhaps we should have used `mediaEncoding` or something like that in JSON Schema (the naming of that keyword has a slightly convoluted history that partially predates me), but we didn't.  I might have only learned about HTTP `Content-Encoding` while researching this PR, come to think of it.

I believe that what you have for the HTTP request is correct.  At this stage with the modifications y'all have made recently, I'm a little unclear on where we ended up on `type` being absent vs `"type": "string"`, but what you have (with `type` absent) is what I would have written.  And is compatible with the latest JSON Schema draft which notes that an application MAY apply a JSON Schema to other media types, including binary media types, and notes that many keywords have no sensible meaning with binary types.

mkistler, line 1436: Second attempt:

```suggestion
In contrast with the 2.0 specification, `file` input/output content in OpenAPI is described with the same semantics as any other schema type.

In contrast with the 3.0 specification, the `format` keyword has no effect on the content-encoding of the schema. JSON Schema offers a `contentEncoding` keyword, which may be used to specify the `Content-Encoding` for the schema. The `contentEncoding` keyword supports all encodings defined in [RFC4648](https://tools.ietf.org/html/rfc4648), including "base64" and "base64url", as well as "quoted-printable" from [RFC2045](https://tools.ietf.org/html/rfc2045#section-6.7). The encoding specified by the `contentEncoding` keyword is independent of an encoding specified by the `Content-Type` header in the request or response or metadata of a multipart body -- when both are present, the encoding specified in the `contentEncoding` is applied first and then the encoding specified in the `Content-Type` header.

JSON Schema also offers a `contentMediaType` keyword.  However, when the media type is already specified by the Media Type Object's key, or by the `contentType` field of an [Encoding Object](#encodingObject), the `contentMediaType` keyword SHALL be ignored if present.
```

darrelmiller (approved):

PR #2360, opened 2020-09-30
Update README.md

Koshal67:


MikeRalphson:
Thanks, but this doesn't look like the kind of change we would merge.

PR #2361, opened 2020-10-01
Add missing schema type in Operation Object YAML Example.

di-wu:


MikeRalphson:
Thanks for both of your recent PRs. The majority of points here relate to both:

* Please see [DEVELOPMENT.md](https://github.com/OAI/OpenAPI-Specification/blob/master/DEVELOPMENT.md#development-guidelines) for an overview
* We do not accept PRs against the spec directly on the `master` (soon to be `main`) branch
* Changes to published versions of the spec are not accepted
* So, your change should affect only `versions/3.1.0.md` and be on the `3.1.0-dev` branch.

As to the intent of this PR, JSON Schema keywords are independent, so `type: object` is not necessary to support the `properties` keyword. Therefore we would need some more context as to why you feel the `type` keyword is "missing" in this case.

di-wu:
Thank you for the comment! Apparently I haven't read the Development Guidelines well enough. 
I'll change branches and remove the changes from the published versions.

The intent of this PR was to address the inconsistency between the JSON and YAML examples of the `Operation Object`.
The JSON example contains `type: object`, this is why I added it to the other example.

```json
"schema": {
          "type": "object",
          "properties": {
```

MikeRalphson:
> Apparently I haven't read the Development Guidelines well enough.

No problem, it's perhaps unfortunate we don't follow the GitHub convention of calling it CONTRIBUTING.md

> The intent of this PR was to address the inconsistency between the JSON and YAML examples of the Operation Object.
The JSON example contains type: object, this is why I added it to the other example.

Ah, excellent, thank you!

MikeRalphson (approved):

karenetheridge:
> No problem, it's perhaps unfortunate we don't follow the GitHub convention of calling it CONTRIBUTING.md

@MikeRalphson You may want to add a `pull_request_template.md` file to the repository, which will be used to populate the body text of new PRs.

darrelmiller (approved):

PR #2362, opened 2020-10-01
Add missing field in Request Body JSON Examples.

di-wu:


MikeRalphson:
See [comments](https://github.com/OAI/OpenAPI-Specification/pull/2361#issuecomment-702667831) in your other PR.

* `text plain` echoes the `text/plain` media-type so I'm not sure changing this to `plain text` is an improvement
* Fixing the capitalisation of `Example` in the `summary` looks like a good change

As to setting the `required` field to `true`, can you expand on why you think this change is beneficial?

di-wu:
Same [comment](https://github.com/OAI/OpenAPI-Specification/pull/2361#issuecomment-702683204) as in the other PR.

I changed the fields so the examples contain the same information.
This includes changing the text and adding the `required` field.

MikeRalphson:
Thanks for the explanation! If you could change the target branch and remove the changes to published versions, I can approve.

MikeRalphson (approved):

darrelmiller (approved):

PR #2378, opened 2020-10-20
Create payment-buyer.yaml

hanbc7743:


MikeRalphson:
Thanks for the PR. Could you expand on why this OAS document forms a good example to have in the spec. repository, above and beyond the examples for version 3.0.x which are already there?

Also, do you own/maintain this API? I'm just wondering about potentially sending traffic to an API which may not be expecting it.

MikeRalphson:
Ping @hanbc7743 re: above.

MikeRalphson:
Closing due to inactivity - but please feel free to reopen the issue if necessary.

Issue #1973, opened 2019-07-15
Add `info.lifecycle` with lifecycle informations.


ioggstream:
## The proposal

A `LifeCycle` object to describe the API lifecycle, eg:

```
info:
  lifecycle:
    maturity: published   # deprecated | retired
    published: 1970-01-01
    deprecated: 2020-01-01
    retired: 2020-06-01
```

## Benefits

Automatic discovery of API status.

## Considerations

- http://apisjson.org/ seems not maintained anymore
- we already have `deprecation` informations in operations
- lifecycle could contain further informations related to eg. versioning (link to previous spec versions)

ioggstream:
@dberlind @dret further consideration on how to improve this feature request are very welcome!

dberlind:
@loggstream, based on our observations over the last **13+ YEARS** of cataloguing APIs and there version, we've made a few observations about API status across API providers and have, for the most part, narrowed a current version's status down to one of the following (trying to keep it very simple):

1. Pre-Release
2. Recommended (active, supported)
3. Active (supported, scheduled for retirement)
4. Retired (still active, but unsupported and scheduled for deactivation)
5. Deactivated

Note that the term "deprecated" is not included here. In our research, we have found that there are different meanings for "deprecated."  And so, in our parlance, the verb "deprecate" is not included as an official term but essentially means the advancement of an API from levels 2-4 to some following level. For example, when you change an API's status from level 2 to 3, that is an act of deprecation. Same would go for changing an API's status from 3-5 (though we wouldn't suggest that API providers ever make this leap. 

For extra credit, when a status indicates some forthcoming deprecation (ie: the  way level 3 indicates that retirement is coming), you could include a date field for when that change is planned to take place.

By the way, one other key use of this "version graph" would be the automation of regression testing such that breaking changes between versions are sussed out. This advantage to my proposal was raised by Alexander Pournelle (I don't have his Github ID yet, but will add it when I hear from him next).

dberlind:
Also, for the sake of "holy cow," just thought I'd point out that the related issue ([#1397](https://github.com/OAI/OpenAPI-Specification/issues/1397)) is just a change in the order of same digits of this issue's number.  Might be time to play the lottery.

dret:
On 2019-07-15 21:54, dberlind wrote:
>  1. Pre-Release
>  2. Recommended (active, supported)
>  3. Active (supported, scheduled for retirement)
[...]

PR #2386, opened 2020-10-28
Draft proposal for experimental field

davidjgoss:
(Not sure if this is the best place/right way to start this conversation, but would be interested in feedback.)

TLDR: a way to include items in an API but marked as "experimental" to set expectations that they could change or go away, and that wouldn't be a "breaking change" in semver terms.


proposals/005_Experimental.md
```diff
@@ -0,0 +1,115 @@
...
+
+- Tools like swagger-ui could surface this in the documentation they generate so consumers are made aware. Experimental items could also be filtered out of the documentation if desired.
+- Tools for detecting and preventing breaking changes could take this into consideration when deciding whether a change is breaking.
+
+## Detailed design
+
+A new boolean field named `experimental`, defaulting to `experimental`, is added to:
+
+- Operation
+- Schema
```
MikeRalphson, line 49: I wonder whether adding the `experimental` field to higher level objects (such as `parameter`, `requestBody` and `response`) and dropping it from `schema` would work better? It might resolve the 'unanswered question' about experimental schema use in non-experimental operations.

proposals/005_Experimental.md
```diff
@@ -0,0 +1,115 @@
...
+## Backwards compatibility
+
+The `experimental` field would default to false, meaning existing behaviour is preserved, and the new field is only used on an opt-in basis.
+
+## Alternatives considered
+
+- _Specification extensions_ - publishers could add an extension in their own domain, but the benefit of the metadata being available to downstream tools (including those used by consumers, not just publishers) would be lost.
+- _Tags_ - as above, but this also gets to mixing other kinds of metadata in with resource taxonomy, which seems wrong.
+- _Different API_ - this would be the least messy from a technical perspective - maintain a completely separate API for experimental items, and then "promote" them to the main API once they are considered stable. This has increased overhead for publishers and consumers, and could also reduce the likelihood of getting feedback on, and early uptake of, experimental items if they are segregated in a different place altogether.
+
```
MikeRalphson, line 115: Can we add a note about use of [overlays](https://github.com/OAI/OpenAPI-Specification/blob/master/proposals/004_Overlays.md#overlays) as a potential alternative? A semver-versioned base OAS document with one-or-more overlays labelled 'experimental' might meet the use-case too.

MikeRalphson (commented):
Thanks for the PR and making it through the proposals template! Just some thoughts.

proposals/005_Experimental.md
davidjgoss: I'd missed parameter - have added that now, with a qualifier that it couldn't be used with `in: path` since a change in path is a breaking change.

davidjgoss: I see your point about `schema`, I agree removing it from there would simplify. I'm not sure I understand how putting on `requestBody` and `response` instead would alleviate though, could you elaborate?

One use case I've had come up is having a subset of an object be experimental, like "we've added some more detail to this response, we're not totally sure about it yet but the rest of it remains stable in the meantime", so it would be a shame to lose that.

proposals/005_Experimental.md
davidjgoss: Overlays looks very interesting! I do think it could work although perhaps more complex to implement against. Added an item to the alternatives section.

davidjgoss:
@MikeRalphson thanks for the feedback and apologies for taking a while to follow up. I've made some edits and responded inline.

MikeRalphson:

davidjgoss:
@MikeRalphson no problem, that's done.

earth2marsh:
Since this PR was submitted, we've created some special interest groups (SIGs), and one of them is planning to address the [lifecycle](https://github.com/OAI/sig-lifecycle), which I think this falls into. There isn't a lot of traction yet, and we're still working out how the group will actually operate, but I'd also suggest we use Discussions, either in the spec repo (this one) or possibly around the new repo linked above. There are some similarities to #1973 worth noting, as well. (Thanks @darrelmiller for flagging this proposal to the nascent lifecycle SIG!)

davidjgoss:
Thanks @earth2marsh, that looks interesting.

What would be the right next step for this PR?

earth2marsh:
TBH, I'm not sure. I think as the lifecycle SIG gets traction, that this would be the basis of a discussion there where the overall lifecycle issues are considered? I'm not sure anyone has stepped up to run it yet, but I plan to stay involved and will flag this (and reach out) once it begins getting momentum.

handrews:
Filed #3257 to revive and clarify our proposals process.  Once we figure it out, we'll resolve this PR accordingly.

handrews:
@OAI/tsc is this proposal, like PR #3286, better started as an `x-` field to test it out?  (@davidjgoss there is a registry for such fields which @miqui is updating as it has been a bit under-utilized).

If so, it would seem like this document should be associated with that, rather than a proposal in this repo.  If not, we will get to looking at the draft proposal process soon and can evaluate this once we've revived that process.

lornajane:
This is already "out there" as an extension (citation: https://github.com/search?q=%22x-experimental%3A+true%22&type=code), and there was quite a bit of engagement around the issues for it. Can we say what is needed for this to merit consideration for 3.2?

handrews:
@lornajane oh nice!  It... did not even occur to me to check 😅 🤦 

Yeah, that seems like a good reason to at least accept the proposal into the repo (meaning merging this PR, assuming it is in alignment with the `x-` usage that's actually happening).  We can then work from the merged proposal document on the actual spec change.  At least I think that's how it's supposed to work?

hudlow:
For what it's worth, we use `x-ibm-release-level` with possible values of `experimental`, `beta`, and `preview`. It seems like it would be worth considering whether to combine this kind of annotation into a single enumeration property with `deprecated` and possibly some other values.

lornajane:
The `x-experimental` also shows up in the extensions dataset from 2021 https://github.com/Mermade/openapi-specification-extensions/blob/main/extensions/2021/extensions.tsv#L154 so it looks like it's actively in use. I also spot `x-deprecated` here and agree with @hudlow that we could consider covering more lifecycle states.

Reports of an API Lifecycle SIG have been greatly exaggerated, it looks like there's a repo and a slack channel but no activity since the setup a couple of years ago in either.

proposals/2020-10-28-Experimental.md
```diff
@@ -0,0 +1,128 @@
...
+
+This kind of requirement is handled for TypeScript libraries by [api-extractor](https://api-extractor.com/pages/tsdoc/doc_comment_syntax/#release-tags) - they have both "alpha" and "beta" markers with a somewhat opinionated flow attached - I'm not sure that level of granularity is necessary. But the "beta" and "public" ones map well to the motivations described here:
+
+> - **beta**: Indicates that an API item has been released as a preview or for experimental purposes. Third parties are encouraged to try it and provide feedback. However, a “beta” API should NOT be used in production, because it may be changed or removed in a future version.
+> - **public**: Indicates that an API item has been officially released, and is now part of the supported contract for a package. If the SemVer versioning scheme is used, then the API signature cannot be changed without a MAJOR version increment.
+
+### Unanswered Questions
+
+- If an operation is not marked as experimental, but it is using a schema which is (i.e. as its request object), then it is implicitly also unstable. Would this usage be considered invalid?
+- Should `experimental` and `deprecated` be mutually exclusive?
```
adamaltman, line 116: No, something can be experimental and become deprecated (for example, the experiment evolved or is abandoned).

lornajane:
@davidjgoss Would you be able to revisit this proposal and make some of the proposed updates? I'm in support of us adopting this change.

davidjgoss:
@lornajane sure, I’ll try and get those changes done over the next few days or so.

(Thanks everyone for input.)

ralfhandl (dismissed):

proposals/2020-10-28-Experimental.md
davidjgoss: I tend to agree and will make this edit. Just want to call out that it could be seen as conflicting with the idea of having a single "lifecycle" enum as mentioned elsewhere.

proposals/2020-10-28-Experimental.md
```diff
@@ -113,12 +113,13 @@ This kind of requirement is handled for TypeScript libraries by [api-extractor](
 ### Unanswered Questions
 
 - If an operation is not marked as experimental, but it is using a schema which is (i.e. as its request object), then it is implicitly also unstable. Would this usage be considered invalid?
```
ralfhandl, line 115: In the request object example only requests using an experimental feature are potentially unstable, so I'd see this as a valid combination.

davidjgoss:
> merging this PR, assuming it is in alignment with the x- usage that's actually happening

Just noting the `x-experimental` usage that I can find in the wild is on operations and schemas. We have included parameters also in this proposal, but I'm not opposed to omitting them if we feel that's a better way to start.

proposals/2020-10-28-Experimental.md
```diff
@@ -113,12 +113,13 @@ This kind of requirement is handled for TypeScript libraries by [api-extractor](
...
-- Should `experimental` and `deprecated` be mutually exclusive?
 
 ## Backwards compatibility
 
 The `experimental` field would default to false, meaning existing behaviour is preserved, and the new field is only used on an opt-in basis.
 
+`experimental` can coexist with `deprecated` - an operation, parameter or schema can be both experimental and deprecated, having never gotten to a stable point before being deprecated.
```
ralfhandl, line 121: Not sure whether I'd bother with deprecating an experimental feature, the Motivation section states that experimental aspects can

> just get removed

and the Proposed solution section repeats that it 

> may change or be removed without a major version bump

What would it mean to deprecate an experimental feature as opposed to not removing it? That it will still work in this major version and disappear with the next major version bump?

ralfhandl (approved):

lornajane (approved):
I think this is a solid proposal that adds something that we do see used "in the wild". I'm approving and accepting this pull request, the changes can be added n 3.1 or any later minor version.

Issue #2392, opened 2020-11-02
3.1.0 - Addition of "summary" to the API info is confusing


InfoSec812:
What does "summary" give us that "description" does not? Perhaps "summary" could just be an alias for "description"?

MikeRalphson:
See #832 and #1779 for the rationale for this field being added and the differences between `title`, `summary` and `description`. The OAS does not have the concept of aliases, and I for one would not like to see them introduced.

What do you find confusing?

InfoSec812:
I think that some of the context from those two issues could help clarify in the 3.1.0 document. Right now, I did not understand that from what I read in the standard. If it is explained somewhere else, then perhaps a link to that explanation?

MikeRalphson:
The line is always between conciseness, clarity, consistency and completeness. The `summary` and `description` fields have their different usage descriptions framed in the same manner as for other objects (such as `pathItem` and `operation`). If you can formulate a wording improvement to either or both field descriptions, a PR would be welcome.

darrelmiller:
@InfoSec812 When presenting information it is common to have a list view of data and a detail view.  `summary` is optimized to represent data in a list view, `description` is optimized for a detail view.  `description` enables rich text which is useful in documentation.  `summary` does not support rich text as that can cause presentation issues when presenting a list of summary values.

`summary` is an optional field so you can choose not to use it if it is not useful for your scenario.  Adding summary to info was a request from the community and it created consistency with other objects in the specification.  This felt like sufficient justification for its inclusion.

InfoSec812:
@darrelmiller Now that I have had it explained by @MikeRalphson I agree. What I would like to do, and thus the PR, is make that obvious to everyone who reads the specification/standard by adding the definition and providing links to that information.

darrelmiller:
@philsturgeon  is going to remove the word "short" from the description of description.  Hopefully this is more descriptive.

philsturgeon:
https://github.com/OAI/OpenAPI-Specification/pull/2408

MikeRalphson:
#2408 was merged, closing this. Thanks all.

PR #2393, opened 2020-11-02
Improve docs about summary and description

InfoSec812:
Addresses #2392 by creating 2 new definitions at the beginning of the specification and linking to those definitions wherever `summary` and `description` are used in the standard.

@MikeRalphson I think that this can safely "balance" between verbosity and clarity. Looking forward to your thoughts.

### DIFF
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -96,6 +96,12 @@ Some examples of possible media type definitions:
 The HTTP Status Codes are used to indicate the status of the executed operation. 
 The available status codes are defined by [RFC7231](https://tools.ietf.org/html/rfc7231#section-6) and registered status codes are listed in the [IANA Status Code Registry](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml).
 
+##### <a name="summary"></a>Summaries
+In various places throughout the OpenAPI Specification, fields may have a `summary`. The summary is intended to be a short overview of that item and SHOULD be less than 120 characters. It MUST be plaintext and not contain formatting like Markdown or HTML.
+
+##### <a name="description"></a>Descriptions
+In various places throughout the OpenAPI Specification, fields may have a `description`. The description is intended to be a long-form explanation of that item and MAY be arbitrary length. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation.
+
 ## Specification
 
 ### Versions
@@ -202,8 +208,8 @@ The metadata MAY be used by the clients if needed, and MAY be presented in editi
 Field Name | Type | Description
 ---|:---:|---
 <a name="infoTitle"></a>title | `string` | **REQUIRED**. The title of the API.
-<a name="infoSummary"></a>summary | `string` | A short summary of the API.
-<a name="infoDescription"></a>description | `string` | A short description of the API. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation.
+<a name="infoSummary"></a>summary | `string` | A short [summary](#summary) of the API.
+<a name="infoDescription"></a>description | `string` | A short [description](#description) of the API. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation.
 <a name="infoTermsOfService"></a>termsOfService | `string` | A URL to the Terms of Service for the API. MUST be in the format of a URL.
 <a name="infoContact"></a>contact | [Contact Object](#contactObject) | The contact information for the exposed API.
 <a name="infoLicense"></a>license | [License Object](#licenseObject) | The license information for the exposed API.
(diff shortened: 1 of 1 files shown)


versions/3.1.0.md
```diff
@@ -96,6 +96,12 @@ Some examples of possible media type definitions:
 The HTTP Status Codes are used to indicate the status of the executed operation. 
 The available status codes are defined by [RFC7231](https://tools.ietf.org/html/rfc7231#section-6) and registered status codes are listed in the [IANA Status Code Registry](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml).
 
+##### <a name="summary"></a>Summaries
+In various places throughout the OpenAPI Specification, fields may have a `summary`. The summary is intended to be a short overview of that item and SHOULD be less than 120 characters. It MUST be plaintext and not contain formatting like Markdown or HTML.
```
MikeRalphson, line 100:
```suggestion
In various places throughout the OpenAPI Specification, objects may have a `summary`. The summary is intended to be a short overview of that item. It MUST be plaintext and not contain formatting like Markdown or HTML.
```

versions/3.1.0.md
```diff
@@ -96,6 +96,12 @@ Some examples of possible media type definitions:
...
+
+##### <a name="description"></a>Descriptions
+In various places throughout the OpenAPI Specification, fields may have a `description`. The description is intended to be a long-form explanation of that item and MAY be arbitrary length. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation.
```
MikeRalphson, line 103:
```suggestion
In various places throughout the OpenAPI Specification, objects may have a `description`. The description is intended to be a long-form explanation of that item and MAY be arbitrary length. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation.
```

LikeLakers2:
You ought to include those headers in the Table of Contents, for consistency with how everything else is in the TOC. :)

MikeRalphson (changes requested):
Please don't reintroduce the 120-character suggested limit, this was deliberately dropped in v3.0.0. I'm not convinced by the value of top level definitions as a solution to your issue, as it requires readers to have two regions of the spec. in mind at the same time.

InfoSec812:
> I'm not convinced by the value of top level definitions as a solution to your issue, as it requires readers to have two regions of the spec. in mind at the same time.
I'm interested in better understanding your concerns @MikeRalphson ... In my mind, the links are there for *if* someone wants clarification, but they can be ignored when users of the Spec are already familiar with the differences between summary/description.

versions/3.1.0.md
InfoSec812: Thanks for that background! I have updated the text.

versions/3.1.0.md
InfoSec812: Thank you for helping with making this consistent! I should have noticed from my review that there had been a concerted effort to use the term "object", sorry for the poor choice of terms.

hkosova:
> It MUST be plaintext and not contain formatting like Markdown or HTML.

This sentence is probaby unnecessary because the ["Rich Text Formatting"](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.1.0.md#rich-text-formatting) section states that only `description` fields support Markdown.

MikeRalphson:
> I'm interested in better understanding your concerns @MikeRalphson ... In my mind, the links are there for if someone wants clarification, but they can be ignored when users of the Spec are already familiar with the differences between summary/description.

I think the fact that they can be ignored points towards the fact that in their current state, they're merely redundant, not adding anything above and beyond the `summary` and `description` field descriptions in the field tables. But, and this is a biggie, I'm just one voice, and the point of the PR is to let others weigh in.

InfoSec812:
@MikeRalphson With regards to the definitions and reciprocal links, I feel like it is useful for those who are **NOT** as deeply familiar with the OAS standards so that they can quickly become familiar without having to seek out an expert to get an explanation. My own case being a potential example. Even though I am a frequent and regular user of OAS, I was not aware and did not quickly see the reasoning of why summary/description were different and how they were different. I would like to think that we could create a standards document which is more inclusive and fosters people at all levels of experience to be able to understand and contribute to the discussion.

InfoSec812:
> This sentence is probably unnecessary because the ["Rich Text Formatting"](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.1.0.md#rich-text-formatting) section states that only `description` fields support Markdown.
Resolved, and agreed that it is well explained further down...

> Plus, "must not contain formatting" may be interpreted as if validators/linters need to explicitly check for the presence of Markdown / HTML tags in summaries and in that case flag the definition as invalid, which looks like an overkill.
Removed.

Issue #2371, opened 2020-10-10
Where is the schema instance for 3.1?


adamaltman:
I left a comment [here](https://github.com/OAI/OpenAPI-Specification/pull/2370/files#r502806631), but since the pull request was already merged, I decided to open this issue.

I wonder if this URL should resolve to some content (instead of a 404): https://spec.openapis.org/oas/3.1/schema-object

I was expecting a meta schema object at that URL.

MikeRalphson:
No, JSON Schema vocabularies are referenced by URI not URLs, so this is not expected to resolved. We can probably make this clearer in the spec.

adamaltman:
When I go here: https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.5 I can access that URI via my browser and download the meta-schema. I think this is what I find confusing.

MikeRalphson:
The current schema **instance** for OAS 3.0 has an `id` of `https://spec.openapis.org/oas/3.0/schema/2019-04-02` (which is a URI not a URL, as required by JSON Schema). It is currently located at https://raw.githubusercontent.com/OAI/OpenAPI-Specification/master/schemas/v3.0/schema.json

We can possibly do better in automatically publishing changed schema instances via our `gh-pages` branch which will cause them to show up under the `spec.openapis.org` domain. (PRs from GitHub Actions wizards gratefully received.)

There is no current schema **instance** for OAS 3.1, as that release is still a work-in-progress.

Because of this confusion between the dialect `$schema` value describing the [schemaObject](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.1.0.md#schema-object) and a schema **instance** I will propose changing the `$schema` value in an upcoming PR.

MikeRalphson:
Linked issue #2314

Relequestual:
> When I go here: https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.5 I can access that URI via my browser and download the meta-schema. I think this is what I find confusing. - @adamaltman

Worth noting that JSON Schema 2020-12, while published, hasn't published the associated meta-schema files yet.

Meta-schemas are informative only, and we may also update them (as per the spec) to fix any issues where it doesn't match the spec.

MikeRalphson:
Thanks to work by @jdesrosiers and @philsturgeon we have a first version here https://github.com/OAI/OpenAPI-Specification/blob/master/schemas/v3.1/schema.json

Issue #2365, opened 2020-10-01
Provide guidance on how to override the implicit OpenAPI vocabulary


darrelmiller:
Words that clarify are required.  Concepts than only Ben and Mike seem to understand.

MikeRalphson:
#2399 is merged.

Issue #2364, opened 2020-10-01
Create an OpenAPI JSON Schema vocabulary


darrelmiller:
Vocab must support these fields.  URI must be created to identify this vocab.

Field Name | Type | Description
-- | -- | --
discriminator | Discriminator Object | Adds support for polymorphism. The discriminator is an object name that is used to differentiate between other schemas which may satisfy the payload description. See Composition and Inheritance for more details.
xml | XML Object | This MAY be used only on properties schemas. It has no effect on root schemas. Adds additional metadata to describe the XML representation of this property.
externalDocs | External Documentation Object | Additional external documentation for this schema.
example | Any | A free-form property to include an example of an instance for this schema. To represent examples that cannot be naturally represented in JSON or YAML, a string value can be used to contain the example with escaping where necessary.Deprecated: The example property has been deprecated in favor of the JSON Schema examples keyword. Use of example is discouraged, and later versions of this specification may remove it.

...and @Relequestual  to help :-)

MikeRalphson:
#2399 is merged.

Issue #2363, opened 2020-10-01
Clarify the relationship between OpenAPI 3.1 and the JSON Schema version


darrelmiller:
The current 3.1 spec says,

> The Schema Object allows the definition of input and output data types. These types can be objects, but also primitives and arrays. This object is a superset of the JSON Schema Specification Draft 2019-09.

We should agree on wording to clarify that future versions of JSON Schema (assuming non-breaking) are also usable with OpenAPI 3.1. (wordsmithing required)

In order to achieve forward compatibility with JSON Schema, we need to define the OpenAPI keywords as a vocabulary.  And define how the OpenAPI vocabulary is applied to the JSON.

darrelmiller:
Wait to see if JSON Schema 2020 makes it out to meet our schedule.  We can decide on wording at the last minute.

MikeRalphson:
#2399 is merged.

PR #2399, opened 2020-11-04
OAS schema dialect clarifications

MikeRalphson:
This PR covers a few outstanding issues from the 3.1-Final milestone:

* #2363 mechanism for use of different JSON Schema drafts
* #2364 add the default `$schema` id before our extension vocabulary
* #2365 overriding the `$schema` dialect
* #2371 change the previous suggested `$schema` value to avoid confusion with schema instances

It also changes a few JSON Schema Draft specification links either to use the `https` scheme or to make them more specific.

The mechanism for setting a default value for `$schema` for schemaObjects within the OAS document could be modelled in several different ways. And while not wanting to over-engineer anything, or provide extension points which may never be needed, it felt right to use a containing `defaults` object with a `schema` defaults object within it. This *does* allow us to have a place to put other things like global default `parameters`, `responses` in the future, which have been asked for repeatedly, and over a long period.

Alternatively to having one `$schema` field which is **REQUIRED** in that schema defaults object, we could possibly allow other JSON Schema core vocabulary keywords using patterned fields (`^\$.*`), but I could not make a definite case for this. I don't think it's a good idea to make the schema defaults object a `schemaObject` itself as the JSON Schema folk have taught us that merging schemas is hard. 

In reviews, I would especially like a check that my terminology (especially around `$schema`, "schema resoources", "dialects", "meta-schemas" and "vocabularies") is correct and clear. @Relequestual could you also weigh in? Wording / sentence ordering changes also welcome of course.

### DIFF
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -35,6 +35,8 @@ An OpenAPI definition can then be used by documentation generation tools to disp
 		- [Server Object](#serverObject)
 		- [Server Variable Object](#serverVariableObject)
 		- [Components Object](#componentsObject)
+		- [Defaults Object](#defaultsObject)
+		- [Schema Defaults Object](#schemaDefaultsObject)
 		- [Paths Object](#pathsObject)
 		- [Path Item Object](#pathItemObject)
 		- [Operation Object](#operationObject)
@@ -133,13 +135,13 @@ In order to preserve the ability to round-trip between YAML and JSON formats, YA
 
 ### <a name="documentStructure"></a>Document Structure
 
-An OpenAPI document MAY be made up of a single document or be divided into multiple, connected parts at the discretion of the user. In the latter case, `$ref` fields MUST be used in the specification to reference those parts as follows from the [JSON Schema](https://json-schema.org) definitions.
+An OpenAPI document MAY be made up of a single document or be divided into multiple, connected parts at the discretion of the user. In the latter case, `$ref` fields MUST be used in the OAS document to reference those parts as follows from the [JSON Schema](https://json-schema.org/draft/2019-09/json-schema-core.html#references) definitions.
 
 It is RECOMMENDED that the root OpenAPI document be named: `openapi.json` or `openapi.yaml`.
 
 ### <a name="dataTypes"></a>Data Types
 
-Data types in the OAS are based on the types supported by the [JSON Schema Specification Draft 2019-09](http://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.4.2). 
+Data types in the OAS are based on the types supported by the [JSON Schema Specification Draft 2019-09](https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.4.2).
 Note that `integer` as a type is also supported and is defined as a JSON number without a fraction or exponent part. 
 Models are defined using the [Schema Object](#schemaObject), which is a superset of JSON Schema Specification Draft 2019-09.
 
@@ -640,6 +642,58 @@ components:
             read:pets: read your pets
 ```
 
+#### <a name="defaultsObject"></a>Defaults Object
+
+Contains globally-applied defaults to specific areas of the OAS document.
+
+##### Fixed Fields
+
+Field Name | Type | Description
+---|:---|---
+<a name="defaultsSchema"></a> schema | [Schema Defaults Object](#schemaDefaultsObject)] | An object to hold defaults for [Schema Objects](#schemaObject).
+
+This object MAY be extended with [Specification Extensions](#specificationExtensions).
+
+##### Defaults Object Example
+
+```json
+"defaults": {
+  "schema": {
+    "$schema": "http://json-schema.org/draft-07/schema#"
+  }
+}
+```
+
+```yaml
+defaults:
+  schema:
+    $schema: 'http://json-schema.org/draft-07/schema#'
+```
+
+#### <a name="schemaDefaultsObject"></a>Schema Defaults Object
+
+Contains globally-applied defaults for [Schema Objects](#schemaObject).
+
+##### Fixed Fields
+
+Field Name | Type | Description
+---|:---|---
+<a name="schemaDefaultsSchema"></a> $schema | `string` | **REQUIRED**. The default value for the `$schema` keyword within [Schema Objects](#schemaObject) contained within this OAS document. MUST be in the form of a URI.
+
+This object MAY be extended with [Specification Extensions](#specificationExtensions).
+
+##### Schema Defaults Object Example
+
+```json
+"schema": {
+  "$schema": "http://json-schema.org/draft-07/schema#"
+}
+```
+
+```yaml
+schema:
+  $schema: 'http://json-schema.org/draft-07/schema#'
+```
 
 #### <a name="pathsObject"></a>Paths Object
 
@@ -2285,7 +2339,7 @@ $ref: definitions.yaml#/Pet
 #### <a name="schemaObject"></a>Schema Object
 
 The Schema Object allows the definition of input and output data types.
-These types can be objects, but also primitives and arrays. This object is a superset of the [JSON Schema Specification Draft 2019-09](http://json-schema.org/).
+These types can be objects, but also primitives and arrays. This object is a superset of the [JSON Schema Specification Draft 2019-09](https://json-schema.org/specification-links.html#2019-09-formerly-known-as-draft-8).
 
 For more information about the properties, see [JSON Schema Core](https://json-schema.org/draft/2019-09/json-schema-core.html) and [JSON Schema Validation](https://json-schema.org/draft/2019-09/json-schema-validation.html).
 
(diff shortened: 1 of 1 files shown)


philsturgeon:
I think "schema resoources" is the Canadian spelling, right @darrelmiller? 

The defaults object feels a little concerning, but if its _just_ for setting default $schema that might not be so bad. Can we discuss this on the call today?

MikeRalphson:
> I think "schema resoources" is the Canadian spelling, right @darrelmiller?

😆 At least it's only in the notes and not in the diff!

> The defaults object feels a little concerning, but if its just for setting default $schema that might not be so bad. Can we discuss this on the call today?

Yep, I was in several minds how to achieve the default for `$schema`. It's on the agenda, so hopefully wise people will have thoughts.

handrews:
I'm very concerned about a default schema object, as it changes the meaning of schemas in a very non-obvious way.  There are exactly two schema keywords for which it is valid to externally specify default _behavior_, which is different from setting default _values_.

The initial base URI, normally set by the top-level `$id`, can be determined in the absence of such an `$id` by external means per RFC 3986.  We have specifically updated JSON Schema in the next draft to state that when JSON Schema is embedded in another media type, then in the absence of `$id` it takes on the base URI set by the enclosing document, according to that media type's rules.  Which of course, ultimately default to the document's retrieval URI if the document does not set it explicitly.

The meta-schema, normally set by `$schema`, is the other keyword where implementations _must_ make an assumption if it is not explicitly present.  Here is what the forthcoming draft says about it:

                         The "$schema" keyword SHOULD be used in the document root schema object,
                         and MAY be used in the root schema objects of embedded schema resources.
                         It MUST NOT appear in non-resource root schema objects.  If absent from
                         the document root schema, the resulting behavior is implementation-defined.

                         If multiple schema resources are present in a single document, then
                         schema resources which do not have a "$schema" keyword in their root
                         schema object MUST be processed as if "$schema" were present with the
                         same value as for the immediately enclosing resource.

The "implementation" here can include whatever implementation _of an OAS processor_ is being used to read the OAS document and hand off the JSON Schema processing to a JSON Schema implementation (which might be embedded in the OAS implementation or might be separate).  Therefore, if OAS has a special rule for determining that value, like "check the `schemaDialect` field of the `defaults` object", then by the time it logically reaches the JSON Schema implementation it knows what to tell that implementation to use in the absence of `$schema`.

Many if not most implementations today actually require you to set this externally anyway even if you do have `$schema` present.  For implementations that do not have such a capability and actually rely on `$schema`, yes you can inject a `$schema` into the JSON structure before handing it off.  But conceptually, that is a workaround and not a real defaulting behavior.  The distinction is important.

Having a schema object in a defaults produces an implicit non-local change in behavior, that I assume applies recursively to nested schemas?  So now everything has to be processed by some sort of recursive schema walker in order to produce something that could be handed off to a JSON Schema implementation, which means that the schema you get is not the schema it looks like you have.  Which is true even if the defaults are not set recursively.

Please don't do this.  Please consider specifying the default meta-schema as its own thing, not as a default schema object.  Applications are allowed to externally determine the processing meta-schema specifically because `$schema`, like `$id`, is needed to bootstrap that processing.  That should not be taken as a reason to offer broader external defaulting.

We can clarify the language in the JSON Schema spec by adding something like:

                         If a schema object is embedded in a document of another media type, then
                         that document MAY specify the meta-schema by which embedded schema
                         objects that lack a "$schema" in the root object are to be processed.

This would parallel the language currently used with the initial base URI and `$id`, and avoid the problems that would arise with non-local defaulting.

MikeRalphson:
@handrews the `schema` object within the `defaults` object is **not** a `schemaObject` type. It is a strictly limited object which only has one requ$ired field: `$schema`, which is an default value to use in place of the implicit OAS 3.1 schemaObject dialect id.

Does this assuage any of your concern, or am I misunderstanding?

MikeRalphson:
> If a schema object is embedded in a document of another media type, then that document MAY specify the meta-schema by which embedded schema objects that lack a "$schema" in the root object are to be processed.

If I understand this correctly, this is exactly what this PR aims to enable.

handrews:
@MikeRalphson 

> I explicitly said above
>
> I don't think it's a good idea to make the schema defaults object a schemaObject itself as the JSON Schema folk have taught us that merging schemas is hard.

I saw this at 6AM unable to go back to sleep and saw that it was on the agenda for this morning's meeting so I replied.  I searched for "merge" trying to find this but didn't think to search for "merg" or "merging."

> Does this assuage any of your concern, or am I misunderstanding?

Yeah, it works.  I think it's a bad idea to make something that looks too much like a schema object, but isn't.  I'd just have a `metaSchema` field or something.

But I'm not going to fight over it, I only answered at all because you specifically assigned me as a reviewer.  So, yes it should work.  It's my opinion that it will cause problems because people will treat a field called "schema" in a section called "defaults" as a default schema object, but if y'all want to deal with that confusion that's fine.

MikeRalphson:
Henry makes very good points, and they were design decisions I did hum and hah about. Happy to rework to avoid the `schema` vs `schemaObject` confusion.

versions/3.1.0.md
```diff
@@ -2333,13 +2391,15 @@ As such, inline schema definitions, which do not have a given id, *cannot* be us
 The [xml](#schemaXml) property allows extra definitions when translating the JSON definition to XML.
 The [XML Object](#xmlObject) contains additional information about the available options.
 
-###### Picking Schema Vocabularies
+###### Specifiying Schema Dialects
+
+It is important for tooling to be able to determine which dialect or meta-schema any given resource wishes to be processed with: JSON Schema Core, JSON Schema Validation, OpenAPI Schema dialect, or some custom meta-schema.
 
-It is important for tooling to be able to detect what meta-schema any given resource wishes to be processed with: JSON Schema Core, JSON Schema Validation, OpenAPI Schema Object, or some custom meta schema.
+The `$schema` keyword MAY be present in any root Schema Object (but MUST NOT occur in subschemas defined directly within those Schema Objects), and if present MUST be used to determine which dialect should be used when processing the schema. This allows use of Schema Objects which comply with other drafts of JSON Schema than the default Draft 2019-09 support. Tooling MUST support the <a href="#dialectSchemaId">OAS dialect schema id</a>, and MAY support additional values of `$schema`.
```
Relequestual, line 2398:
```suggestion
The `$schema` keyword MAY be present in any root Schema Object, and if present MUST be used to determine which dialect should be used when processing the schema. This allows use of Schema Objects which comply with other drafts of JSON Schema than the default Draft 2019-09 support. Tooling MUST support the <a href="#dialectSchemaId">OAS dialect schema id</a>, and MAY support additional values of `$schema`.
```

`$schema` in a subschema is explicitly allowed when bundling a schema to aviod external references. This creates a Compound Schema Document. This is a recently closed PR for 2020-NN.

versions/3.1.0.md
```diff
@@ -2294,14 +2348,18 @@ Where JSON Schema indicates that behavior is defined by the application (e.g. fo
 
 ##### Properties
 
-The OpenAPI Schema Object is a JSON Schema vocabulary which extends JSON Schema Core and Validation vocabularies. As such any keyword available for those vocabularies is by definition available in OpenAPI, and will work the exact same way.
+The OpenAPI Schema Object is described by a JSON Schema [dialect](https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.4.3.3), comprising of a base [vocabulary](https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.4.3.3) which extends the JSON Schema Core and Validation vocabularies. As such any keyword available within those vocabularies is by definition available in a Schema Object, and will work in exactly the same way.
```
Relequestual, line 2351: This reads a little awkwardly and I'm totally sure what you're trying to convey.

There are a few vocabularies, including `format`, and `applicator`, rather than just "core and validation" which are the enclosing specification documents.

Let's talk about this?

Relequestual (changes requested):
Left a few comments.

I may not see any follow up comments. I tend to rely on slack notifying me of github activity for specific projects... or rather just JSON Schema.
You know where to find me =]

versions/3.1.0.md
MikeRalphson: Discussed with Ben and updated.

Relequestual (approved):

darrelmiller:
Proposed @OAI/tsc change:
```
jsonSchemaDialect: http://json-schema.org/draft-2019/schema#

```

karenetheridge:
> jsonSchemaDialect: http://json-schema.org/draft-2019/schema#

with no trailing '#', please. The most recent drafts are stricter about where empty fragments are allowed/preferred.

darrelmiller:
I don't know why I added that trailing #.  Wasn't intentional!

darrelmiller (approved):

versions/3.1.0.md
```diff
@@ -2294,16 +2294,20 @@ Where JSON Schema indicates that behavior is defined by the application (e.g. fo
 
 ##### Properties
 
-The OpenAPI Schema Object is a JSON Schema vocabulary which extends JSON Schema Core and Validation vocabularies. As such any keyword available for those vocabularies is by definition available in OpenAPI, and will work the exact same way.
+The OpenAPI Schema Object [dialect](https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.4.3.3) is defined as requiring the [OAS base vocabulary](#baseVocabulary), in addition to the vocabularies as specified in the JSON Schema draft 2019-09 [general purpose meta-schema](http://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.8).
+
+The OpenAPI Schema Object dialect for this version of the specification is identified by the URI `https://spec.openapis.org/oas/3.1/dialect/base` (the <a name="dialectSchemaId"></a>"OAS dialect schema id").
```
handrews, line 2299: Something that just occurred to me due to the ongoing discussion about `discriminator`, particularly if the `selectBy` idea is adopted: It might be better to put `discriminator` in its own vocabulary since it is complex to implement and (particularly if `selectBy` were added) not strictly necessary.

Putting it in its own vocabulary allows implementations to choose whether to implement it separately from the other OAS extension keywords, and could allow for dialect configurations that require the other OAS extension keywords but leave `discriminator` as optional.  idk if that's really a great idea or not but thought I would mention it.  We moved `unevaluatedProperties` and `unevaluatedItems` into their own vocabulary for similar reasons- they are complex and potentially expensive so some use cases might prefer not to deal with them (streaming implementations, for example, are much more costly with these).

darrelmiller: We believe this issue is one that requires further conversation but don't believe it should hold up merging the current PR that introduces the dialect property.

PR #2401, opened 2020-11-05
Add Device Code authorization to the supported OAuth Flows

### LINKED ISSUE #2381: Add "Device Code" as a supported OAuth Flow
Hi! I'm not in the business of making APIs myself, so I don't know how helpful I can be. However, I found that [RFC8628](https://tools.ietf.org/html/rfc8628) offers a "Device Code" type of OAuth, which I thought might be useful if it were supported by default by OpenAPI. The flow is described at <https://oauth.net/2/grant-types/device-code/> as follows:

> The Device Code grant type is used by browserless or input-constrained devices in the device flow to exchange a previously obtained device code for an access token.

Looking at the RFC myself, it seems like this type of OAuth Flow would need the a device authorization URL (different from a regular `authorizationUrl` in that this is called by the application, not the user -- the RFC has more info under section 3.1), a `tokenUrl`, and potentially a `refreshUrl`. ~~Given how this endpoint works, it may also benefit from a optional `deviceCodeFormat` field, specifying what characters are allowed, how the code can be formatted (uppercase? lowercase? case-insensitive? etc.) and so on.~~ (Turns out I was misreading the RFC -- the authorization server generates the device code, not the client)

Hopefully this isn't too far-fetched of a suggestion! :)

LikeLakers2:
Closes #2381. 

This is a PR for the Device Authorization OAuth Flow, based on [RFC8628](https://tools.ietf.org/html/rfc8628). I suggested this over at #2381 and was asked to make a PR for it.

The URL used for this flow is named `deviceAuthorizationUrl`, separate from `authorizationUrl` as the two have completely different semantics and usages, so I was concerned that using `authorizationUrl` for both might be confusing. That said, I would be happy to merge this into `authorizationUrl` if requested.

Also, I originally made a comment that this flow may benefit from a optional `deviceCodeFormat` field. However, I looked at the RFC again, and found that I originally misread the RFC -- the authorization server is the one that creates the code to be used, not the client wanting authorization.

### DIFF
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -3277,7 +3277,7 @@ Field Name | Type | Applies To | Description
 ---|:---:|---|---
 <a name="oauthFlowAuthorizationUrl"></a>authorizationUrl | `string` | `oauth2` (`"implicit"`, `"authorizationCode"`) | **REQUIRED**. The authorization URL to be used for this flow. This MUST be in the form of a URL. The OAuth2 standard requires the use of TLS.
 <a name="oauthFlowDeviceAuthorizationUrl"></a>deviceAuthorizationUrl | `string` | `oauth2` (`"deviceAuthorization"`) | **REQUIRED**. The device authorization URL to be used for this flow. This MUST be in the form of a URL. The OAuth2 standard requires the use of TLS.
-<a name="oauthFlowTokenUrl"></a>tokenUrl | `string` | `oauth2` (`"password"`, `"clientCredentials"`, `"authorizationCode"`, `"deviceCode"`) | **REQUIRED**. The token URL to be used for this flow. This MUST be in the form of a URL. The OAuth2 standard requires the use of TLS.
+<a name="oauthFlowTokenUrl"></a>tokenUrl | `string` | `oauth2` (`"password"`, `"clientCredentials"`, `"authorizationCode"`, `"deviceAuthorization"`) | **REQUIRED**. The token URL to be used for this flow. This MUST be in the form of a URL. The OAuth2 standard requires the use of TLS.
 <a name="oauthFlowRefreshUrl"></a>refreshUrl | `string` | `oauth2` | The URL to be used for obtaining refresh tokens. This MUST be in the form of a URL. The OAuth2 standard requires the use of TLS.
 <a name="oauthFlowScopes"></a>scopes | Map[`string`, `string`] | `oauth2` | **REQUIRED**. The available scopes for the OAuth2 security scheme. A map between the scope name and a short description for it. The map MAY be empty.
 
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -3161,7 +3161,7 @@ animals:
 
 Defines a security scheme that can be used by the operations.
 
-Supported schemes are HTTP authentication, an API key (either as a header, a cookie parameter or as a query parameter), mutual TLS (use of a client certificate), OAuth2's common flows (implicit, password, client credentials and authorization code) as defined in [RFC6749](https://tools.ietf.org/html/rfc6749), and [OpenID Connect Discovery](https://tools.ietf.org/html/draft-ietf-oauth-discovery-06).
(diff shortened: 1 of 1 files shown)


versions/3.1.0.md
```diff
@@ -3264,6 +3264,7 @@ Field Name | Type | Description
 <a name="oauthFlowsPassword"></a>password| [OAuth Flow Object](#oauthFlowObject) | Configuration for the OAuth Resource Owner Password flow 
 <a name="oauthFlowsClientCredentials"></a>clientCredentials| [OAuth Flow Object](#oauthFlowObject) | Configuration for the OAuth Client Credentials flow.  Previously called `application` in OpenAPI 2.0.
 <a name="oauthFlowsAuthorizationCode"></a>authorizationCode| [OAuth Flow Object](#oauthFlowObject) | Configuration for the OAuth Authorization Code flow.  Previously called `accessCode` in OpenAPI 2.0.
+<a name="oauthFlowsDeviceCode"></a>deviceCode| [OAuth Flow Object](#oauthFlowObject) | Configuration for the OAuth Device Code flow.
```
LikeLakers2, line 3267: Before this gets merged, I have two questions I'd like to ask about this line.

1. I called it "Device Code" because that is how <https://oauth.net/2/grant-types/device-code> names it. However, the RFC is named "OAuth 2.0 Device Authorization Grant". Should I change this to be named "Device Authorization"?

2. Should [RFC8628](https://tools.ietf.org/html/rfc8628) be referenced here? I understand that the standard OAuth Flows don't have a RFC linked with them, but I feel like the Device Code flow may not be as well-known,

LikeLakers2, line 3267: I called it "Device Code" because that is how <https://oauth.net/2/grant-types/device-code> names it. However, the RFC is named "OAuth 2.0 Device Authorization Grant". Should I change this to be named "Device Authorization"?

MikeRalphson: See L:3164/5 for a paragraph which also needs to be updated, and would be a good place to add this link.

MikeRalphson (changes requested):

versions/3.1.0.md
LikeLakers2: Just to make sure I'm understanding that correctly -- you want me to put the RFC link there and *not* here? Or do you want me to put it both there and here?

MikeRalphson: Just in the one place (the earlier paragraph) so that list of flows is expanded, and the link is present in the same part of the spec. as for other flows. Thanks.

LikeLakers2: No problem.

Silly question though, should I refer to it as "OAuth2 device code flow", to keep it consistent with this name?

Or should I refer to it as "OAuth2 device authorization grant flow"? That's what the RFC calls it. If I refer to it as this, I might want to update the property name here too (see my other review, just below this, for more details on that).

MikeRalphson: Though the only place we currently use the term "grant" in the spec is in the recommendation...

> Recommended for most use case is Authorization Code Grant flow with PKCE.

I'm inclined to go with what the RFC calls it.

LikeLakers2: In that case, I think I'll use "OAuth2 device authorization flow", especially since it slides off the tongue a bit better too. I'll update this property name too, just to avoid confusion on which property is for this flow.

LikeLakers2:
@MikeRalphson Let me know if those changes I made upon your request look good to you. :)

MikeRalphson:
@darrelmiller I was going through [the unmerged PRs that were automatically closed when the `v3.1.0-dev` branch was deleted](https://github.com/OAI/OpenAPI-Specification/pulls?q=is%3Apr+is%3Aclosed+base%3Av3.1.0-dev+is%3Aunmerged+closed%3A2021-07-08) and I noticed you self-requested a review on this one. Do you feel we should ask @LikeLakers2 to resubmit against `v3.2.0-dev`?

LikeLakers2:
Let me know if you want me to. I wouldn't mind resubmitting against v3.2.0-dev. :)

koalo:
Any updates on this? I am really looking forward to it!

LikeLakers2:
@webron I'm unsure if that means you want me to file against 3.2.0. Do you want me to, or would you prefer I wait?

In either case, I also want to mention: I have no qualms about someone (whether you, omz13, or someone else) copying and remaking my PR (whether wholesale or edited). So if you feel that might be easier than having me resubmit, then by all means do so.

LikeLakers2:
@webron ...So you want me to resubmit this against the 3.2.0 branch, correct?

webron:
Yes please.

Issue #1514, opened 2018-03-19
Description is not allowed for elements containing $ref


vincnetas:
I know that specification says that when using $ref any other elements are ignored.
So here a real life scenario.

>         location:
>           description: User home address location  <- error
>           $ref: "#/components/schemas/Location"
>         carLocation:
>           description: Where user has left his car  <- error
>           $ref: "#/components/schemas/Location"

So if i have same structure of location but different semantics, only way to communicate this would be through field names, but that is very limiting. 
Also two locations could be created with identical structure, just to distinguish between semantics, but this is code duplication.

I understand that this is not a trivial issue, but this is obvious shortcoming in spec which i noticed after just couple of days of basic real life usage... And i'm not the only one.

There are people with the same problem here
https://github.com/swagger-api/swagger-editor/issues/1184#issuecomment-323070576

darrelmiller:
There are few points regarding this:
- This is a design choice that the OpenAPI specification inherited from JSON Schema.  The JSON Schema folks are working on trying to find a solution to this issue.  That would be the best place to have this conversation.
- This is what appears at first to be a simple issue and then upon deeper inspection it becomes a non-trivial issue to resolve all the possible conflicts arising from merging two overlapping schemas.  This is an issue that has been discussed literally for years.
- The `allOf` operator can be used to define one schema to define the local properties, e.g. description and another schema with the $ref.  It doesn't solve all scenarios and it is not the most concise syntax but it can work.

hkosova:
Related (or duplicate): https://github.com/OAI/OpenAPI-Specification/issues/556

avantipande1:
I was looking for same use case where I want to add description for the $ref field in swagger.  Above answer helped, when we add description with allOf, swagger-ui picks up this description.

greatvovan:
This looks like a serious design flaw in both OpenAPI and JSON Schema.

LikeLakers2:
> * This is a design choice that the OpenAPI specification inherited from JSON Schema. The JSON Schema folks are working on trying to find a solution to this issue. That would be the best place to have this conversation.

[...]

PR #2405, opened 2020-11-10
Improve wording about summary and description fields in Reference Object

greatvovan:
Word "define" is used in two meanings in software development: ~"declare" and ~"assign". In this particular case I would like to avoid possible confusion of a reader caused by understanding it as the former case: "if 'description' field is not set in Reference Object, it will be ignored". Some more details here: https://github.com/OAI/OpenAPI-Specification/issues/1514#issuecomment-720141763

If you can suggest a better wording in English, I would appreciate that.


MikeRalphson (approved):
LGTM, thanks.

PR #2418, opened 2020-12-04
Fix introductory clause comma, hyphen in compound adjective

kernc:
These popped up during a quick reading.

### DIFF
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -159,8 +159,8 @@ The formats defined by the OAS are:
 `string` | `password` | A hint to UIs to obscure input.
 
 ### <a name="richText"></a>Rich Text Formatting
-Throughout the specification `description` fields are noted as supporting CommonMark markdown formatting.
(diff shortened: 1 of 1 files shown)


MikeRalphson (commented):
Thanks for the PR. I find the additional commas somewhat interrupt the sentence flow, especially the second change.

The changes of `out-of band` to `out-of-band` though looks definitely correct.

kernc:
But especially the second one is needed, drawing a distinction between "rich text it MUST support" and "rich text" in general.
_"it must support"_ is not part of _"rich text"_ noun phrase, i.e.

```lisp
; This syntax tree is invalid
; https://www.link.cs.cmu.edu/link/submit-sentence-4.html
(S Where OpenAPI
   (S (NP tooling)
      (VP renders
          (NP rich text it MUST support ,       ; <-- Uh?
              (PP at
                  (NP (NP a minimum)
                      ,
                      (NP (NP markdown syntax)
                          as
                          (VP described
                              (PP by
                                  (NP CommonMark)))))
                  .)))))
```

What, arguably, interrupts the sentence flow is the nonrestrictive adverbial embedded clause _"at a minimum"_. But it's _correct_ to delimit that as well.

PR #2432, opened 2020-12-31
Use https:// for JSON Schema

philsturgeon:
An issue came up at Stoplight which makes me wonder why we've not already
noticed this. https://github.com/stoplightio/spectral/issues/1432

With http:// being used, if anything tries to resolve those schemas, they can
have security warnings for tools that dont like http and https being mixed.
Browsers for example.

I'm sure there will be concerns about modifying old stuff. I'm happy to revert
as many of these changes as you like, but at least the metaschemas for 3.0 and
3.1 should do this.

### DIFF
--- a/schemas/v1.2/apiDeclaration.json
+++ b/schemas/v1.2/apiDeclaration.json
@@ -1,6 +1,6 @@
 {
     "id": "https://raw.githubusercontent.com/OAI/OpenAPI-Specification/master/schemas/v1.2/apiDeclaration.json#",
-    "$schema": "https://json-schema.org/draft-04/schema#",
+    "$schema": "http://json-schema.org/draft-04/schema#",
     "type": "object",
     "required": [ "swaggerVersion", "basePath", "apis" ],
     "properties": {
--- a/schemas/v1.2/apiDeclaration.json
+++ b/schemas/v1.2/apiDeclaration.json
@@ -1,6 +1,6 @@
 {
     "id": "https://raw.githubusercontent.com/OAI/OpenAPI-Specification/master/schemas/v1.2/apiDeclaration.json#",
-    "$schema": "http://json-schema.org/draft-04/schema#",
+    "$schema": "https://json-schema.org/draft-04/schema#",
     "type": "object",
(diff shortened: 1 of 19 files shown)


MikeRalphson:
Those `$schema`values are URIs not URLs and must match the `id`/ `$id` values defined by JSON schema, or nothing will work. Either 2019-09 or 2020-12 was the first draft to use an https `$id`URI (I forget which and I'm on my phone).

karenetheridge:
Not entirely related but I just noticed that http://json-schema.org/latest/... is used throughout -- a link to a specific draft should be used instead, as semantics change between versions.

MikeRalphson:
> Not entirely related but I just noticed that http://json-schema.org/latest/... is used throughout -- a link to a specific draft should be used instead, as semantics change between versions.

@karenetheridge there are no references to http://json-schema.org/latest/... in version 3.1 and we are waiting on Ben and the rest of you fine JSON Schema folk to publish the 2020-12 draft to the JSON Schema website in order to update our links to the latest draft.

Any references in v1.2 of OAS were probably the best links available at that time, and we don't update published versions of the spec after release.

philsturgeon:
v1.2 changes reverted as requested. Let's update URLs in the later specs because its browser based and isnt breaking, and lets update the URL in v3.0 because its just a URL change and more secure is better. :D

MikeRalphson:
> v1.2 changes reverted as requested. Let's update URLs in the later specs because its browser based and isnt breaking...

The same rationale applies to not updating v1.2 as to v2.0 and v3.0.x - we don't amend published versions of the spec.

> and lets update the URL in v3.0 because its just a URL change and more secure is better. :D

If you mean the metaschema change, no this isn't just a URL change. It's a **URI** change, and that URI has to match character for character (including any scheme) with the `id` of the JSON Schema Draft 04 metaschema.

Without those changes I think this PR is a nop.

>

Relequestual:
As Mike pointed out, we do not amend published version of the spec. They should be immutable!
My vote is to close this issue.

PR #2434, opened 2021-01-05
Fixed the webhook example

### LINKED ISSUE #2430: webhook need a paths?
I was reading the changelog of the 3.1.0-rc0 and it says:
``While previous versions required paths, now a valid OpenAPI Document can describe only webhooks, or even only reusable components. Thus, an OpenAPI Document no longer necessarily describes an API.``

But on the file examples/v3.1/webhook-example.yaml it states that:
``OpenAPI documents all need a paths element``

 So can i have a openapi only with a ``webhooks`` in it?

christianfds:
Fixes #2430

Signed-off-by: Christian Franchin dos Santos <christian@corp.kionux.com.br>

### DIFF
--- a/examples/v3.1/webhook-example.json
+++ b/examples/v3.1/webhook-example.json
@@ -4,38 +4,6 @@
     "title": "Webhook Example",
     "version": "1.0.0"
   },
-  "paths": {
-    "/pets": {
-      "get": {
-        "summary": "List all pets",
-        "operationId": "listPets",
-        "parameters": [
-          {
-            "name": "limit",
-            "in": "query",
-            "description": "How many items to return at one time (max 100)",
-            "required": false,
-            "schema": {
-              "type": "integer",
-              "format": "int32"
-            }
-          }
-        ],
-        "responses": {
-          "200": {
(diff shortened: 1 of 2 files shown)


Polo2:
I'm not sure webhook example has to be fixed by removing `/paths`, since it's specified for version 3.1:
> The OpenAPI document MUST contain at least one paths field, a components field or a webhooks field

A documentation may include both `paths` and `webhooks`, ad it's done in this example. WDYT ?

Polo2:
> I'm not sure webhook example has to be fixed by removing `/paths`, since it's specified for version 3.1:
> 
> > The OpenAPI document MUST contain at least one paths field, a components field or a webhooks field
> 
> A documentation may include both `paths` and `webhooks`, ad it's done in this example. WDYT ?

Ooops, my bad, I should have read related issue #2430 before adding this comment... 🙏  Sorry!
Indeed, a webhook example with only `webhooks` , and not extra `paths` could help to understand this new feature

darrelmiller (approved):

darrelmiller:
@christianfds Thank you for your contribution.  Much appreciated.

Issue #1679, opened 2018-08-31
Clarify statement about `Reference obects` in `Responses object`


ricellis:
The description of [Responses object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#responsesObject) (for both fixed and patterned fields) reads:
>A Reference Object can link to a response that the OpenAPI Object's components/responses section defines.

The use of the word "can" is particularly misleading here. Does this mean `MAY` or `MUST`?
That is, can any reference object be used, or is this implying a constraint that the reference object has to be of the form `#/components/responses` when used in a responses object?

darrelmiller:
Will change to  "The Reference Object MUST link to an OpenAPI Response Object."

mkistler:
I think this change will allow a `$ref` to a response object within the "responses" of another operation in the API doc.  Is that the desired result?  If so, should we allow the same kind of behavior for parameters?  I.e., rather than 
> The list can use the Reference Object to link to parameters that are defined at the OpenAPI Object's components/parameters.

Should we say 
> The list can use the Reference Object to link to an OpenAPI Parameter Object

?

MikeRalphson:
Yes, I'll try to mop up all similar wording. There would be no reason to disallow a `$ref` to another operation's `responses` object, but re-use under `components/schemas` is more common and readable.

MikeRalphson:
It was decided to drop these sentences completely in 3.1 for consistency with the rest of the document where `[ TypedObject | ReferenceObject ]` appears in the `type` column. Thanks all.

PR #2435, opened 2021-01-07
Added wording to clarify scope of references

darrelmiller:
Previous wording was ambiguous as to what a $ref could point to.  This PR adds text to be more precise.

### DIFF
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -2235,7 +2235,7 @@ description: Pets operations
 
 #### <a name="referenceObject"></a>Reference Object
 
-A simple object to allow referencing other objects in the OpenAPI document, internally and externally. The target of the reference must match the type required by the context of the reference. Targets of a reference do not need to be contained in a components section and for external references, targets do not need to exist within a valid OpenAPI description. The validity of the target MUST be based on the context of the reference object.
+A simple object to allow referencing other objects in the OpenAPI document, internally and externally. Targets of a reference do not need to be contained in a components section and for external references, targets MAY exist within any compatible resource. Referenced objects are subject to the same constraints as inline objects.
 
 ##### Fixed Fields
 Field Name | Type | Description
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -2235,7 +2235,7 @@ description: Pets operations
 
 #### <a name="referenceObject"></a>Reference Object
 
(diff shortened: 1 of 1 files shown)


versions/3.1.0.md
```diff
@@ -2235,7 +2235,7 @@ description: Pets operations
 
 #### <a name="referenceObject"></a>Reference Object
 
-A simple object to allow referencing other components in the OpenAPI document, internally and externally.
+A simple object to allow referencing other objects in the OpenAPI document, internally and externally. The target of the reference must match the type required by the context of the reference. Targets of a reference do not need to be contained in a components section and for external references, targets do not need to exist within a valid OpenAPI description. The validity of the target MUST be based on the context of the reference object.
```
MikeRalphson, line 2238:
```suggestion
A simple object to allow referencing other objects in the OpenAPI document, internally and externally. The target of the reference must match the type required by the context of the reference. Targets of a reference do not need to be contained in a components section and for external references, targets do not need to exist within a valid OpenAPI document. The validity of the target MUST be based on the context of the reference object.
```

MikeRalphson (changes requested):

MikeRalphson:
See also #1679 to see if we can remove any now-redundant wording or improve elsewhere.

MikeRalphson (approved):

darrelmiller:
@OAI/tsc   Can I get another sign off?

versions/3.1.0.md
```diff
@@ -2235,7 +2235,7 @@ description: Pets operations
 
 #### <a name="referenceObject"></a>Reference Object
 
-A simple object to allow referencing other components in the OpenAPI document, internally and externally.
+A simple object to allow referencing other objects in the OpenAPI document, internally and externally. Targets of a reference do not need to be contained in a components section and for external references, targets MAY exist within any compatible resource. Referenced objects are subject to the same constraints as inline objects.
```
webron, line 2238: Should we stick to a single term here? We use `targets` and `referenced objects` to mean the same thing.

darrelmiller, line 2238:
```suggestion
A simple object to allow referencing other objects in the OpenAPI document, internally and externally. Targets of a reference do not need to be contained in a components section and for external references, targets MAY exist within any compatible resource. Targets are subject to the same constraints as inline objects.
```

PR #2437, opened 2021-01-08
Update JSON Schema Draft to 2020-12 and make $ref resolution rules explicit

MikeRalphson:
Firstly, apologies that this PR does more than one thing, I didn't have time to create multiple branches and keep switching between them when authoring the commits (which sometimes overlap). It may be helpful to review the individual commits first and then the changes as a whole.

This PR:

* Updates the JSON Schema Draft links to Draft 2020-12 using the canonical IETF links
* Makes explicit the relative `$ref` resolution rules (using wording taken from the expired [JSON Reference spec](https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03)
* Makes Reference Objects, pathItem `$ref`s and Example Object `externalValue` fields always relative to the referring document, not the servers array. This makes a distinction between OAS document *authoring* and API *operation* (for example oAuth endpoints).
* Allows Example Object `externalValue` to be a URI and not just a URL - for consistency with other `$ref`-like fields
* Makes it clear pathItem `$ref`s don't now need to be external
* Makes consistent wording around "This MUST be in the **form** of a ..."
* Makes RFC link display text consistent with regard to spacing
* Fixes one stray `http:` link to `spec.commonmark.org` to be `https:` for consistency

### DIFF
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -2247,7 +2247,7 @@ The `$ref` string value contains a URI [RFC3986](https://tools.ietf.org/html/rfc
 
 Resolution of a Reference Object SHOULD yield the referenced value. Implementations MAY choose to replace the reference with the referenced value.
 
-If the URI contained in the `$ref` value is a relative URI, then the Base URI resolution MUST be calculated according to [RFC3986](https://tools.ietf.org/html/rfc3986#section-5.2). Resolution is performed relative to the referring document.
+If the URI contained in the `$ref` value is a [Relative Reference](https://tools.ietf.org/html/rfc3986#section-4.2), then the Base URI resolution MUST be calculated according to [RFC3986](https://tools.ietf.org/html/rfc3986#section-5.2). Resolution is performed relative to the referring document.
 
 If a URI contains a fragment identifier, then the fragment should be resolved per the fragment resolution mechanism of the referrant document.  If the representation of the referrant document is JSON or YAML, then the fragment identifier SHOULD be interpreted as a JSON-Pointer as per [RFC6901](https://tools.ietf.org/html/rfc6901).
 
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -1535,7 +1535,7 @@ When passing complex objects in the `application/x-www-form-urlencoded` content 
 
 It is common to use `multipart/form-data` as a `Content-Type` when transferring request bodies to operations.  In contrast to 2.0, a `schema` is REQUIRED to define the input parameters to the operation when using `multipart` content.  This supports complex structures as well as supporting mechanisms for multiple file uploads.
 
-In a `multipart/form-data` request body, each schema property, or each element of a schema array property, takes a section in the payload with an internal header as defined by [RFC 7578](https://tools.ietf.org/html/rfc7578). The serialization strategy for each property of a `multipart/form-data` request body can be specified in an associated [`Encoding Object`](#encodingObject).
+In a `multipart/form-data` request body, each schema property, or each element of a schema array property, takes a section in the payload with an internal header as defined by [RFC7578](https://tools.ietf.org/html/rfc7578). The serialization strategy for each property of a `multipart/form-data` request body can be specified in an associated [`Encoding Object`](#encodingObject).
 
 When passing in `multipart` types, boundaries MAY be used to separate sections of the content being transferred — thus, the following default `Content-Type`s are defined for `multipart`:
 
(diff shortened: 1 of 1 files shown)


versions/3.1.0.md
```diff
@@ -1943,7 +1945,7 @@ Field Name | Type | Description
 <a name="exampleSummary"></a>summary | `string` | Short description for the example.
 <a name="exampleDescription"></a>description | `string` | Long description for the example. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation.
 <a name="exampleValue"></a>value | Any | Embedded literal example. The `value` field and `externalValue` field are mutually exclusive. To represent examples of media types that cannot naturally represented in JSON or YAML, use a string value to contain the example, escaping where necessary.
-<a name="exampleExternalValue"></a>externalValue | `string` | A URL that points to the literal example. This provides the capability to reference examples that cannot easily be included in JSON or YAML documents.  The `value` field and `externalValue` field are mutually exclusive. 
+<a name="exampleExternalValue"></a>externalValue | `string` | A URI that points to the literal example. This provides the capability to reference examples that cannot easily be included in JSON or YAML documents.  The `value` field and `externalValue` field are mutually exclusive.
```
darrelmiller, line 1948: How would this work with a URI that is not a URL?  A consumer would not know how to resolve the URI to locate the external example.

versions/3.1.0.md
```diff
@@ -2241,10 +2243,18 @@ description: Pets operations
 
 A simple object to allow referencing other components in the OpenAPI document, internally and externally.
 
+The `$ref` string value contains a URI [RFC3986](https://tools.ietf.org/html/rfc3986), which identifies the location of the value being referenced.
```
darrelmiller, line 2246: We need to decide what to do about our use of the term URI.  If the reference is relatative, then it is technically not a URI.

versions/3.1.0.md
```diff
@@ -2241,10 +2243,18 @@ description: Pets operations
...
+
+Resolution of a Reference Object SHOULD yield the referenced value. Implementations MAY choose to replace the reference with the referenced value.
+
+If the URI contained in the `$ref` value is a relative URI, then the Base URI resolution MUST be calculated according to [RFC3986](https://tools.ietf.org/html/rfc3986#section-5.2). Resolution is performed relative to the referring document.
```
darrelmiller, line 2250: As far as a I understand it, there is no such thing as a "relative URI".

karenetheridge: s/relative URI/a URI that is not absolute/ (i.e. a uri-reference)

versions/3.1.0.md
karenetheridge: s/URI/uri-reference/

versions/3.1.0.md
karenetheridge: Then the user would have to make the resource known to the implementation via some other means. This is what JSON Schema specifies.  https://tools.ietf.org/html/draft-bhutton-json-schema-00#section-9.1.2

MikeRalphson: Yes as @karenetheridge says this is for consistency with JSON Schema's `$ref`, and as I amended it, our Reference Object.

versions/3.1.0.md
MikeRalphson: Agreed - "URI Reference" as per https://tools.ietf.org/html/rfc3986#section-4.1

webron: We're already beyond the point of not changing semantics ;)

versions/3.1.0.md
```diff
@@ -2241,10 +2243,18 @@ description: Pets operations
 
 A simple object to allow referencing other components in the OpenAPI document, internally and externally.
 
+The `$ref` string value contains a URI [RFC3986](https://tools.ietf.org/html/rfc3986), which identifies the location of the value being referenced.
+
+Resolution of a Reference Object SHOULD yield the referenced value. Implementations MAY choose to replace the reference with the referenced value.
```
webron, line 2248: `Implementations MAY choose to replace the reference with the referenced value.` - why mention this?

MikeRalphson: Because it is wording directly from https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03 so represents the status quo of what we were referencing in 3.0.x

MikeRalphson: If a concern is that replacing the Reference Object with the referenced value is in conflict with the notion of having sibling properties to the `$ref`, then we can pinch wording from JSON Schema such as:

> Its results are the results of the referenced ~~schema~~ value.

MikeRalphson:
I've taken another swing at this, based on the TSC call feedback. I've also split the `#relativeReferences` definition into separate sections for URIs and URLs - and for the former, I have linked back to it where relevant.

versions/3.1.0.md
```diff
@@ -162,12 +163,20 @@ The formats defined by the OAS are:
...
+Relative references, including those in [`Reference Objects`](#referenceObject), [`PathItem Object`](#pathItemObject) `$ref` fields, [`Link Object`](#linkObject) `operationRef` fields and [`Example Object`](#exampleObject) `externalValue` fields, are resolved using the referring document as the Base URI according to [RFC3986](https://tools.ietf.org/html/rfc3986#section-5.2).
+
+If a URI contains a fragment identifier, then the fragment should be resolved per the fragment resolution mechanism of the referenced document.  If the representation of the referenced document is JSON or YAML, then the fragment identifier SHOULD be interpreted as a JSON-Pointer as per [RFC6901](https://tools.ietf.org/html/rfc6901).
+
+Relative references in [`Schema Objects`](#schemaObject), including any that appear as `$id` values, use the nearest parent `$id` as a Base URI, as described by [JSON Schema Specification Draft 2020-12](https://tools.ietf.org/html/draft-bhutton-json-schema-00#section-8.2).  If no parent schema contains an `$id`, then the Base URI MUST be determined according to [RFC3986](https://tools.ietf.org/html/rfc3986#section-5.1).
```
Relequestual, line 174: This does make it sound like using `$id` isn't in line with RFC3986, but it is, because `$id` is the "Base URI embedded in content" as defined in said RFC. Not super important, but if anyone else reads it that way too, then I don't know if this needs a slight tweak.

darrelmiller (approved):

PR #2442, opened 2021-01-12
Clean-up wording about $refs in responsesObjects, fixes #1679

### LINKED ISSUE #1679: Clarify statement about `Reference obects` in `Responses object`

MikeRalphson:
This is on top of @darrelmiller's clean-up in aisle #2435 

An alternative is to drop these two sentences altogether - as we don't seem to use equivalent language anywhere else we have a `Reference Object | Actual Object` scenario.

### DIFF
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -1650,12 +1650,12 @@ call.
 ##### Fixed Fields
 Field Name | Type | Description
 ---|:---:|---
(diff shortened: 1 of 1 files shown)


MikeRalphson:
TODO: drop these sentences as per TSC meeting.

darrelmiller (approved):

PR #2465, opened 2021-02-16
workflows: yaml2json - update used actions

MikeRalphson:
The version of `peter-evans/create-pull-request` we were using is no-longer supported. Updated to `v3`. This meant updating to `checkout@v2` and removing a deprecated option.


MikeRalphson:
Closing as I will fold this into the PR which updates references from `master` to `main`

PR #2466, opened 2021-02-17
Fix syntax error in map type of webhooks field specification

haumacher:
Syntax error in `webhooks` type description there was an superflous ']' character that does not belong to the `Map[...]` type.


MikeRalphson:
Thanks @haumacher - could you retarget this against `versions/3.1.1.md` on the `v3.1.1-dev` branch?

MikeRalphson (changes requested):
LGTM but target file and branch need to be updated.

PR #1911, opened 2019-04-30
actions: reinstate validation of schemas and examples, fixes #1739

MikeRalphson:
As promised, following the merging of #1270 / #1897

**Note**: *items related to publishing an npm package containing the schemas will be moved to a new issue*.

This PR now reinstates the validation of examples and schemas using GitHub Actions.

PR #2474, opened 2021-02-18
OASv3.1 Schema

### LINKED ISSUE #2314: Create a complete meta schema for 3.1

philsturgeon:
Closes #2314.

Adding a JSON Schema for the whole OpenAPI v3.1 document, as plenty of tooling requires this before they can upgrade to OpenAPI v3.1. 

This is based on a gist from @handrews, and is being tweaked for OpenAPI v3.1. https://gist.github.com/handrews/6dfebd56ef97328f9e4dc7a47a1e8bc7

This is different to the OpenAPI dialect/metaschema work being done first in #2016 but is starting again in another yet unmade PR.

### DIFF
--- a/schemas/v3.1/schema.yaml
+++ b/schemas/v3.1/schema.yaml
@@ -237,6 +237,8 @@ $defs:
               url:
                 type: string
                 format: uri-reference
+              identifier:
+                type: string
             unevaluatedProperties: false
 
       Server:
--- a/schemas/v3.1/schema.yaml
+++ b/schemas/v3.1/schema.yaml
@@ -11,12 +11,11 @@ anyOf:
       - paths
   - required:
       - webhooks
-allOf:
-  - $ref: '#/$defs/Mixins/$defs/Extensible'
+$ref: '#/$defs/Mixins/$defs/Extensible'
 properties:
   openapi:
     type: string
-    pattern: ^3\.1\.\d(-.+)?$
+    pattern: ^3\.1\.\d+(-.+)?$
   jsonSchemaDialect:
     type: string
     format: uri
@@ -36,6 +35,7 @@ properties:
     type: array
     items:
       $ref: '#Tag'
+    uniqueItems: true
   paths:
     $ref: '#Paths'
   webhooks:
@@ -289,7 +289,7 @@ $defs:
                   $ref: '#MediaType'
               links:
                 type: object
-                additionalProperteis:
+                additionalProperties:
                   $ref: '#Link'
             unevaluatedProperties: false
       MediaType:
@@ -372,9 +372,7 @@ $defs:
           parameters:
             type: array
             items:
-              oneOf:
-                - $ref: '#Parameter'
-                - $ref: '#Reference'
+              $ref: '#Parameter'
             uniqueItems: true
       Operation:
         $anchor: Operation
@@ -398,6 +396,7 @@ $defs:
             type: array
             items:
               $ref: '#Parameter'
+            uniqueItems: true
           requestBody:
             $ref: '#RequestBody'
           responses:
@@ -422,7 +421,7 @@ $defs:
           default:
             $ref: '#Response'
         patternProperties:
-          '[1-5](?:\d{2}|XX)':
+          '^[1-5](?:\d{2}|XX)$':
             $ref: '#Response'
         minProperties: 1
         unevaluatedProperties: false
@@ -451,12 +450,11 @@ $defs:
         $anchor: ExternalDocumentation
         allOf:
           - $ref: '#/$defs/Mixins/$defs/Extensible'
+          - $ref: '#/$defs/Mixins/$defs/Describable'
         type: object
         required:
           - url
         properties:
-          description:
-            type: string
           url:
             type: string
             format: uri-reference
@@ -556,7 +554,7 @@ $defs:
               - $ref: '#/$defs/Mixins/$defs/Extensible'
             properties:
               type:
-                string: null
+                type: string
                 enum:
                   - apiKey
                   - http
@@ -564,12 +562,12 @@ $defs:
                   - mutualTLS
                   - oauth2
                   - openIdConnect
-                oneOf:
-                  - $ref: '#APIKey'
-                  - $ref: '#HTTP'
-                  - $ref: '#MutualTLS'
-                  - $ref: '#OAuth2'
-                  - $ref: '#OpenIdConnect'
+            oneOf:
+              - $ref: '#APIKey'
+              - $ref: '#HTTP'
+              - $ref: '#MutualTLS'
+              - $ref: '#OAuth2'
+              - $ref: '#OpenIdConnect'
             unevaluatedProperties: false
         $defs:
           APIKey:
@@ -767,7 +765,7 @@ $defs:
                   wrapped:
                     type: boolean
                     default: false
-            additionalProperties: false
+            unevaluatedProperties: false
       Components:
         $anchor: Components
         type: object
@@ -784,7 +782,7 @@ $defs:
               $ref: '#Schema'
           responses:
             additionalProperties:
-              $ref: '#Responses'
+              $ref: '#Response'
           parameters:
             additionalProperties:
               $ref: '#Parameter'
--- a/schemas/v3.1/schema.yaml
+++ b/schemas/v3.1/schema.yaml
@@ -350,7 +350,7 @@ $defs:
         allOf:
           - $ref: '#/$defs/Mixins/$defs/Extensible'
         patternProperties:
-          ^\/:
+          ^/:
             type: object
             oneOf:
               - $ref: '#Reference'
--- a/schemas/v3.1/schema.yaml
+++ b/schemas/v3.1/schema.yaml
@@ -22,6 +22,8 @@ properties:
   openapi:
     type: string
     pattern: ^3\.1\.\d(-.+)?$
+  jsonSchemaDialect:
+    type: string
   info:
     $ref: '#Info'
   externalDocs:
@@ -372,19 +374,32 @@ $defs:
         patternProperties:
           '^\/':
             type: object
-            patternProperties:
-              '^(get|put|post|delete|options|head|patch|trace)$':
-                $ref: '#Operation'
             properties:
-              servers:
-                type: array
-                items:
-                  $ref: 'Server'
-              parameters:
-                type: array
-                items:
-                  $
+              oneOf:
+                - $ref: '#Reference'
+                - type: '#PathItem'    
         unevaluatedProperties: false
+
+      PathItem:
+        $id: '#PathItem'
+        type: object
+        allOf:
+          - $ref: '#/$defs/Mixins/$defs/Describable'
+          - $ref: '#/$defs/Mixins/$defs/Extensible'
+        properties:
+          summary:
+            type: string
+          servers:
+            type: array
+            items:
+              $ref: '#Server'
+          parameters:
+            type: array
+            items:
+              oneOf:
+                - $ref: '#Parameter'
+                - $ref: '#Reference'
+            uniqueItems: true
 
       Operation:
         $id: '#Operation'
@@ -567,11 +582,13 @@ $defs:
                   - apiKey
                   - http
                   - bearer
+                  - mutualTLS
                   - oauth2
                   - openIdConnect
                 oneOf:
                   - $ref: '#APIKey'
                   - $ref: '#HTTP'
+                  - $ref: '#MutualTLS'
                   - $ref: '#OAuth2'
                   - $ref: '#OpenIdConnect'
             unevaluatedProperties: false
@@ -602,6 +619,11 @@ $defs:
                     const: bearer
                   bearerFormat:
                     type: string
+          MutalTLS:
+            $id: '#MutalTLS'
+            properties:
+              type:
+                const: mutalTLS
           OAuth2:
             $id: '#OAuth2'
             required:
@@ -761,23 +783,23 @@ $defs:
     
           XML:
             $id: '#XML'
-            type: object
-            properties:
-              name:
-                type: string
-              namespace:
-                type: string
-                format: url
-              prefix:
-                type: string
-              attribute:
-                type: boolean
-                default: false
-              wrapped:
-                type: boolean
-                default: false
-            patternProperties:
-              '^x-': true
+            allOf:
+              - $ref: '#/$defs/Mixins/$defs/Extensible'
+              - type: object
+                properties:
+                  name:
+                    type: string
+                  namespace:
+                    type: string
+                    format: url
+                  prefix:
+                    type: string
+                  attribute:
+                    type: boolean
+                    default: false
+                  wrapped:
+                    type: boolean
+                    default: false
             additionalProperties: false
 
       Components:
@@ -802,7 +824,7 @@ $defs:
               $ref: '#Parameter'
           pathItems:
             additionalProperties:
-              $ref: '#PathItems'
+              $ref: '#PathItem'
           examples:
             additionalProperties:
               $ref: '#Example'
--- a/schemas/v3.1/schema.yaml
+++ b/schemas/v3.1/schema.yaml
@@ -1,4 +1,4 @@
-id: 'https://spec.openapis.org/oas/3.1/schema/2021-02-18'
+$id: 'https://spec.openapis.org/oas/3.1/schema/2021-02-18'
 $schema: 'https://json-schema.org/draft/2019-09/schema'
 type: object
 required:
--- a/schemas/v3.1/schema.yaml
+++ b/schemas/v3.1/schema.yaml
@@ -706,7 +706,7 @@ $defs:
           - $ref: '#Reference'
           - type: object
             allOf:
-              - $ref: 'https://json-schema.org/draft/2020-02/schema#'
+              - $ref: 'https://json-schema.org/draft/2020-12/schema'
               - $ref: '#SchemaExtensions'
               - $ref: '#/$defs/Mixins/$def/Extensible'
             unevaluatedProperties: false
--- a/schemas/v3.1/schema.yaml
+++ b/schemas/v3.1/schema.yaml
@@ -1,23 +1,18 @@
-# TODO Components Object now has a new entry pathItems, to allow for reusable Path Item Objects to be defined within a valid OpenAPI document.
-# TODO Added the jsonSchemaDialect top-level field to allow the definition of a default $schema value for Schema Objects.
-
-id: https://spec.openapis.org/oas/3.1/schema/2021-02-18
-$schema: 'https://json-schema.org/draft/2020-02/schema#'
-
+id: 'https://spec.openapis.org/oas/3.1/schema/2021-02-18'
+$schema: 'https://json-schema.org/draft/2019-09/schema'
 type: object
-
 required:
   - openapi
   - info
-
 oneOf:
-  - required: [components]
-  - required: [paths]
-  - required: [webhooks]
-
+  - required:
+      - components
+  - required:
+      - paths
+  - required:
+      - webhooks
 allOf:
   - $ref: '#/$defs/Mixins/$defs/Extensible'
-
 properties:
   openapi:
     type: string
@@ -25,7 +20,7 @@ properties:
   jsonSchemaDialect:
     type: string
   info:
-    $ref: '#Info'
+    $ref: '#/$defs/Info'
   externalDocs:
     $ref: '#ExternalDocumentation'
   servers:
@@ -45,7 +40,6 @@ properties:
   components:
     $ref: '#Components'
 unevaluatedProperties: false
-
 $defs:
   Mixins:
     $defs:
@@ -63,17 +57,14 @@ $defs:
           deprecated:
             type: boolean
             default: false
-
       DescReqDep:
         allOf:
           - $ref: '#/$defs/Mixins/$defs/Describable'
           - $ref: '#/$defs/Mixins/$defs/Requireable'
           - $ref: '#/$defs/Mixins/$defs/Deprecatable'
-
       Extensible:
         patternProperties:
-          '^x-': true
-
+          ^x-: true
       WithSingleExample:
         properties:
           example: true
@@ -89,18 +80,15 @@ $defs:
         oneOf:
           - $ref: '#/$defs/Mixins/$defs/WithSingleExample'
           - $ref: '#/$defs/Mixins/$defs/WithExampleObjects'
-
       WithSchema:
         properties:
           schema:
             $ref: '#Schema'
           content: false
-
       WithSchemaAndExamples:
         allOf:
           - $ref: '#/$defs/Mixins/$defs/WithSchema'
           - $ref: '#/$defs/Mixins/$defs/WithExamples'
-
       WithContent:
         properties:
           schema: false
@@ -110,7 +98,6 @@ $defs:
               $ref: '#MediaType'
             minProperties: 1
             maxProperties: 1
-
       WithNameAndLocation:
         required:
           - name
@@ -125,7 +112,6 @@ $defs:
               - query
               - header
               - cookie
-
       WithExplodeAndReserved:
         properties:
           explode:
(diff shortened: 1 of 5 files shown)


MikeRalphson:
@philsturgeon maybe this will help, I've updated `openapi3-examples` to have a 3.1 directory of pass/fail cases. Not everything would be picked up by a schema-only validator, but 🤷🏼 If you find any mistakes, let me know. https://github.com/Mermade/openapi3-examples/tree/master/3.1

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,843 @@
+# TODO Components Object now has a new entry pathItems, to allow for reusable Path Item Objects to be defined within a valid OpenAPI document.
+# TODO Added the jsonSchemaDialect top-level field to allow the definition of a default $schema value for Schema Objects.
+
+id: https://spec.openapis.org/oas/3.1/schema/2021-02-18
+$schema: 'https://json-schema.org/draft/2020-02/schema#'
```
jdesrosiers: Typo in draft name

```suggestion
$schema: 'https://json-schema.org/draft/2020-12/schema#'
```

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,843 @@
...
+              - $ref: '#/$defs/Mixins/$defs/Extensible'
+            unevaluatedProperties: false
+
+      Schema:
+        $id: '#Schema'
+        oneOf:
+          - $ref: '#Reference'
+          - type: object
+            allOf:
+              - $ref: 'https://json-schema.org/draft/2020-02/schema#'
```
jdesrosiers, line 709: Same typo here

```suggestion
              - $ref: 'https://json-schema.org/draft/2020-12/schema'
```

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,793 @@
+$id: 'https://spec.openapis.org/oas/3.1/schema/2021-02-18'
+$schema: 'https://json-schema.org/draft/2019-09/schema'
```
karenetheridge, line 2: 2020-12?

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,836 @@
...
+              - $ref: '#/$defs/Mixins/$defs/Extensible'
+            unevaluatedProperties: false
+
+      Schema:
+        $id: '#Schema'
+        oneOf:
+          - $ref: '#Reference'
+          - type: object
+            allOf:
+              - $ref: 'http://json-schema.org/draft-04/schema#'
```
karenetheridge: draft/2020-12?

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,836 @@
...
+          - $refs: '#/$defs/Mixins/$defs/StyledFormComplexNoDefaultOrEmpty'
+          - $refs: '#/$defs/Mixins/$defs/WithExplodeReservedAndEmpty'
+        properties:
+          style:
+            default: form
+
+  Objects:
+    $defs:
+      Reference:
+        $id: '#Reference'
```
karenetheridge: This should be $anchor, not $id (and the same for all other uses in this file).

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,836 @@
...
+                  - $ref: '#/$defs/Mixins/$defs/WithEmpty'
+            unevaluatedProperties: false
+
+      Paths:
+        $id: '#Paths'
+        type: object
+        allOf:
+          - $ref: '#/$defs/Mixins/$defs/Extensible'
+        patternProperties:
+          '^\/':
```
karenetheridge: It came up in a previous issue that there is no need to escape the slash here.

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,836 @@
...
+                  - allOf:
+                      - $ref: '#PathParam'
+                      - $ref: '#/$defs/Mixins/$defs/Describable'
+                      - $ref: '#/$defs/Mixins/$defs/Deprecatable'
+                      - required:
+                          - required
+                        properties:
+                          required:
+                            const: true
+                            default: true
```
karenetheridge, line 465: You may want to move this default up to the object level, as `default: { required: true }`, depending on how it is intended to be used (by implementations, as an annotation, rather than informatively by a human).

karenetheridge (changes requested):

schemas/v3.1/schema.yaml
karenetheridge: actually, I think this should be `https://spec.openapis.org/oas/3.1/dialect/base`?

philsturgeon: I'm leaving that for when #2467 is complete, as I'm defining the schema extensions below.

philsturgeon:
Merged some fixes from @jdesrosiers, feedback from @karenetheridge, and pushes some commits that were just laying around on my harddrive waiting to get lost. Will keep trucking along with the oas examples from @MikeRalphson.

If anyone else has suggestions please use the Suggestion feature wherever possible. Makes life a lot easier.

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,810 @@
...
+  - required:
+      - webhooks
+allOf:
+  - $ref: '#/$defs/Mixins/$defs/Extensible'
+properties:
+  openapi:
+    type: string
+    pattern: ^3\.1\.\d(-.+)?$
+  jsonSchemaDialect:
+    type: string
```
MikeRalphson, line 21:
```suggestion
    type: string
    format: uri
```

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,810 @@
...
+            items:
+              oneOf:
+                - $ref: '#Parameter'
+                - $ref: '#Reference'
+            uniqueItems: true
+      Operation:
+        $anchor: Operation
+        type: object
+        required:
+          - responses
```
MikeRalphson, line 382:
```suggestion
```

Operation.responses are no longer required.

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,810 @@
...
+              - properties:
+                  type:
+                    const: bearer
+                  bearerFormat:
+                    type: string
+          MutalTLS:
+            $anchor: MutualTLS
+            properties:
+              type:
+                const: mutalTLS
```
MikeRalphson, line 605:
```suggestion
                const: mutualTLS
```

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,810 @@
...
+            allOf:
+              - $ref: '#/$defs/Mixins/$defs/Extensible'
+            unevaluatedProperties: false
+      Schema:
+        $anchor: Schema
+        oneOf:
+          - $ref: '#Reference'
+          - type: object
+            allOf:
+              - $ref: 'https://json-schema.org/draft/2020-12/schema'
```
MikeRalphson, line 726: Not saying this is definitely a problem, but we would need to make clear that this metaschema only supports the use of the default vocabulary. If this is overridden using `jsonSchemaDialect` or `$schema` within a `schemaObject` - this metaschema may fail to validate the OAS document instance. It is probably preferable to have a "strict" metaschema. If people need a more lax version, it would be trivial to create it from this metaschema.

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,810 @@
...
+        allOf:
+          - $ref: '#/$defs/Mixins/$defs/Extensible'
+        patternProperties:
+          ^\/:
+            type: object
+            oneOf:
+              - $ref: '#Reference'
+              - $ref: '#PathItem'
+        unevaluatedProperties: false
+      PathItem:
```
MikeRalphson, line 358: Does this `PathItem` schema deliberately allow additional/unevaluated properties? Should it not be constrained to `get`, `post`, `put`, `patch`, `delete`, `head`, `options`, `trace`?

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,810 @@
...
+          headers:
+            additionalProperties:
+              $ref: '#Header'
+          links:
+            additionalProperties:
+              $ref: '#Link'
+          callbacks:
+            additionalProperties:
+              $ref: '#Callback'
+        unevaluatedProperties: true
```
MikeRalphson, line 810:
```suggestion
        unevaluatedProperties: false
```

`components` is already `Extensible`, but additional `components` "containers" are not permitted.

MikeRalphson (changes requested):
Good stuff. Few minor comments.

MikeRalphson:
Saw your script for validating the metaschema. I have a set of [scripts](https://github.com/MikeRalphson/OpenAPI-Specification/tree/ci-schemas/scripts) and [GitHub workflows](https://github.com/MikeRalphson/OpenAPI-Specification/tree/ci-schemas/.github/workflows) (in approved but unmerged PR #1911) which you might find useful.

schemas/v3.1/schema.yaml
jdesrosiers: I was thinking about that too. We can use `$dynamicRef` to address this in a not-perfect-but-not-terrible way. I'll elaborate later once I've had a chance to do a POC.

schemas/v3.1/schema.yaml
jdesrosiers: I recommend leaving it here.

jdesrosiers:
I had a thought that might make this work go a little faster. We could create a branch for this work in this repo and merge what we have so far into that branch. Anyone could contribute a fix or a missing piece through PR which can be reviewed by anyone and merged quickly. A full PR to master (like this one) can be kept open in draft status so we always have visibility of the final result. This is basically the process I was using with @philsturgeon's branch, but this way Phil doesn't have to feel like a bottleneck and more people will have visibility on the incremental PRs and agency to contribute.

jdesrosiers:
I've got another update ready, but I'll hold off until we either have a branch on this repo I can PR to or someone tells me we're not going to do that, in which case I'll make another PR on Phil's branch.

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,809 @@
...
+                type: object
+                additionalProperties:
+                  $ref: '#Header'
+              content:
+                type: object
+                additionalProperties:
+                  $ref: '#MediaType'
+              links:
+                type: object
+                additionalProperteis:
```
hkosova, line 292:
```suggestion
                additionalProperties:
```

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,809 @@
...
+          parameters:
+            type: array
+            items:
+              $ref: '#Parameter'
+          requestBody:
+            $ref: '#RequestBody'
+          responses:
+            $ref: '#Responses'
+          callback:
+            $ref: '#Callback'
```
hkosova, line 406:
```suggestion
          callbacks:
            type: object
            additionalProperties:
              $ref: '#Callback'
```
See [`callbacks`](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.1.0.md#user-content-operationcallbacks)

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,809 @@
...
+      Responses:
+        $anchor: Responses
+        type: object
+        allOf:
+          - $ref: '#/$defs/Mixins/$defs/Extensible'
+        properties:
+          default:
+            $ref: '#Response'
+        patternProperties:
+          '[1-5](?:\d{2}|XX)':
```
hkosova, line 425: The regex needs to be anchored, like in the [3.0.x schema](https://github.com/OAI/OpenAPI-Specification/blob/master/schemas/v3.0/schema.yaml#L574)
```suggestion
          '^[1-5](?:\d{2}|XX)$':
```

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,809 @@
...
+          - $ref: '#Reference'
+          - type: object
+            required:
+              - type
+            allOf:
+              - $ref: '#/$defs/Mixins/$defs/Describable'
+              - $ref: '#/$defs/Mixins/$defs/Extensible'
+            properties:
+              type:
+                string: null
```
hkosova, line 559: Is this maybe suposed to be
```suggestion
                type: string
```
?

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,809 @@
...
+                  - bearer
+                  - mutualTLS
+                  - oauth2
+                  - openIdConnect
+                oneOf:
+                  - $ref: '#APIKey'
+                  - $ref: '#HTTP'
+                  - $ref: '#MutualTLS'
+                  - $ref: '#OAuth2'
+                  - $ref: '#OpenIdConnect'
```
hkosova, line 572: This part looks over indented
```suggestion
            oneOf:
              - $ref: '#APIKey'
              - $ref: '#HTTP'
              - $ref: '#MutualTLS'
              - $ref: '#OAuth2'
              - $ref: '#OpenIdConnect'
```

schemas/v3.1/schema.yaml
karenetheridge: I think there was a github issue raised recently asking why the spec should be constrained to the current list of HTTP methods, rather than being open to new ones being added in the future. I would lean towards being open -- the `examples` keyword could be used to list the common/expected values as a hint to humans.

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,809 @@
...
+  - openapi
+  - info
+anyOf:
+  - required:
+      - components
+  - required:
+      - paths
+  - required:
+      - webhooks
+allOf:
```
karenetheridge, line 14: This allOf is not needed - the $ref can just go at the top level.

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,809 @@
...
+  - $ref: '#/$defs/Mixins/$defs/Extensible'
```
karenetheridge, line 15:
```suggestion
$ref: '#/$defs/Mixins/$defs/Extensible'
```

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,809 @@
...
+properties:
+  openapi:
+    type: string
+    pattern: ^3\.1\.\d(-.+)?$
```
karenetheridge, line 19:
```suggestion
    pattern: ^3\.1\.\d+(-.+)?$
```

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,809 @@
...
+        oneOf:
+          - $ref: '#Reference'
+          - type: object
+            allOf:
+              - $ref: '#/$defs/Mixins/$defs/Extensible'
+            unevaluatedProperties: false
+      Schema:
+        $anchor: Schema
+        oneOf:
+          - $ref: '#Reference'
```
karenetheridge, line 722:
```suggestion
```

This reference is redundant, as a draft2020-12 schema will of course accept the $ref keyword itself.

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,809 @@
...
+          - type: object
```
karenetheridge, line 723: Is it intentional that boolean schemas are excluded at this level?

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,809 @@
...
+      Components:
+        $anchor: Components
+        type: object
+        allOf:
+          - $ref: '#/$defs/Mixins/$defs/Extensible'
+          - additionalProperties:
+              $comment: All sub-objects have the same property name constraints
+              type: object
+              propertyNames:
+                pattern: '^[a-zA-Z0-9\.\-_]+$'
```
karenetheridge, line 780:
```suggestion
              $comment: All sub-objects have the same property name constraints
              type: object
              propertyNames:
                pattern: '^[a-zA-Z0-9\.\-_]+$'
              additionalProperties:
                $dynamicAnchor: ComponentsAdditionalProperties
                propertyNames:
                  pattern: '^[a-zA-Z0-9\.\-_]+$'
                $dynamicRef: '#ComponentsAdditionalProperties'
```

This implements the $comment above

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,809 @@
...
+    $defs:
+      Reference:
+        $anchor: Reference
+        type: object
+        required:
+          - $ref
+        properties:
+          $ref:
+            type: string
+            format: uri-reference
```
karenetheridge, line 187:
```suggestion
          $ref:
            type: string
            format: uri-reference
          summary:
            type: string
        $ref: '#/$defs/Mixins/$defs/Describable'
        unevaluatedProperties: false
```

as per https://spec.openapis.org/oas/v3.1.0.html#referenceObject - reference objects also allow the 'summary' and 'description' properties, and do not permit any additional properties.

schemas/v3.1/schema.yaml
karenetheridge: (this is no longer redundant if the '#Reference' definition is extended to include 'summary' and 'description' properties and "unevaluated: false".)

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,809 @@
...
+                    format: url
+                  prefix:
+                    type: string
+                  attribute:
+                    type: boolean
+                    default: false
+                  wrapped:
+                    type: boolean
+                    default: false
+            additionalProperties: false
```
hkosova, line 770:
```suggestion
            unevaluatedProperties: false
```

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,809 @@
...
+                type: string
+            oneOf:
+              - properties:
+                  type:
+                    const: http
+              - properties:
+                  type:
+                    const: bearer
+                  bearerFormat:
+                    type: string
```
hkosova, line 599: This part doesn't look right to me. The logic is supposed to be
> if `scheme` (not `type`!) = `"bearer"`
> then `bearerFormat` may be present

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.1.0.md#fixed-fields-23

Something like:

```suggestion
            properties:
              type:
                const: http
              scheme:
                type: string
            if:
              properties:
                scheme:
                  const: bearer
            then:
              properties:
                bearerFormat:
                  type: string
```

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,809 @@
...
+              $ref: '#Server'
+          parameters:
+            type: array
+            items:
+              oneOf:
+                - $ref: '#Parameter'
+                - $ref: '#Reference'
+            uniqueItems: true
+      Operation:
+        $anchor: Operation
```
karenetheridge, line 380: This definition is not referenced by anything. It used to be referenced by the PathItem definition, like so:

```
    patternProperties:                                                                              
      '^(get|put|post|delete|options|head|patch|trace)$': $ref: '#/definitions/Operation'           
```

Was this removal intentional? If so, this entire definition can be removed as it is unused.

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,809 @@
...
+              type: object
+              propertyNames:
+                pattern: '^[a-zA-Z0-9\.\-_]+$'
+        properties:
+          schemas:
+            additionalProperties:
+              $ref: '#Schema'
+          responses:
+            additionalProperties:
+              $ref: '#Responses'
```
hkosova, line 787:
```suggestion
              $ref: '#Response'
```

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,809 @@
...
+          summary:
+            type: string
+          externalDocs:
+            $ref: '#ExternalDocumentation'
+          operationId:
+            type: string
+          parameters:
+            type: array
+            items:
+              $ref: '#Parameter'
```
hkosova, line 400:
```suggestion
              $ref: '#Parameter'
            uniqueItems: true
```
As per:
> The list MUST NOT include duplicated parameters.

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,809 @@
...
+    items:
+      $ref: '#Server'
+  security:
+    type: array
+    items:
+      $ref: '#SecurityRequirement'
+  tags:
+    type: array
+    items:
+      $ref: '#Tag'
```
hkosova, line 38:
```suggestion
      $ref: '#Tag'
    uniqueItems: true
```
As per:
> Each tag name in the list MUST be unique.

and to match the [3.0 schema](https://github.com/OAI/OpenAPI-Specification/blob/master/schemas/v3.0/schema.yaml#L29).

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,809 @@
...
+        $defs:
+          Extensions:
+            $anchor: SchemaExtensions
+            properties:
+              discriminator:
+                $ref: '#Discriminator'
+              externalDocs:
+                $ref: '#ExternalDocumentation'
+              xml:
+                $ref: '#XML'
```
hkosova, line 738: Schema Object extensions are missing [`example`](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.1.0.md#user-content-schemaexample):
```suggestion
                $ref: '#XML'
              example:
                deprecated: true
```

schemas/v3.1/schema.yaml
jdesrosiers: There's a simpler solution. I'll show it in another suggestion comment.

jdesrosiers, line 780:
```suggestion
        $ref: '#/$defs/Mixins/$defs/Extensible'
        patternProperties:
          '':
            $comment: All sub-objects have the same property name constraints
            propertyNames:
              pattern: '^[a-zA-Z0-9\.\-_]+$'
```

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,809 @@
...
+          servers:
+            type: array
+            items:
+              $ref: '#Server'
+          parameters:
+            type: array
+            items:
+              oneOf:
+                - $ref: '#Parameter'
+                - $ref: '#Reference'
```
hkosova, line 377:
```suggestion
              $ref: '#Parameter'
```
`Parameter` itself includes `Reference` as a `oneOf`.

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,809 @@
...
+      ExternalDocumentation:
+        $anchor: ExternalDocumentation
+        allOf:
+          - $ref: '#/$defs/Mixins/$defs/Extensible'
+        type: object
+        required:
+          - url
+        properties:
+          description:
+            type: string
```
hkosova, line 459: Consider replacing the `description` property with `#/$defs/Mixins/$defs/Describable` for consistency with other definitions.
```suggestion
        allOf:
          - $ref: '#/$defs/Mixins/$defs/Extensible'
          - $ref: '#/$defs/Mixins/$defs/Describable'
        type: object
        required:
          - url
        properties:
```

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,809 @@
...
+              server:
+                $ref: '#Server'
+            unevaluatedProperties: false
+      Callback:
+        $anchor: Callback
+        oneOf:
+          - $ref: '#Reference'
+          - type: object
+            allOf:
+              - $ref: '#/$defs/Mixins/$defs/Extensible'
```
hkosova, line 717: The `Callback` schema is missing the actual contents of callbacks i.e. path items. See the [3.0 schema](https://github.com/OAI/OpenAPI-Specification/blob/master/schemas/v3.0/schema.yaml#L975-L980) for the reference.
```suggestion
            allOf:
              - $ref: '#/$defs/Mixins/$defs/Extensible'
            additionalProperties:
              $ref: '#PathItem'
```

jdesrosiers:
I created a new PR at https://github.com/OAI/OpenAPI-Specification/pull/2489 so we can make progress while @philsturgeon is busy planting trees. All of the comments and suggestions that have been posted here have been incorporated into the new PR.

schemas/v3.1/schema.yaml
```diff
@@ -0,0 +1,809 @@
...
+    items:
+      $ref: '#SecurityRequirement'
+  tags:
+    type: array
+    items:
+      $ref: '#Tag'
+  paths:
+    $ref: '#Paths'
+  webhooks:
+    $comment: 'TODO: Implement meta-schema for webhooks'
```
hkosova, line 42:
```suggestion
  webhooks:
    type: object
    additionalProperties:
      $ref: '#PathItem'
```

philsturgeon:
Ha! Sorry, I just merged all the things but you beat me to it. Brilliant, take it away!

PR #2492, opened 2021-03-05
Isomorphic Request Body Links

luckybroman5:
Feedback from TSC Meeting:

- [x] Rename "Fragments", suggestions: `Mappings`, `Expressions`, etc..
- [ ]  **Reusability**, or the ability to add into `Components`
- [ ] Initial proposal is somewhat too complex. Its usage should be simpler.
- [ ] Representing Form Data
- [ ] More thought Around the `OperationRef` Portions.
- [ ] Ability to ensure tooling can "Validate" these links against Schemas in both Requests and Responses. This might not be feasible with runtime expressions


luckybroman5:
Update Regarding Progress here:

> I am still working on a better pass before asking for another review. I'm in the process of developing some "test" tooling to see what makes the most sense, while also going for what is the most concise.

MikeRalphson:
As a result of #2604 the filename for this proposal should change to use a dated prefix. Please could you update this PR to match? You can use the original PR date. Apologies for the extra work.

lornajane:
This pull request has been open for a long time without any activity. It isn't ready to discuss as-is, so it will be closed. As always, proposals are welcome, so if this still seems relevant, let's re-start the discussion and look at how recent changes such as the Workflows SIG may change some of the ideas.

PR #2502, opened 2021-03-16
Delete 2.0.md

nnn227:


MikeRalphson:
We do not delete older versions of the specification.

PR #2526, opened 2021-04-07
servers is an array of server

SamiAkkawi:


jdesrosiers:
@SamiAkkawi Thanks for this! These schemas went through almost no testing at all and are probably full of bugs like these. Keep these PRs comming, but have a read through the README, especially this part 

> To submit improvements to the schema, modify the schema.yaml file only.
> - https://github.com/OAI/OpenAPI-Specification/tree/master/schemas/v3.1#contributing

I prefer working with JSON as well, but around here it's YAML first. We make the change in the YAML file and then use the following to generate the JSON version.

```
./scripts/yaml2json/yaml2json.js schemas/v3.1/schema.yaml
```

If you have an OpenAPI document that exposed this bug, it would be great if you could add that to the tests in https://github.com/OAI/OpenAPI-Specification/tree/master/tests/v3.1 as well. If not, that's ok too.

SamiAkkawi:
@jdesrosiers  thank you for the clarifications. Changes done.

jdesrosiers (approved):

SamiAkkawi:
@jdesrosiers quick question: do we know how long to pull request need to get accepted? Or Are there any specific dates to go through all open pull requests? Or how is that managed exactly?

jdesrosiers:
@SamiAkkawi I don't know exactly. I'll ask at the TSC meeting next week.

philsturgeon (approved):

philsturgeon:
@MikeRalphson putting this on your radar.

darrelmiller (approved):

PR #2529, opened 2021-04-09
Update Proposal 003 "Clarify Nullable" to Promoted status

tedepstein:
The proposal was adopted in OpenAPI version 3.0.3. This pull request updates the proposal with that status change.


philsturgeon (approved):

proposals/003_Clarify-Nullable.md
```diff
@@ -8,7 +8,7 @@
 |Proposal |[003](https://github.com/OAI/OpenAPI-Specification/tree/master/proposals/003_Clarify-Nullable.md)|
 |Authors|[Ted Epstein](https://github.com/tedepstein)|
 |Review Manager |TBD|
-|Status |Proposal|
```
MikeRalphson, line 11: According to the template, the statuses are:

> Proposal, Draft, Promoted, or Abandoned

MikeRalphson (dismissed):

proposals/003_Clarify-Nullable.md
tedepstein: @MikeRalphson , thanks. I've made the change.

MikeRalphson:
As a result of #2604 the prefix/filename for this proposal has changed. The filename is now `2019-10-31-Clarify-Nullable.md`. Please could you update this PR to match? Apologies for the extra work.

tedepstein:
Hi @MikeRalphson, I merged in the `main` branch it looks like that updated the filename. Please take a look and see if this is ready to merge.

karenetheridge:
> A true value adds "null" to the allowed type specified by the type keyword, only if type is explicitly defined within the same Schema Object. Other Schema Object constraints retain their defined behavior, and therefore may disallow the use of null as a value. A false value leaves the specified or default type unmodified. The default value is false.

Is this proposal still valid? As of OpenAPI 3.1.0, JSON Schema is fully integrated, and the use of a 'nullable' keyword to modify the behaviour of another keyword ('type') would violate the JSON Schema specification (OpenAPI can add new keywords in its vocabulary, but keywords already defined in other vocabularies cannot be modified.)

tedepstein:
@karenetheridge , this proposal was adopted in version 3.0.3, changing the language used to describe `nullable`. It clarifies the meaning of `nullable` in OpenAPI 3.0 and earlier. There is no `nullable` keyword in OpenAPI 3.1, as JSON Schema already allows for type arrays like `type: ["string", "null"]`.

karenetheridge:
thanks for the clarification. I didn't realize that this was already done for 3.0.3, as opposed to being a new proposal for 3.1.x.

darrelmiller (approved):

Roberto-sour (changes requested):
Como puedo quitar la aplicación de macropay app novo

PR #2532, opened 2021-04-11
Security scheme deprecated

### LINKED ISSUE #2506: Deprecating Security Schemes
Case study: GitHub. They support access token through the query string but they're deprecating and removing support for that. They'd like folks to stop using it, and are recommending people move over to other security schemes like HTTP basic or OAuth2. They've been sending out emails when people hit it, writing it in human-docs, and blogs, etc., but OpenAPI could help them out here too.

So far, no way to deprecate a security scheme.

image

We should be able to use `deprecated: true` on a security scheme just like on an operation or property.

Yes? I'll make a PR if some thumbs appear.

philsturgeon:
Fixes #2506

This probably cannot go in v3.1.0 but I'm just trying to start the process somehow. Maybe we need a v3.2.0-dev branch?


webron:
@philsturgeon the new branch has been created with https://github.com/OAI/OpenAPI-Specification/blob/v3.2.0-dev/versions/3.2.0.md as a starting point. Thanks!

philsturgeon:
@webron done!

darrelmiller (approved):

karenetheridge:
Please update the JSON Schemas as well (schemas/v3.1/schema.json and .yaml)

PR #2533, opened 2021-04-12
Make root.servers an array with Server Object items

ibolmo:
I noticed this differed from the spec: https://spec.openapis.org/oas/v3.1.0#openapi-object


ibolmo:
Duplicate of: https://github.com/OAI/OpenAPI-Specification/pull/2526

PR #2535, opened 2021-04-13
fix json example schema

djy0:


MikeRalphson:
@djy0 is this the same change as the other PR, only against the 3.0.2 version? If so, we don't update published versions of the spec. so this PR can be closed.

PR #2536, opened 2021-04-13
fix example in json

djy0:
example should be same level with schema


MikeRalphson:
We now have a `v3.1.1-dev` branch containing a `versions/3.1.1.md` file - please could you retarget your changes? Thank you.

djy0:
thanks, see #2547

PR #2543, opened 2021-04-16
Fixes Inconsistent JSON spacing

ArishSultan:
the whole document uses 2-space indentation but at some places, 4-space indentation is used which is not consistent with the others.

### DIFF
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -117,7 +117,7 @@ For example, if a field has an array value, the JSON array representation will b
 
 ```json
 {
-   "field": [ 1, 2, 3 ]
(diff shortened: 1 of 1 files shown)


ArishSultan:
@MikeRalphson

ArishSultan:
just a ping, anyone here?

MikeRalphson (changes requested):
Thanks, LGTM. There appears to be at least one case of a `json` code block being indented by 3 spaces (the first example for array formatting). Could you also fix that one?

MikeRalphson (approved):
LGTM now, thanks.

webron:
Thanks for the update!

PR #2547, opened 2021-04-19
fix example in json

djy0:
example should be same level with schema

### DIFF
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -1794,7 +1794,7 @@ Plain text response with headers:
     "text/plain": {
       "schema": {
         "type": "string"
-      }
+      },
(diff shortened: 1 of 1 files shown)


versions/3.1.1.md
```diff
@@ -1793,9 +1793,9 @@ Plain text response with headers:
   "content": {
     "text/plain": {
       "schema": {
-        "type": "string",
-        "example": "whoa!"
+        "type": "string"
       }
```
MikeRalphson, line 1797:
```suggestion
      },
```

KenG-pro (approved):

MikeRalphson (approved):

webron:
Thanks for the fix!

PR #2554, opened 2021-04-26
Update TOB.md

ncaidin:
The election cycle for 2021 is complete and the existing TSC-elected members have agreed to serve another term. No additional nominations were received.


MikeRalphson (approved):
Thanks @ncaidin

PR #2576, opened 2021-05-17
Proposal for Server Variable types

erunion:
To allow for the viability of tooling to validate server variables before composing a URL I'm proposing that a `type` property be added onto the [Server Variable Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#serverVariableObject) schema much in the same way parameters and schemas require one.


MikeRalphson:
As a result of #2604 the filename for this proposal should change to use a dated prefix. Please could you update this PR to match? You can use the original PR date. Apologies for the extra work.

erunion:
@MikeRalphson 👍  All good now

darrelmiller:
@erunion Thanks for submitting this.  Apologies for the long delay in responding.  Would you be available to present on this proposal in one of our Thursday TDC calls?

darrelmiller:
@erunion We are closing this for the moment as it has been inactive for a while.  Please feel free to re-open if you want to follow up on this.

PR #2582, opened 2021-05-19
Initial proposal for New Security Schemes

whitlockjc:
This purpose of this proposal is to rethink how [Security Schemes](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#security-scheme-object) are defined in OpenAPI.  This proposal would separate the credential location(s) from the Security Scheme _type_ and would allow for complete control over dictating where the credential(s) are provided for a Security Scheme.  This proposal will add a generic way to retrieve a credential from anywhere in a request, and allow any Security Scheme to dictate where its supported credential(s) are instead of relying on OpenAPI-supported conventions/opinions.  This proposal also separates the credential location(s) details from the other configurations associated with the Security Scheme _(like JWKS details for JWT)_.

### DIFF
--- a/proposals/2021-05-19-New-Security-Definitions.md
+++ b/proposals/2021-05-19-New-Security-Definitions.md
@@ -1,0 +1,529 @@
+# New Security Definitions
+
+## Metadata
+
+| Tag | Value |
+| ---- | ---------------- |
+| Proposal | [2021-05-19-New-Security-Definitions](https://github.com/OAI/OpenAPI-Specification/tree/main/proposals/2021-05-19-New-Security-Definitions.md) |
+| Authors | [Jeremy Whitlock](https://github.com/whitlockjc) |
+| Review Manager | TBD |
+| Status | Proposal |
+| Implementations | N/A |
+| Issues | [{issueid}](https://github.com/OAI/OpenAPI-Specification/issues/{Issueid}) |
+| Previous Revisions | N/A |
+
+## Change Log
+
+| Date | Responsible Party | Description |
+| 2021/05/13 | Jeremy Whitlock | Initial proposal |
+
+## Introduction
+
+This purpose of this proposal is to rethink how
+[Security Schemes](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#security-scheme-object) are
+defined in OpenAPI.  This proposal would separate the credential location(s) from the Security Scheme _type_ and would
+allow for complete control over dictating where the credential(s) are provided for a Security Scheme.  This proposal
+will add a generic way to retrieve a credential from anywhere in a request, and allow any Security Scheme to dictate
+where its supported credential(s) are instead of relying on OpenAPI-supported conventions/opinions.  This proposal also
+separates the credential location(s) details from the other configurations associated with the Security Scheme
+_(like JWKS details for JWT)_.
+
+**Note:** The proposal isn't to remove `type` from the Security Scheme object but to allow for more control over where
+the credential(s) are provided for a Security Scheme.  This proposal would still have reasonable defaults for the
+already-supported `type` values and will be documented as such.
+
+## Motivation
+
+[Security Schemes](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#security-scheme-object)
+_(formerly known as "Security Definitions")_ provide a way to describe the security scheme(s) used by your API.
+Unfortunately, the _types_ supported by OpenAPI are not only limited but they often tie the credential location to the
+Security Scheme type.  This proposal will add a generic way to define where a credential is and its format so that the
+consumer can have complete control over where credentials are used in their API.  Not only would this allow consumers to
+specify their own Security Schemes without OpenAPI having to be the gatekeeper, but it also allows a level of
+flexibility not available now.  There have also been extra Security Scheme properties that have been _tacked on_ to
+support the growing needs of consumers, which would be alleviated _(or moved in this case)_.
+
+For example, if a consumer wants to use a JSON Web Key _(JWT)_ in their API, OpenAPI will allow this but only under
+certain circumstances:
+
+* `type: apiKey`: This will allow you to provide your JWT anywhere you would like, but it could be confusing to overload
+  `type: apiKey` since you are describing a Security Scheme using a JWT and not an API Key
+* `type: http`: This will only work if your JWT is in a header and in a supported
+  [HTTP Auth](https://datatracker.ietf.org/doc/html/rfc7235#section-5.1) format
+
+OpenAPI being so opinionated about these kinds things does not serve any real benefit to anyone involved, and this
+proposal would make things much more flexible.
+
+## Proposed solution
+
+This proposal suggests the following changes to the Security Scheme object:
+
+* Add a new `config` property that is type specific to contain the scheme-specific configuration not related to
+  credential configuration
+* Add a new `credentials` property that is an array of [Credential Objects](#credential-object) that describe a
+  credential _(where it is located, what its format is, etc.)_
+* Remove the credential-specific properties _(`bearerFormat`, `flows`, `in`, `openIdConnectUrl` and `scheme`)_ and move
+  them into a type-specific `config` property
+* Update `type` to allow for any value instead of the hard-coded list of supported types
+
+## Detailed design
+
+The bulk of this proposal is defining a way to separate the properties used to describe the credential(s) for a Security
+Scheme from the root properties of the Security Scheme.  This proposal will also allow you to define multiple
+credentials _(multiple locations, formats, ...)_ for the same Security Scheme for cases where you might support
+multiple formats/locations for effectively the same credential.  The new objects being suggested as part of this
+proposal are defined below.
+
+**Note:** While not specifically mentioned below for brevity, JSON References and Extensions should be allowed within
+every complex type documented below.
+
+### Value Getter Object
+
+The _Value Getter Object_ describes a place in the request where something exists that you want to retrieve.  The name
+and purpose of this object is generic, as in I could see other places in OpenAPI where you could point to a value in the
+request.  The _Value Getter Object_ is the base object to be _extended_ by **all**
+[Credential Object](#credential-object)s.  Below is its schema:
+
+| Field Name | Type | Applies To | Description |
+| ---------- | ---- | ---------- | ----------- |
+| in | `string` | Any | **REQUIRED.** Specifies where in the request the credential is provided.  Valid values are: `body`, `cookie`, `header`, `path` and `query` |
+| format | `regex` | Any | Specifies the format of the credential.  Since in many cases, the credential is a subset of the raw value, `format` must be a regular expression that matches the complete value, and contains a single capturing group to match the actual value within the raw value. |
+| name | `string` | Any | **REQUIRED.** Specifies the name used to resolve the credential in the location provided by `in`. |
+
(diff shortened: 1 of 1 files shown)


proposals/005_New_Security_Definitions.md
```diff
@@ -0,0 +1,531 @@
...
+| format | `regex` | Any | Specifies the format of the credential.  Since in many cases, the credential is a subset of the raw value, `format` must be a regular expression that matches the complete value, and contains a single capturing group to match the actual value within the raw value. |
+| name | `string` | Any | **REQUIRED.** Specifies the name used to resolve the credential in the location provided by `in`. |
+
+**Note:** For `body`, `name` should be a JSON Pointer _(or something similar)_ that lets you define what part of the
+body you are interested in.  For example, you could use `#` to dictate that body in its entirety is the credential or
+you could use something like `#/some/path` to dictate that a portion of the body is the credential.  I understand that
+JSON Pointer usage implies that the body is JSON but that doesn't have to be the case and could be used to address any
+structured data.
+
+**Note:** For `format`, I realize that it being a regular expression might seem a bit offputting but there is no way to
```
savage-alex, line 99: off-putting

proposals/005_New_Security_Definitions.md
```diff
@@ -0,0 +1,531 @@
...
+
+**Note:** The proposal isn't to remove `type` from the Security Scheme object but to allow for more control over where
+the credential(s) are provided for a Security Scheme.  This proposal would still have reasonable defaults for the
+already-supported `type` values and will be documented as such.
+
+## Motivation
+
+[Security Schemes](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#security-scheme-object)
+_(formerly known as "Security Definitions")_ provide a way to describe the security scheme(s) used by your API.
+Unfortunately, the _types_ supported by OpenAPI are not only limited but they often tie the credential location to the
```
MikeRalphson, line 39: > Unfortunately, the _types_ supported by OpenAPI ... often tie the credential location to the Security Scheme type

Can you expand on why this is "unfortunate"? Type `apiKey` can already be located `in` any of `query`, `header`,`cookie`, and arguably if you try to put your `Bearer` token anywhere other than in an `Authorization` header, you **are not doing oAuth2**.

proposals/005_New_Security_Definitions.md
```diff
@@ -0,0 +1,531 @@
...
+
+For example, if a consumer wants to use a JSON Web Key _(JWT)_ in their API, OpenAPI will allow this but only under
+certain circumstances:
+
+* `type: apiKey`: This will allow you to provide your JWT anywhere you would like, but it could be confusing to overload
+  `type: apiKey` since you are describing a Security Scheme using a JWT and not an API Key
+* `type: http`: This will only work if your JWT is in a header and in a supported
+  [HTTP Auth](https://datatracker.ietf.org/doc/html/rfc7235#section-5.1) format
+
+OpenAPI being so opinionated about these kinds things does not serve any real benefit to anyone involved, and this
```
MikeRalphson, line 54: > OpenAPI being so opinionated about these kinds things does not serve any real benefit to anyone involved

Unsure this is totally true. Tooling authors may appreciate the limited set of possibilities they have to support?

proposals/005_New_Security_Definitions.md
```diff
@@ -0,0 +1,531 @@
...
+* Remove the credential-specific properties _(`bearerFormat`, `flows`, `in`, `openIdConnectUrl` and `scheme`)_ and move
+  them into a type-specific `config` property
+* Update `type` to allow for any value instead of the hard-coded list of supported types
+
+## Detailed design
+
+The bulk of this proposal is defining a way to separate the properties used to describe the credential(s) for a Security
+Scheme from the root properties of the Security Scheme.  This proposal will also allow you to define multiple
+credentials _(multiple locations, formats, ...)_ for the same Security Scheme for cases where you might support
+multiple formats/locations for effectively the same credential.  The new objects being suggested as part of this
```
MikeRalphson, line 74: These multiple locations, are they AND or OR?

proposals/005_New_Security_Definitions.md
```diff
@@ -0,0 +1,531 @@
...
+
+The _Value Getter Object_ describes a place in the request where something exists that you want to retrieve.  The name
+and purpose of this object is generic, as in I could see other places in OpenAPI where you could point to a value in the
+request.  The _Value Getter Object_ is the base object to be _extended_ by **all**
+[Credential Object](#credential-object)s.  Below is its schema:
+
+| Field Name | Type | Applies To | Description |
+| ---------- | ---- | ---------- | ----------- |
+| in | `string` | Any | **REQUIRED.** Specifies where in the request the credential is provided.  Valid values are: `body`, `cookie`, `header`, `path` and `query` |
+| format | `regex` | Any | Specifies the format of the credential.  Since in many cases, the credential is a subset of the raw value, `format` must be a regular expression that matches the complete value, and contains a single capturing group to match the actual value within the raw value. |
```
MikeRalphson, line 90: Rather than express how to **retrieve** the credential using a regex, it may be more OAS-like to provide a *template* for how the credential should be constructed. Thus 

> `format: ^[B|b][E|e][A|a][R|r][E|e][R|r] (.*)$`

would become something like:

> `format: Bearer {credential}`

PS: Header names are case-insensitive, but the "Bearer" string is not, so this regex can be simplified.

proposals/005_New_Security_Definitions.md
```diff
@@ -0,0 +1,531 @@
...
+structured data.
+
+**Note:** For `format`, I realize that it being a regular expression might seem a bit offputting but there is no way to
+reasonably support raw values _(as-is)_ and taking a raw value and only grabbing the important part
+_(like when using HTTP Auth where you don't want the scheme details)_.  We could also have _named formats_ that would
+alleviate the need for consumers to author regular expressions themselves for things like HTTP Auth, Bearer Tokens, etc.
+
+**Note:** I could see a case for a `description` field.
+
+**Note:** I could see a case for supporting an `in` for mTLS
```
MikeRalphson, line 106: How would the certificate be presented in a `query` parameter, `header`, `cookie` or `body`? Unless presented as part of a TLS handshake, you're not doing mutual TLS.

MikeRalphson (changes requested):
Basically I feel this proposal is attempting to do several things (and gets caught up trying to be backwards compatible as well):

* allow (some) credentials in `body` and `path`
* allow JWTs with some `type`-specific config
* genericise credential locations and config

It's not clear to me why all these things are or need to be combined.

orubel:
Bringing it up again folks. Since when a token is supplied during a client request (note I said, the token is supplied so this is AFTER the OAUTH handshake), the ROLE can be checked but it is NOT checked vs the request data being sent or the response data being returned that is defined in the doc.

As an ADMIN ROLE can request the same endpoint entirely differently than a USER ROLE and get an entirely different response dataset, all prior itterations as well as the current iteration of the OpenAPI spec are in direct violation of OWASP API3:2019 (https://apisecurity.io/encyclopedia/content/owasp/api3-excessive-data-exposure.htm)

without knowing what the roles are, this can be alleviated simply by allowing a STRING to be the KEY association for request/response datasets like so:
```
"REQUEST": {			            
    "permitAll":["username","password","email"],                                   
     "ROLE_ADMIN":["id"]			       
 },			        
"RESPONSE": {			        	
    "permitAll":["id","version"]			       
}
```
permitAll defines all roles defined in ROLE_GROUP that have access to this endpoint (the catchAll ) whereas Role_ADMIN is a specific RULE for ROLE_ADMIN allowing a concatenation onto the 'permitAll' data if your ROLE==ROLE_ADMIN

This document establishes these RULES for the request response so that ANYWHERE in the call flow, a device/service can now use this document to check these rules.

orubel:
and BTW, you can separate what you need for an intial handshake but you can't separate what you need for basic RBAC association with endpoints (as pointed out above). Every service would need to do this at the same time... not separately or else error can occur.

Example: gateway checks your role and lets you in, but role is not associated with endpoint so it give generic endpoint call (which does not differentiate between USER/ADMIN). This could cause client to send too much data to server and server to return too much data to client per **OWASP API3:2019**

MikeRalphson:

proposals/005_New_Security_Definitions.md
whitlockjc: The point was more about separating the credential location(s) from the Security Scheme type.  API Key is the closest to allowing free-form credential locations while others do not.  I can see where some Security Schemes where these are locked down, like OAuth 2.0 _(while it uses Bearer Tokens, I don't think it supports them being in all locations a Bearer Token could be represented)_, but this separation is really a flexibility thing than just trying to support JWT.

proposals/005_New_Security_Definitions.md
whitlockjc: It's subjective for sure but is focusing on the tooling authors more important than the consumers of OpenAPI?  I think with a clean implementation, it shouldn't be too hard to support something like this as a fellow tooling author to another.

proposals/005_New_Security_Definitions.md
whitlockjc: I would suggest we do something like how the Security Requirements work now, which would allow for an AND or an OR.  But in the case of having an AND of the same credential location, I can't think of a case where you'd say I want the same to credential to be supplied in multiple places at the same time.

proposals/005_New_Security_Definitions.md
whitlockjc: I don't disagree.  The main reason for using a regex example was to show how one might pick out a portion of a larger HTTP value, I'm not married to any specific implementation.

proposals/005_New_Security_Definitions.md
whitlockjc: I get what you mean.  mTLS is done at the protocol level so actually providing the certificate as part of the request wouldn't be useful.  I'll remove this.

proposals/005_New_Security_Definitions.md
whitlockjc: It would also be nice to have a way to describe a type not supported by OAS, where our locked-down types don't fit and a customer wants to still describe their security needs.

whitlockjc:
A few notes after presenting this to the Security SIG:

* I could see a case where the `Credential Object` might also allow for a `credential` property in place of a `credentials` property.
* I could see a case where OAS might have some pre-defined credential _(or credential formats, like HTTP Auth)_
* There may be a case where we might need to allow for more JWT claims described for the `jwt` Security Scheme _(I know on the backend this could be useful but seeing as this isn't the intent, I am not sure there is a case for this but it came up in the meeting so we'll wait for examples.)_

proposals/2021-05-19-New-Security-Definitions.md
```diff
@@ -0,0 +1,529 @@
...
+```yaml
+# ...
+components:
+  securitySchemes:
+    apiKey:
+      type: apiKey
+      credentials:
+      - in: header
+        name: authorization
+        format: ^[B|b][E|e][A|a][R|r][E|e][R|r] (.*)$
```
ioggstream, line 199: if it's a regexp, I'd use `pattern`.

ioggstream (commented):
Added some preliminary notes.

whitlockjc:
I've dropped the ball on this, but since it's never been reasoned about, what should we do here?

handrews:
Filed #3257 to revive and clarify our proposals process.  Once we figure it out, we'll resolve this PR accordingly.

darrelmiller:
Closing because the conversation is now happening in Moonwalk discussions https://github.com/OAI/sig-moonwalk/discussions/46

PR #2593, opened 2021-05-26
Update 3.1.0.md

aowss:
Small English correction


MikeRalphson:
Thanks for the PR! The content looks good, but as per [DEVELOPMENT.md](https://github.com/OAI/OpenAPI-Specification/blob/main/DEVELOPMENT.md) could you please retarget or resubmit against the `versions/3.1.1.md` file in the `v3.1.1-dev` branch please, as we don't amend published versions of the spec?

aowss:
I see that it has already been fixed there.
Sorry for the confusion on which branch to use.
Thanks for your prompt feedback.

PR #2598, opened 2021-06-01
Allow booleans in place of any top-level schema field

MikeRalphson:
It has been pointed out that we didn't achieve 100% JSON Schema compatibility with the release of OAS v3.1.0. JSON Schema allows spelling the schema `{}` as `true` and the schema `not: {}` as `false`. 

These are currently allowed in subschemas, but not where our `schema` field indicates an OpenAPI `schemaObject` (literally of type `object` as expected by the metaschema for example) is allowed.

This may only be of academic interest, as it is possible to spell the empty schema either as `{}` as above, or by omitting the `schema` keyword entirely (where allowed), and the `false` / `not: {}` schema has limited utility at the top-level in OAS - but for the sake of spec. and tooling compatibility, here is this PR.

It allows the `schema` fields to contain `boolean`s in the same way we used to allow `schemaObject | referenceObject` in v3.0.x

This PR does not touch the metaschema as it is still somewhat in flux. I propose we pick up metaschema changes in the RC / post-release phase as with previous versions.

**Edit**: an alternative is to add wording similar to that we had in `v3.0.x`

> Alternatively, anywhere a Schema Object can be used, a ~~[Reference Object](#referenceObject)~~ `boolean` MAY be used in its place.

Signed-off-by: Mike Ralphson <mike.ralphson@gmail.com>

### DIFF
--- a/versions/3.2.0.md
+++ b/versions/3.2.0.md
@@ -452,7 +452,7 @@ All objects defined within the components object will have no effect on the API 
 
 Field Name | Type | Description
 ---|:---|---
-<a name="componentsSchemas"></a> schemas | Map[`string`, [Schema Object](#schemaObject)] | An object to hold reusable [Schema Objects](#schemaObject).
+<a name="componentsSchemas"></a> schemas | Map[`string`, [Schema Object](#schemaObject) \| `boolean`] | An object to hold reusable [Schema Objects](#schemaObject).
 <a name="componentsResponses"></a> responses | Map[`string`, [Response Object](#responseObject) \| [Reference Object](#referenceObject)] | An object to hold reusable [Response Objects](#responseObject).
 <a name="componentsParameters"></a> parameters | Map[`string`, [Parameter Object](#parameterObject) \| [Reference Object](#referenceObject)] | An object to hold reusable [Parameter Objects](#parameterObject).
 <a name="componentsExamples"></a> examples | Map[`string`, [Example Object](#exampleObject) \| [Reference Object](#referenceObject)] | An object to hold reusable [Example Objects](#exampleObject).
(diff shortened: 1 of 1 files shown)


karenetheridge:
Wouldn't this require updates to the metaschemas as well?  specifically

```
$defs:
  components:
    type: object
    properties:
      schemas:
        type: object
        additionalProperties:
          $dynamicRef: '#meta'
```

should change `type: object` to `type: [ object, boolean ]`

karenetheridge:
for those following along by email, I have added an edit above

MikeRalphson:
> Wouldn't this require updates to the metaschemas as well?

@karenetheridge I explicitly said...

> This PR does not touch the metaschema as it is still somewhat in flux. I propose we pick up metaschema changes in the RC / post-release phase as with previous versions.

... but thank you for pointing out that the v3.1 metaschemas may be out of step with the markdown source-of-truth. Ping @jdesrosiers

char0n:
Great, thanks for clarification! Would it be possible to explicitly mention in 3.1 that [Boolean JSON Schemas](https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.4.3.2) are not supported as top level Schema Objects? It's kind of implied because the name of the spec object is Schema Object but, when implementing tooling it gives space for interpretation and I was actually inclined to support Boolean JSON Schemas in 3.1 tooling (as top level Schema Objects) but eventually decided not to go for it.

MikeRalphson:
> I was actually inclined to support Boolean JSON Schemas in 3.1 tooling (as top level Schema Objects) but eventually decided not to go for it.

Thanks also, @char0n for the data-point.

jdesrosiers:
Instead of adding "| `boolean`" everywhere, I'd suggest just redefining "Schema Object" to include boolean. It allows the spec to only be changed in one place and is more correct (boolean schemas are a JSON Schema concept, not an OpenAPI concept).

I agree with @char0n that it's not explicitly stated in the spec that this value must be an object ("Schema Object" is a name, it's not specifying anything). Therefore, I think it's reasonable to fix this in the 3.1.1 patch rather than 3.2.0. I think it's clarifying an ambiguity, not changing a specified behavior.

MikeRalphson:
Will replace with alternative PR on the `v3.1.1-dev` branch as discussed.

PR #2609, opened 2021-06-04
Allow booleans in place of any top-level schema field

MikeRalphson:
Replacement PR for #2598 

This targets `v3.1.1` as we have agreed this is a clarification. The change spells out that `boolean` values MAY be used wherever a `schemaObject` is allowed.

A separate PR (#2614) updates the `tests` folder with passing and failing cases for this issue.

 Plus one bonus fix.

### DIFF
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -2306,7 +2306,7 @@ $ref: definitions.yaml#/Pet
 #### <a name="schemaObject"></a>Schema Object
 
 The Schema Object allows the definition of input and output data types.
-These types can be objects, but also primitives and arrays. This object is a superset of the [JSON Schema Specification Draft 2020-12](https://tools.ietf.org/html/draft-bhutton-json-schema-00).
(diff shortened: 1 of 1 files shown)


jdesrosiers (approved):

magicmatatjahu:
There can be also added some example with `true/false` schemas like in https://github.com/asyncapi/spec/pull/550:

```json
{
  "type": "object",
  "required": [
    "anySchema"
  ],
  "properties": {
    "anySchema": true,
    "cannotBeDefined": false
  }
}
```

```yaml
type: object
required:
- anySchema
properties:
  anySchema: true
  cannotBeDefined: false
```

MikeRalphson:
@magicmatatjahu rather than add examples to the spec, I have updated the `tests` folder with passing and failing examples. Still open to adding spec examples if people feel that is warranted.

magicmatatjahu:
Good to know, let's wait for people's opinion. Feel free to use these examples anywhere you want.

darrelmiller (approved):

PR #2618, opened 2021-06-12
3.1.1 discriminator improvements

jdesrosiers:
Here are some updates for the `discriminator` keyword specification. Mostly, this update is to make it clear that `discriminator` shouldn't change the validation behavior of the `oneOf`/`anyOf` keywords. `discriminator` can only be used to improve efficiency or messaging of evaluating `oneOf`/`anyOf` I also fixed bugs and clarified some confusing parts.

- The language in the table introducing `discriminator` seems to describe the value of the `discriminator` keyword to be an "object name" (whatever that means) rather than an object. I tweaked the wording to avoid that contradiction.

- The `discriminator` keyword uses the concept of "schema names", but what that means is never defined. There is no concept of a schema having a name in JSON Schema. They only have URIs. I defined "schema name" and added some clarifications where the use of schema names got confusing.

- The `discriminator` keyword can't be a shortcut, only a hint. I tried to tweak any language that might suggest that `discriminator` overrides the behavior of `anyOf`/`oneOf` in any way.

- The last example was super confusing, which is mostly because it doesn't indicate which schema the payload examples are being applied to. So, I added the schema and an explanation of what is happening.

### DIFF
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -2702,10 +2702,6 @@ schema at `#/components/schemas/Cat` is "Cat". Therefore, when using
 `discriminator`, _inline_ schemas will not be considered because they don't have
 a "name".
 
-Schema names are scoped to the OpenAPI document they are defined in. That means
-if OpenAPI document "A" references a schema in OpenAPI document "B", the schema
-names in "A" aren't considered when evaluating the referenced schema in "B".
-
 ##### Fixed Fields
 Field Name | Type | Description
 ---|:---:|---
@@ -2776,11 +2772,6 @@ For example:
 ```yaml
 components:
   schemas:
-    MyResponseType:
-      oneOf:
-      - $ref: '#/components/schemas/Cat'
-      - $ref: '#/components/schemas/Dog'
-      - $ref: '#/components/schemas/Lizard'
     Pet:
       type: object
       required:
@@ -2818,12 +2809,7 @@ components:
             type: boolean
 ```
 
-The `MyResponseType` schema will use the discriminator defined by the `Pet`
-schema because it is part of the `Cat`, `Dog`, and `Lizard` schemas in the
-`oneOf`. The behavior if not all schemas define a `discriminator` and they are
-not all the same is undefined.
-
-Validated against the `MyResponseType` schema, a payload like this:
+Validated against the `Pet` schema, a payload like this:
 
 ```json
 {
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -2772,6 +2772,11 @@ For example:
 ```yaml
 components:
   schemas:
+    MyResponseType:
+      oneOf:
+      - $ref: '#/components/schemas/Cat'
+      - $ref: '#/components/schemas/Dog'
+      - $ref: '#/components/schemas/Lizard'
     Pet:
       type: object
       required:
@@ -2809,7 +2814,12 @@ components:
             type: boolean
 ```
 
-a payload like this:
+The `MyResponseType` schema will use the discriminator defined by the `Pet`
+schema because it is part of the `Cat`, `Dog`, and `Lizard` schemas in the
+`oneOf`. The behavior if not all schemas define a `discriminator` and they are
+not all the same is undefined.
+
+Validated against the `MyResponseType` schema, a payload like this:
 
 ```json
 {
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -2332,7 +2332,8 @@ The OpenAPI Specification's base vocabulary is comprised of the following keywor
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="schemaDiscriminator"></a>discriminator | [Discriminator Object](#discriminatorObject) | Adds support for polymorphism. The discriminator is an object name that is used to differentiate between other schemas which may satisfy the payload description. See [Composition and Inheritance](#schemaComposition) for more details.
+<a name="schemaDiscriminator"></a>discriminator | [Discriminator
+Object](#discriminatorObject) | Adds support for polymorphism. The discriminator is used to determine which of a set of schemas a payload is expected to satisfy. See [Composition and Inheritance](#schemaComposition) for more details.
 <a name="schemaXml"></a>xml | [XML Object](#xmlObject) | This MAY be used only on properties schemas. It has no effect on root schemas. Adds additional metadata to describe the XML representation of this property.
 <a name="schemaExternalDocs"></a>externalDocs | [External Documentation Object](#externalDocumentationObject) | Additional external documentation for this schema.
 <a name="schemaExample"></a>example | Any | A free-form property to include an example of an instance for this schema. To represent examples that cannot be naturally represented in JSON or YAML, a string value can be used to contain the example with escaping where necessary.<br><br>**Deprecated:** The `example` property has been deprecated in favor of the JSON Schema `examples` keyword. Use of `example` is discouraged, and later versions of this specification may remove it.
@@ -2345,13 +2346,8 @@ The OpenAPI Specification allows combining and extending model definitions using
 `allOf` takes an array of object definitions that are validated *independently* but together compose a single object.
 
 While composition offers model extensibility, it does not imply a hierarchy between the models.
-To support polymorphism, the OpenAPI Specification adds the `discriminator` field.
-When used, the `discriminator` will be the name of the property that decides which schema definition validates the structure of the model.
-As such, the `discriminator` field MUST be a required field.
-There are two ways to define the value of a discriminator for an inheriting instance.
-- Use the schema name.
-- Override the schema name by overriding the property with a new value. If a new value exists, this takes precedence over the schema name.
-As such, inline schema definitions, which do not have a given id, *cannot* be used in polymorphism.
+To support polymorphism, the OpenAPI Specification adds the `discriminator` keyword.
+When used, the `discriminator` will indicate the name of the property that decides which schema definition validates the structure of the model.
 
 ###### XML Modeling
 
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -2332,8 +2332,7 @@ The OpenAPI Specification's base vocabulary is comprised of the following keywor
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="schemaDiscriminator"></a>discriminator | [Discriminator
-Object](#discriminatorObject) | Adds support for polymorphism. The discriminator is used to determine which of a set of schemas a payload is expected to satisfy. See [Composition and Inheritance](#schemaComposition) for more details.
+<a name="schemaDiscriminator"></a>discriminator | [Discriminator Object](#discriminatorObject) | Adds support for polymorphism. The discriminator is used to determine which of a set of schemas a payload is expected to satisfy. See [Composition and Inheritance](#schemaComposition) for more details.
 <a name="schemaXml"></a>xml | [XML Object](#xmlObject) | This MAY be used only on properties schemas. It has no effect on root schemas. Adds additional metadata to describe the XML representation of this property.
 <a name="schemaExternalDocs"></a>externalDocs | [External Documentation Object](#externalDocumentationObject) | Additional external documentation for this schema.
 <a name="schemaExample"></a>example | Any | A free-form property to include an example of an instance for this schema. To represent examples that cannot be naturally represented in JSON or YAML, a string value can be used to contain the example with escaping where necessary.<br><br>**Deprecated:** The `example` property has been deprecated in favor of the JSON Schema `examples` keyword. Use of `example` is discouraged, and later versions of this specification may remove it.
@@ -2727,7 +2726,7 @@ MyResponseType:
   - $ref: '#/components/schemas/Lizard'
 ```
 
-which means the payload _MUST_, by validation, match exactly one of the schemas described by `Cat`, `Dog`, or `Lizard`. Evaluating a `oneOf` can be a costly operation, so `discriminator` MAY be used as a "hint" to improve the efficiency of selection of the matching schema. The `discriminator` keyword can not change the validation result of the `oneOf`, it can only help make the evaluation more efficient and provide better error messaging. We can then describe exactly which field tells us which schema is expected to match the instance:
+which means the payload _MUST_, by validation, match exactly one of the schemas described by `Cat`, `Dog`, or `Lizard`. Evaluating a `oneOf` can be a costly operation, so `discriminator` MAY be used as a "hint" to improve the efficiency of selection of the matching schema. The `discriminator` keyword cannot change the validation result of the `oneOf`, it can only help make the evaluation more efficient and provide better error messaging. We can then describe exactly which field tells us which schema is expected to match the instance:
 
 ```yaml
 MyResponseType:
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -2710,7 +2710,7 @@ names in "A" aren't considered when evaluating the referenced schema in "B".
 ##### Fixed Fields
 Field Name | Type | Description
 ---|:---:|---
-<a name="propertyName"></a>propertyName | `string` | **REQUIRED**. The name of the property in the payload that will hold the discriminator value.
+<a name="propertyName"></a>propertyName | `string` | **REQUIRED**. The name of the property in the payload that will hold the discriminator value. This property MUST be required in the payload schema.
 <a name="discriminatorMapping"></a> mapping | Map[`string`, `string`] | An object to hold mappings between payload values and schema names or references.
 
 This object MAY be extended with [Specification Extensions](#specificationExtensions).
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -2696,7 +2696,11 @@ components:
 
 When request bodies or response payloads may be one of a number of different schemas, a `discriminator` object can be used to aid in serialization, deserialization, and validation.  The discriminator is a specific object in a schema which is used to inform the consumer of the document of an alternative schema based on the value associated with it.
 
-When using the discriminator, _inline_ schemas will not be considered.
+`discriminator` uses a schema's "name" to automatically map a property value to
+a schema. The schema's "name" is the property name used when declaring the
+schema as a component. For example, the name of the schema at
+`#/components/schemas/Cat` is "Cat". Therefore, when using `discriminator`,
+_inline_ schemas will not be considered because they don't have a "name".
 
 ##### Fixed Fields
 Field Name | Type | Description
@@ -2758,7 +2762,7 @@ MyResponseType:
       monster: 'https://gigantic-server.com/schemas/Monster/schema.json'
 ```
 
-Here the discriminator _value_ of `dog` will map to the schema `#/components/schemas/Dog`, rather than the default (implicit) value of `Dog`.  If the discriminator _value_ does not match an implicit or explicit mapping, no schema can be determined and validation SHOULD fail. Mapping keys MUST be string values, but tooling MAY convert response values to strings for comparison.
+Here the discriminator _value_ of `dog` will map to the schema `#/components/schemas/Dog`, rather than the default (implicit) `#/components/schemas/dog`.  If the discriminator _value_ does not match an implicit or explicit mapping, no schema can be determined and validation SHOULD fail. Mapping keys MUST be string values, but tooling MAY convert response values to strings for comparison.
 
 When used in conjunction with the `anyOf` construct, the use of the discriminator can avoid ambiguity where multiple schemas may satisfy a single payload.
 
@@ -2815,7 +2819,7 @@ a payload like this:
 }
 ```
 
-will indicate that the `Cat` schema be used.  Likewise this schema:
+will indicate that the `#/components/schemas/Cat` schema be used.  Likewise this payload:
 
 ```json
 {
@@ -2824,7 +2828,7 @@ will indicate that the `Cat` schema be used.  Likewise this schema:
 }
 ```
 
-will map to `Dog` because of the definition in the `mapping` element.
+will map to `#/components/schemas/Dog` because the `dog` entry in the `mapping` element maps to `Dog` which is the schema name for `#/components/schemas/Dog`.
 
 
 #### <a name="xmlObject"></a>XML Object
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -2698,9 +2698,10 @@ When request bodies or response payloads may be one of a number of different sch
 
 `discriminator` uses a schema's "name" to automatically map a property value to
 a schema. The schema's "name" is the property name used when declaring the
-schema as a component. For example, the name of the schema at
-`#/components/schemas/Cat` is "Cat". Therefore, when using `discriminator`,
-_inline_ schemas will not be considered because they don't have a "name".
+schema as a component in an OpenAPI document. For example, the name of the
+schema at `#/components/schemas/Cat` is "Cat". Therefore, when using
+`discriminator`, _inline_ schemas will not be considered because they don't have
+a "name".
 
 ##### Fixed Fields
 Field Name | Type | Description
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -2703,6 +2703,10 @@ schema at `#/components/schemas/Cat` is "Cat". Therefore, when using
 `discriminator`, _inline_ schemas will not be considered because they don't have
 a "name".
 
+Schema names are scoped to the OpenAPI document they are defined in. That means
+if OpenAPI document "A" references a schema in OpenAPI document "B", the schema
+names in "A" aren't considered when evaluating the referenced schema in "B".
+
 ##### Fixed Fields
 Field Name | Type | Description
 ---|:---:|---
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -2727,7 +2727,7 @@ MyResponseType:
   - $ref: '#/components/schemas/Lizard'
 ```
 
-which means the payload _MUST_, by validation, match exactly one of the schemas described by `Cat`, `Dog`, or `Lizard`. Evaluating a `oneOf` can be a costly operation, so `discriminator` MAY be used as a "hint" to improve the efficiency of validation and selection of the matching schema. The `discriminator` keyword can not change the validation result of the `oneOf`, it can only help make the evaluation more efficient and provide better error messaging. We can then describe exactly which field tells us which schema is expected to match the instance:
+which means the payload _MUST_, by validation, match exactly one of the schemas described by `Cat`, `Dog`, or `Lizard`. Evaluating a `oneOf` can be a costly operation, so `discriminator` MAY be used as a "hint" to improve the efficiency of selection of the matching schema. The `discriminator` keyword can not change the validation result of the `oneOf`, it can only help make the evaluation more efficient and provide better error messaging. We can then describe exactly which field tells us which schema is expected to match the instance:
 
 ```yaml
 MyResponseType:
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -2347,7 +2347,7 @@ The OpenAPI Specification allows combining and extending model definitions using
 
 While composition offers model extensibility, it does not imply a hierarchy between the models.
 To support polymorphism, the OpenAPI Specification adds the `discriminator` keyword.
-When used, the `discriminator` will indicate the name of the property that decides which schema definition validates the structure of the model.
+When used, the `discriminator` will indicate the name of the property that hints which schema definition is expected to validate the structure of the model.
 
 ###### XML Modeling
 
@@ -2694,7 +2694,7 @@ components:
 
 #### <a name="discriminatorObject"></a>Discriminator Object
 
-When request bodies or response payloads may be one of a number of different schemas, a `discriminator` object can be used to aid in serialization, deserialization, and validation.  The discriminator is a specific object in a schema which is used to inform the consumer of the document of an alternative schema based on the value associated with it.
+When request bodies or response payloads may be one of a number of different schemas, a `discriminator` object can be used to aid in serialization, deserialization, and validation. The `discriminator` keyword is used to inform the consumer of the document which of the alternatives is expected or preferred.
 
 `discriminator` uses a schema's "name" to automatically map a property value to
 a schema. The schema's "name" is the property name used when declaring the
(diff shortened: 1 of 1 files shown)


MikeRalphson:
On a first reading, this looks to be a fantastic set of cleanups, improvements and fixes.

One thing that might still be outstanding is to clarify the schema name resolution process for the case of multiple (externally reffed) resources. If the "shortcut" reference was always scoped to the same resource, that would seem to make sense, but would have a subtle impact on bundling. Effectively we would be treating the schema name as if it was a schema `$id`. Tooling would have to provide some mechanism to avoid schema name clashes.

MikeRalphson:
Also, if the `discriminator`is used in a pure JSON Schema resource, it is likely that there will be a `$defs` section, not a `components/schemas` section. Do we allow those shortcuts?

jdesrosiers:
@MikeRalphson You're highlighting some of the major problems with the way `discriminator` is defined. Let's start with,

> if the `discriminator` is used in a pure JSON Schema resource, it is likely that there will be a `$defs` section, not a `components/schemas` section. Do we allow those shortcuts?

The "schema name" concept doesn't exist in JSON Schema and isn't defined in OAS, so I had to make up a definition based on context and the existing examples. However, It doesn't look to me like the existing spec considers schemas in `$defs` to be "named" schemas. If it did there would have to be consideration for how to deal with naming conflicts that would then be possible.

Given the way things are defined, I think that "schema names" can only be declared in the `/components/schemas` section of an OAS document. Therefore, the "schema name" addressing feature of `discriminator` _can't_ be used in a pure JSON Schema. It also means that the schema in `/components/schemas` can't have an `$id` or the `#/components/schemas/{schemaName}` URI that is constructed will not point to an OAS document.

Which brings us to,

> If the "shortcut" reference was always scoped to the same resource, that would seem to make sense, but would have a subtle impact on bundling.

Bundling is only a concern if you are working with external schemas and since the "schema name" feature can't work with external schemas, that means bundlers never have to worry about schema names colliding because the schemas it's bundling can't have them.

I don't know how much of that should go in the spec, but it's probably worth adding that schemas defined in `$defs` are not considered named schemas.

MikeRalphson:
I think I agree with the thrust of what you say there @jdesrosiers 

> Therefore, the "schema name" addressing feature of discriminator can't be used in a pure JSON Schema

Obviously a 'pure' JSON Schema resource could still have a `components` section masquerading as an unknown keyword, so let me re-phrase my question as: should we allow shorthand references to the `components/schemas` section in a resource which does not identify itself as an OAS document? Is this overcomplicating the parsing expected of the `mapping` keyword?
Should we deprecate the shorthand name form as soon as possible?

> Bundling is only a concern if you are working with external schemas 

I was using bundling in a generic sense to include the cases where one OAS document includes another, or includes a bare fragment of JSON/YAML, e.g.:

```yaml
openapi: 3.1.1
info:
  title: Music API
  version: 1.0.0
paths:
  /bands:
    get:
      responses:
        '200':
          content:
            'application/json':
              schema:
                $ref: `./subresource.yaml#/components/schemas/Punk'
components:
  schemas:
    Clash: { not: {} }
```

```yaml
components:
  schemas:
     Punk:
      oneOf:
        - $ref: '#/components/schemas/Clash'
        - $ref: '#/components/schemas/Pistols'
        - $ref: '#/components/schemas/Damned'
      discriminator:
        propertyName: bandName
        mapping:
          clash: Clash
          pistols: Pistols
          damned: Damned
     Clash: ...
     Pistols: ...
     Damned: ...
```

in the above example, you can see where the clash (or `Clash`) occurs.

MikeRalphson:
Very open to dealing with additional issues outside of this PR, no need for `mapping` considerations to block this.

jdesrosiers:
> should we allow shorthand references to the `components/schemas` section in a resource which does not identify itself as an OAS document?

I'm trying to be careful not to add or change anything in this patch update so I think keeping "schema names" as an OpenAPI-only concept is the safest choice. Since this concept was never defined in the spec, I think the best thing to do is to get feedback from vendors who have implemented `discriminator` and try to match what existing tooling does. That's assuming different tools even behave the same. Every JSON Schema validator I've seen that implements `discriminator` does it differently and none follow the OpenAPI spec exactly. Interestingly, none that I've seen use schema names.

However, I hadn't considered an OpenAPI document like in your example that is just a container for schemas that another OpenAPI document can reference. I definitely think "schema names" would apply in that case and bundling would become an issue. I would expect that schema names should be scoped to the OpenAPI document where they are defined. Because you can't embed an OpenAPI document in another like you can with pure JSON Schema, you would have to merge `/components/schemas` which could require rewriting/introducing the `mapping` sub-keyword to deal with naming conflicts. You would have to deal with naming conflicts for `$ref`s as well.

> Is this overcomplicating the parsing expected of the `mapping` keyword?

No, I think you are identifying important edge cases that need to be addressed. I don't think we will be able to fix much in this patch update, but it's definitely a conversation that needs to happen.

> Should we deprecate the shorthand name form as soon as possible?

That would be my recommendation. At a minimum, `mapping` shouldn't allow schema names. There's no unambiguous way for implementations to distinguish between value that is intended to be a schema name and a value that is intended to be a URI. My preference would be to deprecate `discriminator` entirely in favor of a new JSON Schema keyword I've proposed, https://github.com/json-schema-org/json-schema-spec/issues/1082. Instead of ...

```yaml
components:
  schemas:
    Punk:
      oneOf:
        - $ref: '#/components/schemas/Clash'
        - $ref: '#/components/schemas/Pistols'
        - $ref: '#/components/schemas/Damned'
      discriminator:
        propertyName: bandName
        mapping:
          clash: Clash
          pistols: Pistols
          damned: Damned
    Clash: ...
    Pistols: ...
    Damned: ...
```
... it would be ...

```yaml
components:
  schemas:
    Punk:
      propertyDependencies:
        bandName:
          clash:
            $ref: '#/components/schemas/Clash'
          pistols:
            $ref: '#/components/schemas/Pistols'
          dammed:
            $ref: '#/components/schemas/Damned'
    Clash: ...
    Pistols: ...
    Damned: ...
```

But that's a whole other conversation.

versions/3.1.1.md
```diff
@@ -2762,9 +2766,9 @@ MyResponseType:
       monster: 'https://gigantic-server.com/schemas/Monster/schema.json'
 ```
 
-Here the discriminator _value_ of `dog` will map to the schema `#/components/schemas/Dog`, rather than the default (implicit) value of `Dog`.  If the discriminator _value_ does not match an implicit or explicit mapping, no schema can be determined and validation SHOULD fail. Mapping keys MUST be string values, but tooling MAY convert response values to strings for comparison.
+Here the discriminator _value_ of `dog` will map to the schema `#/components/schemas/Dog`, rather than the default (implicit) `#/components/schemas/dog`.  If the discriminator _value_ does not match an implicit or explicit mapping, no schema can be determined and validation SHOULD fail. Mapping keys MUST be string values, but tooling MAY convert response values to strings for comparison.
```
sm-Fifteen, line 2769: Doesn't this line now contradict the constraint added a few lines above?

> The `discriminator` keyword can not change the validation result of the `oneOf`, it can only help make the evaluation more efficient and provide better error messaging.

jdesrosiers: In this case the validation result of the `discriminator` keyword is "fail". The validation result of the `oneOf` keyword is not affected. The validation result of the schema as whole is affected by this behavior, but not the validation result of the `oneOf` keyword.

In #2621, I remove this constraint because it doesn't seem necessary to cause validation to fail for a keyword that doesn't otherwise affect the validation result. I couldn't make that change here to avoid backward incompatible changes since this is a patch release.

versions/3.1.1.md
```diff
@@ -2722,8 +2727,7 @@ MyResponseType:
   - $ref: '#/components/schemas/Lizard'
 ```
 
-which means the payload _MUST_, by validation, match exactly one of the schemas described by `Cat`, `Dog`, or `Lizard`.  In this case, a discriminator MAY act as a "hint" to shortcut validation and selection of the matching schema which may be a costly operation, depending on the complexity of the schema. We can then describe exactly which field tells us which schema to use:
-
+which means the payload _MUST_, by validation, match exactly one of the schemas described by `Cat`, `Dog`, or `Lizard`. Evaluating a `oneOf` can be a costly operation, so `discriminator` MAY be used as a "hint" to improve the efficiency of validation and selection of the matching schema. The `discriminator` keyword can not change the validation result of the `oneOf`, it can only help make the evaluation more efficient and provide better error messaging. We can then describe exactly which field tells us which schema is expected to match the instance:
```
darrelmiller, line 2730:
```suggestion
which means the payload _MUST_, by validation, match exactly one of the schemas described by `Cat`, `Dog`, or `Lizard`. Evaluating a `oneOf` can be a costly operation, so `discriminator` MAY be used as a "hint" to improve the efficiency of selection of the matching schema. The `discriminator` keyword can not change the validation result of the `oneOf`, it can only help make the evaluation more efficient and provide better error messaging. We can then describe exactly which field tells us which schema is expected to match the instance:
```
This needs some more discussion...  @jdesrosiers @mkistler

mikekistler: This looks good to me, and I made a small suggestion above to preserve the statement about the discriminator property being required.

versions/3.1.1.md
```diff
@@ -2722,8 +2727,7 @@ MyResponseType:
   - $ref: '#/components/schemas/Lizard'
```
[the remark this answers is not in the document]
jdesrosiers: Done. Thanks.

darrelmiller (approved):

darrelmiller:
Please review and merge @OAI/tsc

versions/3.1.1.md
```diff
@@ -2332,7 +2332,8 @@ The OpenAPI Specification's base vocabulary is comprised of the following keywor
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="schemaDiscriminator"></a>discriminator | [Discriminator Object](#discriminatorObject) | Adds support for polymorphism. The discriminator is an object name that is used to differentiate between other schemas which may satisfy the payload description. See [Composition and Inheritance](#schemaComposition) for more details.
+<a name="schemaDiscriminator"></a>discriminator | [Discriminator
+Object](#discriminatorObject) | Adds support for polymorphism. The discriminator is used to determine which of a set of schemas a payload is expected to satisfy. See [Composition and Inheritance](#schemaComposition) for more details.
```
ralfhandl, line 2336:
```suggestion
<a name="schemaDiscriminator"></a>discriminator | [Discriminator Object](#discriminatorObject) | Adds support for polymorphism. The discriminator is used to determine which of a set of schemas a payload is expected to satisfy. See [Composition and Inheritance](#schemaComposition) for more details.
```
Remove unnecessary line-break within "Discriminator Object"

ralfhandl (changes requested):

versions/3.1.1.md
```diff
@@ -2722,8 +2727,7 @@ MyResponseType:
   - $ref: '#/components/schemas/Lizard'
 ```
 
-which means the payload _MUST_, by validation, match exactly one of the schemas described by `Cat`, `Dog`, or `Lizard`.  In this case, a discriminator MAY act as a "hint" to shortcut validation and selection of the matching schema which may be a costly operation, depending on the complexity of the schema. We can then describe exactly which field tells us which schema to use:
-
+which means the payload _MUST_, by validation, match exactly one of the schemas described by `Cat`, `Dog`, or `Lizard`. Evaluating a `oneOf` can be a costly operation, so `discriminator` MAY be used as a "hint" to improve the efficiency of selection of the matching schema. The `discriminator` keyword can not change the validation result of the `oneOf`, it can only help make the evaluation more efficient and provide better error messaging. We can then describe exactly which field tells us which schema is expected to match the instance:
```
ralfhandl, line 2730:
```suggestion
which means the payload _MUST_, by validation, match exactly one of the schemas described by `Cat`, `Dog`, or `Lizard`. Evaluating a `oneOf` can be a costly operation, so `discriminator` MAY be used as a "hint" to improve the efficiency of selection of the matching schema. The `discriminator` keyword cannot change the validation result of the `oneOf`, it can only help make the evaluation more efficient and provide better error messaging. We can then describe exactly which field tells us which schema is expected to match the instance:
```

ralfhandl (approved):

versions/3.1.1.md
```diff
@@ -2722,8 +2726,7 @@ MyResponseType:
   - $ref: '#/components/schemas/Lizard'
 ```
 
-which means the payload _MUST_, by validation, match exactly one of the schemas described by `Cat`, `Dog`, or `Lizard`.  In this case, a discriminator MAY act as a "hint" to shortcut validation and selection of the matching schema which may be a costly operation, depending on the complexity of the schema. We can then describe exactly which field tells us which schema to use:
-
+which means the payload _MUST_, by validation, match exactly one of the schemas described by `Cat`, `Dog`, or `Lizard`. Evaluating a `oneOf` can be a costly operation, so `discriminator` MAY be used as a "hint" to improve the efficiency of selection of the matching schema. The `discriminator` keyword cannot change the validation result of the `oneOf`, it can only help make the evaluation more efficient and provide better error messaging. We can then describe exactly which field tells us which schema is expected to match the instance:
```
skryukov, line 2729: The description of the PR:
> The discriminator keyword can't be a hint and a shortcut.

but here in the spec I see:
> `discriminator` MAY be used as a "hint" to improve the efficiency of selection of the matching schema

Probably it's more about the PR description, but I think it still worth pointing out 😅

versions/3.1.1.md
```diff
@@ -2810,7 +2818,12 @@ components:
             type: boolean
 ```
 
-a payload like this:
+The `MyResponseType` schema will use the discriminator defined by the `Pet`
```
skryukov, line 2821: This usage of `discriminator` deep inside inheritance tree seems quite complicated to implement and even reason about 🤔 Like, why should we use `MyResponseType.oneOf` and not `Cat.allOf`.

Shouldn't we require having `oneOf`/`anyOf`/`allOf` on the same level? We already enforce it in json schema btw, see https://github.com/OAI/OpenAPI-Specification/pull/3137.

skryukov (commented):
Hey, @jdesrosiers I don't know the status of this PR, but I found the last example quite confusing even after added explanation, and I strongly believe we should reiterate on it.

versions/3.1.1.md
jdesrosiers: I'm not following what you're pointing out. My goal with this change is to make it clear that `discriminator` doesn't change the validation behavior of the `oneOf`/`anyOf` it applies to. At best, it can be used internally by the validator to improve validation performance. For example, if you have a discriminated `anyOf` and the schema you expect to pass is the fifth schema listed, then `discriminator` can be used to tell the implementation to check the fifth schema first deferring the first, second, third, and fourth schemas that aren't expected to pass. But, if that fifth schema fails validation, the validator still needs to check the rest of the schemas to look for a match. So, the validation isn't shortcutted. You still have to do the same validation checks as you would if `discriminator` wasn't present. `discriminator` allows the validator to _prioritize_ the evaluation of a schema that is expected to pass, it doesn't _eliminate_ the other schemas from consideration.

versions/3.1.1.md
jdesrosiers: > Shouldn't we require having `oneOf`/`anyOf`/`allOf` on the same level? We already enforce it in json schema btw, see https://github.com/OAI/OpenAPI-Specification/pull/3137.

Good point! It seems that change was incorrect and should be backed out.

versions/3.1.1.md
skryukov: Yup-yup, I see, and since `discriminator` MAY be used as a hint, description of the PR should be changed from:
> The discriminator keyword can't be a hint and a shortcut.
to maybe smth like this:
> The discriminator keyword can't be a shortcut, only a hint.

versions/3.1.1.md
skryukov: Added a PR with the revert: https://github.com/OAI/OpenAPI-Specification/pull/3405

karenetheridge:
> mapping shouldn't allow schema names. There's no unambiguous way for implementations to distinguish between value that is intended to be a schema name and a value that is intended to be a URI.

Right now this is an ambiguity in the spec. One example uses a simple string mapping, which is intended to be resolved by prefacing with `#/components/schemas`. Another example uses a uri-reference directly. Can we continue to support both (without making changes that would require a larger version bump)? I think perhaps for 3.1.1 we can say that if the mapping string is a fragment-only uri-reference, then it's resolved against the containing document (e.g. for '#/components/schemas/Dog'), but otherwise (no leading `#`) it's treated as a "named schema", where we go to look for that schema under /components/schemas.  This allows for using the discriminator keyword in a plain old json schema (not an openapi document), where we can use a reference to a local $defs entry for mapping schemas.

handrews:
@karenetheridge 

>  I think perhaps for 3.1.1 we can say that if the mapping string is a fragment-only uri-reference, then it's resolved against the containing document (e.g. for '#/components/schemas/Dog'

This is one of the ambiguous cases I go into in my [detailed document](https://github.com/OAI/oascomply/blob/main/reports/processing-model-details.md) of processing models.  Using the containing document would make sense to me, but is probably not how most 3.0 implementations handle it.

karenetheridge:
@handrews I think you missed the main part of my point -- or at least, you veered in a direction that I wasn't going. I wasn't ever implying that there was an ambiguity about how to treat a uri-reference (I would have assumed it was obvious that that should be relative to the current document -- how else could it be interpreted?? what are you referring to with "is probably not how most 3.0 implementations handle it"?).

I was speaking to what to do when the mapping isn't a fragment-only uri-reference -- such as the last example in the spec, where it's just `dog`. This is a valid uri-reference as well, but the example makes it clear that it's intended instead to be interpreted as a "named schema" (which as was pointed out in a different PR that this is a concept introduced in the openapi spec), i.e. one looks for the local `/components/schemas/{mapping value}`.

handrews:
@karenetheridge 

> I was speaking to what to do when the mapping isn't a fragment-only uri-reference -- such as the last example in the spec, where it's just `dog`.

Yes, that's the case I was discussing. (and yes, these cases could be interpreted as URI-references but presumably the name option takes precedence, although that should be clarified).

> i.e. one looks for the local `/components/schemas/{mapping value}`, i.e. `/components/schemas/dog`.

That would be true according to this new working from this PR (and is how I would expect it to work):

> Schema names are scoped to the OpenAPI document they are defined in. That means
if OpenAPI document "A" references a schema in OpenAPI document "B", the schema
names in "A" aren't considered when evaluating the referenced schema in "B".

However, according to @MikeRalphson and @webron, that is not how it has historically worked.  Only the Components Object from the entry document is in scope for name resolution.  Or at least, I am told that only direct `$ref` targets are incorporated in to the resolve OpenAPI Description, and since you cannot `$ref` the _entire_ Components Object (including the names) there is no way to make names from Components Objects in documents other than the entry document available as named targets.  This is because there is no guarantee that anything other than the `$ref` target in the other document is even in a valid OpenAPI format, and the specification does not offer any ways to detect such things.

I have a detailed example about this, with commentary, in the document that I linked in my last comment.

jdesrosiers:
> I think perhaps for 3.1.1 we can say that if the mapping string is a fragment-only uri-reference, then it's resolved against the containing document (e.g. for '#/components/schemas/Dog'), but otherwise (no leading `#`) it's treated as a "named schema", where we go to look for that schema under /components/schemas.

I don't have an opinion here, but I'd point out that this excludes the possibility of an external reference in `discriminator`. That may or may not be a good thing. I think a lot depends on what existing implementations support and what users are expecting.

> However, according to @MikeRalphson and @webron, that is not how it has historically worked.

This PR is specifically not trying to change anything, just clarify and fill some gaps. I'm far from an expert how this has historically worked, so I might have inadvertently introduced a change. I'm happy to rewrite anything that needs fixing or better yet, hand this over to someone more actively involved with OpenAPI to make the corrections.

handrews:
> I don't have an opinion here, but I'd point out that this excludes the possibility of an external reference in discriminator. That may or may not be a good thing. I think a lot depends on what existing implementations support and what users are expecting.

Yeah, that's pretty murky.  The final 2023 milestone for the OASComply project is starting a set of test cases around referencing ambiguities (including both literal `$ref` and things like implicit mappings in the Discriminator Object).  The spec ambiguities mean that there are several possible outcomes that are at least _arguably_ within the spec.  So this "test suite" will be more of an "assessment suite" in the sense that it will specify several potentially valid outcomes.  So we can use that to find out what tools actually support.

I think this PR is a solid effort at clarification, which will be a lot easier to resolve once we figure out the current tooling landscape and the TSC makes a ruling on how to handle the ambiguities.  That could be done by leaving this open and revisiting it, or just filing an issue (or updating an existing one?) to revisit it after we've assessed the tools.

earth2marsh:
Discussed today, this generally seems like a good idea, but since Moonwalk plans to revisit discriminator as part of its focus on signatures, it seems wiser to see how that plays out before we try to resolve the ambiguities here.

However, there are a number of good wording improvements (thank you), it seems useful to separate those as a separate PR? Then we can incrementally improve the spec in the short term. @jdesrosiers , if you are able to do that, that would be great, and if not, folks seemed positive on today's call that we could find help for that.

jdesrosiers:
I could do a revision, but it's not clear to me which parts you'd like to keep and which to revert. If someone wants to submit a review pointing out the parts they'd like changed, that would be helpful. Or, I'm totally fine if someone want to just take the content and submit a new PR if that's easier.

versions/3.1.1.md
```diff
@@ -2698,14 +2693,23 @@ components:
...
+`discriminator` uses a schema's "name" to automatically map a property value to
+a schema. The schema's "name" is the property name used when declaring the
+schema as a component in an OpenAPI document. For example, the name of the
+schema at `#/components/schemas/Cat` is "Cat". Therefore, when using
+`discriminator`, _inline_ schemas will not be considered because they don't have
+a "name".
+
+Schema names are scoped to the OpenAPI document they are defined in. That means
+if OpenAPI document "A" references a schema in OpenAPI document "B", the schema
+names in "A" aren't considered when evaluating the referenced schema in "B".
```
handrews, line 2707: This is the main part that's controversial.  In Swagger 2.0, the references were strongly implied to be scoped to document "A", and many tools implement that behavior.  There are other component name connections that have similar ambiguities and we should address them all at once.  (FWIW, I strongly agree with what you've written here, but it would be a de-facto breaking change for some and we recently decided not to do "clarifications" that are de-facto breaking changes).

versions/3.1.1.md
```diff
@@ -2773,6 +2776,11 @@ For example:
 ```yaml
 components:
   schemas:
+    MyResponseType:
+      oneOf:
+      - $ref: '#/components/schemas/Cat'
+      - $ref: '#/components/schemas/Dog'
+      - $ref: '#/components/schemas/Lizard'
```
handrews, line 2783: My understanding is that the `"allOf"` form specifically exists to avoid having to list the subschemas in a `"oneOf"` or `"anyOf"`, so this should not be changed.  I don't personally understand how this even works, and [I'm not the only one](https://github.com/OAI/OpenAPI-Specification/issues/3591).  But it's related to the controversial thing- if all schema names resolve to the Components Object of what we now call the ["entry document"](https://learn.openapis.org/glossary.html), you can just search that one list.  I guess?  Apparently we need to figure that out.

versions/3.1.1.md
```diff
@@ -2810,7 +2818,12 @@ components:
             type: boolean
 ```
 
-a payload like this:
+The `MyResponseType` schema will use the discriminator defined by the `Pet`
+schema because it is part of the `Cat`, `Dog`, and `Lizard` schemas in the
+`oneOf`. The behavior if not all schemas define a `discriminator` and they are
+not all the same is undefined.
```
handrews, line 2824: This would need to be removed if the `oneOf` addition to the example is removed.

handrews (commented):
@jdesrosiers I saw your question about what could and couldn't be kept - I've added some comments based on my understanding and the discussions we're having elsewhere related to how referencing and component names interact in general.

AFAICT I philosophically agree with what you're trying to do here, I'm just conveying what I've been told about what's complex and ambiguous right now.

Up to you whether you want to edit this or close it.  I'm pretty sure all of these have issues related to them so they'd get revisited eventually.

jdesrosiers:
Thanks @handrews. That looks like what I need to do a revision. It'll be a couple more days still until I can get to it, but I wanted to acknowledge that I saw it and plan to get to it soon.

versions/3.1.1.md
```diff
@@ -2698,14 +2693,23 @@ components:
 
 #### <a name="discriminatorObject"></a>Discriminator Object
 
-When request bodies or response payloads may be one of a number of different schemas, a `discriminator` object can be used to aid in serialization, deserialization, and validation.  The discriminator is a specific object in a schema which is used to inform the consumer of the document of an alternative schema based on the value associated with it.
+When request bodies or response payloads may be one of a number of different schemas, a `discriminator` object can be used to aid in serialization, deserialization, and validation. The `discriminator` keyword is used to inform the consumer of the document which of the alternatives is expected or preferred.
```
adamaltman, line 2696: I find the phrase "expected or preferred" strange.  Also, I'm not sure how "consumer" is defined in this sentence but it can be used by both the consumer and server.

```suggestion
When request bodies or response payloads may be one of a number of different schemas, a `discriminator` object gives a hit about the expected schema of the document. It can be used to aid in serialization, deserialization, and validation.
```
Also, why interchange "keyword" and "object"?

versions/3.1.1.md
adamaltman:
```suggestion
which means the payload _MUST_, by validation, match exactly one of the schemas described by `Cat`, `Dog`, or `Lizard`. Evaluating a `oneOf` can be a costly operation, so `discriminator` MAY be used as a "hint" to improve the efficiency of selection of the matching schema. The `discriminator` keyword cannot change the validation result of the `oneOf`, it can only help make the evaluation more efficient and provide better error messaging. We can specify the exact field that tells us which schema is expected to match the instance:
```

versions/3.1.1.md
```diff
@@ -2744,7 +2747,7 @@ The expectation now is that a property with name `petType` _MUST_ be present in
 }
 ```
 
-Will indicate that the `Cat` schema be used in conjunction with this payload.
+Will indicate that the `Cat` schema is the alternative that is expected to match this payload.
```
adamaltman, line 2750:
```suggestion
Will indicate that the `Cat` schema is expected to match this payload.
```

versions/3.1.1.md
```diff
@@ -2345,13 +2345,8 @@ The OpenAPI Specification allows combining and extending model definitions using
...
 While composition offers model extensibility, it does not imply a hierarchy between the models.
-To support polymorphism, the OpenAPI Specification adds the `discriminator` field.
-When used, the `discriminator` will be the name of the property that decides which schema definition validates the structure of the model.
-As such, the `discriminator` field MUST be a required field.
-There are two ways to define the value of a discriminator for an inheriting instance.
-- Use the schema name.
-- Override the schema name by overriding the property with a new value. If a new value exists, this takes precedence over the schema name.
-As such, inline schema definitions, which do not have a given id, *cannot* be used in polymorphism.
+To support polymorphism, the OpenAPI Specification adds the `discriminator` keyword.
+When used, the `discriminator` will indicate the name of the property that hints which schema definition is expected to validate the structure of the model.
```
adamaltman, line 2349:
```suggestion
When used, the `discriminator` indicates the name of the property that hints which schema definition is expected to validate the structure of the model.
```

versions/3.1.1.md
jdesrosiers: > why interchange "keyword" and "object"?

"Keyword" refers to the key and "object" refers to the value. The `discriminator` keyword has an object as its value.

versions/3.1.1.md
jdesrosiers: The problem I was trying to solve here is that the example payloads never say which schema is being used to evaluate those payloads. Based on my understanding of the `allOf` usage, none of those schemas could produce the results the documentation says it does, so I introduced something that could.

At this point I can only assume that the "Pet" schema is what these payloads are intended to be evaluated against and the behavior described is correct in that case. So, I've reverted my change other than to specify that "Pet" is the schema these payloads are evaluated against. Let me know if that's the right thing to do. I can't see how that makes sense, so I wouldn't be surprised if it's still not correct. Let me know.

handrews (approved):
Thanks! LGTM

versions/3.1.1.md
```diff
@@ -2722,8 +2722,7 @@ MyResponseType:
   - $ref: '#/components/schemas/Lizard'
 ```
 
-which means the payload _MUST_, by validation, match exactly one of the schemas described by `Cat`, `Dog`, or `Lizard`.  In this case, a discriminator MAY act as a "hint" to shortcut validation and selection of the matching schema which may be a costly operation, depending on the complexity of the schema. We can then describe exactly which field tells us which schema to use:
-
+which means the payload _MUST_, by validation, match exactly one of the schemas described by `Cat`, `Dog`, or `Lizard`. Evaluating a `oneOf` can be a costly operation, so `discriminator` MAY be used as a "hint" to improve the efficiency of selection of the matching schema. The `discriminator` keyword cannot change the validation result of the `oneOf`, it can only help make the evaluation more efficient and provide better error messaging. We can specify the exact field that tells us which schema is expected to match the instance:
```
mikekistler, line 2725: I'm a little hung up on this paragraph's use of "evaluation" -- does that mean "validation" and if so will that be clear to (knowledgable) readers? Also, it's not clear to mean that the discriminator makes _validation_ of a `oneOf` schema more efficient -- you still have to check _all_ of the schemas to know that exactly one matches. The value of the discriminator is that it can make serialization/deserialization more efficient.

Here's a revision that I think is clearer:
```suggestion
which means the payload _MUST_, by validation, match exactly one of the schemas described by `Cat`, `Dog`, or `Lizard`. Deserialization of a `oneOf` can be a costly operation, as it requires determining which schema matches the payload and thus should be used in deserialization. This problem also exists for `anyOf` schemas. A `discriminator` MAY be used as a "hint" to improve the efficiency of selection of the matching schema. The `discriminator` keyword cannot change the validation result of the `oneOf`, it can only help make the deserialization more efficient and provide better error messaging. We can specify the exact field that tells us which schema is expected to match the instance:
```

mikekistler (commented):
This is a nice improvement, but I suggested some alternate wording for one paragraph that I think is a tad better.

versions/3.1.1.md
jdesrosiers: > I'm a little hung up on this paragraph's use of "evaluation" -- does that mean "validation"

Generally, in JSON Schema land, we use the term "evaluation" rather than "validation" to acknowledge that there are multiple ways to evaluate a schema and not all of them are validation. Other types of evaluations include code generation, documentation generation, form generation, etc. However, I think in this case we're really just talking about validation, so it might be better to just say "validation".

> it's not clear to mean that the discriminator makes validation of a `oneOf` schema more efficient -- you still have to check all of the schemas to know that exactly one matches.

In most cases, you need to validate every option. The one case where you can short circuit is when you find two matching schemas. Once you find the second passing schema, you know validation will fail and you don't need to check the rest. `disciminator` can help you get there more quickly by checking the one you expect to fail first.

* Pass
* Fail
* Fail
* Fail
* Pass <- discriminator

Assuming the schemas will be evaluated in order, without a discriminator, you would have to check every schema. With a disciminator, you could check that one first and then when evaluating the rest in order, you would find the extra match right away and could skip evaluating the rest of the schemas.

It requires a bit of luck and a very specific scenario, but it's possible. The only reason I included this line was to keep the spirit of the original text, but I'm happy to take it out. It's technically correct, but the possibility of optimization is so small and so rare, it's not really worth mentioning. `anyOf` could provide a level of optimization worth mentioning, but no one uses it, which is why I used `oneOf` instead.

mikekistler: > With a disciminator, you could check that one first and then when evaluating the rest in order, you would find the extra match right away and could skip evaluating the rest of the schemas.

Ah, great point. I did not think of this, but now I see how a discriminator can also make validation more efficient.

mikekistler (approved):
Looks good! 👍

PR #2627, opened 2021-06-23
Backfill CONTRIBUTORS.md as per all-contributors spec

MikeRalphson:
As discussed, this PR backfills the `CONTRIBUTORS.md` file according to the [All-Contributors](https://allcontributors.org/) specification, moving us from our current rather high bar of 25 commits/PRs.

The `all-contributors-cli` package was used to create the commits (to avoid having 150+ PRs to merge), based on a reformatted `git log` output of the `main` branch, where email addresses were looked up to their GitHub accounts.

Going forward we can install the `All Contributors` GitHub app and comment as follows on PRs/issues:

`@all-contributors please add @whoever for [doc|example|code]`

to keep this up-to-date.


MikeRalphson:
@OAI/tsc reaction was lukewarm, in any case the list of contributors would have to be regenerated if it was merged after all this time.

Issue #2471, opened 2021-02-18
Limiting OAS3.1 to RFC7231 status codes?


ioggstream:
## I expect

To reference all registered status codes

## Instead

The spec says

> The available status codes are defined by RFC7231

## Note

It is not clear  to whether the current formulation limits the available status codes to the one in 7231

philsturgeon:
What else would it mean? You can use any status codes defined in RFC 7231. Using custom status codes is undefined and therefore undefined behavior is to be expected.

Was there a particular status code you were looking to use thats not in 7231 or are you trying to use custom codes? 

For extra context, the JSON Schema validates the following:
```
patternProperties:
          '[1-5](?:\d{2}|XX)':
```

So you could go using custom status codes if you wanted to, but if its outside of those rules some validators will barf errors at you.

ioggstream:
429 status code is rfc6585, but not in 7231 ;) 

I'm just saying that there are http-extensions beyond 7231 and they are standards too.

hkosova:
OAS refers not only to RFC 7231 but also to the IANA registry:

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.1.0.md#http-status-codes
> **HTTP Status Codes**
> The available status codes are defined by [RFC7231](https://tools.ietf.org/html/rfc7231#section-6) and registered status codes are listed in the [IANA Status Code Registry](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml).

RFC 7231 itself [points to](https://tools.ietf.org/html/rfc7231#page-50) the IANA registry for the complete list of codes:

>  Note that this list is not exhaustive -- it does not include extension status codes defined in other specifications.  The complete list of status codes is maintained by IANA.

<br/>

Maybe the intended meaning here is that _definition_ of "HTTP status code" is taken from RFC 7231 (i.e. it's a 3-digit integer in the 100-599 range), and for specific registered status codes the reader should refer to the IANA registry?

philsturgeon:
@hkosova sounds good!

ioggstream:
@hkosova exactly, the text just need a slight rewording. Feel free to PR with your text... or I'll do it tomorrow if you don't have time.

MikeRalphson:
@ioggstream @hkosova woudl one of you like to take a go at that proposed PR?

ioggstream:
[...]

PR #2630, opened 2021-06-25
Fix: #2471. Clarify status code text.

ioggstream:
## This PR

- clarifies text around supported status codes.

Fixes: #2471

### DIFF
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -97,7 +97,7 @@ Some examples of possible media type definitions:
 ```
 ##### <a name="httpCodes"></a>HTTP Status Codes
 The HTTP Status Codes are used to indicate the status of the executed operation. 
-Status codes are defined in [RFC7231](https://tools.ietf.org/html/rfc7231#section-6) and the available status codes are registered in the [IANA Status Code Registry](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml).
+Status codes SHOULD be selected from the available status codes registered in the [IANA Status Code Registry](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml).
 
 ## Specification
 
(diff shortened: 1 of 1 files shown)


ioggstream:
@MikeRalphson as you requested :)

MikeRalphson (dismissed):
Thanks @ioggstream! Please could you retarget against `versions/3.1.1.md` on the `v3.1.1-dev` branch?

versions/3.1.1.md
```diff
@@ -97,7 +97,7 @@ Some examples of possible media type definitions:
 ```
-The available status codes are defined by [RFC7231](https://tools.ietf.org/html/rfc7231#section-6) and registered status codes are listed in the [IANA Status Code Registry](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml).
+Status codes are defined in [RFC7231](https://tools.ietf.org/html/rfc7231#section-6) and the available status codes are registered in the [IANA Status Code Registry](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml).
```
darrelmiller, line 100:
```suggestion
Status codes SHOULD be selected from the available status codes registered in the [IANA Status Code Registry](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml).
```

karenetheridge:
This suggests the JSON Schema should be softened to support more than just `[1-5][0-9][0-9]`. Can you make that change as well? (schemas/v3.1/schema.json and .yaml)

ioggstream:
Hi @karenetheridge,  maybe I'm missing something. 
The IANA table covers the range `[100; 599]` as specified in https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#status.codes

> All valid status codes are within the range of 100 to 599, inclusive

iiuc all IANA status codes should comply with the above spec, and that there should be no change in  https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.1/schema.yaml#L529


If its not  the case, how would you modify the above file?

thx++, R

darrelmiller:
@karenetheridge There are no status codes in the IANA registry outside of the 100-599 range, so the regex in JSON Schema is fine.

darrelmiller (approved):

karenetheridge:
The specification says SHOULD -- so doesn't this mean that it is still permissible for values to be outside that range (for some unusual usecase)?

Issue #2635, opened 2021-06-29
Ambiguity between Reference Object and PathItem Object


char0n:
Hi everybody, not sure if this is the right place to ask this, but I've intercepted an ambiguity while building tooling for OAS 3.1

```json
{
  "openapi": "3.1.0",
  "components": {
    "pathItems": {
      "pathItem1": {
        "$ref": "#/components/pathItems/pathItem2",
      },
      "pathItem2": {
        "summary": "path item summary",
        "description": "path item description",
        "get": {}
      }
    }
  }
}
```

Given following document, pathItem1  can be interpreted either as Reference Object or Path Item Object. It's hard to tell what the author intended. He could have intended for pathItem1 to be a Reference Object. Or he may have intended pathItem1 to be a Path Item Object. As Path Item Object can have exactly the same shape as Reference Object the  intention of the author is lost. 

Can you please advice how to approach that?

Instead of having this ambiguity and `$ref` field in Path Item Object, wouldn't it be cleaner to use Reference Object explicitly in every place where Path Item Object is allowed?

Current **workaround** could be assert the presence of `$ref` field and one of the other fixed fields from Path Item Object (like `get`, `post`, etc...). If assertion like that passes, we can say with limited certainty that author intended to define a Path Item Object.

jdesrosiers:
I didn't notice this before. The schema doesn't validate this properly. It only allows for a Path Item Object (with no Reference Object properties) _OR_ a Reference Object. A Path Item Object with Reference Object properties would fail validation.

What is the reason for including Reference Object properties in Path Item Object as opposed to just using a Reference Object? It's the only type that does that. It also doesn't seem very well specified. It doesn't specify how a reference works in this context, but implies that there differences from a normal Reference Object. A Reference Object has rules for merging "summary" and "description", but other properties are not merged in the referenced object. Path Item Object implies that some merging takes place, but doesn't specify how (e.g. deep or shallow merge). Is it assumed that Reference Object behaviors apply here unless otherwise specified? Do the merging rules for "summary" and "description" in a Reference Object apply?

[...]

PR #2656, opened 2021-07-23
Deprecate Path Item Object ref field (#2635)

char0n:
The  field has been deprecated in favor of
using Reference Object in OAS 3.2.0 and forward.

Signed-off-by: Vladimir Gorej <vladimir.gorej@gmail.com>


char0n:
Not really sure if my wording in deprecation message is ideal. I'm very open to any suggestions ;]

MikeRalphson:
Hmm. I think `$ref` within a `pathItem` object is a useful construct for OAS document composition, and it is only the `$ref` field with sibling `operationObject` properties which appears to be almost unused and I feel should be deprecated. As your PR #2655 removes the ambiguous use of `referenceObjects` to refer to `pathItems`, the `$ref` field remains the only way to refer to `pathItems` in external resources.

As to wording, I feel we should not "throw forward" to other versions of the spec, for two reasons:
* it may imply an ordering in which we need to release `v3.2.0` and `v3.1.1`
* versions of the spec beyond `v3.2.0` (such as a `v4.x`) may handle this in different ways

so personally, I would prefer the wording to be self-contained as to the situation within the version of the spec the markdown file refers to.

char0n:
Yep, fully agree. I'll try to come up with self-contained wording deprecating just use-case of `$ref` accompanied with operation fields.

char0n:
We've came up with 3 possible versions of deprecation message (thanks @frankkilcommins)

**1.)**
```
The usage of `$ref` property has been deprecated when accompanied with other properties different from `summary` and `description`.
```

**2.)**
```
Usage of the `$ref` property has been deprecated when accompanied with properties other than `summary` and `description`.
```

**3.)**
```
Usage of the `$ref` property has been deprecated for use cases where properties other than `summary` and/or `description` are specified within the Path Item Object.
```

@MikeRalphson please let me know which one works best for you and I'll amend the PR.

char0n:
@MikeRalphson just pinging here, regarding my last comment.

karenetheridge:
Maybe this should be in a separate PR, but should this bit of the schema be changed?

```
 $id: 'https://spec.openapis.org/oas/3.1/schema/2021-05-20'
 $schema: 'https://json-schema.org/draft/2020-12/schema'
 type: object
 ...
 $defs:
 ...
   paths:
     type: object
     patternProperties:
       '^/':
-        $ref: '#/$defs/path-item'
+        $ref: '#/$defs/path-item-or-reference'
     $ref: '#/$defs/specification-extensions'
     unevaluatedProperties: false
```

ivank:
Doesn't path-item always allow "reference"? At least that's what the wording on it seems to imply.
https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#path-item-object

Its been weird that some tooling are allowing openapi paths to be reference, for easier splitting into separate files, but this validation actually prevents this. Maybe I'm reading this wrong but the spec actually state that all paths should be able to be references?

PR #2657, opened 2021-07-23
Allow referencing Path Item only via Reference Object (#2635)

char0n:
Signed-off-by: Vladimir Gorej <vladimir.gorej@gmail.com>

### DIFF
--- a/versions/3.2.0.md
+++ b/versions/3.2.0.md
@@ -661,7 +661,7 @@ The path is appended to the URL from the [`Server Object`](#serverObject) in ord
 
 Field Pattern | Type | Description
 ---|:---:|---
-<a name="pathsPath"></a>/{path} | [Path Item Object](#pathItemObject) | A relative path to an individual endpoint. The field name MUST begin with a forward slash (`/`). The path is **appended** (no relative URL resolution) to the expanded URL from the [`Server Object`](#serverObject)'s `url` field in order to construct the full URL. [Path templating](#pathTemplating) is allowed. When matching URLs, concrete (non-templated) paths would be matched before their templated counterparts. Templated paths with the same hierarchy but different templated names MUST NOT exist as they are identical. In case of ambiguous matching, it's up to the tooling to decide which one to use.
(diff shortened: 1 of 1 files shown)


MikeRalphson (approved):

gimbimloki:
Hello. When OAI 3.2.0 is released, could I use Reference Objects in Paths Object like below?
```yaml
paths:
  $ref: foo.yaml
  $ref: bar.yaml
```

handrews:
@gimbimloki YAML [does not allow for duplicate keys](https://yaml.org/spec/1.2.2/#node-comparison), so no (and that's not something OAS can change).  

Duplicate keys in JSON are a [more complex topic](https://www.rfc-editor.org/rfc/rfc8259.html#section-4), but the TL;DR is that most JSON implementations don't support them in a usable way.

gimbimloki:
@handrews Thank you for your reply. 
So, Only one '$ref' can be in Paths Object, is it right?

karenetheridge:
This change feels incomplete, as it's now allowing a path-item entry to be a regular reference, but then not changing the definition of a path-item itself to disallow `$ref` from being present adjacent to other path-item properties like summary, description, parameters, get etc.

So now, this is still legal:

```yaml
paths:
  /pets/{petId}:
    $ref: "#/components/pathItems/one_pet"
components:
  path-items:
    one_pet:
      $ref: "#/components/pathItems/more_pet_things"
      summary: a single pet
      get:
        ... operation things here
    more_pet_things:
      $ref: "#/components/pathItems/even_more_pet_things"
      description: ...
      post:
        ... operation things here
    even_more_pet_things:
      put:
        ...
```

...which means that when processing a request for `PUT /pets/123` we end up getting bits of the request description from all of /components/path-items/*, which is the mess that we had in 3.1.0.

To fix this, I think we need to remove the `$ref` property from /$defs/path-item (which is both a spec and a schema change). That way we can still have multiple $refs all chained together (just as we allow for other entity types like response, header etc) but no non-$ref fields can appear anywhere along the chain, only at the leaf.

PR #2669, opened 2021-08-05
Generate YAML versions of these schema documents

karenetheridge:
generated via:
perl -MYAML::XS -MJSON::MaybeXS -we'$YAML::XS::Boolean="JSON::PP"; print Dump(JSON::MaybeXS->new->decode(do { local $/; <> }))' input.json > output.yaml


jdesrosiers (approved):
I didn't think it was necessary to include a yaml version of these, but if people find it useful, I'm happy to include it.

darrelmiller (approved):

PR #2671, opened 2021-08-06
minor simplification of v3.1 schema

karenetheridge:
"if the X property is present, then apply this constraint to property X" can
be simplified by simply stating the constraint for property X.

### DIFF
--- a/schemas/v3.1/schema.yaml
+++ b/schemas/v3.1/schema.yaml
@@ -634,9 +634,6 @@ $defs:
       deprecated:
         default: false
         type: boolean
-      allowEmptyValue:
-        default: false
-        type: boolean
       schema:
         $dynamicRef: '#meta'
       content:
@@ -654,9 +651,6 @@ $defs:
             enum:
               - simple
           explode:
-            default: false
-            type: boolean
-          allowReserved:
             default: false
             type: boolean
         $ref: '#/$defs/examples'
--- a/schemas/v3.1/schema.yaml
+++ b/schemas/v3.1/schema.yaml
@@ -637,6 +637,10 @@ $defs:
       allowEmptyValue:
         default: false
         type: boolean
+      schema:
+        $dynamicRef: '#meta'
+      content:
+        $ref: '#/$defs/content'
     dependentSchemas:
       schema:
         properties:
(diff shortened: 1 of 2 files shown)


jdesrosiers (changes requested):
You're right, this doesn't work like I intended.

Would you mind also making a couple other corrections I noticed when reviewing?

1. The "allowEmptyValue" property doesn't apply to headers. That can be removed.
2. The schema doesn't include the constraint that one and only of "schema" or "content" is required.

karenetheridge:
yes, give me a day or two

karenetheridge:
> The "allowEmptyValue" property doesn't apply to headers. That can be removed.

Are you sure? https://spec.openapis.org/oas/v3.1.0#headerObject specifies that it's the same as parameter, except no 'name' or 'in' -- and 'parameter' has 'allowEmptyValue'.

> The schema doesn't include the constraint that one and only of "schema" or "content" is required.

done.

karenetheridge:
since the header definition has 'unevaluatedProperties: false' in it, does this mean that no one has been testing this schema yet against headers (or surely they would have noticed the allowEmptyValue discrepancy)??!

jdesrosiers:
> > The "allowEmptyValue" property doesn't apply to headers. That can be removed.
>
> Are you sure?

Sorry, I meant the "allowReserved" property doesn't apply to headers. :facepalm: 

I came to this conclusion because the `in` value is implicitly `header` and the `allowReserved` description says, "This property only applies to parameters with an `in` value of `query`." Therefore, I concluded that `allowReserved` doesn't apply to the Header Object.

Actually, now that I look at it again, it looks like the same is true of `allowEmptyValue`. It's description says, "This is valid only for `query` parameters". This is for a header parameter, so it doesn't apply either.

schemas/v3.1/schema.yaml
```diff
@@ -650,13 +656,7 @@ $defs:
           allowReserved:
             default: false
             type: boolean
```
jdesrosiers, line 658: Let's remove this one too.
```suggestion
```

jdesrosiers (changes requested):

schemas/v3.1/schema.yaml
karenetheridge: done.

jdesrosiers (approved):
Thanks!

jdesrosiers (approved):

darrelmiller (approved):

Issue #2666, opened 2021-08-02
OAuth Flow required fields + validation


char0n:
Hi everybody,

I'm currently working on some tooling regarding OAS 3.0.0 and found a situation where I'd need some clarification. Spec defines `OAuth Flow Object` as an object with following fields required:

- **authorizationUrl**
- **tokenUrl**
- **scopes**

This is crystal clear so far. Next I'm using [JSON Schema](https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.0/schema.yaml#L871) provided in this repo to validate the OAS 3.0.0 definition. But looking into the JSON Schema document reveals that none of the `OAuth Flow` objects require above define fields to be required. Some OAuthFlow objects requires just one field, other two. I'm not sure if this is intended (for particular type of OAuth Flow to have only certain fields required) or this is something we should fixed.

**Update**
I completely missed **Applies To** column of OAuth Flow Object. That clarifies things out. But `scopes` are required in `ImplicitOAuthFlow` only, even though they should be required in all flow types.  In this case is the spec or the JSON Schema the source of truth?

```yaml
  ImplicitOAuthFlow:
    type: object
    required:
      - authorizationUrl
      - scopes
    properties:
      authorizationUrl:
        type: string
        format: uri-reference
      refreshUrl:
        type: string
        format: uri-reference
      scopes:
        type: object
        additionalProperties:
          type: string
    patternProperties:
      '^x-': {}
    additionalProperties: false

  PasswordOAuthFlow:
    type: object
    required:
      - tokenUrl
    properties:
      tokenUrl:
        type: string
        format: uri-reference
      refreshUrl:
        type: string
        format: uri-reference
      scopes:
        type: object
        additionalProperties:
          type: string
    patternProperties:
      '^x-': {}
    additionalProperties: false

  ClientCredentialsFlow:
    type: object
    required:
      - tokenUrl
    properties:
      tokenUrl:
        type: string
        format: uri-reference
      refreshUrl:
        type: string
        format: uri-reference
      scopes:
        type: object
        additionalProperties:
          type: string
    patternProperties:
      '^x-': {}
    additionalProperties: false

  AuthorizationCodeOAuthFlow:
    type: object
    required:
      - authorizationUrl
      - tokenUrl
    properties:
      authorizationUrl:
        type: string
        format: uri-reference
      tokenUrl:
        type: string
        format: uri-reference
      refreshUrl:
        type: string
        format: uri-reference
[...]

PR #2158, opened 2020-02-25
#2157: Update schema to validate refs in headers of encoding objects

alasdairhurst:
See https://github.com/OAI/OpenAPI-Specification/issues/2157

PR #2673, opened 2021-08-10
Add scopes as required field of OAuth Flow Object

char0n:
This makes metaschema consistent with the 3.0.x spec.

Refs #2666

### DIFF
--- a/schemas/v3.0/schema.yaml
+++ b/schemas/v3.0/schema.yaml
@@ -1,4 +1,4 @@
-id: https://spec.openapis.org/oas/3.0/schema/2020-02-25
+id: https://spec.openapis.org/oas/3.0/schema/2021-08-12
 $schema: http://json-schema.org/draft-04/schema#
 description: Validation schema for OpenAPI Specification 3.0.X.
 type: object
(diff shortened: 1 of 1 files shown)


karenetheridge (changes requested):
Please update schema.yaml as well.

char0n:
@karenetheridge schema.yaml is updated as well now.

hkosova:
Should the schema `id` also be updated to the current date?

```diff
-id: https://spec.openapis.org/oas/3.0/schema/2020-02-25
+id: https://spec.openapis.org/oas/3.0/schema/2021-08-10
```

This was done the last time when the 3.0 schema was updated:
https://github.com/OAI/OpenAPI-Specification/pull/2158/files

webron:
Update procedure is in the [README](https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.0/README.md) 😃

webron:
That was also a nudge to @MikeRalphson as he likes automating things ;) You can keep the json version too, I don't see it as a problem.

char0n:
Ok now it's done precisely according to the README ;]

darrelmiller (approved):

PR #2681, opened 2021-08-23
tighten up regexes for path items

karenetheridge:
https://spec.openapis.org/oas/v3.1.0#pathTemplating
"The value for these path parameters MUST NOT contain any unescaped “generic syntax” characters described by [[!RFC3986]]: forward slashes (/), question marks (?), or hashes (#)."

### DIFF
--- a/schemas/v3.1/schema.json
+++ b/schemas/v3.1/schema.json
@@ -502,6 +502,9 @@
               },
               "then": {
                 "properties": {
+                  "name": {
+                    "pattern": "[^/#?]+$"
+                  },
                   "style": {
                     "default": "simple",
                     "enum": [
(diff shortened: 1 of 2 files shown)


schemas/v3.1/schema.json
```diff
@@ -277,7 +277,7 @@
     "paths": {
       "type": "object",
       "patternProperties": {
-        "^/": {
+        "^/[^#?]*$": {
```
MikeRalphson, line 280: As mentioned on Slack, I believe this restriction would be going beyond what the markdown source-of-truth dictates (solely that the path key MUST start with a forward slash). Though unescaped `#` and '?' characters cannot appear within path template variables, there is afaik no restriction on them occurring **outside** those parameters. I make no comment on whether that is fortunate or not.

MikeRalphson (changes requested):

jdesrosiers:
This looks good to me if @MikeRalphson confirms that it accurately reflects the specification.

darrelmiller (approved):

PR #2682, opened 2021-08-24
Update and rename petstore.yaml to test1.1.yaml

StefanBigler:


MikeRalphson:
Could you provide an description of why this OAS document is a better replacement for the petstore example?

darrelmiller:
This PR appears to overwrite the Petstore example with a different example.  Please feel free to reopen if you want to submit a different example along side the Petstore example.

PR #2690, opened 2021-09-01
per the spec, the only allowed ranges are [12345]XX

karenetheridge:
https://spec.openapis.org/oas/v3.1.0#patterned-fields-0

### DIFF
--- a/schemas/v3.1/schema.json
+++ b/schemas/v3.1/schema.json
@@ -774,7 +774,7 @@
         }
       },
       "patternProperties": {
-        "^[1-5][0-9X]{2}$": {
+        "^[1-5](?:[0-9]{2}|XX)$":
           "$ref": "#/$defs/response-or-reference"
         }
       },
--- a/schemas/v3.1/schema.yaml
+++ b/schemas/v3.1/schema.yaml
@@ -520,7 +520,7 @@ $defs:
       default:
         $ref: '#/$defs/response-or-reference'
     patternProperties:
-      '^[1-5][0-9X]{2}$':
+      '^[1-5](?:[0-9]{2}|XX)$':
         $ref: '#/$defs/response-or-reference'
     $ref: '#/$defs/specification-extensions'
     unevaluatedProperties: false


jdesrosiers (approved):

darrelmiller (approved):

undera:
The schema is broken after this change. The problem is removed closing curly brace. I guess it was not intentional...

undera:
Specifically, the JSON file of schema is broken. Unfortunately, automated tests don't even check that file to be a valid JSON.

webron:
@undera - thanks for pointing that out. Ideally the process is to update the YAML version and automatically convert it to JSON, which just not quite there yet.

@karenetheridge, @jdesrosiers - would you mind taking a look?

jdesrosiers:
Unfortunately, you caught us in the middle of a multi-step process of merging several PRs and updating the version for a new release #2718. This would have been caught in the next steps that include generating the JSON from the YAML and running tests to ensure everything still works. A better process would help, but people shouldn't be using this as the source of truth. They should be using what is published on the website instead: (repo: https://github.com/OAI/OpenAPI-Specification/tree/gh-pages/oas/3.1/schema, URI: https://spec.openapis.org/oas/3.1/schema/2021-05-20). These are the officially _released_ versions of the schema.

undera:
For the URLs @jdesrosiers have mentioned, especially URI variant, where would I find it mentioned? I  looked through https://spec.openapis.org/oas/v3.1.0 and there's no mention of official JSONSchema of OAS. The file name of `2021-05-20` missing `.json` suffix also causes some issues to view the spec and use it with other tools...

jdesrosiers:
> where would I find it mentioned?

The spec includes the schema's URI at https://spec.openapis.org/oas/v3.1.0#schema-object.

> The OpenAPI Schema Object dialect for this version of the specification is identified by the URI `https://spec.openapis.org/oas/3.1/dialect/base` (the “OAS dialect schema id”).

It's not made clear where to find that schema, but I think it's reasonable to assume that you can download the schema from that URI. The latest schema should be downloadable from that URI. We should definitely add that, but we also want people to be able to pin to a specific version if they need to using the URI that includes the release date. Thanks for bringing this up.

> The file name of `2021-05-20` missing `.json` suffix also causes some issues to view the spec and use it with other tools...

This isn't really provided for viewing. It's provided to be plugged into a JSON Schema validation library. If you want to view or modify the schema, you'd probably prefer the yaml version on the main branch. This is just the place where the finished product is deployed. It doesn't have a file extension because then the URI would have a file extension and that's bad form for URIs. It should respond with a `Content-Type: application/schema+json` header, but unfortunately GitHub Pages doesn't support that. You could always save the file you download with the `.json` suffix if that's important, but maybe I've missed your point.

undera:
Thanks for the clarification @jdesrosiers

PR #2696, opened 2021-09-02
Add "type" qualifier to ensure that "if" clauses fail properly

karenetheridge:
If the wrong type is used for a particular piece of data, then a bare
"required" will validate as true, causing the "if" clause to be true, which
can lead to confusing errors when the "then" schema then fires, instead of the
error occurring at a higher position in the schema.


jdesrosiers (approved):
Thanks!

darrelmiller (approved):

Issue #2703, opened 2021-09-08
Errata and/or Spec Clarification Required for Security Scheme Object (OAS 3.0 and OAS 3.1)


hornworm61:
(This applies to both 3.0 and 3.1, but I'm referencing 3.1 sections)

Section 4.8.27 defines the "scheme" field of the Security Scheme Object as "The name of the HTTP Authorization scheme to be used in the Authorization header as defined in [RFC7235]. The values used SHOULD be registered in the IANA Authentication Scheme registry."

It isn't clear whether this field is case-insensitive. RFC7235 implies that in the context of the Authorization header, the value is case insensitive, but it is confusing whether it is also case insensitive here. The IANA Authentication Scheme registry defines several values, including "Basic" and "Bearer" with an initial upper case character in each case. The examples in the OAS use "basic" and "bearer", which does imply that is should be case-insensitive in this context as well.

But confusing matters further in the OAS for Security Scheme Objects is the definition of "bearerFormat", which implies that it is valid only if "scheme" is "bearer". The official JSON Schema for OpenAPI files only allows "bearerFormat" if the scheme is literally "bearer". Should it not also be valid is the scheme is "Bearer" (or any other combination of upper and lower case characters)?

Specifically, the following OpenApi fragment does not pass schema validation, but I believe the specification implies that it should be OK:

```
    "securitySchemes": {
      "global": {
        "type": "http",
        "description": "A bearer token granting access to the operation is required.",
        "scheme": "Bearer",
        "bearerFormat": "JWE"
      }
    }
```

Some words of clarification in the specification would be helpful, as well as a confirmation that the JSON Schema is either correct or is in error.

For reference, the JSON Schema that fails this is:

```
    "HTTPSecurityScheme": {
      "type": "object",
      "required": [
        "scheme",
        "type"
      ],
      "properties": {
        "scheme": {
          "type": "string"
        },
        "bearerFormat": {
          "type": "string"
        },
        "description": {
          "type": "string"
        },
        "type": {
          "type": "string",
          "enum": [
            "http"
          ]
        }
      },
      "patternProperties": {
        "^x-": {
        }
      },
      "additionalProperties": false,
      "oneOf": [
        {
          "description": "Bearer",
          "properties": {
            "scheme": {
              "enum": [
                "bearer"
              ]
            }
          }
        },
        {
          "description": "Non Bearer",
          "not": {
            "required": [
              "bearerFormat"
            ]
          },
          "properties": {
[...]

PR #2706, opened 2021-09-09
Treat security scheme as case insensitive in JSON schema [Issue #2703]

hornworm61:
This change updates the JSON Schemas for both 3.0 and 3.1 to allow for the security scheme to be case insensitive when checking for whether the bearerFormat propery is allowed.

### DIFF
--- a/schemas/v3.0/schema.yaml
+++ b/schemas/v3.0/schema.yaml
@@ -806,7 +806,8 @@ definitions:
       - description: Bearer
         properties:
           scheme:
-            enum: [bearer]
+            type: string
+            pattern: ^[Bb][Ee][Aa][Rr][Ee][Rr]$
 
       - description: Non Bearer
         not:
@@ -814,7 +815,8 @@ definitions:
         properties:
           scheme:
             not:
-              enum: [bearer]
+              type: string
+              pattern: ^[Bb][Ee][Aa][Rr][Ee][Rr]$
 
   OAuth2SecurityScheme:
     type: object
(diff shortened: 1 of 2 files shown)


jdesrosiers (approved):

darrelmiller (approved):

jdesrosiers:
@hornworm61 I'm doing final checks before releasing the new version of the schema and I noticed that you removed `required: [scheme]` from the http type. This doesn't look correct. Did you remove this on purpose?

webron:
@jdesrosiers doesn't https://github.com/hornworm61/OpenAPI-Specification/blob/638d7eb06b65f1171e1f5dc196b0f1ef78b80c49/schemas/v3.1/schema.yaml#L770-L772 cover it?

jdesrosiers:
@webron No, that `required` is in the `if` schema. It's only used to determine whether or not to apply the `then` schema.

hornworm61:
I think the `type-http-bearer` subschema originally said: "If the object has a property 'type' of 'http' and a 'scheme' of 'bearer' then it is allowed to have a property of bearerFormat and must have a property 'scheme'

But if the object passes the `if` part of this, then it definitely has a property 'scheme', so the "required" portion of the `then` branch is redundant because it can never fail.

Also, the `type-http` subschema already says that if the object has a property 'type' of 'http' then it must have a property of 'scheme' (and `type-http` and `type-http-bearer` are used in an "all-of" construct).

jdesrosiers:
> the `type-http` subschema already says that if the object has a property 'type' of 'http' then it must have a property of 'scheme' 

Ahh, yes, you're right. That's what I was missing.

PR #2746, opened 2021-10-13
more schema fixes

karenetheridge:
schema fixes for the "parameter" and "header" objects
    
- name is required (for parameter)
- the map under content must contain only one entry
    
as per https://spec.openapis.org/oas/v3.1.0#fixed-fields-9

### DIFF
--- a/schemas/v3.1/schema.json
+++ b/schemas/v3.1/schema.json
@@ -437,10 +437,13 @@
           "$dynamicRef": "#meta"
         },
         "content": {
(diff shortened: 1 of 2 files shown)


jdesrosiers:
Spec Reference: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#parameterObject

In the Parameter Object, the "name" property should be required and the "content" property object map should have only one entry.

darrelmiller (approved):

PR #2748, opened 2021-10-14
Update api-with-examples.yaml

somethingInMe:


darrelmiller:
This PR appears to update our V3 example with a V2 description which is not appropriate.  I will close this and if you want to reopen with fixes, please feel free to.

PR #2553, opened 2021-04-26
Add Schemathesis to IMPLEMENTATIONS.md

Stranger6667:
As a replacement for #2206

PR #2787, opened 2021-11-11
IMPLEMENTATIONS.md: Restore table header format

waldyrious:
Following the standardization of the table headers' formatting in #2683, a slight deviation was reintroduced in #2553.
This PR re-normalizes the headers so they're consistent again.


darrelmiller (approved):

darrelmiller:
Thanks for the fix!

PR #2791, opened 2021-11-15
Complete the example of discriminator mapping

johnlinp:
Before this fix, the example contains types of Cat and Lizard, but the
discriminator mapping for them is missing.

### DIFF
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -2783,7 +2783,9 @@ components:
       discriminator:
         propertyName: petType
         mapping:
+          cat: Cat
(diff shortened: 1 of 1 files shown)


Thanayaby:
Stay

darrelmiller:
@johnlinp Thanks for bringing this up.  The mapping for cat and lizard are not included intentionally to demonstrate that a complete set of mappings is not necessary.  Mappings are there for when the discriminator property value does not map to a schema name.

johnlinp:
I see. Thank you for the explanation.

PR #2793, opened 2021-11-17
Change server-variable property to 'description'

ahoglund:
According to the spec: https://spec.openapis.org/oas/v3.1.0#server-variable-object

This should validate a `description` field, not `descriptions`.


darrelmiller (approved):

darrelmiller:
@ahoglund  Thanks for catching this and creating the PR.  Much appreciated.

Issue #2796, opened 2021-11-18
Broken markdown link referencing "media type range"


saibotsivad:
On the OpenAPI specs page for [Request Body Object](https://spec.openapis.org/oas/latest.html#request-body-object) what is loaded in my browser looks like this (note the highlighted text):

<img width="812" alt="image" src="https://user-images.githubusercontent.com/91254/142489647-c065dd57-262f-4960-a861-2227344eb399.png">

I believe that comes from [/versions/3.1.0.md](https://raw.githubusercontent.com/OAI/OpenAPI-Specification/a114313cf1558877bce771bdad7d3b095a4f48f5/versions/3.1.0.md) but the Markdown looks correct to me:

```markdown
The key is a media type or [media type range](https://tools.ietf.org/html/rfc7231#appendix-D) and the value describes it.
```

I tried copy+pasting the surrounding section over to the [markdown-it online demo](https://markdown-it.github.io/) to see if maybe the table syntax was buggy, but it renders correctly there.

Is there a possibility that the published version (at `spec.openapis.org/oas/latest.html`) is not actually generated from the latest markdown file?

I'm not sure what else to do to be helpful, sorry!

webron:
Thanks for bringing this up. My guess is that this is a rendering issue related to anchored URL. We'll need to take a look at the generation process in order to solve it.

saibotsivad:
Yeah, that must be it. I thought it was only that one place, but now I see it all over the docs, e.g. in the Operation Object `requestBody` text:

<img width="804" alt="image" src="https://user-images.githubusercontent.com/91254/142686867-a024c7c7-6ae9-4440-a4c7-a145e1fcf868.png">

which has this markdown text:

```markdown
... (such as [GET](https://tools.ietf.org/html/rfc7231#section-4.3.1), ...
```

saibotsivad:
I believe the problem is to do with this regex-cleanup stuff here: https://github.com/OAI/OpenAPI-Specification/blob/main/scripts/md2html/md2html.js#L262

@webron is this something you'd appreciate a pull request for? If so, from/into what branch?

webron:
We always welcome PRs, and in this case against `main`. Not sure how to test it though, but will be happy to give it a shot.

karenetheridge:
[...]

PR #2798, opened 2021-11-19
fix: some broken IETF links

saibotsivad:
From #2796 some IETF links were correct in the markdown syntax, but the generated HTML had broken links.

For example, the markdown for the [Request Body Object > Fixed Fields (4.8.13.1)](https://spec.openapis.org/oas/latest.html#fixed-fields-10) looks like this:

```markdown
The key is a media type or [media type range](https://tools.ietf.org/html/rfc7231#appendix-D) and
```

But the rendered HTML looks like this:

```html
The key is a media type or [media type range]appendix-D) and
```

The problem occurs due to the markdown pre-processor code used to adjust RFC links.

It looks like the original intent of lines 267-269 was to transform links that looked like this:

```markdown
[RFC2119](https://tools.ietf.org/html/rfc2119)
```

Into this format:

```markdown
[[!RFC2119]]
```

I'm not sure what the reasoning was for that, but if I simply *delete* lines 267-269, it does resolve the issue.

To be thorough, I inspected the output for instances of lines containing `https://tools.ietf.org/html/rfc` at the end, aka what lines 267-269 *would have* operated on, and it's all links where the text is not `RFC\d+`. Specifically:

* In versions 3.0.0-3.1.0 the "Media Type Object" has two instances of `<a href="https://tools.ietf.org/html/rfc7231#appendix-D">media type range</a>`
* In versions 3.0.0-3.1.0 the "Runtime Expressions" section has `<a href="https://tools.ietf.org/html/rfc5234">ABNF</a>`
* In version 3.0.0 only, in the "Operation Object" section under "Fixed Fields", the `requestBodies` section has one for each method, e.g. `<a href="https://tools.ietf.org/html/rfc7231#section-4.3.1">GET</a>`

Since that's the only instances affected by removing lines 267-269, this change seems very safe.

### DIFF
--- a/scripts/md2html/md2html.js
+++ b/scripts/md2html/md2html.js
@@ -263,10 +263,6 @@ for (let l in lines) {
             //return (group1 ? group1 : '')+']]';
             return ']]';
         });
-
-        while (line.indexOf('https://tools.ietf.org/html/rfc')>=0) {
-            line = line.replace(/.https:..tools.ietf.org.html.rfc[0-9]{1,5}./g,'');
-        }
     }
 
     // minor fixup to get bibliography link to work


saibotsivad:
@webron any progress on this? Just wanting to make sure it didn't fall through the cracks. 😅

webron (approved):

webron:
Thanks, @saibotsivad. It looks pretty straight forward but would like someone else from the TSC to look into it before the merging. I added it to this week's [meeting agenda](#2821). I can't attend to it this week, but hopefully they'll be able to review it. If you can join the meeting, it could help with the review process but not required.

saibotsivad:
Neat! I'll see if I can make it to the meeting. 👍

darrelmiller (approved):

PR #2801, opened 2021-11-26
The value field and externalValue field are mutually exclusive.

karenetheridge:
https://spec.openapis.org/oas/v3.1.0#example-object

### DIFF
--- a/schemas/v3.1/schema.yaml
+++ b/schemas/v3.1/schema.yaml
@@ -607,6 +607,10 @@ $defs:
       externalValue:
         type: string
         format: uri
+    not:
+      required:
+      - value
+      - externalValue
     $ref: '#/$defs/specification-extensions'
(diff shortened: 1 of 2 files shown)


schemas/v3.1/schema.yaml
```diff
@@ -607,6 +607,12 @@ $defs:
       externalValue:
         type: string
         format: uri
+    not:
+      allOf:
+      - required:
+        - value
+      - required:
+        - externalValue
```
hkosova, line 615: Can this be simplified like so?
```suggestion
      required:
        - value
        - externalValue
```

karenetheridge: yes. good call!

jdesrosiers (approved):

darrelmiller (approved):

PR #2802, opened 2021-11-26
add description to all schemas

karenetheridge:
in v3.1 the number of schemas increased considerably, which can be a bit confusing.

### DIFF
--- a/schemas/v3.0/schema.yaml
+++ b/schemas/v3.0/schema.yaml
@@ -1,6 +1,6 @@
 id: https://spec.openapis.org/oas/3.0/schema/2021-09-28
 $schema: http://json-schema.org/draft-04/schema#
-description: Validation schema for OpenAPI Specification 3.0.X.
+description: The description of OpenAPI v3.0.x documents, as defined by https://spec.openapis.org/oas/v3.0.3
 type: object
 required:
   - openapi
--- a/schemas/v3.1/dialect/base.schema.yaml
+++ b/schemas/v3.1/dialect/base.schema.yaml
@@ -1,4 +1,6 @@
 ---
+title: OpenAPI 3.1 Schema Object Dialect
+description: A JSON Schema dialect describing schemas found in OpenAPI documents
 $dynamicAnchor: meta
 $id: https://spec.openapis.org/oas/3.1/dialect/base
 $schema: https://json-schema.org/draft/2020-12/schema
@@ -14,4 +16,3 @@ $vocabulary:
 allOf:
 - $ref: https://json-schema.org/draft/2020-12/schema
 - $ref: https://spec.openapis.org/oas/3.1/meta/base
-title: OpenAPI 3.1 Schema Object Dialect
--- a/schemas/v3.1/meta/base.schema.yaml
+++ b/schemas/v3.1/meta/base.schema.yaml
@@ -1,4 +1,6 @@
 ---
+title: OAS Base vocabulary
+description: A JSON Schema Vocabulary used in the OpenAPI Schema Dialect
 $defs:
   discriminator:
     $ref: '#/$defs/extensible'
@@ -57,7 +59,6 @@ properties:
     $ref: '#/$defs/external-docs'
   xml:
     $ref: '#/$defs/xml'
-title: OAS Base vocabulary
 type:
 - object
 - boolean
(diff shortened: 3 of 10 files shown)


schemas/v3.1/dialect/base.schema.yaml
```diff
@@ -1,4 +1,5 @@
 ---
+description: a dialect describing schemas found in OpenAPI documents
```
jdesrosiers, line 2: I suggest being a little more specific and calling this a "JSON Schema dialect" instead of just "dialect". Not everyone is going to know what a dialect is, but at least this way it's clear that it's a JSON Schema concept. I suggest doing the same for the use of "vocabulary" in the vocabulary schema: "JSON Schema vocabulary" instead of "vocabulary".

schemas/v3.1/meta/base.schema.yaml
```diff
@@ -1,4 +1,5 @@
 ---
+description: a vocabulary to use in schemas found in OpenAPI documents
```
jdesrosiers, line 2: It's not necessarily used by all schemas in an OpenAPI document. I suggest something more along the lines of, "A JSON Schema Vocabulary used in the OpenAPI Schema Dialect."

schemas/v3.1/schema-base.yaml
```diff
@@ -1,3 +1,4 @@
+description: specifies a dialect to use for schemas within OpenAPI documents
```
jdesrosiers, line 1: I think this can be more clear. Here's my suggestion.
"A schema for validating an OpenAPI 3.1 document where all Schema Objects use the OpenAPI Schema Dialect."

schemas/v3.0/schema.yaml
```diff
@@ -1,6 +1,6 @@
 id: https://spec.openapis.org/oas/3.0/schema/2021-09-28
 $schema: http://json-schema.org/draft-04/schema#
-description: Validation schema for OpenAPI Specification 3.0.X.
+description: description of OpenAPI v3.0.x documents, as defined by https://spec.openapis.org/oas/v3.0.3
```
jdesrosiers, line 3: Just as a style thing, I usually prefer descriptions that are full sentences including starting with a capital letter. I don't feel to strongly about it if no one cares, but I wanted to make the suggestion anyway.

jdesrosiers (commented):
I think this is a great idea. I have a couple suggestions that I hope will make some of the descriptions even more helpful.

karenetheridge:
ok, this might be a bit better now.

schemas/v3.1/schema-base.yaml
```diff
@@ -1,3 +1,4 @@
+description: A dialect to use for schemas within OpenAPI documents
```
jdesrosiers, line 1: This isn't a dialect schema. It describes an OpenAPI document where the schemas all use the OpenAPI Schema dialect.

jdesrosiers (changes requested):
Looks good, but one description is still not quite right.

schemas/v3.1/schema-base.yaml
karenetheridge: pushed a revision.

schemas/v3.0/schema.yaml
karenetheridge: pushed a revision.

darrelmiller (approved):

jdesrosiers (changes requested):
One of the descriptions is not quite right. In case it's still unclear what that schema is for, there's a better description here, https://json-schema.org/blog/posts/validating-openapi-and-json-schema

karenetheridge:
done.

schemas/v3.1/schema-base.json
```diff
@@ -1,4 +1,5 @@
 {
+  "description": "A dialect to use for JSON Schemas within OpenAPI documents",
```
jdesrosiers, line 2:
```suggestion
  "description": "The description of OpenAPI v3.1.x documents using the OpenAPI JSON Schema dialect, as defined by https://spec.openapis.org/oas/v3.1.0",
```

schemas/v3.1/schema.json
```diff
@@ -1,4 +1,5 @@
 {
+  "description": "The description of OpenAPI v3.1.x documents, as defined by https://spec.openapis.org/oas/v3.1.0",
```
jdesrosiers, line 2:
```suggestion
  "description": "The description of OpenAPI v3.1.x documents without schema validation, as defined by https://spec.openapis.org/oas/v3.1.0",
```

jdesrosiers (changes requested):
The JSON versions need to be updated too.

jdesrosiers (approved):

webron:
Thanks for all the effort, @karenetheridge, and @jdesrosiers for the additional reviews!

PR #2831, opened 2021-12-19
fix a markdown error

karenetheridge:


webron:
Thanks for the fix. This needs to be submitted against the `v3.1.1-dev` branch as we don't make changes in published versions.

webron (approved):

darrelmiller (approved):

karenetheridge:
> This needs to be submitted against the v3.1.1-dev branch as we don't make changes in published versions.

..but the published specifications were just changed 
in PRs https://github.com/OAI/OpenAPI-Specification/pull/2798
and https://github.com/OAI/OpenAPI-Specification/pull/2836 -- so this markup fix should be valid there too?

webron:
@karenetheridge the source of truth of the file is the one found in the `main` branch. Once it reaches that branch, it is published and no longer touched (active work is done on `*-dev` branches). The PRs you referenced only affect the rendered version of the spec that appears at spec.openapis.org. That's just a reflection of the one in the `main` branch, but has the same contents.

PR #2842, opened 2022-01-04
Add missing type: object field in example schema

adriangb:
I think this is missing

### DIFF
--- a/versions/3.0.3.md
+++ b/versions/3.0.3.md
@@ -943,6 +943,7 @@ requestBody:
   content:
     'application/x-www-form-urlencoded':
       schema:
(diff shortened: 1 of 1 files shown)


hkosova:
This was fixed in the 3.1.0 spec (PR #2361).

adriangb:
I take it then that typo fixes like this don't get "backported"? If that's the case we can just close this PR

hkosova:
> I take it then that typo fixes like this don't get "backported"?

Yes. Published versions of the spec are not amended.

adriangb:
Gotcha. Closing this then.

Issue #1706, opened 2018-10-11
Support deep objects for query parameters with deepObject style


bajtos:
**Background**

Many applications expect deeply nested objects in input parameters, see the discussion in swagger-ui starting from this comment: https://github.com/swagger-api/swagger-ui/issues/4064#issuecomment-357417513 In [LoopBack](https://loopback.io), we are running into this problem too, see https://github.com/strongloop/loopback-next/pull/1679.

Consider a `filter` parameter defined as follows:

```yaml
parameters:
 filterParam:
   in: query
   name: filter
   schema:
     type: object
   style: deepObject
   explode: true
   description: Options for filtering the results
   required: false
```

Let's say the user wants to provide the following value, for example by entering a JSON into the text area rendered by swagger-ui:

```json
{
  "name": "ivan",
  "birth-date": {
    "gte": "1970-01-01"
  }
}
```

At the moment, the OpenAPI Specification v3 does not describe how to encode such value in a query string. As a result, OpenAPI clients like swagger-ui don't know how to handle this input - see the discussion in https://github.com/swagger-api/swagger-js/issues/1385 for an example.

**Proposal**

The following query string should be created by swagger-js client for the input value shown above.

```
filter[name]=ivan&filter[birth-date][qte]=1970-01-01
```

The proposed serialization style is supported by https://www.npmjs.com/package/qs, which is used by http://expressjs.com as the default query parser, which means that a vast number of Node.js API servers are already expecting this serialization style.

I am not sure if there is any existing formal specification of this format, I am happy to look into that once my proposal gets considered as acceptable in principle.

**Additional information**

Existing issues that are touching a similar topic:
 - style: form + object looses the parameter name? OAI/OpenAPI-Specification#1006
 - Are indexes in the query parameter array representable? OAI/OpenAPI-Specification#1501

Two older comments from swagger-js that may be relevant:

https://github.com/swagger-api/swagger-js/pull/1140
> Limitations:
> deepObject does not handle nested objects. The specification and swagger.io documentation does not provide an example for serializing deep objects. Flat objects will be serialized into the deepObject style just fine.

https://github.com/swagger-api/swagger-js/pull/1140#issuecomment-333017825
> As for deepObject and nested objects - that was explicitly left out of the spec, and it's ok to just Not Support It™.

earth2marsh:
[...]

PR #2846, opened 2022-01-05
Clarify deepObject applicability

hkosova:
This PR changes the description of the `deepObject` serialization style to mention "simple non-nested objects" instead of "nested objects". I think "nested objects" is a typo because @webron [previously said](https://github.com/swagger-api/swagger-js/issues/1385#issuecomment-423768114) that the `deepObject` behavior is not actually defined for nested objects.

> when we defined `deepObject` in the spec, we explicitly chose to not mention what happens when the object has several levels in it, but in our conversations we went with 'not supported'. If you're looking for supporting that use case, it needs to be addressed by the spec first.

The "nested objects" phrase is confusing some tooling developers and users. I hope this change will help clarify things until #1706 is added to the spec.

### DIFF
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -1059,7 +1059,7 @@ form |  `primitive`, `array`, `object` |  `query`, `cookie` | Form style paramet
 simple | `array` | `path`, `header` | Simple style parameters defined by [RFC6570](https://tools.ietf.org/html/rfc6570#section-3.2.2).  This option replaces `collectionFormat` with a `csv` value from OpenAPI 2.0.
 spaceDelimited | `array`, `object` | `query` | Space separated array or object values. This option replaces `collectionFormat` equal to `ssv` from OpenAPI 2.0. 
 pipeDelimited | `array`, `object` | `query` | Pipe separated array or object values. This option replaces `collectionFormat` equal to `pipes` from OpenAPI 2.0.
-deepObject | `object` | `query` | Allows simple non-nested objects to be serialized using form parameters.
+deepObject | `object` | `query` | Allows objects with scalar properties to be serialized using form parameters. 
 
 
 ##### Style Examples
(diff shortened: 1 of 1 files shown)


darrelmiller (approved):

versions/3.1.1.md
```diff
@@ -1059,7 +1059,7 @@ form |  `primitive`, `array`, `object` |  `query`, `cookie` | Form style paramet
 simple | `array` | `path`, `header` | Simple style parameters defined by [RFC6570](https://tools.ietf.org/html/rfc6570#section-3.2.2).  This option replaces `collectionFormat` with a `csv` value from OpenAPI 2.0.
 spaceDelimited | `array`, `object` | `query` | Space separated array or object values. This option replaces `collectionFormat` equal to `ssv` from OpenAPI 2.0. 
 pipeDelimited | `array`, `object` | `query` | Pipe separated array or object values. This option replaces `collectionFormat` equal to `pipes` from OpenAPI 2.0.
-deepObject | `object` | `query` | Provides a simple way of rendering nested objects using form parameters.
+deepObject | `object` | `query` | Allows simple non-nested objects to be serialized using form parameters.
```
darrelmiller, line 1062:
```suggestion
deepObject | `object` | `query` | Allows objects with scalar properties to be serialized using form parameters. 
```

versions/3.1.1.md
```diff
@@ -1059,7 +1059,7 @@ form |  `primitive`, `array`, `object` |  `query`, `cookie` | Form style paramet
 simple | `array` | `path`, `header` | Simple style parameters defined by [RFC6570](https://tools.ietf.org/html/rfc6570#section-3.2.2).  This option replaces `collectionFormat` with a `csv` value from OpenAPI 2.0.
 spaceDelimited | `array`, `object` | `query` | Space separated array or object values. This option replaces `collectionFormat` equal to `ssv` from OpenAPI 2.0. 
 pipeDelimited | `array`, `object` | `query` | Pipe separated array or object values. This option replaces `collectionFormat` equal to `pipes` from OpenAPI 2.0.
-deepObject | `object` | `query` | Provides a simple way of rendering nested objects using form parameters.
+deepObject | `object` | `query` | Allows objects with scalar properties to be serialized using form parameters.
```
darrelmiller, line 1062:
```suggestion
deepObject | `object` | `query` | Allows objects with scalar properties to be represented using form parameters. The representation of array or object properties is not defined.
```

PR #2857, opened 2022-01-14
Added formats to the registry as defined by OpenAPI v3

darrelmiller:
The following types as defined in the OpenAPI v3 specification have been added to the format registry.  The OpenAPI 3.1 specification no-longer defines these formats as the goal is to decouple the registry of formats from specific specifications to enable easier re-use of formats.

Common Name | type | format | Comments
-- | -- | -- | --
integer | integer | int32 | signed 32 bits
long | integer | int64 | signed 64 bits
float | number | float |  
double | number | double |  
byte | string | byte | base64 encoded characters
binary | string | binary | any sequence of octets
date | string | date | As defined by full-date - [!RFC3339]
dateTime | string | date-time | As defined by date-time - [!RFC3339]

Password was not included as that is not a serialization format and I believe it is mixing concerns by using the `format` property.


karenetheridge:
shouldn't `dateTime` be `date-time, as in https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#data-types ?

darrelmiller:
@karenetheridge  Yes. Thanks for catching that.

ioggstream:
@darrelmiller where will be the new registry hosted? json-schema?

webron (approved):

DavidBiesack:
See https://spec.openapis.org/registry/format/

PR #2861, opened 2022-01-20
document the default for /servers

karenetheridge:
https://spec.openapis.org/oas/v3.1.0#fixed-fields


darrelmiller:
Hey Karen, thanks for the PR.  Can you help us understand what problem this addresses?

karenetheridge:
`default` is mostly for documentation purposes but some implementations may use it to fill in a value if not specified.

We use the `default` keyword in many other places in the schema where the specification indicates there is a default value to be assumed.

PR #2869, opened 2022-01-31
Added array type in openapi-specifications 3.1.0.md

soumyajit-samanta:


webron:
Thank you for the PR. Generally speaking, we do not accept PRs that change the published specs as they are immutable. There's a future version branch available for such changes.

That said, this change does not make sense. The changed table is meant to cover `format` and not `type`. Not entirely sure what you were trying to do, but feel free to start a new discussion to cover it.

soumyajit-samanta:
I have deleted the branch to tidy up. I was comparing swaggers documentation and open api documentation and saw that the types of data in type field mentioned in OAI does not have array type and the date thing so I thought maybe adding them will make it easy for other users.

But I think that can be mentioned somewhere in docs.

PR #2881, opened 2022-02-10
Websocket spec proposal

### LINKED ISSUE #55: Swagger for WebSocket services
Is there a way to use the Swagger Specification for WebSockets? It seems to be quite bound to HTTP.

bali182:
Proposal for Websocket spec, resolves #55

### DIFF
--- a/proposals/2022-02-08-Websockets.md
+++ b/proposals/2022-02-08-Websockets.md
@@ -1,0 +1,230 @@
+# WebSockets
+
+## Metadata
+
+| Tag                | Value                                                                                                                |
+| ------------------ | -------------------------------------------------------------------------------------------------------------------- |
+| Proposal           | [2022-02-08-Websockets](https://github.com/OAI/OpenAPI-Specification/tree/main/proposals/2022-02-08-Websockets.md) |
+| Authors            | [Balazs Edes](https://github.com/bali182)                                                                            |
+| Review Manager     | TBD                                                                                                                  |
+| Status             | Proposal                                                                                                             |
+| Implementations    | -                                                                                                                    |
+| Issues             | [#55](https://github.com/OAI/OpenAPI-Specification/issues/55)                                                        |
+| Previous Revisions | -                                                                                                                    |
+
+## Change Log
+
+| Date       | Responsible Party | Description         |
+| ---------- | ----------------- | ------------------- |
+| 2022-02-08 | Balazs Edes       | First draft created |
+
+## Introduction
+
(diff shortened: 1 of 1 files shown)


proposals/2022-02-08-Websockets.md
```diff
@@ -0,0 +1,230 @@
...
+### Extension of the current OpenAPI spec
+
+- Add a `WebSocket Payload Object` type (very similar to [Request Body Object](https://swagger.io/specification/#request-body-object), except the `required` field - which makes no sense here). It has the following fields:
+  - `description`: `string` - A brief description of the payload.
+  - `content`: `Map[string, `[Media Type Object](https://swagger.io/specification/#media-type-object)`]` The content of the payload.
+- Add an optional field in the [Components Object](https://swagger.io/specification/#components-object) type, so payloads can be reused:
+  - `webSocketPayloads`: `Map[string, WebSocket Payload Object]` - Reuseable payloads
+- Add 2 optional fields on the [Operation Object](https://swagger.io/specification/#operation-object) type:
+  - `publish`: [Reference Object](https://swagger.io/specification/#reference-object) | `WebSocket Payload Object` - describes the message(s) the client can publish/push to the server.
+  - `subscribe`: [Reference Object](https://swagger.io/specification/#reference-object) | `WebSocket Payload Object` - describes the message(s) the client can receive from the server.
```
brunogirin, line 88: If both fields can be optional, how can a tool that reads an OpenAPI document identify a web socket endpoint with certainty? If the OpenAPI document author did not supply either field, the endpoint is unrecognisable from a standard endpoint.

What should happen if those fields are added to an operation other than `get`? Is that possible?

bali182: 1. I'd say if either fields `publish` or `subscribe` is present, then it's a socket enabled path. If neither is supplied, it should fall back to be recognized as regular http path. Are you suggesting something more explicit (extra boolean field maybe)?

2. I'd say error. Similar situation, as to what should happen if a `GET` request has a `requestBody`. 

What do you think?

brunogirin: OK, that would work. If anyone needs more, they can always create a proposal for it!

brunogirin (approved):

Lcote71 (approved):

linghengqian (approved):

darrelmiller:
We appreciate the submission but as discussed in the issue this is out of scope for the OpenAPI specification.

PR #2891, opened 2022-03-04
Fixed a broken link to RFC 7231

melvinsh:


webron:
Thanks for the PR, @melvinsh! Unfortunately, we can't make any changes to published versions, even if it's fixing broken links. Since version 2.0 is very old, we also don't have a mechanism today to publish patch versions for it with such fixes (as opposed to later versions). It looks like the current link in the spec redirects to the one you put in the PR, so hopefully it keeps working.

PR #2901, opened 2022-03-23
Create sig-create-template.yml

namdeirf:
Based on initial conversations here is my initial SIG creation form, happy to incorporate edits. etc. This will help formalize our process as we move forward. I think the initial creation form should exist at the high-level Spec repo and each individual will get their own SIG. A separate Community Repo to house this and other aspects also makes sense. Happy to discuss going forward.


namdeirf:
Moving my PR from the main repo, to the Community Repo

PR #2956, opened 2022-06-26
Remove analytics/tracking pixel from README.md

Susheel4115:


earth2marsh:
This has been broken for a while, so removing the pixel seems like a good move. Times have changed, and now GitHub offers better analytics on repos, so this approach no longer makes sense. See: https://github.com/igrigorik/ga-beacon/commit/6acd8627bb7be36f24f5516e9873c92719a50e55 for more info.

earth2marsh (approved):
This has been broken for a while, so removing the pixel seems like a good move, and this commit LGTM. Times have changed, and now GitHub offers better analytics on repos, so this approach no longer makes sense. See: https://github.com/igrigorik/ga-beacon/commit/6acd8627bb7be36f24f5516e9873c92719a50e55 for more info.

Thanks for the PR!

PR #2964, opened 2022-07-02
Add Device Code authorization to the supported OAuth Flows

### LINKED ISSUE #2381: Add "Device Code" as a supported OAuth Flow

LikeLakers2:
Clone of PR #2401, resubmitted against the `v3.2.0-dev` branch [as requested by webron](https://github.com/OAI/OpenAPI-Specification/pull/2401#issuecomment-1170594516). Note that this is not submitted against the `main` branch, but I will happily resubmit it against the `main` branch upon request.

Description of the original PR follows, with some extraneous details edited out:

----

Closes #2381. 

This is a PR for the Device Authorization OAuth Flow, based on [RFC8628](https://tools.ietf.org/html/rfc8628). I suggested this over at #2381 and was asked to make a PR for it.

The URL used for this flow is named `deviceAuthorizationUrl`, separate from `authorizationUrl` as the two have completely different semantics and usages, so I was concerned that using `authorizationUrl` for both might be confusing. That said, I would be happy to merge this into `authorizationUrl` if requested.

### DIFF
--- a/versions/3.2.0.md
+++ b/versions/3.2.0.md
@@ -3188,7 +3188,7 @@ animals:
 
 Defines a security scheme that can be used by the operations.
 
-Supported schemes are HTTP authentication, an API key (either as a header, a cookie parameter or as a query parameter), mutual TLS (use of a client certificate), OAuth2's common flows (implicit, password, client credentials and authorization code) as defined in [RFC6749](https://tools.ietf.org/html/rfc6749), and [OpenID Connect Discovery](https://tools.ietf.org/html/draft-ietf-oauth-discovery-06).
+Supported schemes are HTTP authentication, an API key (either as a header, a cookie parameter or as a query parameter), mutual TLS (use of a client certificate), OAuth2's common flows (implicit, password, client credentials and authorization code) as defined in [RFC6749](https://tools.ietf.org/html/rfc6749), OAuth2 device authorization flow as defined in [RFC8628](https://tools.ietf.org/html/rfc8628), and [OpenID Connect Discovery](https://tools.ietf.org/html/draft-ietf-oauth-discovery-06).
 Please note that as of 2020, the implicit flow is about to be deprecated by [OAuth 2.0 Security Best Current Practice](https://tools.ietf.org/html/draft-ietf-oauth-security-topics). Recommended for most use case is Authorization Code Grant flow with PKCE.
 
 ##### Fixed Fields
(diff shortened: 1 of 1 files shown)


karenetheridge:
I see that schemas/v3.2 doesn't exist yet, but if this PR is accepted, some schema changes should be made as well.

LikeLakers2:
@karenetheridge Apologies for the late response, but are you suggesting I should create (and edit appropriately) the v3.2 schema folder? Or is this simply a note for whomever reviews this PR?

karenetheridge:
Yes, I think the v3.2 schema folder (with schemas) should be created in the v3.2.0-dev branch (by the maintainers), so you can merge your changes on top.

versions/3.2.0.md
```diff
@@ -3292,6 +3292,7 @@ Field Name | Type | Description
 <a name="oauthFlowsPassword"></a>password| [OAuth Flow Object](#oauthFlowObject) | Configuration for the OAuth Resource Owner Password flow 
 <a name="oauthFlowsClientCredentials"></a>clientCredentials| [OAuth Flow Object](#oauthFlowObject) | Configuration for the OAuth Client Credentials flow.  Previously called `application` in OpenAPI 2.0.
 <a name="oauthFlowsAuthorizationCode"></a>authorizationCode| [OAuth Flow Object](#oauthFlowObject) | Configuration for the OAuth Authorization Code flow.  Previously called `accessCode` in OpenAPI 2.0.
+<a name="oauthFlowsDeviceAuthorization"></s>deviceAuthorization| [Oauth Flow Object](#oauthFlowObject) | Configuration for the OAuth Device Authorization flow.
```
ericallam, line 3295: Small typo, should be `OAuth Flow Object` not `Oauth Flow Object`

ericallam (commented):
Checked this change against [https://datatracker.ietf.org/doc/html/rfc8628](rfc8628) and also a real world usage on [GitHub REST Device Flow](https://docs.github.com/en/developers/apps/building-oauth-apps/authorizing-oauth-apps#device-flow) and it all looks good.

versions/3.2.0.md
LikeLakers2: I just want to say that I read this a few days ago, and I mean to fix it, but it keeps slipping my mind. >_< I am so sorry.

LikeLakers2: Fixed. I apologize for the delay.

MikeRalphson:
@karenetheridge as the markdown is the source of truth and the schema a (mere) artifact (though a useful one), it's normal that the creation / update of the schema can lag the creation of a new dev branch and initial changes.

MikeRalphson (approved):
LGTM - thanks.

PR #2991, opened 2022-08-02
v3.1 license schema url and identifier mutually exclusive

### LINKED ISSUE #2975: The mutually-exclusive url and identifier properties of License object are optional in specification and required in schema
In the definition of the [License object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#licenseObject) of OpenAPI 3.1: the `url` and `identifier` properties are not marked as REQUIRED, so optional, and are defined as mutually exclusive. 

The [schema of the License object](https://github.com/OAI/OpenAPI-Specification/blob/aa91a19c43f8a12c02efa42d64794e396473f3b1/schemas/v3.1/schema.json#L133) does not match with the specification making `url` or `identifier` required using the following construction at the root of of the object definition:

```json
"oneOf": [
  {"required": ["identifier"] },
  {"required": ["url"]}
]
```
If we compare both behaviors, that means the following (note that `name` is required):

<table>
<tr><th>Specification</th><th>Schema</th></tr>
<tr>
<td>
<ul>
  <li>We can have
    <ul>
       <li><code>name</code> alone</li>
       <li><code>name</code> and <code>url</code></li>
      <li><code>name</code> and <code>identifier</code></li>
    </ul>
  </li>
  <li>But we can't have 
   <ul>
      <li><code>name</code>, <code>url</code> and <code>identifier</code></li>
   </ul></li>
</td>
<td>
<li>We can have
    <ul>
       <li><code>name</code> alone</li>
       <li><code>name</code> and <code>url</code></li>
      <li><code>name</code> and <code>identifier</code></li>
    </ul>
  </li>
  <li>But we can't have 
   <ul>
      <li><code>name</code> alone</li>
      <li><code>name</code>, <code>url</code> and <code>identifier</code></li>
   </ul></li>
</td>
<tr>
</table>

The following document which is valid according to the specification is considered invalid when using the schema with the validation script or directly using it with ajv:

```yaml
openapi: 3.1.0
info:
  title: Dummy Bookshop
  version: '1.0'
  license:
    name: Apache 2.0
paths: {}
```

Before proposing a pull request to fix that issue, I would like to confirm that my interpretation of the specification is correct and to know if there are any recommendations to follow regarding when modifying the schema. Besides sticking to the JSON Schema version already used, there are maybe some features to not use to ensure compatibility? (I already found 4 different ways to say "url and identifier are optional and mutually exclusive", I'll put them in the discussion if fixing the schema is the solution).

arno-di-loreto:
Fixes #2975

### DIFF
--- a/schemas/v3.1/schema.yaml
+++ b/schemas/v3.1/schema.yaml
@@ -105,11 +105,11 @@ $defs:
         format: uri
     required:
       - name
-    oneOf:
-      - required:
-        - identifier
-      - required:
-        - url
+    dependentSchemas:
+      identifier:
+        not:
+          required:
+            - url
     $ref: '#/$defs/specification-extensions'
     unevaluatedProperties: false
 


schemas/v3.1/schema.yaml
```diff
@@ -105,11 +105,11 @@ $defs:
...
-    oneOf:
-      - required:
-        - identifier
-      - required:
-        - url
+    dependentSchemas:
+      identifier:
+        not:
+          required:
+            - url
```
karenetheridge, line 112: Doesn't it go the other way as well? if `url` exists, then we can't have `identifier`.

```suggestion
    dependentSchemas:
      identifier:
        not:
          required:
            - url
      url:
        not:
          required:
            - identifier            
```

karenetheridge (changes requested):

schemas/v3.1/schema.yaml
karenetheridge: This could also be done more succinctly with just `maxProperties: 2`, since `name` is required and x- properties aren't permitted here.

arno-di-loreto: It's not necessary. What you propose was one of my initial options, but @jdesrosiers proposed that shorter variation:
> I didn't include the inverse because it isn't necessary and produces unnecessary error results.  

See [discussion here](https://github.com/OAI/OpenAPI-Specification/issues/2975#issuecomment-1192824618)

karenetheridge (approved):

darrelmiller:
@jdesrosiers This is only updating the yaml file.  Shouldn't the json file get the update too?

jdesrosiers:
@darrelmiller It's fine. The README instructs that contributors should only update the yaml version and that the json version will be updated as part of the publishing process. @arno-di-loreto is one of the very few that actually followed those instructions.

https://github.com/OAI/OpenAPI-Specification/tree/main/schemas/v3.1#contributing

darrelmiller (approved):

PR #2998, opened 2022-08-11
per the spec, the only allowed ranges are [12345]XX

karenetheridge:
For example, 50X is NOT permitted.

This was fixed in v3.1 in PR#2690 but the change did not make it to v3.2.0.

### DIFF
--- a/schemas/v3.1/schema.json
+++ b/schemas/v3.1/schema.json
@@ -774,7 +774,7 @@
         }
       },
       "patternProperties": {
-        "^[1-5][0-9X]{2}$": {
+        "^[1-5](?:[0-9][0-9]|XX)$": {
           "$ref": "#/$defs/response-or-reference"
         }
       },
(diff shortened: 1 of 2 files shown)


jdesrosiers (approved):

schemas/v3.1/schema.json
```diff
@@ -774,7 +774,7 @@
         }
       },
       "patternProperties": {
-        "^[1-5][0-9X]{2}$": {
+        "^[1-5](?:[0-9]{2}|XX)$": {
```
kaznovac, line 777:
```suggestion
        "^[1-5](?:[0-9][0-9]|XX)$": {
```

this might be a bit more readable IMHO

kaznovac (approved):

schemas/v3.1/schema.json
karenetheridge: ok, taken.

darrelmiller (approved):

PR #3017, opened 2022-09-04
Update schemas for v1.2 to draft-07

HoldYourWaffle:
Title pretty much says it all :)

This allows differentiating between the `uri` and `uri-reference` format. Draft-04 was [ambiguous](https://json-schema.org/understanding-json-schema/reference/string.html#resource-identifiers) in that regard, causing false-positive validation failures with some validator implementations.


darrelmiller:
From our TDC call the sentiment is that this change could cause consumers to break.  We would rather not make this change.

PR #3088, opened 2022-12-02
chore: add openapi to the example spec names

bufferoverflow:
A while ago I've implemented a OpenAPI viewer within GitLab, see https://docs.gitlab.com/ee/user/project/repository/#openapi-viewer or https://gitlab.com/gitlab-org/gitlab/-/issues/19515.
As parsing each `yaml` or `json` file to figure out if it is a OpenAPI spec is not feasible we agreed on using a name based pattern to render the OpenAPI specs, see e.g., https://gitlab.com/gitlab-org/gitlab/-/blob/master/doc/api/openapi/openapi.yaml

To improve testing of that feature using your examples, I thought let's try to contribute upstream instead of having a local copy with some example specs to test.


darrelmiller:
Hey @bufferoverflow , thank you for this submission.  We discussed this issue in the March 9th TDC call.  There was much enthusiasm about this new capability in GitLab.  However, we do not want to constrain the authors of OpenAPI descriptions to specific filenames or filename patterns.  Users should be in control of their own naming conventions.  The limitations of naming conventions can be seen by that fact that the majority of users name their OpenAPI files with a .yaml or .json extension because if they were to use a `.oas` extension most tooling wouldn't recognize it as yaml or json.  Unfortunately the only practical solution when reading from filesystems is for tooling to sniff the content.

bufferoverflow:
Thanks for discussing this within the maintainer crew! No problem at all with your decision, the maintainer should decide on this, feel free to close the PR.

Of course, I would be ready to create PRs for the examples if needed, but I'm fine with your decision.

btw. I've seen a similar practice with [draw.io](https://draw.io) , people are saving `.png` files with embedded data as `my-fancy-drawing.drawio.png` and the tooling such as vscode or drawio will handle that smooth.

/cc @rhofer

rhofer:
@bufferoverflow thanks for notifying me.

I wouldn't necessarily change the openapi file name specification. You've brought up the similarity to draw.io .  There the original definition comes from diagrams.net who creates draw.io. Their file ending for the original tool is `*.drawio`.

In the example for visual studio code extension [Draw.io Integration by Henning Dietrichs](https://marketplace.visualstudio.com/items?itemName=hediet.vscode-drawio), he took the decision by wrapping draw.io into a VSC extension to allow automatic detection of compatible files by according file endings. This means if somebody decides to embed draw.io raw as meta data into a `*.png` file and is willing to name it `*.drawio.png`, then the extension will detect it automatically and opens draw.io directly to allow editing it, instead of only viewing a png.

Hence this support could be delegated to gitlab in accordance to the above similarity. I.e. if somebody is willing to name his file e.g. `*.openapi.yml` or `*.openapi.json` then gitlab will render it automatically. This should then go into gitlab's user documentation - a feature you can use for convenience but you don't have to. (maybe not only for viewing, may it even opens an editor then).

This would then be still compatible to any other tool / consumer/ whatever, which is only looking at the "top-level" file ending as `*.yml` or `*.json`.

On the other hand, if openapi would define and integrate the content detection to a file ending specification (as e.g. `*.openapi.yml`) then any tool could rely on it, not only what a specific tool provides as support locally (standardization over local decision).

Potentially this could also be a "recommendation" instead of "mandatory specification" in sense of: openapi recommends a file ending pattern for easily identifying a yaml or json content as an openapi content.

Anyhow, I would prefer "standardization over local decision" in this case.

These are just my 2 cents on this topic, what may could flow into decision making of @darrelmiller and team.

rhofer:
Maybe I need to precise my preference for "standardization over local decision" in this case:

draw.io native has `*.drawio` file ending which is unique. `*.png`, `*.svg`, et.al are only considered as an export. Still you can open again this "export" in draw.io (what is an explicit action) and when it detects draw.io raw in the meta data section it makes it editable.

VSC cannot decide on this just based on top-level file ending as `*.png`, whether this needs to be handed over to an Extension. This brings in the local need to register special file endings in VSC as `*.drawio.png`.

Furthermore, on the file ending e.g. `*.png` there is an ambiguity in sense of how it could be opened. And it is even allowed to edit it in gimp, what will then diverge from the draw.io raw section in the meta data.

I do not see this variability for openapi files. Irrespective whether it is a `*.yaml` or a `*.json` it remains an openapi spec. ... then why not to mark it more clearly by adding a file name suffix as e.g. `*.openapi.yaml`?

Silvanoc:
@bufferoverflow contribution is only about the filenames of the examples, right? It looks to me as if @darrelmiller and @rhofer think that the proposal goes in the direction of global file naming rules, what is not the case. But @bufferoverflow seem to be fine with the comments. I'm puzzled... I might be on the wrong page too 😁

PR #3107, opened 2022-12-26
Add new extension namespace registry

darrelmiller:
In order to encourage the creation and registration of extensions in a central registry, this PR proposes creating a new namespace registry for extensions so that OpenAPI users can reserve an extension prefix for their organization to ensure uniqueness.


jeremyfiel:
where can we reserve our namespace?

darrelmiller:
@jeremyfiel As with all the registries, doing a PR to the gh-pages branch is all that is needed to add a an entry.

PR #3110, opened 2023-01-03
Refactor markdown specification to fix broken internal links

SethAngell:
Hello! While referencing the v3.1.0 spec, I noticed that internal links to sections didn't seem to be working (such as in the table of contents, clicking **Parameter Object** didn't navigate the page to the **Parameter Object** section. Upon closer examination, I saw that the links were pointing to `<a>` tags on each header which followed a naming camelCase naming convention. To the best of my knowledge, Markdown tends to generate snake-case anchor tags when rendered as html. This seems to be supported by the rendered markdown on github and the official OpenAPI website. 

### Problem

The table of contents and other internal links would try and reference page sections using a camelCase naming convention when instead they should be referenced by their snake-case representations. For example, the **Parameter Object** section, was being referenced with the link `#parameterObject` when said section was actually being rendered with the link `#parameter-object`.

### Solution

I refactored all internal links to use the snake-case naming convention that would be generated by markdown renderers. However, I didn't want to break any third party viewers which may be relying on the `<a>` tags within the headers, so I also updated those tags to now use a snake-case naming convention instead.


SethAngell:
Was cluelessly working against the release branch instead of `v3.1.1-dev`. Need to duplicate efforts there to avoid overwriting previous work

PR #3128, opened 2023-01-05
Internal link refactor to be more resilient across different markdown renderers

SethAngell:
Hello!

While working on an OpenAPI spec for the first time the other day, I noticed that internal links on github didn't seem to be working. Upon closer inspection, I noticed that the links were trying to reference an anchor tag within the header which used a camelCase convention as opposed to a snake-case one. I believe when Github renders markdown, it removes these explicit anchor tags and instead uses the traditional snake-case ones which are generated by markdown. 

For example, rather than the line `#### <a name="parameterObject"></a> Parameter Object` being rendered as `<h4 id="parameterObject">Parameter Object</h4>`, it was instead rendered as `<h4 id="parameter-object">Parameter Object</h4>`.

To rectify this, I replaced all of the old camelCase references with snake-case versions. However, I didn't want to disrupt any downstream workflows which utilized this `<a>` tag, so I also converted that value found there from camelCase to snake-case. I believe this should improve the viewing experience on Github while not breaking anything downstream. 

I'd love to hear opinions on this solution, or if it is even one y'all are interested in!

Thanks

### DIFF
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -16,69 +16,69 @@ An OpenAPI definition can then be used by documentation generation tools to disp
 <!-- TOC depthFrom:1 depthTo:3 withLinks:1 updateOnSave:1 orderedList:0 -->
 
 - [Definitions](#definitions)
-	- [OpenAPI Document](#oasDocument)
-	- [Path Templating](#pathTemplating)
-	- [Media Types](#mediaTypes)
-	- [HTTP Status Codes](#httpCodes)
+	- [OpenAPI Document](#openapi-document)
+	- [Path Templating](#path-templating)
+	- [Media Types](#media-types)
+	- [HTTP Status Codes](#http-status-codes)
 - [Specification](#specification)
 	- [Versions](#versions)
 	- [Format](#format)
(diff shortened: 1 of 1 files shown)


kexoth:
FYI today also needed the specification, whilst reading noticed that the section links are broken & updated them as well.
Noticed now that my PR is a duplicate of Yours.
https://github.com/OAI/OpenAPI-Specification/pull/3160

handrews:
@kexoth @SethAngell I did not see this when I made the comment just now on @kexoth 's PR.

If y'all could also target this fix to `versions/3.0.4.md` on the `v3.0.4-dev` branch (as it seems to apply there as well), that would be fantastic.  However you arrange it among yourselves :-)  Or perhaps wait and see if this approach gets approved and then apply it to 3.0.4.

SethAngell:
@handrews Apologies, just now seeing your comments. Best to keep this target at 3.1 or switch it to 3.0.4?

handrews:
The link fixes have been consolidated for all versions in PR #3408 (and follow-up PRs as noted in its description), so I am closing all other link fix PRs in favor of that one.

PR #3136, opened 2023-01-12
Fix anchors/links in markdown [3.1.0.md]

vksepm:
Broken link for `Path Item Object`

### DIFF
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -37,7 +37,7 @@ An OpenAPI definition can then be used by documentation generation tools to disp
 		- [Server Variable Object](#serverVariableObject)
 		- [Components Object](#componentsObject)
 		- [Paths Object](#pathsObject)
-		- [Path Item Object](#pathItemObject)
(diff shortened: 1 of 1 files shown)


ralfhandl:
Basically all links are now broken.

Problem is that GitHub changed the way they translate Markdown into HTML.

The OpenAPI spec uses this pattern to define explicit section anchors:

~~~md
#### <a name="pathItemObject"></a>Path Item Object
~~~

This worked fine until a few months ago, since then the anchor element is no longer copied into the output HTML when it is embedded within the headline.

A systematic fix would be to change *all* of these header lines with embedded anchors into anchors immediately preceding the header line:

~~~md
<a name="pathItemObject"></a>
#### Path Item Object
~~~

This style of anchors is still copied verbatim into the output HTML, and the links work fine.

I did this in two other GitHub repos for a bunch of Markdown files.

handrews:
The link fixes have been consolidated for all versions in PR #3408 (and follow-up PRs as noted in its description), so I am closing all other link fix PRs in favor of that one.

PR #3137, opened 2023-01-15
enforce existence of composite applicator keyword adjacent to "discriminator"

karenetheridge:
"The discriminator object is legal only when using one of the composite keywords oneOf, anyOf, allOf."
https://spec.openapis.org/oas/v3.1.0#discriminator-object

### DIFF
--- a/schemas/v3.1/meta/base.schema.json
+++ b/schemas/v3.1/meta/base.schema.json
@@ -18,6 +18,15 @@
         "externalDocs": { "$ref": "#/$defs/external-docs" },
         "xml": { "$ref": "#/$defs/xml" }
     },
+    "dependentSchemas": {
+      "discriminator": {
+        "anyOf": [
+          { "required": [ "oneOf" ] },
+          { "required": [ "anyOf" ] },
+          { "required": [ "allOf" ] }
+        ]
+      }
+    },
 
     "$defs": {
         "extensible": {
--- a/schemas/v3.1/meta/base.schema.yaml
+++ b/schemas/v3.1/meta/base.schema.yaml
@@ -20,6 +20,12 @@ properties:
     $ref: '#/$defs/external-docs'
   xml:
     $ref: '#/$defs/xml'
+dependentSchemas:
+  discriminator:
+    anyOf:
+    - required: [oneOf]
+    - required: [anyOf]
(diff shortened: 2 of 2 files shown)


schemas/v3.1/meta/base.schema.json
```diff
@@ -18,6 +18,16 @@
...
+    "if": {
+      "required": [ "discriminator" ]
+    },
+    "then": {
+      "anyOf": [
+        { "required": [ "oneOf" ] },
+        { "required": [ "anyOf" ] },
+        { "required": [ "allOf" ] }
+      ]
+    },
```
jdesrosiers, line 30: I suggest going with `dependentSchemas` for this. The end result is the same, but it's a little more concise.

```suggestion
    "dependentSchemas": {
      "discriminator": {
        "anyOf": [
          { "required": [ "oneOf" ] },
          { "required": [ "anyOf" ] },
          { "required": [ "allOf" ] }
        ]
      }
    },
```

jdesrosiers (approved):
I wish the spec said that `discriminator` should be ignored if there is no `anyOf`/`oneOf`/`allOf`. That would be more inline with JSON Schema conventions, but that's not what it says, so this change should be included.

schemas/v3.1/meta/base.schema.json
karenetheridge: > I wish the spec said that `discriminator` should be ignored if there is no `anyOf`/`oneOf`/`allOf`. That would be more inline with JSON Schema conventions, but that's not what it says, so this change should be included.

Indeed. This would be worthy of reconsidering for v3.2.

> I suggest going with `dependentSchemas` for this. The end result is the same, but it's a little more concise.

Heh, I considered `dependentRequired` but totally forgot about `dependentSchemas`! :p

jdesrosiers (approved):

darrelmiller (approved):

darrelmiller:
@OAI/tsc Can we get another set of eyes and a merge?

MikeRalphson:
@darrelmiller I'll reread up on dependentSchemas tomorrow unless someone else gets there first. I'm sure it's OK.

MikeRalphson (approved):
LGTM, thanks.

PR #3145, opened 2023-01-23
fix(petstore): Set to correct api url

xorima:
Petstore api is now listening at: https://petstore.swagger.io/

### DIFF
--- a/examples/v3.0/petstore-expanded.yaml
+++ b/examples/v3.0/petstore-expanded.yaml
@@ -12,7 +12,7 @@ info:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
 servers:
-  - url: http://petstore.swagger.io/api
+  - url: https://petstore.swagger.io
 paths:
   /pets:
     get:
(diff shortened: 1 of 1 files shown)


examples/v3.0/petstore-expanded.yaml
```diff
@@ -12,7 +12,7 @@ info:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
 servers:
-  - url: http://petstore.swagger.io/api
+  - url: https://petstore.swagger.io
```
MikeRalphson, line 15:
```suggestion
  - url: https://petstore.swagger.io/v2
```

I believe this is required too, though I'm not sure what endpoints if any have changed between v1 and v2 of the API. Ping @hkosova

hkosova: `petstore.swagger.io` API has and always had `/v2` only.

There's a newer version at https://petstore3.swagger.io/, but I don't recall if or how it's different (apart from using the OAS3 format).

ping @ponelat

hkosova (approved):

MikeRalphson (approved):
Thanks @hkosova !

Issue #3170, opened 2023-02-24
Encoding Object examples have invalid schema property


jeremyfiel:
Reviewing the [Encoding Object Examples](https://spec.openapis.org/oas/v3.0.3#special-considerations-for-multipart-content)

The keyword `type` should be `$ref`. 

```yaml
         addresses:
            # default Content-Type for arrays is based on the `inner` type (object shown, so `application/json` in this example)
            type: array
            items:
              type: '#/components/schemas/Address'
```

jeremyfiel:
I noticed [3.1.0](https://spec.openapis.org/oas/v3.1.0#special-considerations-for-multipart-content) has a slightly different schema, albeit, valid.  Lmk if you want that updated too. 

```yaml
          addresses:
            # default Content-Type for arrays is based on the _inner_ type (object shown, so `application/json` in this example)
            type: array
            items:
              type: object
              $ref: '#/components/schemas/Address'
```

handrews:
@jeremyfiel I'm not sure what would need to be fixed in the 3.1 schema?  I might be missing something.

jeremyfiel:
the `type: object` is redundant with a `$ref` sibling

handrews:
@jeremyfiel I suspect I put it in there just to emphasize that you can put things next to `$ref` now :-)  I don't much care one way or the other, but I think that's why it's like that.

jeremyfiel:
i'm impartial.

handrews:
Let's leave it as-is and just fix it for 3.0.4.

handrews:
This has been fixed in 3.0.4 and we decided not to change it in 3.1.1.

PR #3171, opened 2023-02-24
fix(docs): Update Encoding Object example

jeremyfiel:
closes https://github.com/OAI/OpenAPI-Specification/issues/3170


webron:
Thanks for the fix, @jeremyfiel. 
Unfortunately, we can't modify files that have already been published and can only modify minor versions.
It looks like we don't have a 3.0.4 branch but I can create one if you want.

jeremyfiel:
Hi Ron,

This is not a patch update, only a docs issue to explain the example
correctly. I don't believe that would require anew version, would it?

webron:
Unfortunately, it would. Anything under the `versions` directory cannot be updated once published, we can only add new versions.

jeremyfiel:
are there any other errata in the works?

webron:
I don't see anything for 3.0 (there is for 3.1), but happy to start one if needed.

jeremyfiel:
since it was updated in 3.1.x.  I guess there's no burning reason to create a new patch version. There is no behavior change.

handrews:
I have a couple of proposals for a quick 3.0.4.  @webron unless you'd rather I did not, I will tag the issue for this PR with the 3.0.4 label and also the other clarifications/bugfixes I've noticed recently.

webron:
I don't see a technical reason not to release 3.0.4. We'll potentially retire 3.0.X after the next minor/major version is released.

handrews:
@jeremyfiel there's now a v3.0.4-dev branch with a `versions/3.0.4.md` file.  This PR has been re-targeted to that branch which is why it shows a bunch of extra commits (those are the post-3.0.3 commits on main, which are not relevant to `3.0.4.md`).

You'll want to either close this and open a new PR, or edit the commit to only apply to `3.0.4.md`, rebase it on top of the v3.0.4-dev branch, and force-push.

jeremyfiel:
ok. probably easier and cleaner to close and make a new one. will do so

PR #3179, opened 2023-03-01
Make registry internal links relative [gh-pages]

handrews:
This allows [renderings of forks](https://handrews.github.io/OpenAPI-Specification/) to work.  Otherwise the links all 404 because forks are rendered under a directory name.

### DIFF
--- a/registry/index.md
+++ b/registry/index.md
@@ -15,7 +15,7 @@ Please raise a [Pull-Request](https://github.com/OAI/OpenAPI-Specification/pulls
 #### API access
 
 * [registries.json](./api/registries.json) - Registries meta-registry
-{% for registry in site.collections %}{% unless registry.hidden %}* <a href="..//api/{{ registry.slug }}.json">{{ registry.slug }}.json</a>{% endunless %}
+{% for registry in site.collections %}{% unless registry.hidden %}* <a href="../api/{{ registry.slug }}.json">{{ registry.slug }}.json</a>{% endunless %}
 {% endfor %}
 
 #### RSS feed
--- a/registry/index.md
+++ b/registry/index.md
@@ -9,16 +9,16 @@ Please raise a [Pull-Request](https://github.com/OAI/OpenAPI-Specification/pulls
 
 ### Contents
 
-{% for registry in site.collections %}{% unless registry.hidden %}* <a href="/registry/{{ registry.slug }}">{{ registry.name }}{% endunless %}
+{% for registry in site.collections %}{% unless registry.hidden %}* <a href="./{{ registry.slug }}">{{ registry.name }}{% endunless %}
 {% endfor %}
 
 #### API access
 
-* [registries.json](/api/registries.json) - Registries meta-registry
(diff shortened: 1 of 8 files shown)


MikeRalphson:
Thanks @handrews - are we sure this won't break the main site which is deployed without a directory name? I seem to recall I had difficulty making it work for both cases? Not that I hate jekyll or anything.

handrews:
@MikeRalphson I actually missed some links, the most recent commit should be correct (I could have sworn I tested those pages... blah).

There could be jekyll weirdness that might prevent this from working, I have no idea how to test it, though.  OpenAPI is using a custom domain, and I do not, so I don't know if that's a concern.

I can't think of any reason why it _wouldn't_ work, as the `./` should just go relative to the root, but... jekyll (I unabashedly hate jekyll).

MikeRalphson:
It looks like I set up https://spec.mermade.org.uk/ in the dim and distant past for testing, so will pull your changes. Thanks.

registry/index.md
```diff
@@ -9,16 +9,16 @@ Please raise a [Pull-Request](https://github.com/OAI/OpenAPI-Specification/pulls
...
-{% for registry in site.collections %}{% unless registry.hidden %}* <a href="/registry/{{ registry.slug }}">{{ registry.name }}{% endunless %}
+{% for registry in site.collections %}{% unless registry.hidden %}* <a href="./{{ registry.slug }}">{{ registry.name }}{% endunless %}
 {% endfor %}
 
 #### API access
 
-* [registries.json](/api/registries.json) - Registries meta-registry
-{% for registry in site.collections %}{% unless registry.hidden %}* <a href="/api/{{ registry.slug }}.json">{{ registry.slug }}.json</a>{% endunless %}
+* [registries.json](./api/registries.json) - Registries meta-registry
+{% for registry in site.collections %}{% unless registry.hidden %}* <a href="..//api/{{ registry.slug }}.json">{{ registry.slug }}.json</a>{% endunless %}
```
MikeRalphson, line 18:
```suggestion
{% for registry in site.collections %}{% unless registry.hidden %}* <a href="../api/{{ registry.slug }}.json">{{ registry.slug }}.json</a>{% endunless %}
```

registry/index.md
```diff
@@ -9,16 +9,16 @@ Please raise a [Pull-Request](https://github.com/OAI/OpenAPI-Specification/pulls
...
 {% endfor %}
 
 #### RSS feed
 
-* [feed.xml](/rss/feed.xml)
+* [feed.xml](./rss/feed.xml)
```
MikeRalphson, line 23:
```suggestion
* [feed.xml](../rss/feed.xml)
```

MikeRalphson (changes requested):
LGTM apart from two tiny nits. Tested at https://spec.mermade.org.uk/

handrews:
Thanks for the thorough testing @MikeRalphson — I committed both of your suggestions.

MikeRalphson (approved):
LGTM Thanks.

whitlockjc:
Thanks for everyone involved!

PR #3167, opened 2023-02-20
adds missing formats in the registry

baywet:
fixes #845
closes #1811

Issue #3174, opened 2023-02-28
Clarify "format": "byte" vs "format": "base64" in OAS 3.0.x


handrews:
See PR #3167 for a lot of context.

In OAS 3.0.3:

* [§4.4 Data Types](https://spec.openapis.org/oas/v3.0.3#data-types) shows `format: byte` for base64-encoded binary data, matching the [same section in OAS 2.0](https://spec.openapis.org/oas/v2.0#data-types)
* [§4.7.14.3 Considerations for File Uploads](https://spec.openapis.org/oas/v3.0.3#considerations-for-file-uploads) has a very minimal example showing `format: base64` alongside an example of `format: binary`; there is no sign of `format: byte`
* [§4.7.14.5 Special Considerations for `multipart` Content](https://spec.openapis.org/oas/v3.0.3#special-considerations-for-multipart-content) mentions `format: base64` alongside `format: binary` when talking about default `Content-Type` values, but does not mention `format: byte`

It's clear that `format: byte` was intended to carry over from OAS 2.0, but that `base64` is more intuitive and got used everywhere outside of the table of formats.

@OAI/tsc I feel like the most sensible thing to do here would be to define both `byte` and `base64` as having the same effect, and note that `byte` is compatible with OAS 3.0, while `base64` is analogous to using `contentEncoding: base64` in OAS 3.1.

I'm guessing that adding a format would not be allowed by SemVer, but we could make the argument that the current situation is so confusing that defining both is just a clarification.

PR #3184, opened 2023-03-03
- adds a base64url format description

baywet:
related #3167

### DIFF
--- a/registries/_format/base64url.md
+++ b/registries/_format/base64url.md
@@ -1,0 +1,29 @@
+---
+owner: baywet
+issue:
+description: Binary data encoded as a url-safe string as defined in [RFC4648](https://www.rfc-editor.org/rfc/rfc4648#section-5)
+base_type: string
+layout: default
+deprecated_note: '3.1'
+remarks: "When using OpenAPI 3.1 it's recommended not to use this format and instead use [`contentEncoding` with a value of `base64url`](https://json-schema.org/draft/2020-12/json-schema-validation.html#name-contentencoding)."
+---
+
+# <a href="..">{{ page.collection }}</a>
+
+## {{ page.slug }} - {{ page.description }}
+
+Base type: `{{ page.base_type }}`.
+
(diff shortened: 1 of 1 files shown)


handrews:
@baywet I think you need to update this to remove the deprecated version info.

I'm also wondering if, now that we have definitively settled on `byte` rather than `base64`, if it makes sense to align with RFC 4648 terminology for this, or call it `byteUrl` or something.  Neither feels very satisfactory to me.

baywet:
Can you expand on why do you think we should remove the deprecated aspect please? I might have missed part of the conversation here.

If it were only up to me, byte would be base64 and base64url would be base64url. byte only being an alias to byte64 for compatibility reasons. Those different names are simply adding confusion IMHO.

handrews:
Edit to my previous comment: formats are kebab-case so `byte-url`, not `byteUrl`

Re: `deprecated_note` — see https://github.com/OAI/OpenAPI-Specification/pull/3167#issuecomment-1456845756

In #3174 I suggested having a `base64` alias for `byte` but we discussed it in the last call and decided not to do that (as seen in #3182).  So we're stuck with `byte`.

baywet:
@handrews so are we good with this PR as is or do you suggest we make other changes?

handrews:
I'm still feeling meh about whether to use `base64url` for compatibility with OAS 3.1 vs `byte-url` for compatibility with OAS 3.0, the actual targeted version here.  As much as I dislike `byte`, it's what we're stuck with and this is a URL-safe version of it.  Although I suspect that very few tools will implement support for this so I'm not sure how much it matters.

@MikeRalphson, @darrelmiller, anyone else - any preference for `base64url` vs `byte-url` in the context of OpenAPI 3.0?

MikeRalphson:
+1 for base64url, let's not be held back by our past missteps.

handrews:
@MikeRalphson thanks - in that case this looks good to go.

handrews (approved):

MikeRalphson (approved):
LGTM

PR #3185, opened 2023-03-03
- adds a char format description

baywet:
related #3167

### DIFF
--- a/registries/_format/char.md
+++ b/registries/_format/char.md
@@ -1,0 +1,27 @@
+---
+owner: baywet
+issue:
+description: A single character
+base_type: string
+layout: default
+---
+
+# <a href="..">{{ page.collection }}</a>
+
+## {{ page.slug }} - {{ page.description }}
+
+Base type: `{{ page.base_type }}`.
+
+The `{{page.slug}}` format is a single character.
+
+{% if page.issue %}
+### GitHub Issue
+
+* [#{{ page.issue }}](https://github.com/OAI/OpenAPI-Specification/issues/{{ page.issue }})
+{% endif %}
+
+{% if page.remarks %}
+### Remarks
+
+{{ page.remarks }}
+{% endif %}


handrews (approved):
I have an impulse to note that it's more reliable for validation to use this alongside `minLength` and `maxLength`, but you can say similar things about all uses of `format` so it's probably best to not complicate things.

baywet:
@handrews people could use both:
- min set to 1
- max set to 1
- format char

This way they'd have validation from both tooling that doesn't understand the format and from the tooling that does. Do you think we should add a note?

handrews:
@baywet no, it's probably best not to as we'd then want to add notes to all of the formats about how to compensate for format support being a dumpster fire in general.  People use `format` at their own risk, and (having thought about it a bit more since writing that comement), the registry is not the right place to try to mitigate that.

baywet:
makes sense. I do think there should be a place for "things that are not normatize but that people should consider". Kind of a collection of best practices if that makes sense.

baywet:
Thanks. This is a bit buried, does it get published anywhere? Should we move it to the GitHub pages branch?

MikeRalphson:
> This is a bit buried, does it get published anywhere? Should we move it to the GitHub pages branch?

A separate repo was considered preferable, I think partly due to different licensing (CC versus Apache-2.0). It is published at https://oai.github.io/Documentation and linked to from the website, but there are plans afoot to create a https://learn.openapis.org subdomain.

baywet:
Thanks, let log an issue to document that "best practice" should this format get merged.

MikeRalphson:
@baywet https://github.com/OAI/Documentation/issues/42

baywet:
@darrelmiller I believe this one is good to be merged as we've been through all the review comments.

MikeRalphson (approved):
LGTM

Issue #3175, opened 2023-03-01
Clarify the utility of base64-encoding of whole payloads or multipart parts


handrews:
As seen in [recent PR comments](https://github.com/OAI/OpenAPI-Specification/pull/3167#discussion_r1118451051) with @ralfhandl and @baywet, and in an earlier [thread I started on the HTTP APIs Slack](https://httpapis.slack.com/archives/C1C0J9X7Z/p1677253124382239), base64-encoding an entire payload, or an entire part in a `multipart` payload, doesn't make much sense in HTTP.

Encoding binary to text as standardized in [RFC 2045 Multipurpose Internet Mail Extensions (MIME) Part One:  Format of Internet Message Bodies](https://www.rfc-editor.org/rfc/rfc2045.html) and revised and expanded in [RFC 4648 The Base16, Base32, and Base64 Data Encodings](https://www.rfc-editor.org/rfc/rfc4648) was designed for email systems that often restricted bodies to 7-bit ASCII data.   One of RFC 4648's revisions was to remove the line length restrictions on encoded data in RFC 2045, but the RFC still targeted environments with a 7-bit ASCII limitation.  The MIME [`Content-Transfer-Encoding`](https://www.rfc-editor.org/rfc/rfc2045.html#section-6) header indicated the use of such encoding.

HTTP does not have a 7-bit ASCII limitation on message bodies.  Encoding binary data into ASCII expands it by [33% to 300%](https://developer.mozilla.org/en-US/docs/Glossary/Base64#encoded_size_increase) (although compressing the result complicates the calcuations).

In 2015, [RFC 7578 Returning Values from Forms: `multipart/form-data`](https://www.rfc-editor.org/rfc/rfc7578) replaced RFC 2388 and [deprecated the `Content-Transfer-Encoding` header in HTTP usage](https://www.rfc-editor.org/rfc/rfc7578#section-4.7) as it is not needed and no active use of it had been seen in the wild.  It is still allowed in MIME.

I would suggest that we do the following in 3.0.4 and 3.1.1:

* Note that there are no HTTP headers which communicate base64 or similar encoding for an entire message payload, and that because of that and the data size increase, attempting such a thing is NOT RECOMMENDED
* Reference RFC 7578 §4.7's deprecation of `Content-Transfer-Encoding` in `multipart/form-data` and note that attempting to use an Encoding Object to induce it is unlikely to be supported by relevant software
[...]

PR #3187, opened 2023-03-05
Clarify the use of "binary" and "byte" formats

### LINKED ISSUE #1544: Clarify: "type": "string", "format": "binary" in non-entity-body

The choices I see:

handrews:
Fixes #1544 and #3175.

Add a brief explanation as a subsection of Data Types, and update the relevant examples.  Specifically:

* Don't use `format: byte` for a whole message body
* Fix `multipart/mixed` example to be `multipart/form-data` (because `multipart/mixed` does not name its parts and therefore cannot be represented with an object schema)
* Add an `application/x-www-form-urlencoded` example, which is where `format: byte` needs to be used

### DIFF
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -162,12 +162,20 @@ The formats defined by the OAS are:
 `integer` | `int64` | signed 64 bits (a.k.a long)
 `number` | `float` | |
 `number` | `double` | |
-`string` | `byte` | base64 encoded characters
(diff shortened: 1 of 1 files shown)


CALIFAS95 (approved):

ralfhandl (approved):

darrelmiller (approved):

MikeRalphson (approved):
LGTM thanks

PR #3188, opened 2023-03-05
Fix broken RFC links

### LINKED ISSUE #3146: The links to RFC3339 stopped working
The target of the RFC3339 link on the page https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#data-types does not exist anymore.

There is only a 404 page.

RFC3339 is available under https://www.rfc-editor.org/rfc/rfc3339

handrews:
Fixes #3146.

The IETF apparenlty decided to redo everything about publishing RFCs, and while most of the old URLs redirect, these ones for RFC 3339 did not.  Also, there was a stray trailing '/' that broke another RFC link.

### DIFF
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -164,8 +164,8 @@ The formats defined by the OAS are:
 `number` | `double` | |
 `string` | `byte` | base64 encoded characters
 `string` | `binary` | any sequence of octets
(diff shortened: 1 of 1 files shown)


MikeRalphson (approved):
Thanks.

darrelmiller (approved):

Issue #2940, opened 2022-06-09
rfc6570 template clarification requested for the parameter styles, cookie use case unclear


spacether:
Hello there. Can we get some clarification on:
- cookie parameter example values using form serialization
- rfc6570 template clarification for all style and parameter type combos?

I am asking because the swagger docs show examples:
https://swagger.io/docs/specification/serialization/
And in the query parameter examples we see:

| style | explode | rfc6570 template | array_value |
| ----- | ------- | ------------- | ----------- |
| form | true | {?id*} | ?id=3&id=4&id=5 |
| form | false | {?id} | ?id=3,4,5 |

Yet for cookie parameters values lack the ? prefix.
This is not consistent with the examples given in rcf6570 for form style {?var} or {&var} expansion
Those expansions require the ? or & prefix with values like:
{?var} -> {?id} ->
{?id*} ?id=3&id=4&id=5
{?id} -> ?id=3,4,5

{&var} -> {&id} ->
{&var*} &id=3&id=4&id=5
{&var} &id=3,4,5
If we are meant to exclude those prefixes, and use form styles please provide the uri templates and direction.

The openapi spec omits the ? prefix from the form examples of serialized data though they are needed in style form query parameter serialization.

## Header Question:
Also, what should None serialize to in a header? In a path it is empty string. In a header should the value be empty string or should it be omitted?

Related issues:
- https://github.com/OAI/OpenAPI-Specification/issues/1528
- https://github.com/OAI/learn.openapis.org/issues/100

MaceWindu:
`explode=false` doesn't even make sense for cookies, as it itroduce comma, not allowed by cookie value

handrews:
@spacether OAS 3.x does not use RFC 6570 URI Template syntax.  That Swagger documentation is just saying that certain RFC 6570 template syntaxes are equivalent to certain `style` and `explode` combinations.

spacether:
If they are identical in implementation, openapi gives no specific step by step directions on how to serialized the data and rfc 6570 is mentioned then it looks to me like the spec does use it.

handrews:
@spacether the specification [already cites](https://spec.openapis.org/oas/v3.1.0#style-values) the parts of RFC 6570 that is uses, in the places where it is used.  This usage is not the full RFC 6570 syntax.  OAS URL Templating syntax is [defined in the OAS itself](https://spec.openapis.org/oas/v3.1.0#path-templating).

[...]

Issue #1528, opened 2018-04-06
[oas3] Default 'explode' for cookie parameters?


jwalton:
According to the [spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#parameterExplode), "When `style` is `form`, the default value [for explode] is true. For all other styles, the default value is false."  However, according to [this documentation](https://swagger.io/docs/specification/serialization/), in the "Cookie Parameters" section, there are "*"s indicating the default serialization, and it is style: form, explode: false.

So which is it?  Do cookies use explode by default or not?

hkosova:
The Specification is the source of truth - so the cookies are not exploded by default.

I've fixed the swagger.io doc. If you notice any other info on the Swagger website that contradicts the info in the OpenAPI Specification, please file an issue in the https://github.com/swagger-api/swagger.io repository. Thanks!

jwalton:
What does explode even mean for a cookie, though?  If I have a list in a cookie "foo", does this mean I have a Cookie header with multiple copies of the cookie?

    Cooke: foo=1; foo=2

RFC 6265 doesn't disallow this, and some clients will even send multiple copies of the same cookie in certain weird circumstances (if you set a cookie "foo" with the domain "*.myserver.com", and then set a second cookie "foo" with the domain "baz.myserver.com").  But...  That'd be a weird way to pass parameters around.  :P

jwalton:
And, cookies default to "form", but they aren't *really* "form".  An array encoded in "form" with `explode: true` is encoded as: 'color=blue&color=black&color=brown' (this is the example [from the spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#style-examples)) but if you tried to literally set the cookie header to:

```
Cookie: color=blue&color=black&color=brown
```

A cookie parser would turn that into `{color: "blue&color=black&color=brown"}`, which is... ugh.

webron:
Does this discourage you from using cookies as parameters?

jwalton:
Using cookies as parameters is weird, and I know there's a history of debate about whether to include them in the spec or not in the first place.  I'm not using cookies as parameters; I'm writing an implementation of the spec.  If someone is using my implementation, they expect it to work correctly, even if they are using cookie parameters.  So I need to work out what "correctly" is.

[...]

PR #3193, opened 2023-03-08
Don't allow explode: true for cookie parameters

ahx:
This changes the 3.1 schema to not allow `explode: true` for cookie parameters.  Supporting `explode: true` for cookie parameters with array values would end up in a header like ~~`Cookie: ids=1; ids=2`~~ `[UPDATE: I don't know]`, right? I was asking myself if that would be allowed and could not find an explicit "yes" in the spec. I assume it should not be supported.

The description of the 3.0 version on the Swagger site implies that `explode: true` is not supported for cookie parameters:  https://swagger.io/docs/specification/serialization/#cookie

### DIFF
--- a/schemas/v3.1/schema.yaml
+++ b/schemas/v3.1/schema.yaml
@@ -438,7 +438,9 @@ $defs:
                 style:
                   default: form
                   const: form
-
+                explode:
+                  default: false
+                  const: false
           styles-for-form:
             if:
               properties:


jdesrosiers:
This doesn't look right to me. The [official documentation](https://spec.openapis.org/oas/latest.html#parameter-object) says this about the `explode` field.

> When `style` is `form`, the default value is `true`.

... and this about the `style` field.

> Default values (based on value of `in`): [...] for `cookie - form`.

So, if `in` is `"cookie"`, then `style` is `"form"`. If `style` is `form` then the default value for `explode` is `true`. Not only does this seem to be allowed, it appears to be the default behavior. It's been awhile since I've worked with cookies, but as I recall this is the normal syntax for cookies so I don't see why it wouldn't be allowed.

ahx:
@jdesrosiers Thanks for pointing out the default behaviour (style: form, explode: true). This does not fit with this change request. But how do we encode arrays in Cookies? The [example](https://spec.openapis.org/oas/latest.html#style-examples) says `color=blue&color=black&color=brown, but that is only applicable to query parameters, right?

hkosova:
@ahx 
> But how do we encode arrays in Cookies?

The exploded form of `param` = `[3, 4, 5]` as a cookie would be:
```
Cookie: param=3&param=4&param=5
```
which is [not the normal syntax](https://stackoverflow.com/a/73708186/113116) for cookies.

The current OpenAPI cookie serialization options don't make much sense, and this is discussed here:
* [[oas3] Default 'explode' for cookie parameters?](https://github.com/OAI/OpenAPI-Specification/issues/1528)
* [rfc6570 template clarification requested for the parameter styles, cookie use case unclear](https://github.com/OAI/OpenAPI-Specification/issues/2940)

It's not enough to just drop the `explode` option for cookies. We need to come up with proper cookie serialization behavior that makes sense and results in valid cookie syntax.

PR #3200, opened 2023-03-14
docs: fix anchor links in rendered specs

MikeRalphson:
Fixes the annoyance whereby clicking on a section link (such as to the `responseObject`) took you to the ToC, requiring a further click to take you to the section you actually wanted.

The `md2html` script lives in the `main` branch, but its target is the `gh-pages` branch.

Note: the ToC doesn't auto-scroll the linked section into view, but AFAIK ReSpec does not have an option for that.

Tested here: https://mermade.github.io/static/v3.1.0.html

Should not break other peoples' links, as tested with both camelCase and kebab-case anchors:

* https://mermade.github.io/static/v3.1.0.html#responseObject
* https://mermade.github.io/static/v3.1.0.html#response-object


webron (approved):
:shipit:

PR #3226, opened 2023-04-01
Use `api-key` instead of `api_key` as example header name

### LINKED ISSUE #3225: `api_key` is not a good example for an HTTP header, because of the underscore
Several HTTP servers strip HTTP headers when they contain `_` for security reasons. Using something like `api-key` instead of `api_key` in the examples  would be a more realistic example.

For references, nginx drops headers with underscores by default (  see https://nginx.org/en/docs/http/ngx_http_core_module.html#underscores_in_headers or https://dev.to/thesameeric/dont-use-underscores-in-your-http-headers-gfp ) django project's blog ( https://www.djangoproject.com/weblog/2015/jan/13/security/ ) has a good description of the problem. Other servers such as waitress ( https://github.com/Pylons/waitress/pull/80 ) followed by dropping headers with underscore. I think apache http also drop such headers ( https://httpd.apache.org/docs/trunk/en/new_features_2_4.html#:~:text=Header%20names%20containing%20invalid%20characters%20(including%20underscores)%20are%20no%20longer%20converted%20to%20environment%20variables ) 

---

What I'm thinking is that in places where `api_key` is used to give an example of a header, for example in

https://github.com/OAI/OpenAPI-Specification/blob/24088857068802d5fd4e4f8bef3f95ef06fa8830/versions/3.1.0.md?plain=1#L566-L567

using something like `api-key` might be better.

perrinjerome:
Several HTTP servers drop http headers with underscores.

Fixes #3225

### DIFF
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -563,7 +563,7 @@ my.org.User
   "securitySchemes": {
     "api_key": {
       "type": "apiKey",
-      "name": "api_key",
+      "name": "api-key",
       "in": "header"
     },
     "petstore_auth": {
@@ -640,7 +640,7 @@ components:
   securitySchemes:
     api_key:
       type: apiKey
-      name: api_key
+      name: api-key
       in: header
     petstore_auth:
       type: oauth2
@@ -3226,14 +3226,14 @@ scheme: basic
 ```json
 {
   "type": "apiKey",
-  "name": "api_key",
+  "name": "api-key",
   "in": "header"
 }
 ```
 
 ```yaml
 type: apiKey
-name: api_key
+name: api-key
 in: header
 ```
 


versions/3.1.1.md
```diff
@@ -563,7 +563,7 @@ my.org.User
   "securitySchemes": {
     "api_key": {
       "type": "apiKey",
-      "name": "api_key",
+      "name": "api-key",
```
perrinjerome, line 566: Here I have only changed the `name` and kept the `id` to be "api_key", therefore later in "Security Requirement Object" section, it stays using "api_key"
https://github.com/OAI/OpenAPI-Specification/blob/9bf45ceb368f3fe14d3f7e18ace19ce4b3f24ad5/versions/3.1.1.md?plain=1#L3370-L3381

I first replaced all `api_key` by `api-key` but I thought it would be better to keep consistent `api_key` to have consistent naming with `petstore_auth`

MikeRalphson (approved):
LGTM

PR #3233, opened 2023-04-06
Content-Type for properties with type string and contentEncoding

### LINKED ISSUE #3230: v3.1 - contentEncoding
The [example](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#special-considerations-for-multipart-content) from the spec:
```
          profileImage:
            # Content-Type for application-level encoded resource is `text/plain`
            type: string
            contentMediaType: image/png
            contentEncoding: base64
```            
contradicts the specification:
  * `If the property is a type: string with a contentEncoding, the default Content-Type is application/octet-stream`

[Henry Andreews agrees](https://stackoverflow.com/questions/67388636/in-a-json-schema-what-is-the-correct-way-to-describe-a-default-example-or-enu)

commonism:
fixes #3230

### DIFF
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -1574,7 +1574,7 @@ requestBody:
             type: object
             properties: {}
           profileImage:
-            # Content-Type for application-level encoded resource is `text/plain`
+            # Content-Type for properties with type string and contentEncoding is `application/octet-stream`
             type: string
             contentMediaType: image/png
             contentEncoding: base64


handrews:
I should probably agree with past-me :-)  Sadly, my current state (attempting to recover from COVID, which is going veerrrryyyyy slowly) has paused my effort to revamp the `content*` aspects of 3.1.1 along the same lines as PR #3187 did for `format` in 3.0.4.  

I'm fine with merging this, but be aware that once my brain is back online I'm probably going to rewrite a lot of this section.  Because even before I came down with COVID I struggled to follow it and I wrote most of it 😅 Plus, as we discussed #3187 we made some decisions that clarify all of this, as we realized that some of these combinations don't even make sense.

I have too much COVID-brain right now to figure out where this particular problem fits in all of that.

MikeRalphson:
Get well soon @handrews

handrews (approved):
My apologies for the extended delay... "recovering from COVID" ended up being a more extended process than anticipated and I forgot about this PR.  Although since 3.1.1 has yet to ship the practical impact is small.

But let's get this approved and merged now!

asbjornu (approved):

darrelmiller (approved):

PR #3234, opened 2023-04-06
Add templated-url format to the registry

MikeRalphson:
As discussed with @jdesrosiers on Slack

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ deploy/
 node_modules/
 scripts/
 package-lock.json
+.bundle/
+vendor/bundle/
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,15 +2,15 @@
 
 ## Building the site
 
-The registy site uses [jekyll](https://jekyllrb.com/), a Ruby based static site generator, with the Dinky theme.
+The registy site uses [jekyll](https://jekyllrb.com/), a Ruby based static site generator, with the Just-the-Docs theme.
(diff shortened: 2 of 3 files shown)


handrews:
(pardon muddled thoughts - I'm recovering from COVID) is there a reason to not make this `oas-templated-url`?  There might be a wide range of formats that fit the vague description here, making the format difficult to support.

MikeRalphson:
There's no reason not to, is it just for clarity? AsyncAPI might use something very similar though.

handrews:
trying to avoid a repeat of `decimal` where there are different meanings depending on context

handrews:
i think the question to ask is: what guarantees can tools assume/implement when they see this format?  Currently, not much.  URI + `{` and `}`, but that is a broader constraint than OAS defines - OAS puts constraints on what parts of a URI can be replaced by a variable, and might constrain the variable syntax (I can't recall)

handrews:
See also #3256 (Formally define templating behavior)

handrews:
I think if we make this `oas3-templated-url` we could merge it.  That avoids confusion with other URL templating systems, of which there are many, including ones that differ across OAS versions.

handrews:
Also, I think this needs rebasing.

handrews:
Since Mike's not available to rebase and discuss this, and we're already tracking the need for a format for this in #3530 (which already links to this PR), I'm just going to close this.  When someone is available to figure it out they can submit a new PR.

PR #3235, opened 2023-04-06
fix: update format for servers.url and links.operationRef in v3.1 schema

MikeRalphson:
`uri-reference` is not the right format for these properties, as they can contain unescaped `{` and `}` characters. As @jdesrosiers points out, they are not `uri-template` either, as they do not support RFC6570 URI templates in full.

### DIFF
--- a/schemas/v3.1/schema.json
+++ b/schemas/v3.1/schema.json
@@ -168,7 +168,7 @@
       "properties": {
         "url": {
           "type": "string",
-          "format": "uri-reference"
+          "format": "templated-url"
         },
         "description": {
           "type": "string"
@@ -961,7 +961,7 @@
       "properties": {
         "operationRef": {
           "type": "string",
-          "format": "uri-reference"
+          "format": "templated-url"
         },
         "operationId": {
           "type": "string"
--- a/schemas/v3.1/schema.yaml
+++ b/schemas/v3.1/schema.yaml
@@ -119,7 +119,7 @@ $defs:
     properties:
       url:
         type: string
-        format: uri-reference
+        format: templated-url
       description:
         type: string
       variables:
@@ -664,7 +664,7 @@ $defs:
     properties:
       operationRef:
         type: string
-        format: uri-reference
+        format: templated-url
       operationId:
         type: string
       parameters:


jdesrosiers (approved):

handrews:
can't `operationRef` have embedded runtime expressions in addition to the regular templating?

MikeRalphson:
@handrews yes, but does that mean it requires a new `format` value?

handrews:
@MikeRalphson it depends on what you want tooling to be able to do with the `format`.  If all you're going for is "the old `format` was wrong and we should have a not-wrong `format`", then pretty much anything you want is fine.  If you want schema tools to be able to recognize exactly what sort of templating is happening and validate that (either within JSON Schema if the `format-assertion` vocabulary is in use, or as a post-schema-validation step if the `format-annotation` vocabulary is in use, which it is the default), then your `format` value(s) need to be specific enough to convey that information.

It's hard for me to judge whether this `format` value is what we'd want in either or both situations without knowing the expectations for implementation.  This, btw, is part of the problem of `format`, as we saw with `decimal` — it seems to encourage underspecification and over-expectation.  In other words, `format` consistently over-promises to schema authors while allowing/encouraging implementation to under-deliver.

But really, if all you want this to do is indicate "`uri-reference` with balanced `{` and `}` allowed" then sure this is fine.  I did not notice yesterday (COVID-brain) but now I see that in PR #3234 you are referencing the Paths Object for "usage" (rather than syntax?) and not the Operation Reference.  Does the operation reference allow syntax that is not allowed in the Paths Object?  I'm definitely a little unclear on how runtime expressions fit into the templating for the operation reference - I need to study that when I'm feeling better.

handrews:
See also #3256 (Formally define templating behavior)

handrews:
While I agree that we need to do something like this, I'm going to close this particular PR so that:

* We can merge #3455 to at least get the wrong format out of the schema
* We can resolve the question of what to call the format in #3234 and/or #3530
* We can resolve the question of how many formats we need in #3530 

We can open a new PR for this once the number and name of formats is resolved.

PR #3238, opened 2023-04-07
Add media-range format to registry

jdesrosiers:
I used a custom format in the 3.1 schemas. Now that we have a registry, it should probably be included.

https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.1/schema.yaml#L497-L503


MikeRalphson (approved):
LGTM

PR #3249, opened 2023-04-16
Update servers description on Path Item Object

### LINKED ISSUE #3248: Clarification on Path Item Servers
If "servers" are specified at the Path Item level, do they override Root level servers?

On an Operation Object it is explicitly stated that:
 _"If an alternative server object is specified at the Path Item Object or Root level, it will be overridden by this value."_

Do Path Item servers override Root level servers in the same manner? If not I assume it must add to the list of available servers (for that path only).

I would request that the PathItem servers Description be more explicit on whether it overrides, or adds to the Root level servers.

I'd be happy to put in a PR for this if I know what the correct behaviour is.

charjr:
close #3248 

- Explicitly state that Path Item servers override Root level servers
- More consistent with the servers description for Operations


MikeRalphson (approved):
LGTM - thank you.

darrelmiller (approved):

PR #3258, opened 2023-04-26
Use $ref to definitions in oneOf for Parameter locations

handrews:
I'd like to use a JSON Merge Patch (RFC 7386) document to add keywords for the OAS compliance parser project.

JSON Merge Patch does not handle arrays well, so moving these to their own definitions and having the "oneOf" directly in the "Parameter" definition (avoiding duplicate "description" fields on the same instance location) allows merge-patch to work where I need it.

Otherwise, JSON Patch (RFC 6902) will be needed, which is less intuitive.

The 3.1 schema already uses "$defs" here.

### DIFF
--- a/schemas/v3.0/schema.yaml
+++ b/schemas/v3.0/schema.yaml
@@ -706,46 +706,51 @@ definitions:
     allOf:
       - $ref: '#/definitions/ExampleXORExamples'
       - $ref: '#/definitions/SchemaXORContent'
-      - $ref: '#/definitions/ParameterLocation'
-
-  ParameterLocation:
-    description: Parameter location
     oneOf:
-      - description: Parameter in path
-        required:
-          - required
-        properties:
-          in:
-            enum: [path]
-          style:
-            enum: [matrix, label, simple]
-            default: simple
-          required:
-            enum: [true]
+      - $ref: '#/definitions/PathParameter'
+      - $ref: '#/definitions/QueryParameter'
+      - $ref: '#/definitions/HeaderParameter'
+      - $ref: '#/definitions/CookieParameter'
 
-      - description: Parameter in query
-        properties:
-          in:
-            enum: [query]
-          style:
-            enum: [form, spaceDelimited, pipeDelimited, deepObject]
-            default: form
+  PathParameter:
+    description: Parameter in path
+    required:
+      - required
+    properties:
+      in:
+        enum: [path]
+      style:
+        enum: [matrix, label, simple]
+        default: simple
+      required:
+        enum: [true]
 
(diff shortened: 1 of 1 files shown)


MikeRalphson:
Just to be clear, you're wanting to patch the current OAS schema document in order not to have redundancy in the compliance parser chain?

handrews:
@MikeRalphson sorry I did not notice your question when you asked it!  It's not so much avoiding redundancy as it is making the tools more user-friendly.  I'm perfectly comfortable with JSON Patch (RFC 6902), but [it is not incredibly intuitive to read](https://github.com/OAI/oascomply/blob/main/patches/oas/v3.0/preliminary-patch.json) compared to JSON Merge Patch (RFC 7396).  For folks wanting to modify `oascomply` in the future, it would be slightly nicer to have everything be JSON Merge Patch.

handrews:
@OAI/openapi-maintainers one more approval on this and I can merge it 🙏

ralfhandl:
Will the sibling `schema.json` file be auto-generated from the changed `schema.yaml` file?

schemas/v3.0/schema.yaml
```diff
@@ -331,7 +331,12 @@ definitions:
     patternProperties:
       '^x-': {}
     additionalProperties: false
-
+    dependencies:
+      discriminator:
+        anyOf:
+          - required: [allOf]
+          - required: [anyOf]
+          - required: [oneOf]
```
notEthan, line 339: I'm not sure you meant for this to go in this PR, it seems to not fit with the title/description.

If you did: I don't know `discriminator` well, but the example [Models with Polymorphism Support](https://github.com/OAI/OpenAPI-Specification/blob/3.0.3/versions/3.0.3.md#models-with-polymorphism-support) has `discriminator` in `/components/schemas/Pet` without a `*Of` keyword. But then `/components/schemas/Pet` is never used in the example in a way that would invoke `discriminator`, so that's confusing.

handrews: @notEthan oops- looks like I need to rebase this (there was a whole thing where this got added and then removed), doing so momentarily.

notEthan: It's rebased but seems to be the same as what I commented on

handrews: wat... oh %$@#! I fixed it locally and forgot to add that commit.  I'm still not sure how it got there in the first place.  Fixed for real now, sorry about that 🤦

ralfhandl (approved):

darrelmiller (approved):

PR #3259, opened 2023-04-27
Servers description 3.0.4

charjr:
close https://github.com/OAI/OpenAPI-Specification/issues/3248

    Explicitly state that Path Item servers override Root level servers
    More consistent with the servers description for Operations


MikeRalphson (approved):
Also LGTM, thanks!

darrelmiller (approved):

PR #3262, opened 2023-04-28
Update list of Special Interest Groups

duncanbeevers:
The Special Interest Groups document was created, but no content was ever added.

This PR updates the document to list each Special Interest Group, along with a description of the group and a link to its Slack channel and Github repository where available.

### Methodology

This list is constructed from surveying Slack channels on the open-api.slack.com instance, specifically checking for channels prefixed `#sig-`.

After gathering the list of channels I checked each channel's topic and description, and surveyed (briefly) its history.

If there are groups I missed, please let me know.
If this document should contain different information, or be formatted differently, that's doable too.
If this document shouldn't really exist, that's fine too; I'd be happy to PR that removal in-lieu.

### DIFF
--- a/SPECIAL_INTEREST_GROUPS.md
+++ b/SPECIAL_INTEREST_GROUPS.md
@@ -1,1 +1,43 @@
+## Codegen
+> code-gen vocabulary special interest (working) group
 
+* Slack channel: #sig-codegen
+
+## Formats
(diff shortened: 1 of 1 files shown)


MikeRalphson (approved):
LGTM - thank you for this!

PR #3269, opened 2023-05-08
Add link support from urls

mikekistler:
This PR adds the `url` field to the Link Object. This allows a Link to express the relationship between a URL returned in a response, such as `Location` to a corresponding operation, which may use the URL directly or as the source of path or query parameter values.

### DIFF
--- a/versions/3.2.0.md
+++ b/versions/3.2.0.md
@@ -2031,6 +2031,10 @@ Unlike _dynamic_ links (i.e. links provided **in** the response payload), the OA
 
 For computing links, and providing instructions to execute them, a [runtime expression](#runtimeExpression) is used for accessing values in an operation and using them as parameters while invoking the linked operation.  
 
+Clients follow all links at their discretion.
+Neither permissions, nor the capability to make a successful call to that link, is guaranteed
+solely by the existence of a relationship.
+
 ##### Fixed Fields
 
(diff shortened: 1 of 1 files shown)


versions/3.2.0.md
```diff
@@ -2031,6 +2031,10 @@ Unlike _dynamic_ links (i.e. links provided **in** the response payload), the OA
 
 For computing links, and providing instructions to execute them, a [runtime expression](#runtimeExpression) is used for accessing values in an operation and using them as parameters while invoking the linked operation.  
 
+Clients follow all links at their discretion.
+Neither permissions, nor the capability to make a successful call to that link, is guaranteed
+solely by the existence of a relationship.
+
```
mikekistler, line 2037: I moved this text from within the first example description to the main description of the Link Object to indicate that it applies generally and not just to that specific example.

handrews:
@darrelmiller @mikekistler what was the status and context of this PR?  Was the change agreed-to in principle already, or is this a PR-as-proposal?  Is it something we can finish up quickly or should it be moved back to an issue for deeper discussion?

mikekistler:
Sadly I think this PR is dead. Based on discussions on this in the TDC meetings, I think it is fair to say that OpenAPI v3 links are too ill-defined to make any attempt to extend them fruitful. I'll close this out.

PR #3285, opened 2023-05-26
Add format for imf-fixdate used in HTTP headers

mikekistler:
This PR adds a new format for the date-time format used in HTTP headers, as described in [RFC 7231, Section 7.1.1.1](https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.1.1).

### DIFF
--- a/registries/_format/imf-fixdate.md
+++ b/registries/_format/imf-fixdate.md
@@ -1,0 +1,28 @@
+---
+owner: mikekistler
+issue: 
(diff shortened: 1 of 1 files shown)


MikeRalphson (approved):
LGTM, thanks! As I wasn't familiar with this RFC, I thought maybe a rendered example of a date-time in this format might be beneficial, but I realise we don't have them for most of the other formats (that's my bad, it's missing on the original template).

handrews (approved):

Issue #2239, opened 2020-05-25
Clarification on JWT Bearer access token from OAuth2 Client Credentials Grant...


cadethacker:
Hello,
First time building a spec with OpenAPI and appreciate the thoughtfulness of this spec.  The changes in OpenAPI3 are really nice. 

I'm lost on one point:  The endpoints will be secured via Bearer JWT token, that was retrieved via OAuth2 Client Credentials grant.  The JWT token will contain a correct audience and optional scopes for more granular authorization. The audience is properly validated when requested and rejected if not authorized already. 

Reading this very closely and have some question/clarification, apologies if these are basic :D Be nice, I'm new :D 
https://swagger.io/docs/specification/authentication/

**Clarification 1.** This page above shows a security scheme for both Bearer token AND OAuth2, but to me, those don't feel like different things.  They feel like the same thing.  The client uses OAuth2 Client Credentials grant to get the jwt access token (aka bearer token) , and passes it on the Authorization header to the API which validates it.   So do I put *both* security schemes? At this point, just leaning toward bearer token. But it is OAuth2.  

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#jwt-bearer-sample

**Clarification 2.** The audience claim in the JWT is a key. We require the API, but also envoy (among many other in transit systems) to check the audience.  But I am not sure where to specify this except in the comments in the top  as to what the value of the audience is required to be.  Can I use an extension? I have not quite wrapped my head around those yet... but working on it.  `x-audience`? 

https://tools.ietf.org/html/rfc7519#section-4.1.3
https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/jwt_authn_filter#config-http-filters-jwt-authn

**Clarification 3.** If I need to specify the OAuth2 section the `authorizationUrl` but that changes per lifecycle.  I knew so maybe this is obvious, but not sure how to specify different values per lifecycles.  Just two really, QA and PR. 

**Clarification 4.** Is there a spot to list the JWKS? is that what the `x-jwks-token-keys` maybe? 

Anyway, huge thanks!  Love the work you all have done.  I just hate guessing on stuff like this and figured quicker to ask and get clarification than, guess and get it wrong.

handrews:
See also:

* #3286 

which was only closed because it was suggested to try an `x-audience` field before adding something to 3.2.  The closure is not a rejection on the merits!

Here is @emmanuelgautier's PR comment:

[...]

PR #3286, opened 2023-05-28
Proposal: Add JWT audience in security scheme

emmanuelgautier:
As described in the [JWT RFC](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3) and in the [OpenId Connect spec](https://openid.net/specs/openid-connect-core-1_0.html) for example, `audience` must be verified.

A client can request the audience during an OAuth 2.0 authorization flow, so having `audience` in the OpenAPI can be convenient. This audience documentation can also be used to add more documentation about the API security strategy and ensure the security is enforced.


marzkud:
..!

robjtede (approved):

darrelmiller:
This would be a good suggestion to trial as an `x-` to see the demand for this and the value it can bring.  It would also be good to bring this proposal along with other JWT related issues.

handrews:
@emmanuelgautier since @darrelmiller suggested trying this out as an `x-` field first, I'm going to close this for now.  There is an issue already tracking this topic:

* #2239 

which mentions an `x-audience`, so if that's already being done and there's evidence that it's being used and working, please feel free to re-submit this PR and point to where it's being used.  We will also keep the issue open for consideration in 3.2 or some later 3.x with other JWT issues, as we have several that have been filed.  But it does not seem like there's enough consensus to add this to the spec _yet_.

emmanuelgautier:
Thanks @darrelmiller and @handrews both for your review and sorry if I missed a submission process. This is my first contribution to the spec.

As suggested, more comments will be added then in the following issue about tokens here: https://github.com/OAI/OpenAPI-Specification/issues/2239.

handrews:
@emmanuelgautier no worries about missing the process- we're coming out of a bit of a hibernation period and we're working on documenting our current processes through our [Contributor Guidance Project](https://github.com/orgs/OAI/projects/5).  So you didn't miss anything that you should have been able to figure out right now- we're working on it!

PR #3348, opened 2023-08-14
Updating Link RFC-3339 for date-time and date format

pinalbaldha:
Noticed broken link while reading data-type section in 3.0.3 spec: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#data-types

Hence fixing date-time & date format RFC-3339 reference link.


whitlockjc (approved):
👍

PR #3354, opened 2023-08-17
add string data type, add float data type details.

david-w-t:
Elucidate string data type and document the float data types.

### DIFF
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -155,8 +155,9 @@ The formats defined by the OAS are:
 ------ | -------- | --------
 `integer` | `int32` | signed 32 bits
 `integer` | `int64` | signed 64 bits (a.k.a long)
-`number` | `float` | |
-`number` | `double` | |
+`number` | `float` | single precision IEEE floating point.
+`number` | `double` | double precision IEEE floating point.
+`string` | `UTF-8 chars` | text using UTF-8.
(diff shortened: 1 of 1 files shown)


handrews:
UTF-8 strings are the default, it's not necessary to have a `format` for them.

LasneF:
@david-w-t  we should add as well 

`string` | `date` | As defined by `full-date` - [RFC3339](https://xml2rfc.ietf.org/public/rfc/html/rfc3339.html#anchor14)
`string` | `date-time` | As defined by `date-time` - [RFC3339](https://xml2rfc.ietf.org/public/rfc/html/rfc3339.html#anchor14)

as it was part of the spec in 3.0.3 
https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md?plain=1#L170

handrews:
Somehow I missed that this was a PR against the spec.  We are not adding more formats to the specification itself.  Instead you want the [format registry](https://spec.openapis.org/registry/format/).  

@david-w-t you'll find more details on the numeric formats in the registry, and as previously noted, we don't document missing formats (strings being UTF-8 is already covered by the JSON specification itself, and by the JSON-compatibility text in the forthcoming YAML RFC).

@LasneF the date-time formats are in the registry already.

I'm going to close this.  If there are missing formats, please open an issue to discuss, or open a PR against the registry (you can check closed PRs to see how ot do this).

PR #3355, opened 2023-08-20
Fix 3.1 Path Item schema

### LINKED ISSUE #2635: Ambiguity between Reference Object and PathItem Object

### LINKED ISSUE #3298: OpenAPI 3.1 JSON Schema incorrect for Paths Item in Paths
I believe this line https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.1/schema.json#L294

ie:

```json
"$ref": "#/$defs/path-item"
```

should actually be

```json
"$ref": "#/$defs/path-item-or-reference"
```

according to the docs https://spec.openapis.org/oas/v3.1.0#pathItemObject

as when I am trying to use a json schema validator on my OpenAPI document I am getting an error like:

`OpenAPI document invalid: not allowed - https://spec.openapis.org/oas/3.1/schema/2022-10-07#/$defs/path-item/unevaluatedProperties`

for the below snippet

```yaml
paths:
  /vehicles/{id}/pedals/accelerator:
    $ref: "./paths/acceleratorv3.yaml"
```

handrews:
Fixes #2635 (schema aspects, not spec aspects); fixes #3298

Path Item Objects allow "$ref", and despite the syntactic overlap, it is neither accurate nor semantically meaningful to parse a Path Item Object as a Reference Object, even if such parsing would be successful.

### DIFF
--- a/schemas/v3.1/schema.yaml
+++ b/schemas/v3.1/schema.yaml
@@ -25,7 +25,7 @@ properties:
   webhooks:
     type: object
     additionalProperties:
-      $ref: '#/$defs/path-item-or-reference'
+      $ref: '#/$defs/path-item'
   components:
     $ref: '#/$defs/components'
   security:
@@ -192,7 +192,7 @@ $defs:
       pathItems:
         type: object
         additionalProperties:
-          $ref: '#/$defs/path-item-or-reference'
+          $ref: '#/$defs/path-item'
     patternProperties:
       '^(schemas|responses|parameters|examples|requestBodies|headers|securitySchemes|links|callbacks|pathItems)$':
         $comment: Enumerating all of the property names in the regex above is necessary for unevaluatedProperties to work as expected
@@ -214,6 +214,9 @@ $defs:
     $comment: https://spec.openapis.org/oas/v3.1.0#path-item-object
     type: object
     properties:
+      $ref:
+        type: string
+        format: uri-reference
       summary:
         type: string
       description:
@@ -244,16 +247,6 @@ $defs:
         $ref: '#/$defs/operation'
     $ref: '#/$defs/specification-extensions'
     unevaluatedProperties: false
-
-  path-item-or-reference:
-    if:
-      type: object
-      required:
-        - $ref
-    then:
-      $ref: '#/$defs/reference'
-    else:
-      $ref: '#/$defs/path-item'
 
   operation:
     $comment: https://spec.openapis.org/oas/v3.1.0#operation-object
@@ -617,7 +610,7 @@ $defs:
     type: object
     $ref: '#/$defs/specification-extensions'
     additionalProperties:
-      $ref: '#/$defs/path-item-or-reference'
+      $ref: '#/$defs/path-item'
 
   callbacks-or-reference:
     if:


karenetheridge:
Thank you for this.

But is this really the first actual PR we've seen for this? (sincere question) It's been talked about as an issue for years.

handrews:
@karenetheridge I didn't notice another one ¯\\_(ツ)_/¯

skryukov:
I noticed that `path-item` doesn't include the `$ref` property in `schema.json`, should we update `schema.json` in this PR as well?

handrews:
@skryukov I believe the JSON form is generated from the YAML form, so the PR only targets YAML directly.  @webron would probably know for sure.

webron:
That's correct. Before publishing a new version of the Schema we generate the json from the yaml version.

TristanSpeakEasy:
Okay just ran into a customer using `$ref` in a path-item which currently according to the documentation here: https://spec.openapis.org/oas/v3.1.0#pathItemRef seems to be perfectly valid.

Its just the json schema here: https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.1/schema.json#L294 being `#/$defs/path-item` instead of `#/$defs/path-item-or-reference` that seems to be the problem, as json-schema validators are complaining about it.

But this PR and the discussion around it seems to be removing the ability to use references for path items, is that correct? Should I go back to the customer and tell them what they are doing isn't valid even though the documentation states otherwise?

Will the documentation be updated as part of this PR?

The use-case the customer has seems totally valid (individual teams manage different endpoints that are documented in different documents and they are trying to pull it all together into one document via the references), what would someone do to organize their document in this distributed manner if this feature goes away?

handrews:
@TristanSpeakEasy apologies for the delay in replying, I lost track of stuff during the holiday season.

This is _not_ changing anything about the specification.  The schema is not normative (meaning it does not have the force of the spec itself), so nothing changed in the schema impacts the spec.

This PR is confusing because Path Item's `$ref` is not a normal Reference Object for Historical Reasons™.  Instead, there's a `"$ref"` property in the Path Item Object schema, and it is _not_ combined with the Reference Object schema the way it is for other Objects that can be referenced.

ralfhandl (approved):

darrelmiller (approved):

PR #3364, opened 2023-08-30
Update schema.json

Bhush3092:


jdesrosiers:
This change is incorrect. Someone with the power to do so, please close this PR.

handrews:
@Bhush3092 the `type` keyword is not required by the OpenAPI specification.  I realize that some tools require it for certain things, but that is enforced by tools and not by the schema.  You can, of course, use a custom schema that is more strict if you want to use JSON Schema validation to check for such things.

PR #3380, opened 2023-09-24
the chart in section 4.8.12.4 shows that primitive and object types arere also valid for style=simple

karenetheridge:
this presumably should also be synced to the 3.1.1-dev and 3.2.0-dev branches.


darrelmiller (approved):

earth2marsh:
As discussed on today's TDC, merging based on approvals above.

PR #3381, opened 2023-09-25
Add double-int to formats registry

mikekistler:
This PR adds a new format, "double-int" to the OAI formats registry.

The double-int format should be used to specify an integer that can be stored in an IEEE 754 double-precision number without loss of precision.

Naming is one of the hardest problems in computer science. We considered many names other than "double-int" -- "int53", "jsonint", "safest", others. We even asked ChatGPT for suggestions.

In the end "double-int" seems most descriptive and accurate.

### DIFF
--- a/registries/_format/double-int.md
+++ b/registries/_format/double-int.md
@@ -1,0 +1,29 @@
+---
+owner: mikekistler
+issue: 
+description: an integer that can be stored in an IEEE 754 double-precision number without loss of precision
+base_type: integer
+layout: default
+---
+
+# <a href="..">{{ page.collection }}</a>
+
+## {{ page.slug }} - {{ page.description }}
+
+Base type: `{{ page.base_type }}`.
+
+The `{{page.slug}}` format represents an integer that can be stored in an IEEE 754 double-precision number without loss of precision. The range of values is -(2<sup>53</sup>)+1 to (2<sup>53</sup>)-1.
+
+This format is useful for systems that need to support languages (such as JavaScript) that store all numeric values as IEEE 754 double-precision numbers.
+
+{% if page.issue %}
+### GitHub Issue
+
+* [#{{ page.issue }}](https://github.com/OAI/OpenAPI-Specification/issues/{{ page.issue }})
+{% endif %}
+
+{% if page.remarks %}
+### Remarks
+
+{{ page.remarks }}
+{% endif %}


handrews:
I'm not a fan of "double-int". What's wrong with sticking with the existing naming pattern and using "int53"?

bterlson:
Far from an expert here but as someone who originally advocated for int53 and was convinced otherwise, I think there are two main problems with it:

1. A double can store integers in a larger range than a hypothetical signed int53 and very slightly less than a signed int54. A signed int53 would have a range of -2^52 to 2^52-1, a double is -2^53+1 to -2^53, and a 54 bit signed integer is -2^53 to 2^53-1. Note that int54 is actually closer, but unfortunately -2^53 is just outside its safe range.
2. int53 seems to indicate that the point of the integer is to store a number in 53 bits, but it's not - the point is to actually store it in 64 bits, just encoded as a double with some bits wasted.

mikekistler:
My recollection of the discussion on this item in the Sep 28 meeting is that no one really liked the "double-int" name but all the other alternatives suggested, "int53", "int54", "safeint", all seemed worse (save the half-serious suggestion of "doubloon"!). I think we agreed to give folks a week or so to suggest other alternatives and if none come forward then we'd go with "double-int".

Sorry I waited a month to capture this but hopefully this matches other folks memory. If not please respond here. But if this is accurate I'd like to make one last call for suggestions and if none come forward that we can agree is better than "double-int" we'll call this done.

baywet:
here is my suggestion: why don't we define a pattern instead by prefixing with the standard?
here the format would be ieee754-int53.
This way we only have to document the "valid" standard prefixes, and anybody can use the range under that.

mikekistler:
This PR was discussed in the TDC meeting on 11/9.

One point that was raised is that this type can be described using existing assertions, e.g.:
```
    type: integer
    minimum: -9007199254740991
    maximum: 9007199254740991
```

This is certainly true and works well for validation, but the intent of the format is not to express actual minimum and maximum values but rather as a guide for code generators on what type can be used to hold the value, in this case most especially in JavaScript.

Using format as a hint to generators is a well-established practice -- int32 and int64 are prime examples.

And using format rather than assertions for this purpose leaves the assertions available to express actual constraints on the value that are unrelated to how it is stored in a program.

ralfhandl:
> using format rather than assertions for this purpose leaves the assertions available to express actual constraints on the value that are unrelated to how it is stored in a program

This seems to indicate that the existing `format:"double"` is sufficient for expressing the storage requirement, and the value constraint "integer" can be expressed with the assertion `multipleOf:1`.

darrelmiller (approved):

Issue #1980, opened 2019-07-19
Support for structured-headers de/serialization


ioggstream:
## I expect

A way to express structured-headers serialization

- https://httpwg.org/http-extensions/draft-ietf-httpbis-header-structure.html

Here are some examples:

```
Example-IntegerHeader: 42
Example-BoolHdr: ?1
Example-StringHeader: "hello world"
Example-BinaryHdr: *cHJldGVuZCB0aGlzIGlzIGJpbmFyeSBjb250ZW50Lg==*
Example-StrListHeader: "foo", "bar", "It was the best of times."
Example-DictHeader: en="Applepie", da=*w4ZibGV0w6ZydGU=*

```
## Note

In the near future, all http headers will probably be de/serialized using structured-header specs.

handrews:
@ioggstream JSON Schema is defined over a [data model](https://tools.ietf.org/html/draft-handrews-json-schema-02#section-4.2), so as long as the structured header format can be unambiguously parsed into that data model, it would make sense to just use JSON Schema for that.

@darrelmiller @webron is this something we should note or clarify for 3.0.3 and/or 3.1?  I'm going to flag this under #2099 even though its pretty tangential to the JSON Schema draft update.  It is related to JSON Schema stuff that we have tried to make more explicit in recent drafts.

ioggstream:
Can define `structured-headers` as an alternative schema help?

darrelmiller:
It might be worth introducing a new `style` value called `structured-header` or something like that in the future.  The Schema object can describe the header value content and the style will clarify how it should be serialized.  I am still not seeing very many uses of structured headers so I don't think there is an urgency.  It is worth considering for 3.2.

ioggstream:
structured-headers is still WIP, so 3.2 is fine.

OTOH it could be interesting some feedback from the OAS community to  https://github.com/httpwg/http-extensions/blob/master/draft-ietf-httpbis-header-structure.md

Budda0ne:
> ## I expect
> 
> 
> 
> A way to express structured-headers serialization
> 
> 
> 
> - https://httpwg.org/http-extensions/draft-ietf-httpbis-header-structure.html
> 
> 
> 
> Here are some examples:
> 
> 
> 
> ```
> 
> Example-IntegerHeader: 42
> 
> Example-BoolHdr: ?1
> 
> Example-StringHeader: "hello world"
> 
[...]

PR #3390, opened 2023-10-03
Add structured fields primitive types to format registry

mikekistler:
This PR adds six new formats to the format registry for the primitive types defined in [RFC 8641] Structured Field Values for HTTP.

[RCF 8641]: https://www.rfc-editor.org/rfc/rfc8941

### DIFF
--- a/registries/_format/sf-binary.md
+++ b/registries/_format/sf-binary.md
@@ -2,6 +2,8 @@
 owner: mikekistler
 issue: 
 description: structured fields byte sequence as defined in [RFC 8941]
+source: https://www.rfc-editor.org/rfc/rfc8941#name-byte-sequences
+source_label: RFC 8941
 base_type: string
 layout: default
 ---
--- a/registries/_format/sf-binary.md
+++ b/registries/_format/sf-binary.md
@@ -1,0 +1,37 @@
+---
+owner: mikekistler
+issue: 
+description: structured fields byte sequence as defined by `sf-binary` in [RFC 8941]
+base_type: string
+layout: default
+---
+
+# <a href="..">{{ page.collection }}</a>
+
+## {{ page.slug }} - {{ page.description }}
+
+Base type: `{{ page.base_type }}`.
+
+The `{{page.slug}}` format represents a structured fields byte sequence as defined by `sf-binary` in [RFC 8941].
+
+```abnf
+sf-binary = ":" *(base64) ":"
+base64    = ALPHA / DIGIT / "+" / "/" / "="
+```
+
+A Byte Sequence is delimited with colons and encoded using base64 ([RFC 4648], Section 4).
+
+{% if page.issue %}
+### GitHub Issue
+
+* [#{{ page.issue }}](https://github.com/OAI/OpenAPI-Specification/issues/{{ page.issue }})
+{% endif %}
+
+{% if page.remarks %}
+### Remarks
+
+{{ page.remarks }}
+{% endif %}
+
+[RFC 8941]: https://www.rfc-editor.org/rfc/rfc8941#name-byte-sequences
+[RFC 4648]: https://www.rfc-editor.org/rfc/rfc4648#section-4
--- a/registries/_format/sf-binary.md
+++ b/registries/_format/sf-binary.md
@@ -1,7 +1,7 @@
 ---
 owner: mikekistler
-issue: 
-description: structured fields byte sequence as defined in [RFC 8941]
+issue:
+description: structured fields byte sequence as defined in [RFC8941]
 source: https://www.rfc-editor.org/rfc/rfc8941#name-byte-sequences
 source_label: RFC 8941
 base_type: string
@@ -14,14 +14,16 @@ layout: default
 
 Base type: `{{ page.base_type }}`.
 
-The `{{page.slug}}` format represents a structured fields byte sequence as defined in [RFC 8941].
+The `{{page.slug}}` format represents a structured fields byte sequence as defined in [RFC8941].
 
 ```abnf
 sf-binary = ":" *(base64) ":"
 base64    = ALPHA / DIGIT / "+" / "/" / "="
 ```
 
-A Byte Sequence is delimited with colons and encoded using base64 ([RFC 4648], Section 4).
+A Byte Sequence is delimited with colons and encoded using base64 ([RFC4648], Section 4).
+
+This format is appropriate for a header value that must conform to the {{page.slug}} structured field definition.
 
 {% if page.issue %}
 ### GitHub Issue
@@ -35,5 +37,5 @@ A Byte Sequence is delimited with colons and encoded using base64 ([RFC 4648], S
 {{ page.remarks }}
 {% endif %}
 
-[RFC 8941]: https://www.rfc-editor.org/rfc/rfc8941#name-byte-sequences
-[RFC 4648]: https://www.rfc-editor.org/rfc/rfc4648#section-4
+[RFC8941]: https://www.rfc-editor.org/rfc/rfc8941#name-byte-sequences
+[RFC4648]: https://www.rfc-editor.org/rfc/rfc4648#section-4
--- a/registries/_format/sf-binary.md
+++ b/registries/_format/sf-binary.md
@@ -1,7 +1,7 @@
 ---
 owner: mikekistler
 issue: 
-description: structured fields byte sequence as defined by `sf-binary` in [RFC 8941]
+description: structured fields byte sequence as defined in [RFC 8941]
 base_type: string
 layout: default
 ---
@@ -12,7 +12,7 @@ layout: default
 
 Base type: `{{ page.base_type }}`.
 
-The `{{page.slug}}` format represents a structured fields byte sequence as defined by `sf-binary` in [RFC 8941].
+The `{{page.slug}}` format represents a structured fields byte sequence as defined in [RFC 8941].
 
 ```abnf
 sf-binary = ":" *(base64) ":"
--- a/registries/_format/sf-boolean.md
+++ b/registries/_format/sf-boolean.md
@@ -1,0 +1,36 @@
+---
+owner: mikekistler
+issue: 
+description: structured fields boolean as defined by `sf-boolean` in [RFC 8941]
+base_type: string
+layout: default
+---
+
+# <a href="..">{{ page.collection }}</a>
+
+## {{ page.slug }} - {{ page.description }}
+
+Base type: `{{ page.base_type }}`.
+
+The `{{page.slug}}` format represents a structured fields boolean as defined by `sf-boolean` in [RFC 8941].
+
+```abnf
+sf-boolean = "?" boolean
+boolean    = "0" / "1"
+```
+
+A Boolean is indicated with a leading "?" character followed by a "1" for a true value or "0" for false.
+
+{% if page.issue %}
+### GitHub Issue
+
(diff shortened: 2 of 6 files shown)


registries/_format/sf-binary.md
```diff
@@ -0,0 +1,37 @@
+---
+owner: mikekistler
+issue: 
+description: structured fields byte sequence as defined by `sf-binary` in [RFC 8941]
```
mikekistler, line 4: I think I've put too much emphasis on the ABNF here, since RFC 8941 explicitly states
> If there is disagreement between the parsing algorithms and ABNF, the specified algorithms take precedence.
```suggestion
description: structured fields byte sequence as defined in [RFC 8941]
```

registries/_format/sf-binary.md
```diff
@@ -0,0 +1,37 @@
...
+layout: default
+---
+
+# <a href="..">{{ page.collection }}</a>
+
+## {{ page.slug }} - {{ page.description }}
+
+Base type: `{{ page.base_type }}`.
+
+The `{{page.slug}}` format represents a structured fields byte sequence as defined by `sf-binary` in [RFC 8941].
```
mikekistler, line 15:
```suggestion
The `{{page.slug}}` format represents a structured fields byte sequence as defined in [RFC 8941].
```

registries/_format/sf-boolean.md
```diff
@@ -0,0 +1,36 @@
+---
+owner: mikekistler
+issue: 
+description: structured fields boolean as defined by `sf-boolean` in [RFC 8941]
```
mikekistler, line 4:
```suggestion
description: structured fields boolean as defined in [RFC 8941]
```

registries/_format/sf-boolean.md
```diff
@@ -0,0 +1,36 @@
...
+layout: default
+---
+
+# <a href="..">{{ page.collection }}</a>
+
+## {{ page.slug }} - {{ page.description }}
+
+Base type: `{{ page.base_type }}`.
+
+The `{{page.slug}}` format represents a structured fields boolean as defined by `sf-boolean` in [RFC 8941].
```
mikekistler, line 15:
```suggestion
The `{{page.slug}}` format represents a structured fields boolean as defined in [RFC 8941].
```

registries/_format/sf-decimal.md
```diff
@@ -0,0 +1,36 @@
+---
+owner: mikekistler
+issue: 
+description: structured fields decimal as defined by `sf-decimal` in [RFC 8941]
```
mikekistler, line 4:
```suggestion
description: structured fields decimal as defined in [RFC 8941]
```

registries/_format/sf-decimal.md
```diff
@@ -0,0 +1,36 @@
...
+layout: default
+---
+
+# <a href="..">{{ page.collection }}</a>
+
+## {{ page.slug }} - {{ page.description }}
+
+Base type: `{{ page.base_type }}`.
+
+The `{{page.slug}}` format represents a structured fields decimal as defined by `sf-decimal` in [RFC 8941].
```
mikekistler, line 15:
```suggestion
The `{{page.slug}}` format represents a structured fields decimal as defined in [RFC 8941].
```

registries/_format/sf-integer.md
```diff
@@ -0,0 +1,37 @@
+---
+owner: mikekistler
+issue: 
+description: structured fields integer as defined by `sf-integer` in [RFC 8941]
```
mikekistler, line 4:
```suggestion
description: structured fields integer as defined in [RFC 8941]
```

registries/_format/sf-integer.md
```diff
@@ -0,0 +1,37 @@
...
+layout: default
+---
+
+# <a href="..">{{ page.collection }}</a>
+
+## {{ page.slug }} - {{ page.description }}
+
+Base type: `{{ page.base_type }}`.
+
+The `{{page.slug}}` format represents a structured fields integer as defined by `sf-integer` in [RFC 8941].
```
mikekistler, line 15:
```suggestion
The `{{page.slug}}` format represents a structured fields integer as defined in [RFC 8941].
```

registries/_format/sf-string.md
```diff
@@ -0,0 +1,41 @@
+---
+owner: mikekistler
+issue: 
+description: structured fields string as defined by `sf-string` in [RFC 8941]
```
mikekistler, line 4:
```suggestion
description: structured fields string as defined in [RFC 8941]
```

registries/_format/sf-string.md
```diff
@@ -0,0 +1,41 @@
...
+layout: default
+---
+
+# <a href="..">{{ page.collection }}</a>
+
+## {{ page.slug }} - {{ page.description }}
+
+Base type: `{{ page.base_type }}`.
+
+The `{{page.slug}}` format represents a structured fields string as defined by `sf-string` in [RFC 8941].
```
mikekistler, line 15:
```suggestion
The `{{page.slug}}` format represents a structured fields string as defined in [RFC 8941].
```

registries/_format/sf-token.md
```diff
@@ -0,0 +1,35 @@
+---
+owner: mikekistler
+issue: 
+description: structured fields token as defined by `sf-token` in [RFC 8941]
```
mikekistler, line 4:
```suggestion
description: structured fields token as defined in [RFC 8941]
```

registries/_format/sf-token.md
```diff
@@ -0,0 +1,35 @@
...
+layout: default
+---
+
+# <a href="..">{{ page.collection }}</a>
+
+## {{ page.slug }} - {{ page.description }}
+
+Base type: `{{ page.base_type }}`.
+
+The `{{page.slug}}` format represents a structured fields token as defined by `sf-token` in [RFC 8941].
```
mikekistler, line 15:
```suggestion
The `{{page.slug}}` format represents a structured fields token as defined in [RFC 8941].
```

mikekistler (commented):
De-emphasize the ABNF.

registries/_format/sf-integer.md
```diff
@@ -0,0 +1,37 @@
+---
+owner: mikekistler
+issue: 
+description: structured fields integer as defined in [RFC 8941]
+base_type: integer
```
darrelmiller, line 5: Should be number

registries/_format/sf-decimal.md
```diff
@@ -0,0 +1,36 @@
+---
+owner: mikekistler
+issue: 
+description: structured fields decimal as defined in [RFC 8941]
+base_type: integer
```
darrelmiller, line 5: number

darrelmiller:
@mikekistler  These look good but we should also talk about what to do with sf-list and sf-dictionary and more complex headers.

handrews:
@mikekistler I'd like to propose an alternative that encompasses sf-list and sf-dictionary with a different approach.  I should have the proposal up by next week's TSC meeting, although probably not in the next few days.

registries/_format/sf-integer.md
mikekistler: Why base type for `sf-integer` should be `number`? I know it is restricted to integers that fit in a double, but semantically it is an integer. If "format" is just "extra information", then shouldn't we make the base type as specific as possible?

registries/_format/sf-binary.md
```diff
@@ -0,0 +1,37 @@
+---
+owner: mikekistler
+issue: 
+description: structured fields byte sequence as defined in [RFC 8941]
```
mikekistler, line 4: Is there a convention for references to RFCs? I'm using "RFC 8941" with a space between "RFC" and the number, but I see many places that omit that space, e.g. "RFC8941". Should I be using this instead?

handrews: I slightly prefer the space, but I don't have strong feelings.  It is clear either way.

registries/_format/sf-integer.md
handrews: @mikekistler `integer` is not a type in the JSON data model, it is just a convenience option in  JSON Schema's `type` keyword.  Formats are tied to a data model type, not the `type` keyword (this is a common misunderstanding).

mikekistler: I've added `number` but kept `integer` because it is an allowed type in the OAS vocabulary.

handrews:
@mikekistler My apologies for the extremely long delay while I dealt with health issues (and end-of-year holidays).

My concern here is that we have a smooth path to full RFC 8941 in OAS 3.2, as suggested by #1980.  I have some thoughts on what that might look like, although I'm still working out the details.  I will add them to #1980 when I get it sorted out.  Now that I'm able to work properly again, and have mostly caught up on my backlog, I expect to finish that proposal in the next week.

If these are intended only for use with headers with single-item values, I might be OK with them.  There will need to be some wording around how you can't just say `type: array, items: {format: sf-integer}` and have that work on a List-valued structured field.  Because my concern with these is that people will think they can automatically use JSON Schema object/list structures with header values that are Structured Field Lists or Dictionaries, and that is not going to be the case.

mikekistler:
@handrews My hope was to have a solution for primitive types in headers -- particularly strings -- without creating any impediments to a solution for full RFC 8941 support.

Are there things in the current PR that would be an impediment to the full solution?

handrews:
@mikekistler I think I'm mostly worried about other people running wild with these, doing something unexpected, and then we get stuck having to either support that (because it's "in the wild") or annoy people by breaking something that they think ought to work.

We could just add another sentence to these:

> The `{{page.slug}}` format represents a structured fields byte sequence as defined in [RFC 8941].

along the lines of "This format is currently only intended for use describing single-Item headers; any other usage may not be compatible with future full support for structured fields."

(replacing "single-Item headers" with whatever terminology RFC 8941 mandates).

That way we set boundaries around this and they don't become general-purpose formats that people start using because they're close enough to some other use case they want.  Or at least they know not to expect that to work in the long run.

Does that sound reasonable?

registries/_format/sf-binary.md
ralfhandl: According to the [RFC Style Guide](https://www.ietf.org/archive/id/draft-flanagan-7322bis-07.html#name-citations) it's with a space in plain text, and without space in a citation label within square brackets:

> Example: "See RFC 2119 [[BCP14](https://www.ietf.org/archive/id/draft-flanagan-7322bis-07.html#BCP14)] for more information."

handrews:
@mikekistler ping!  Do you have thoughts on the proposed resolution in my last comment?

mikekistler:
I'm struggling with this because it's just not clear to me that it's necessary. We don't have language like this in any of the other format registry entries.

But for the sake of moving this forward, let me propose a slightly different wording that I think I could live with:

> This format is appropriate for a header value that must conform to the {{page.slug}} structured field definition.

Can we compromise on this language?

handrews:
@mikekistler Yes, all I'm trying to do is ensure it's used only for the appropriate kind of structured header and not treated as a general-purpose format, and your wording does that.  I wouldn't even call it a compromise- I don't much care about the wording, just the results.

mikekistler:
Excellent! I'll make the updates and also fix the RFC references and I think that will address all the outstanding concerns.

mikekistler:
I think all PR comments are addressed. Tagging reviewers for re-review.

ralfhandl (approved):

lornajane (approved):
Thank you @mikekistler  and @handrews for getting this done!

PR #3398, opened 2023-10-08
switch the order of these styles in the tables

karenetheridge:
This is a more natural grouping of similar types, making the data much easier to read.


darrelmiller:
Thank you for the update, however, these changes need to be applied to the 3.1.1.md file, not the 3.1.0.md file.

karenetheridge:
Files fixed. (view with whitespace changes ignored.)

darrelmiller (approved):

PR #3416, opened 2023-10-23
extra keywords in the reference object are permitted (v3.1.0)

karenetheridge:
see https://spec.openapis.org/oas/v3.1.0#fixed-fields-18


handrews (approved):
Yup, this change matches how 3.0 interprets "SHALL be ignored".

darrelmiller (approved):

PR #3435, opened 2023-11-06
- updates agenda workflow so it closes older agenda issues automatically

baywet:
related #3420

### DIFF
--- a/.github/workflows/agenda.yaml
+++ b/.github/workflows/agenda.yaml
@@ -23,6 +23,7 @@ jobs:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       TITLE_PREFIX: "Open Community (TDC) Meeting, "
       LABEL: "Housekeeping"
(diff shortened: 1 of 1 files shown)


lornajane:
How long will it leave the issues open for? I would suggest something over 2 weeks, so we can always look at the previous meeting notes, even if we skip a meeting one week.

baywet:
@lornajane right now it closes the issue as soon as the date in the title is passed, but runs only every week, so that should give us a week effectively.
We could introduce a delay if the TDC decides it.

baywet:
Notes from the meeting:
- wait 10 days before close the issue.
- look into unpinning the issue as we close them.

baywet:
@lornajane @darrelmiller @MikeRalphson @webron : 
- added the delay of 10 days before closing the agenda item
- added pinning/unpinning the issues

I think this is ready to merge.

darrelmiller (approved):

earth2marsh (approved):

PR #3437, opened 2023-11-06
- adds a workflow to handle stale issues

baywet:
related #3420
TODO:
- [x] update the delay after which we send a warning to the author
- [x] update the delay after which we close the issue (after sending the warning)
- [x] update and/or create the labels
- [x] update the message to be sent as the reminder
- [x] decide whether we want the needs author feedback label to be added automatically based on a list of *experts*.

### DIFF
--- a/.github/workflows/inactive-issues.yml
+++ b/.github/workflows/inactive-issues.yml
@@ -1,0 +1,34 @@
+on:
+  issues:
+    types: labeled
+  workflow_dispatch:
+  schedule:
+    - cron: '*/5 * * * *'
+
+permissions:
+  issues: write
+
+name: Label and close issues with no recent activity
+
+env:
+  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  NEEDS_ATTENTION_LABEL: "Needs attention"
+  NEEDS_AUTHOR_FEEDBACK_LABEL: "Needs author feedback"
+  NO_RECENT_ACTIVITY_LABEL: "No recent activity"
+  NO_RECENT_ACTIVITY_DURATION: "00:00:01"
+  NO_RECENT_ACTIVITY_DURATION_CLOSE: "00:00:01"
+  ORG_NAME: ${{ github.repository_owner }}
+  REPO_NAME: ${{ github.event.repository.name }}
+  NO_RECENT_ACTIVITY_COMMENT: "This issue has been labeled because there has been no recent activity. It will be closed if no further activity occurs within 1 day."
+
+
+jobs:
+  run:
+    runs-on: ubuntu-latest
+    name: Label issues with no recent activity
+    steps:
+      - uses: actions/checkout@v4
+      - run: scripts/label-no-recent.ps1
+        shell: pwsh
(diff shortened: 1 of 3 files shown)


darrelmiller:
Delay after which we send a warning to the author:  7 days
Delay after which we close the issue (after sending the warning): 28 days
Update and/or create the labels: "Needs author feedback" "No  recent activity" "Needs attention"
Message to be sent as the reminder: Add the name of the label.  Update the days. Add comment to acknowledge that author can reopen if they want to.
Do we want the needs author feedback label to be added automatically based on a list of experts?  Let's not do this for the moment. We will manually add the labels.

baywet:
@darrelmiller @lornajane @webron @MikeRalphson : I updated the delays and the messages. This is ready to merge.
After this gets merged, the labels will need to be created (let me know if you need a script for that), and we'll need to start labeling older issues (again scripting can help with this).

darrelmiller (approved):

earth2marsh:
Per today's TDC discussion, it's worth having these labels even if we don't have the label names "just right" so far (we can revisit later). 

Adding my own thoughts... 

- `Needs author feedback` works, as it indicates who and what is needed
- `No recent activity`  seems more like a quality than an actionable label. I think it means that without more community interest, this is likely to be parked or closed? Ideally the name suggests either that this has entered an issues lifecycle category or that an specific action is looming.  Some names that come to mind are "stale" (noted on the call has a negative connotation), "Auto-close candidate",  or even "Low priority" or "Un-prioritized". 
- `Needs attention` invites the question of whose attention (not the author, presumably) or what kind of attention is necessary ("triage", "review", for "discussion" etc.).

baywet:
@earth2marsh Thanks for reviewing and merging the workflow
-  `No recent activity`: is additive to `Needs author feedback` when it gets added. So hopefully the later still prompts people to do something. Also when `No recent activity` is added a comment is also added to the issue asking the author to reply or else the issue will be closed.
- `Needs attention`: it's deliberate not to use triage because at this point the issue as been triaged and replied to by the team at least once. We could specialize it further to something like `Needs reviewers attention` if you want to make it clearer.

earth2marsh:
So `No recent activity` then seems like a kind of escalation that requires the author's participation? And is it removed automatically when the author comments? What about `Auto-close candidate` in that case, as it suggests what will happen in the event the author stays absent?

`Needs review` or `TDC ready` or something that indicates better who-must-do-what would help me. :)

One more thought... we have a decent number of labels, so might it help to prefix these workflow labels in order to group them together? Are there good practices we can borrow from other projects here?

baywet:
Yes, no recent activity is an escalation of needs author feedback. I can't remember whether you were on the TDC last week, but you can watch the recording where I walked the audience through the process.

Yes, I'm not dead set on the labels, this is why they can be changed in the workflow simply through changing the environment variables. My only recommendation is not to mix those labels with the triage ones to avoid overcrowding triage.

In terms of good practices, these initial suggestions are borrowed from what we use at Microsoft, and I'm sure they went through thorough political correctness review, as well diversity review (so they are clear to non-native English speakers like me).

PR #3453, opened 2023-11-16
- fixes a bug where auto close issues would be too eager

baywet:
follow up to #3437 because date math is hard


darrelmiller (approved):

baywet:
@earth2marsh for review and merge, apologies about the defect here.

PR #3454, opened 2023-11-17
- adds documentation for issues automation

### LINKED ISSUE #3430: Investigate automation of closing issues that are labeled as "enhancement" but have not gained sufficient community interest
- We need to determine what heuristic to use to identify "not sufficient interest"  e.g.  How long since the last comment? How many thumbs up vs thumbs down? By who? etc.

### LINKED ISSUE #3452: Update DEVELOPMENT.md to explain labels and processes
Following the merge of  #3437 , might we want to update the DEVELOPMENT.md to explain our labeling system and process, including what is automated and what is manual?

baywet:
fixes #3452

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -20,11 +20,52 @@ The specification *will evolve over time*. Changes  may be made when any of the 
 
 * Impact. A change will provide impact on a large number of use cases. We should not be forced to accommodate every use case. We should strive to make the *common* and *important* use cases both well supported and common in the definition of the OAI Spec. We cannot be edge-case driven.
 
+## Automated closure of issues Process
+
+In an effort to keep the list of issues up to date and easier to navigate through, issues get closed automatically when they become inactive.
+
+The following flowchart describes the automation process:
+
+```mermaid
+flowchart TD
+  openIssue["anyone opens an issue"]
+  reviewerReplies["a reviewer replies to the issue"]
+  authorRepliesWithin7Days{"author of the issue replies within 7 days"}
+  authorRepliesWithin28Days{"author of the issue replies within 28 days"}
+  reviewerAddsLabel["reviewer adds the Needs author feedback label"]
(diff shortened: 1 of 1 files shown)


lornajane (changes requested):
Thanks for adding this information! I think it needs to go a lot further down the page since the automation (while excellent) is not key to our mission - perhaps between Transparency and Participation? Also is there a way we can condense the diagram, it's more than a whole screen tall for me and while it's a good representation of how the code works, I'm not sure this is the information that users need.

I'd be tempted to add a list of the labels and what they mean.

baywet:
just pushed some changes to implement the suggestions.
as for the flowchart, I'm not sure we can control mermaid diagram sizes besides with CSS which GitHub is not going to allow us to use on a markdown page. (but would if this ended up statically generated on a site).
maybe we could create a new markdown document for technical diagrams of automations or processes?

baywet:
ping @lornajane

lornajane:
Thanks for moving the information to a better location in the document. I'm still not sure the mermaid diagram adds a lot for the amount of screen space it takes - the label descriptions are very good though!

baywet:
@lornajane did you see my suggestion to adding a new page dedicated to the diagram?

lornajane:
I don't feel strongly about it - probably nobody would look at the diagram if we moved it, so maybe we can just remove it and see how we get on.

One question, when I reply to an issue that has "needs attention", do I manually remove that label when I do that?

baywet:
To answer your question: yes when you reply you manually need to remove the "needs attention" label and add the "needs author feedback" one. We could also automate that, but I'd need a list of "experts" GitHub handle for the people who qualify for this automation (we don't want it to trigger for everyone).

Ok I'll remove the diagram, I found it's usually a good way to help people understand the logic of a workflow rather than lots of text.

baywet:
updated. I also added information to when labels are added removed in their descriptions since we don't have the diagram anymore.

darrelmiller (approved):

Issue #3256, opened 2023-04-24
Formally define (ABNF) 3.x URL templating behavior


handrews:
As noted in Slack discussions, there is no formal specification for where variables can appear in server URL templates and path templates.  There does not seem to be any reason to place restrictions on it.

Ideally, we could leverage the ABNF of RFC 6570 URI Templates even though OAS 3.x templating is _not_ the same as RFC 6570 in general.  (@darrelmiller @webron @MikeRalphson any concerns here?)

handrews:
Issue #2119 brings up a lot of points relevant to this.

char0n:
@handrews I created a implementation with formal ANBF grammar in https://github.com/char0n/openapi-path-templating.

I did a couple of assumption, because of the fact that the templating is not defined very precisely. Hope it helps a bit.

handrews:
@char0n thank you!  I don't know if we're going to get to this in the current patch releases or not (I'm unassigning it for onw as I have too many things in my backlog), but this will help when we do get there!

handrews:
@char0n I took a look at your ABNF- it looks like it restricts path template variables to whole segments.  This issue got filed because in a discussion with TSC folks on slack, it was decided that template variables are _not_ restricted to whole segments.  You can have a template like `/foo{bar}/whatever`.  I don't know how many API descriptions actually do that, but I figured you might want to know since the initial comment in this issue was not sufficiently clear.

char0n:
@handrews I've created implementation with formal ABNF grammar for Server URL Templating in https://github.com/char0n/openapi-server-url-templating.

It's distinct from https://github.com/char0n/openapi-path-templating, as the Server URL Templating is based on RFC 6570, but Path Templating is more based on RFC 3986.

char0n:
@handrews, 

There is one un-clarity related to `template-expression-param-name` rule from `openapi-path-templating`. OpenAPI spec doesn't technically constrain `Parameter Object`.`name` fixed field in any way. It just says it's a string. This means that people can use `"parameter name"` as a value.

This also means that the path template will need to be defined as `/{parameter name}`

It's just another example how OpenAPI Path Templating differs from RFC 6570. 

---

@handrews taking into account the above, `template-expression-param-name` should be less constrained and in theory should allow any non empty string value. What do you think?

LasneF:
The key point here is that [RFC ](https://datatracker.ietf.org/doc/html/rfc6570)  is defining a way to represent template 
[...]

Issue #151, opened 2014-10-13
Host schema publicly


webron:
In order to accomplish #103 elegantly, we should host the json schema for the 2.0 spec under the swagger.io domain. Probably something like http://swagger.io/schemas/v2.0/schema (based on json-schema's id declaration `"id": "http://json-schema.org/draft-04/schema#"`).

handrews:
@webron I think we agreed this should be the issue tracking how/when we deploy schema updates.  Including deploying new revisions of the OAS 3.1 JSON Schema vocabulary and dialect  meta-schemas which don't currently have dates in their `$id`s.  Which I think happened because I wrote them as tentative proposals and then by the time OAS 3.1 shipped I wasn't around to catch that sort of thing.

jdesrosiers:
I've been handling schema deployment. I don't usually prepare a deployment for every change, but tend to wait a little while in case more changes come in (they usually come in waves). There are currently two changes that have been merged but not deployed yet and one open PR.

Here's the process I've been following for deployment
1. Update release dates in
    * schema.yaml
    * schema-base.yaml (Don't forget the one in the `$ref`)
    * tests/v3.1/test.js
2. Generate the JSON versions from the YAML versions
    * The README says to only modify the YAML version, but contributors usually include changes to both. So, this step is usually just a sanity check.
3. Run the tests
    * The tests are very minimal and haven't been added to since they were first added, but I make sure that we at least don't break what we have.
3. Deploy the schemas
    * Add the schemas to the `gh-pages` branch
    * Update the `latest` symlink to point to the newest version
4. Other stuff I'm forgetting about ???

Here's an example of a deploy I've done in the past
* https://github.com/OAI/OpenAPI-Specification/pull/3042
* https://github.com/OAI/OpenAPI-Specification/pull/3043

>  Including deploying new revisions of the OAS 3.1 JSON Schema vocabulary and dialect meta-schemas which don't currently have dates in their `$ids`. Which I think happened because I wrote them as tentative proposals and then by the time OAS 3.1 shipped I wasn't around to catch that sort of thing.

Actually, I wrote those meta-schemas. I didn't include a date in the schema for the same reasons we don't include one in JSON Schema meta-schemas. The `$id`s of meta-schemas are what schemas authors use in `$schema` to indicate the version of JSON Schema their schema is using. If we version meta-schemas, users would need to update all of their schemas that declare this dialect every time we release a bugfix. There are lots of reasons that's not a good user experience.

[...]

PR #3455, opened 2023-11-20
Drop `uri-reference` format for template URLs

davishmcclurg:
`Server.url` and `Link.operationRef` both allow variable substitution with {brackets}, which means they're not always valid URI references.

For example, the [current specification][0] shows
`https://{username}.gigantic-server.com:{port}/{basePath}` as a Server Object `url`, but it's not a valid URI reference because the host includes curly brackets.

[`operationRef`][1] similarly includes
`https://na2.gigantic-server.com/#/paths/~12.0~1repositories~1{username}/get` as an example that isn't valid using the `uri-reference` format.

I looked into the other uses of `uri-reference` and they seemed ok.

I'm not sure how to update the actual JSON schemas but [readme](https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.1/README.md#contributing) says to just update the yaml files, so let me know if there's anything else to do.

Related:
- https://github.com/OAI/OpenAPI-Specification/pull/2586
- https://github.com/OAI/OpenAPI-Specification/pull/3235
- https://github.com/OAI/OpenAPI-Specification/issues/3256
- https://github.com/davishmcclurg/json_schemer/issues/158

[0]: https://spec.openapis.org/oas/v3.1.0#server-object-example
[1]: https://spec.openapis.org/oas/v3.1.0#operationref-examples

### DIFF
--- a/schemas/v3.0/schema.yaml
+++ b/schemas/v3.0/schema.yaml
@@ -975,7 +975,6 @@ definitions:
         type: string
       operationRef:
         type: string
-        format: uri-reference
       parameters:
         type: object
         additionalProperties: {}
--- a/schemas/v3.1/schema.yaml
+++ b/schemas/v3.1/schema.yaml
@@ -119,7 +119,6 @@ $defs:
     properties:
       url:
         type: string
-        format: uri-reference
       description:
         type: string
       variables:
@@ -632,7 +631,6 @@ $defs:
     properties:
       operationRef:
         type: string
-        format: uri-reference
       operationId:
         type: string
       parameters:


handrews:
Thanks, @davishmcclurg !  This is more or less a duplicate of #3235, which got a bit bogged down in questions over what format to use (and add to the format registry) for these things.

@MikeRalphson it might be good to go ahead and accept this one to at least stop implementations that try to enforce the existing format from breaking on this.  Assuming we can actually work out the logistics of publishing a schema update soon.

ekzobrain:
This also enables support for IDNs. Current format "uri-reference" does not allow them.

handrews:
@ekzobrain OAS 3.x does not have direct IRI support (I am hopeful it will be in OAS 4 "Moonwalk").  IRIs are only supported by being encoded down to URIs.

handrews (approved):
I think it's worth merging this even though there's another PR that overlaps with it.  This one is simpler.  Although we still need to solve the schema deployment issues discussed in #151.

darrelmiller (approved):

karenetheridge:
Thanks for this! I just ran into this issue when implementing servers url matching in my implementation.

It would be good to see this published so I can make use of it :)

handrews:
For those following the schema publishing topic who haven't otherwise seen it, I pulled the still-relevant schema publishing challenges out of #151 (which I have now closed) and filed them as #3715 (see also #3716 for related schema development process concerns).

PR #3491, opened 2024-01-07
fix types to use an array rather than a string when multiple types are valid

karenetheridge:
Programmatic users of api/format.json will expect to see an array here, not a string containing an invalid type.


handrews (approved):
Thanks!

Issue #3513, opened 2024-01-22
Meta schema 3.1 is wrong for path / pathItem


LasneF:
Accordinf to  https://spec.openapis.org/oas/latest.html#fixed-fields-6 

Path item Object should support $ref as in the snippet below 

```
paths:
    /orders:
         $ref: paths/orders.yaml
```

it looks not supported in the meta schema 
https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.1/schema.yaml#L211 

to me the line 209 is wrong 
```
      '^/':
        $ref: '#/$defs/path-item'
```
        
 but should be 
```
      '^/':
        $ref: '#/$defs/path-item-or-reference:'
```

this has been found here : https://github.com/daveshanley/vacuum/issues/422

handrews:
Thanks, @LasneF.  This is a dup of #3298 and is addressed by PR #3355, which (like all not-yet-deployed schema PRs — technically only the Schema Object portion is a metaschema) is awaiting proper resolution to https://github.com/OAI/OpenAPI-Specification/issues/151#issuecomment-1556268486.  Which we expect to actually have fairly soon as it's part of the big clean-up we're doing on the issue/PR/process backlog as discussed in the TDC calls this month.

LasneF:
not 100 % sure read the PR correctly path-item-or-reference: is removed 
so does it means 

* that $ref syntax provided is not be supported  , ie having path item definition in dedicated files is forbidden  
using $ref more or less like an include the targeted files containing only content of path-item 

or 

* the $ref is by default supported no need to have dedicated syntax for it 


should the docuementation  https://spec.openapis.org/oas/latest.html#fixed-fields-6  being updated ?

handrews:
@LasneF PathItem is a special case for referencing which uses a different syntax/semantics for adjacent properties.  So PR #3355 [deletes `#/$defs/path-item-or-references`](https://github.com/OAI/OpenAPI-Specification/pull/3355/files#diff-74cf27ff990ec1185141e72e3fd8ec6f251530e156309460ca8d9253abf234bfL248), [adds `$ref` directly to `#/$defs/path-item`](https://github.com/OAI/OpenAPI-Specification/pull/3355/files#diff-74cf27ff990ec1185141e72e3fd8ec6f251530e156309460ca8d9253abf234bfR217), and replaces `#/$defs/path-item-or-reference` with `#/$defs/path-item` wherever it had appeared.

PR #3514, opened 2024-01-22
handle$ref for path item

LasneF:
handle path item in reference 

intend to fix 
https://github.com/OAI/OpenAPI-Specification/issues/3513


handrews:
This is already addressed by PR #3355, which handles the unusual requirements for how `$ref` functions in the Path Item Object.  Unlike other Objects, you can't actually use the `-or-reference` pattern for Path Items.

PR #3521, opened 2024-01-25
- removes the mention of the agenda backlog from the template

baywet:
@lornajane here is the change you suggested :)


handrews:
Superseded by #3522

PR #3207, opened 2023-03-20
docs(3.1.0): :memo: Fix relative broken links and address some of the markdownlint complaints

CandiedCode:
- [x] fix relative links
- [x] address some of the markdownlint complaints

PR #3408, opened 2023-10-19
fix: GitHub changed markdown rendering for headings/anchors

MikeRalphson:
Closes #3207 #3160 #3136 #3128 #3029 #3197 #3132

This PR mechanically corrects for the change made by GitHub to its markdown renderer which no longer honours the mixed markdown/html form:

`### <a name="myAnchor"></a> My Anchor`

(the use of `name` or `id` makes no difference), and only links to the **implicit** anchor `#my-anchor` now work.

This PR affects the `main` branch, so published versions of the OAS only. If merged, follow-up PRs will be raised for in-flight `-dev` branch versions.

An example of the result of this PR can be previewed [here](https://github.com/MikeRalphson/OpenAPI-Specification/blob/anchor-fixes/versions/3.1.0.md).

Issue #3197, opened 2023-03-09
Documentation links not working


vcarballo:
None of the local links in the https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md file work.

MikeRalphson:
Thanks, we're aware and we were discussing it on today's call. We have agreed what we need to do to fix.

gevial:
Same for version 3.0.1.

MikeRalphson:
It affects all the markdown files, same cause.

handrews:
We discussed this in the [March 9th meeting](https://github.com/OAI/OpenAPI-Specification/issues/3183#issuecomment-1456609127) and decided to fix these things in place.  @webron was going to figure out the exact logistics, which has not happened by the time I commented in the agenda for the [April 27th meeting](https://github.com/OAI/OpenAPI-Specification/issues/3253#issuecomment-1523800012).  AFAIK, this last detail was not resolved.

One difficulty was the range of PRs from various people that needed tweaks, but the PR submitters had moved on due to lack of activity.

Since it's been a good 6 months without movement, I'm going to consolidate the PRs myself.

handrews:
Consolidating all broken link issues into issue #3054 and PR #3408

PR #3160, opened 2023-02-13
Update Relative Links to Sections

kexoth:
The relative links to sections in the markdown document are using camel case.

As I have checked the section links generated from Github use kebab case.

In the current contribution updated all the references in the document to use `kebab-case`.

PR #3548, opened 2024-02-02
fix: GitHub changed markdown rendering for headings/anchors

handrews:
This is a rebase/re-submit of @MikeRalphson's exact commit from PR #3408 as he is not currently available to rebase it (the markdown checks were broken at that point on main- they should be fixed now).

Here is that PR's text:

-----

Closes #3207 #3160 #3136 #3128 #3029 #3197 #3132

This PR mechanically corrects for the change made by GitHub to its markdown renderer which no longer honours the mixed markdown/html form:

`### <a name="myAnchor"></a> My Anchor`

(the use of `name` or `id` makes no difference), and only links to the **implicit** anchor `#my-anchor` now work.

This PR affects the `main` branch, so published versions of the OAS only. If merged, follow-up PRs will be raised for in-flight `-dev` branch versions.

An example of the result of this PR can be previewed [here](https://github.com/handrews/OpenAPI-Specification/blob/anchor-fixes/versions/3.1.0.md). _[Note: points to @handrews's fork because of the re-submit]_

### DIFF
--- a/versions/2.0.md
+++ b/versions/2.0.md
@@ -24,10 +24,10 @@ Version | Date | Notes
 
 ## Definitions
 
-##### <a name="pathTemplating"></a>Path Templating
+##### Path Templating
 Path templating refers to the usage of curly braces ({}) to mark a section of a URL path as replaceable using path parameters.
 
-##### <a name="mimeTypes"></a>Mime Types
+##### Mime Types
 Mime type definitions are spread across several resources. The mime type definitions should be in compliance with [RFC 6838](http://tools.ietf.org/html/rfc6838).
 
 Some examples of possible mime type definitions:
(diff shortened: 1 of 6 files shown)


handrews:
So, what's going on here is this (all done by a script written by Mike Ralphson):

* camelCase anchors that were created within section headings, like `name="pathItemObject"`, no longer work, so:
    * they have been removed from the headings
    * uses have been converted e.g. from `href="#pathItemObject"` to `href="#path-item-object"`
* anchors that are declared as camelCase _outside_ of headings, like `name="dialectSchemaId"`, still work with `href="#dialectSchemaId"` and are left alone

You can see the rendered result at https://github.com/handrews/OpenAPI-Specification/tree/anchor-fixes/versions

I've skimmed through these and it looks like the updates are correct.  I skimmed slowly enough for 3.1.1 that I am fairly sure nothing else was touched (except trimming a bit of trailing whitespace on some lines, which is fine).

darrelmiller (approved):

PR #3551, opened 2024-02-05
Update SIGs list, improve formatting and links

handrews:
This updates the list with newer SIGs, which I also checked against the list of "sig-" channels on Slack.  It also adds some links to things that are mentioned, and makes what I consider to be formatting improvements.

I added a title and a bit of text at the top, but if someone wants to more properly define SIGs or whatever processes we have around them, I'd be happy to replace the text (or someone else can do it as a follow-on PR).

### DIFF
--- a/SPECIAL_INTEREST_GROUPS.md
+++ b/SPECIAL_INTEREST_GROUPS.md
@@ -1,43 +1,78 @@
-## Codegen
-> code-gen vocabulary special interest (working) group
+# OpenAPI Special Interest Groups (SIGs)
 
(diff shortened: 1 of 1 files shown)


LasneF:
wondering if the sig group on finance should not be archived as no activity on github as well on slack

handrews:
@LasneF we don't have a process for archiving SIGs yet- it would be worth discussing.  For now let's just get the list accurate, then we can make changes.

darrelmiller (approved):

PR #3554, opened 2024-02-07
Always checkout JavaScript environment from main

### LINKED ISSUE #3542: vallidate-markdown failure: How should we maintain scripts across branches?
The validate-markdown check is currently failing on all branches but main because main now has a _very_ different JavaScript configuration from the branches.  Those branches should have GitHub actions that match their environments, but currently on v3.1.1-dev and v3.2.0-dev, the JavaScript package that validate-markdown runs fails even if you just ask it for help output:

```console
(.venv) handrews@ixat OpenAPI-Specification % npx mdv versions/3.*.md
Unexpected token '.'
(.venv) handrews@ixat OpenAPI-Specification % npx mdv --help         
Unexpected token '.'
(.venv) handrews@ixat OpenAPI-Specification %
```

handrews:
Fixes #3542 

Workflows run from the workflow file on main, even when they run on branches.  But just running "npm i" or a command like "npx" that also installs packages uses the package.json file from the branch.

Rather than attempt to keep the branch package files up-to-date (which would _quadruple_ the number of update PRs, adding three branch update PRs for every dependabot PR), let's just checkout the package.json from main whenver we run.  The other workflows only run on main - these two are the ones that run on branches.

### DIFF
--- a/.github/workflows/schema-tests.yaml
+++ b/.github/workflows/schema-tests.yaml
@@ -20,12 +20,16 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v1 # checkout repo content
+    - uses: actions/checkout@v4 # checkout repo content
+      with:
+        fetch-depth: 0
     - uses: actions/setup-node@v4 # setup Node.js
       with:
         node-version: '20.x'
-    - name: Install dependencies
-      run: npm i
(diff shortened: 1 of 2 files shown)


darrelmiller (approved):

earth2marsh:
merging as discussed on today's TDC call

Issue #3561, opened 2024-02-08
Use community repo for governance, info, & user/dev help questions


handrews:
_[**NOTE**: I'm filing this issue in this repo because it's about things that currently happen here.]_

In today's call we rediscovered the [OAI/community](https://github.com/OAI/community) repository.

* It has a lot of useful information, but some of it is out of date, and there's significant overlap with things in _this_ repository.
* It would be the ideal place to route user / implementor questions that are a huge percentage of the issues and discussions here in the spec repo (@lornajane kind of like our discussion of moving them to the learn repo, but even better because then we can just use learn issues for actual docs changes)
* It seems like a better home for our general governance and process work.  Perhaps the `Process` label can move over there, and we keep `Housekeeping` here for the day-to-day repo/project specific stuff?  (paging @earth2marsh from our Process vs Housekeeping discussion)

Note that @webron transferred OAI/community#5 over to that repo as a community question, so apparently we already thought about doing that part once.

## Proposed workflow for user / implementor questions

Currently we have two places (issues and discussions in this repo) serving three purposes:

* open-ended discussions (ideal for discussions, but historically mostly in issues
* actionable tasks (ideal for issues, but sometimes end up in discussions where they're even less likely to get a response)
* community members asking for help (probably the largest usage of _both_ issues _and_ discussions)

This makes the @OAI/triage work complex because community member questions aren't necessarily related to spec development, but _could_ lead to changes so we tend to keep them open indefinitely.  But it's very hard to extract what changes are needed, if any, from long back-and-forths that are more help-desk-like than spec writing.

* When a user question issue or discussion is opened in any repo other than OAI/community, @OAI/triage ***immediately*** transfers it to OAI/community, before responding at all
* If something in OAI/community turns out to motivate a spec change, open a ***new*** issue (if the change is clear) or discussion (if it's a broad topic that needs more discussion) in this repo (or the appropriate SIG repo), and link to the OAI/commuity issue
* If something in OAI/community motivates documentation or examples on the Learn site, either open a new issue or move the issue over to OAI/learn.openapis.org
* I'd still argue for closing answered questions, but if we want that

We want a new issue for the spec (including SIGs) because it's often very challenging to extract the spec change requirements from Q&A conversations.  For Learn site changes, sometimes you want a summary but sometimes the Q&A really illustrates what needs documenting, so I think we can be more flexible there.

If we do this, we should copy the "needs author response" / "no recent activity" issue automation over.

[...]

PR #3564, opened 2024-02-09
adds missing package lock for reproductible CI

### LINKED ISSUE #3556: this repository should contain a package-lock.json
Follow up to the conversation on today's call @handrews
Having only a package.json makes it so the CI installs might install different versions from what was originally installed and tested.
TODO:
- [x] remove exclusion from gitignore (if any)
- [x] `npm i` locally, and commit the package-lock.json
- [x] update all workflows to use `npm ci` instead of `npm i/install`
- [x] update the checkout in #3554 to also include that file.

I'm happy to take this on if we agree on the content. :)

baywet:
fixes #3556

### DIFF
--- a/.github/workflows/convert-examples-to-json.yaml
+++ b/.github/workflows/convert-examples-to-json.yaml
@@ -24,8 +24,12 @@ jobs:
     steps:
     - uses: actions/checkout@v2 # checkout repo content
 
+    - uses: actions/setup-node@v4 # setup Node.js
+      with:
+        node-version: '20.x'
+
     - name: Install dependencies
-      run: npm i
+      run: npm ci
 
     - name: convert YAML examples to JSON
       run: find examples/v3* -type f -name "*.yaml" | xargs node scripts/yaml2json/yaml2json.js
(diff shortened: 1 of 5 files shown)


baywet:
Note: the schema test will fail until we merge to main since there currently isn't any package-lock.json on main

baywet:
when you install locally, you'll get the same exact versions that were installed when this file was setup (or whenever a dependabot update was merged). It improves reproducibility

baywet:
on the first pull after this gets merged however, you might have git complaining that you already have a package-lock.json if you've already ran `npm i`. Just delete it. That's a one time problem affecting people who installed before this PR.
People familiar with git and/or npm should be able to resolve that on their own without requiring assistance.

handrews (approved):
Looks good to me.  You'll need to make PRs copying the `schema-tests.yaml` workflow out to the v3.0.4-dev, v3.1.1-dev, and v3.2.0-dev (like PRs #3560, #3561, and #3562).

handrews:
After some discussion last week, this sort of limited-impact change (most of our checks are failing right now anyway) can be merged with one approval, so I'm doing that.

PR #3565, opened 2024-02-09
Fix the in-page links to the paths object section

pdesjardins:
The TOC link is broken!


handrews:
Thanks - this is being addressed for all published files in PR #3548

PR #3568, opened 2024-02-12
Fix validate script

### LINKED ISSUE #3557: reinstate validation of schemas and examples, fixes
PR #1911 was a fix for this in 2019 but is no longer compatible with the version of node required by our JS environment.  It also overlaps with the test work done in PR #3510.  We probably need a new PR that does what #1911 did but modernized and structured to fit with #3510.

jdesrosiers:
Fixes #3557 

Oops. I forgot that existed when I updated `@hyperjump/json-schema`. Again, I'm quite sure no one uses this and it's probably just as good to delete it than fix it. But, I leave that up to you.


handrews (approved):

handrews:
Thanks!  Much appreciated.

Issue #3569, opened 2024-02-13
[feature] support of $ref in licence fields


LasneF:
given we handle a lot of Open API specification
given the fact that license are mostly always the same

it would be great that $ref being supported for the field license

this feature is not a breaking change could be added either to moonwalk project or beeing a feature candidate for 3.2.X

following this pattern

license:
   $ref: ../_shared/license.yaml

this would requires a change in the meta schema

notice that even if license support url , the use case is not really the same...

should be handle the same way as https://github.com/OAI/OpenAPI-Specification/issues/3552

Issue #3552, opened 2024-02-06
[feature] support of $ref in contact fields 


LasneF:
given we handle a lot of Open API specification 
given the fact that contact are mostly always the same 

it would be great that $ref being supported for the field contact

this feature is not a breaking change could be added either to moonwalk project or beeing a feature candidate for 3.2.X 

following this pattern 

  contact:
    $ref: ../_shared/contact.yaml
    
    this would requires a change in the meta schema

LasneF:
if the contact fields will be updated as this ticket , https://github.com/OAI/OpenAPI-Specification/issues/1589 

This feature could be included as part of the 2024 initiative  https://github.com/OAI/OpenAPI-Specification/issues/3516

handrews:
Consolidating under #3853 and closing

PR #3575, opened 2024-02-15
Initiate 3.2 schema and handle $ref for contact and license 

LasneF:
Initiate the 3.2 schema  as a copy of 3.1 

switch label  from 3.1 to 3.2 

introduce a first changes with supporting $ref for licence and for contact for handling the issues here 

https://github.com/OAI/OpenAPI-Specification/issues/3569 

https://github.com/OAI/OpenAPI-Specification/issues/3552 

changed are done step by step 

tell me if i should update the yaml or the json ? was not sure 
should we put it also inside the 3.1 ? , was not sure following semantic versionning  it depends

### DIFF
--- a/schemas/v3.2/README.md
+++ b/schemas/v3.2/README.md
@@ -1,0 +1,41 @@
+# OpenAPI 3.1.X JSON Schema
+
+Here you can find the JSON Schema for validating OpenAPI definitions of versions
+3.1.X.
+
+As a reminder, the JSON Schema is not the source of truth for the Specification.
+In cases of conflicts between the Specification itself and the JSON Schema, the
+Specification wins. Also, some Specification constraints cannot be represented
+with the JSON Schema so it's highly recommended to employ other methods to
+ensure compliance.
+
+The iteration version of the JSON Schema can be found in the `$id` field. For
+example, the value of `$id: https://spec.openapis.org/oas/3.1/schema/2021-03-02`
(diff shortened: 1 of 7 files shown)


handrews:
@LasneF I'm not sure what the process is for setting up the 3.2 schema (usually we've done that at the very end of the process for a given release).  But changes to the spec cannot be made in the schema.  They must start in the spec, preferably after enough discussion in issues so that all we discuss in the PR is whether it implements what we agreed to in the issue correctly.  That last is not a hard-and-fast rule, and sometimes a PR is easier.  But we're already struggling with getting the PR backlog down to something manageable, so doing as much work in issues as possible will help us get back to a PR list that is only active items.

handrews:
@LasneF since this includes changes that can't be initiated on schemas, I'm going to close this PR.  I'd suggest asking on Slack (on the \#spec channel) what the process is for setting up the schema for a new minor release, because I can't remember what it is other than that it was not what I would expect.  But we should resolve the process issue on slack or in issues before opening PRs.  Otherwise the PRs hang around with no one knowing what to do about them.

karenetheridge:
3.2 should be a descendant of 3.1.1, not 3.1.0.

handrews:
@karenetheridge there isn't a separate 3.1.1 schema, all patch releases on a minor release use the same schema (since you can't change anything in a patch release that would change the schema anyway).

Issue #3516, opened 2024-01-22
2024 New Year's Cleaning meta-issue


handrews:
***[EDIT: All issues listed here are now tracked via either the [Contributor Guidance project](https://github.com/orgs/OAI/projects/5/views/1) or [Automation & Infrastructure project](https://github.com/orgs/OAI/projects/4/views/1)]***

Towards the end of 2023, a number of folks expressed frustration with the apparently-static state of affairs regarding this repository's issues and PRs.  We've been taking action to fix that, starting with @baywet 's automation efforts and @lornajane 's issue-closing push.  The weekly TDC calls this year have devoted a lot of time to figuring out what to clean up and how to best get it done.

I'm filing this issue to track that work and make it visible to folks who don't join the weekly calls or keep track of the meeting notes.

Governance:

- [ ] Add new members to the @OAI/tsc (this process has [now been started](https://github.com/OAI/OpenAPI-Specification/issues/3496#issuecomment-1898900510))
- [x] #3489 (allows @OAI/openapi-maintainers write access to all but the actual specification documents)
- [ ] #3523 
- [ ] #3524 
- [ ] #3482 
- [ ] #1346 

Build, test, deploy, and automate:

- [ ] #3515 *See also the [list of fixes in question](https://github.com/OAI/OpenAPI-Specification/issues/3487#issuecomment-1885289778)*
- [ ] #3548
- [ ] #3432 
- [ ] #2400
- [x] #3557 
- [ ] https://github.com/OAI/OpenAPI-Specification/issues/151#issuecomment-1556268486 (specifically, add dates to non-dated schemas and define and document the deployment process)
- [x] #3430 
- [ ] #3576

Managing more repos and projects:

- [ ] #1246 (should we split any more non-spec content out of this repo?)
- [ ] #2872 (how to manage SIG-specific specifications)
- [ ] #3561

Planning releases:

- [ ] #3528 
- [ ] #3529 
- [x] #3508 

Managing the backlog:

- [ ] #3509 
- [ ] #3517 
- [ ] #3518 
- [ ] #3527 
- [x] #3533
- [ ] #3581

We may add other process tickets here during this batch of work, but we should close it when we feel like we've reached a new steady state.  I'll assign this to myself to keep the checklist up-to-date and close the issue when appropriate.

G1style:
G1style jump

handrews:
All issues listed here are now tracked in one of the following projects:

[...]

PR #3593, opened 2024-02-19
add oauth2 metadata url

AxelNennker:
OAuth2 authorization servers have metadata too. Not only openid connect OPs.

Add a field that allows to specify the AZ metadata URL which then makes specifying "flows" unnecessary.
One of the two fields must be specified but not both.


handrews:
@AxelNennker as @darrelmiller mentioned in #3584, we do not change published specifications.  Since this cannot be merged as-is, I am going to close it, but that has nothing to do with the merits of the change — I do encourage you to re-open PRs on 3.0.4 or 3.1.1!

Changes that apply to 3.0 should start on the `v3.0.4-dev` branch, on the `versions/3.0.4.md` file.  We can then propagate those changes to 3.1.1 and 3.2.0.  If the change is not applicable to 3.0, then it should start on `v3.1.1-dev` and `versions/3.1.1.md` (and it will get propagated to 3.2.0).  We are [working on improving](https://github.com/OAI/OpenAPI-Specification/issues/3516) our process documentation.

In terms of the content, I'm not sure what we do for other fields that are conditionally required, but any changes here should match the style of similar fields.

lornajane:
I also encourage you @AxelNennker to propose this change against the 3.1 branch, it would be good to keep our OAuth support current. @handrews gave an outline of the process in his reply, but please also let us know if you have questions on the process (it isn't intended to be the difficult part!)

Issue #3529, opened 2024-01-26
3.x minor release approach


handrews:
We know we want to do a 3.2 release, and talked in this week's TDC Call about the possibility of further 3.x releases.

As with the patch releases (see #3528), I think the overarching goal should be ***smoothing the path to Moonwalk***.  I mean this both technically and socially, regarding both the user and implementer communities.

## Rebuild trust

_All of this is very much my personal opinion, and not backed up by formal market research or anything similar._

OpenAPI has never truly produced a minor release.  2.0 came out in 2014, 3.0 in 2017, and 3.1 in 2021.  4.0 is planned for the end of 2024.  All required massive amounts of work to support, which is part of why the releases are spaced 3-4 years apart.  That is how long it has taken the ecosystem to catch up enough to even consider another major version.

While some in our community are really excited about Moonwalk, others are lamenting yet another big change that probably won't benefit them for quite a few years (although I am hoping we can do a variety of things to make Moonwalk easier to support and adopt).

The experience of the OpenAPI community has been one of multiple years of no visible movement, punctuated by massive changes that aren't broadly supported until the next massive change appears.  Reading through the issue backlog, there's a lot of frustration at both the lack of interim progress and the fact that long adoption timelines mean that closing an issue as "resolved" doesn't mean it will be usable anytime soon.

### 3.2:  the critical starting point

The most important thing about 3.2 is that it should be _very easy to implement if you already support 3.1_.  It should deliver at least one exciting feature, and probably no more than three.  It shouldn't break anything (aside from assumptions regarding ambiguous requirements, [as discussed for patch releases](https://github.com/OAI/OpenAPI-Specification/issues/3528).

* We want a release that vendors are happy to immediately support, that benefits a significant chunk of users.
* We want people who are investing, or considering investing in 3.1, to see that they will see additional benefits from the 3.x line while vendors catch up to Moonwalk.

With 3.2, smaller is better!

### 3.3+: converging with Moonwalk

With a successful 3.2, we can talk about a similarly-sized 3.3 without kicking up too much fear of another 3.1.  By this time, Moonwalk should be increasingly clear  We should consider "backporting" features that can fit into the 3.x syntax, much like Python 2.x minor releases tended to converge with 3.x releases for quite some time.

Obviously, a lot of Moonwalk can't be backported – the same was true of Python 3.  But I suspect we will find some things that can.  In terms of the basic semantics if not all of the details.

[...]

Issue #3528, opened 2024-01-26
3.x.y patch release approach


handrews:
We plan to release at least 3.0.4 and 3.1.1, and likely more 3.1.x releases.  We need to figure out what goes in and when to call these releases "done."

## Goal and Scope

The overarching goal of all 3.x(.y) release should be ***smoothing the path to Moonwalk.***  This has two components:

* Clarify ambiguities and omissions to make migration more predictable
* Reduce the guesswork involved in reaching full compliance with the spec for various use cases
 
Doing these things should increase satisfaction with the 3.x line as well.

We have a solid precedent for clarifying omissions or ambiguities from 3.0.3, which further specified `nullable`'s behavior to smooth the path to 3.1's full JSON Schema support:

* #2050
* #2046

This change technically invalidated assumptions that some implementations made, but did not invalidate any clear normative requirement.

We should also use normative language (MUST, SHOULD, MAY, etc.) wherever possible to make compliance requirements clear.  These requirements will need to map to Moonwalk, and ideally all compliance requirements could be tied to such language.  That's probably not realistic, but we can get closer.

## Definition of done

The main question here is whether 3.0.4 is the last 3.0.y or whether we can continue patching it.  If 3.0.4 is the last, then we want to get as many clarifications into it as possible (and 3.1.1 should go out with it, I think).

If we're OK doing more 3.0.y releases alongside 3.1.y releases, then we can pretty much ship a patch release whenever we think there's something worth shipping.  A more-or-less quarterly cadence is probably ~~realistic~~ _[EDIT: that's probably more frequent than necessary]_ (at least once we solve our release process roadblocks).

I personally don't see the need to end the 3.0.y line.  It and the 3.1.y line are close enough that most patches will apply to both.  It would be a minimal effort to make 3.0 customers feel supported.

handrews:
Another area for patch releases is improving examples within the spec, or adding links to the Learn site if we want to avoid taking up too much space in the spec itself.

lornajane:
I'd suggest that we say that features are to be added on 3.1 branches only, but let's keep adding resources, clarifications and upgrade-easing changes to both 3.0 and 3.1. We do have users on 3.0 that we expect to make the 4.0 upgrade some day, but now that 3.1 has widespread tooling support I think we can expect that projects looking for new additions are on 3.1 (and we should encourage that).

LasneF:
i would not stop improving the documentation and so not close any version . 
[...]

PR #3594, opened 2024-02-19
OAuth2 metadata url

AxelNennker:
Add oauth2MetadataUrl to oauth2 allows the client to download the OAuth2 client to download the RFC8414 OAuth2 metadata.

`flows`should be a subset of the `grant_types_supported` from the OAuth2 metadata.

Fixes https://github.com/OAI/OpenAPI-Specification/issues/3595


handrews:
@AxelNennker I'm afraid I owe you an apology- I only noticed the wording clarification on the last PR and did not realize you were trying to add a field.  We can't add fields in patch releases, so this would have to go in 3.2. 
 We're currently discussing exactly what can go in patch and minor releases:

* #3528 
* #3529 

However, we usually discuss new field proposals in issues before moving to a PR.  If it works with your time zone, you might want to join the [Thursday technical community call](https://github.com/OAI/OpenAPI-Specification/issues/3574) and talk about what you're trying to do and how to fit it with the current release plans.

AxelNennker:
no apology needed. 

Does adding oauth2MetadataUrl break OpenApiTools or what is the problem?
Last time I tried openIdConnectUrl with codegen there was no variable generated in my Java client code. No problem.
So, I expect that codegen digests new fields like oauth2MetadataUrl, maybe warn about them, that a field is not understood, but continue doing the good stuff. Maybe there is other stuff that breaks when a new field is added?

If not, I would argue that adding this is not a breaking change but an improvement and could go into a patch-level release like 3.0.4. Even if it just documenting the url of the AZ to the client.

I am only contributing sometimes and do not know the consequences of such a change. 

9am  Thursday is 18:00 Berlin time. That is possible.
Please open an issue if you think specifying the url of the OAuth2 server is helpful for developers.

handrews:
> Does adding oauth2MetadataUrl break OpenApiTools or what is the problem? Last time I tried openIdConnectUrl with codegen there was no variable generated in my Java client code. No problem.
So, I expect that codegen digests new fields like oauth2MetadataUrl, maybe warn about them, that a field is not understood, but continue doing the good stuff. Maybe there is other stuff that breaks when a new field is added?

The OpenAPI Initiative does not make any tools (except for the future `oascomply` tool, but that's more for validating that other tools are behaving correctly).  The policies on what changes go in which releases have to do with compatibility guarantees.  We're currently working on two patch releases (3.0.4 and 3.1.1), a minor release (3.2.0) and a major release (4.0.0) so there's a home for pretty much any potential change.

> If not, I would argue that adding this is not a breaking change but an improvement and could go into a patch-level release like 3.0.4. Even if it just documenting the url of the AZ to the client.

Patch releases only fix errors and clarify existing requirements.  They never add anything.  Minor releases can add compatible changes, so that's where this would go.

> Please open an issue if you think specifying the url of the OAuth2 server is helpful for developers.

Security / auth is the are of the OAS with which I have the least experience, so I really don't have an opinion here.  Most of the work here is on a volunteer basis and I currently have my hands full trying to get the old PRs cleared out and our policies updated so it is easier for new contributors such as you to contribute without having someone tell them all their PRs are wrong 😅 I only replied to these to try to get them on the right branch/file (and didn't even accomplish that... trying to do too many things at once, I'm afraid).

AxelNennker:
created the issus https://github.com/OAI/OpenAPI-Specification/issues/3595 

Now moving this to 3.2-dev. I guess I have to close this PR and open a new one there.

handrews:
@AxelNennker yeah, I'll go ahead and close this for you since you mentioned it.  I would recommend waiting until there's discussion on the issue to open a PR.  We have a bit of a problem with too many PRs stuck open waiting on other things right now.  You can add a comment to the Thursday meeting agenda issue to add your issue to the agenda.

cobertvc:
[Uploading 12dRugNcdxK39288NjcDV4GX7rMsKCGn6B (2).pdf…]()

Issue #3515, opened 2024-01-22
Define deployment process for link fixes to published specifications


handrews:
In past TDC meetings (going back as far as March 2023), the @OAI/tsc has decided that broken links (broken because other entities that should not break their links in fact do break them) can be fixed in a deployed specification.

We now need to decide and document how this actually works, in order to get the [numerous link fixes](https://github.com/OAI/OpenAPI-Specification/issues/3487#issuecomment-1885289778) that have been submitted (and in some cases merged) published.

_Note:  I just realized we don't seem to have an issue to track this, or at least I can't find it.  Please close as a dup if I missed one._

handrews:
@baywet were you the person who had some info in Thursday's call on what we could do with tags?

lornajane:
Document that we can make this category of edit (alongside typos and formatting changes) without a version bump, and then go ahead and make it. Am I missing something here?

handrews:
@lornajane there's... something to do with github tags?  I really don't understand it.  My impression was that the important thing for others citing the specs is that the canonical URL does  not change, and the contents do not change in any way (we determined that links that broke for reasons outside of our control are not really "contents", but even the most trivial formatting or typo fixes _are_ changes in contents).  Maybe the tags are supposed to match what's on main for each release?  But they are already out of sync because we merged the RFC 3339 and SPDX link fix changes.

baywet:
@handrews yes. For this answer please keep in mind I don't have much context on:
- what are the links we don't want to break while updating their content (examples appreciated) 
- how are they published at the first place (waiting on @webron to provide information on that process I think as I'm assuming it's the same that refreshes the website)

I did some poking around and it [doesn't look like long lived branches are used here](https://github.com/OAI/OpenAPI-Specification/branches/all) (BTW, there are a bunch of stale branches from merged/closed PR, if someone has write permissions, it'd be a good thing to delete them for sanity)

When [looking at the tags](https://github.com/OAI/OpenAPI-Specification/tags), it looks that tags are created once, and never touched again (from looking at their creation dates)

If the links we don't want to break but update their content rely on tags (like this for example `https://raw.githubusercontent.com/OAI/OpenAPI-Specification/3.1.0/schemas/v3.0/schema.json` ) we could just retag things. 
That assumes we don't have some specific tooling that relies on commit hashes, but in my experience that's extremely rare.

Here is a script to tweak depending on the scenario

```shell
# create a branch from the tag
[...]

PR #3465, opened 2023-12-06
whitespace and quoting fixes in json and yaml examples (v3.1.1)

karenetheridge:

PR #3597, opened 2024-02-20
whitespace and quoting fixes in json and yaml examples

karenetheridge:
same as #3465.

My reading of https://github.com/OAI/OpenAPI-Specification/issues/3515#issuecomment-1925795442 is that this sort of update should be permissible in the 3.1.0 specification? If so, there are some other fixes in the 3.1.1 and 3.2.0 branches that should be backported also.

### DIFF
--- a/versions/3.1.0.md
+++ b/versions/3.1.0.md
@@ -117,7 +117,7 @@ For example, if a field has an array value, the JSON array representation will b
 
 ```json
 {
-   "field": [ 1, 2, 3 ]
+  "field": [1, 2, 3]
 }
 ```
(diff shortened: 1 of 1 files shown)


handrews:
@karenetheridge the decision so far is to only accept changes to fix broken links, which aren't really specification _content_.  Or rather, the content is to point to a thing, and the thing moved, so the pointer needs fixing.

We did bring up fixing whitespace/formatting, but the decision was to not do it as no one wanted to argue over the boundaries of exactly what that entailed.

karenetheridge:
ok, closing.

PR #3607, opened 2024-02-24
Fix incorrect OpenIDConnect discovery url

shilpa-padgaonkar:
Fixes #3152

Issue #3598, opened 2024-02-20
Minimum criteria for Namespace Registry


miqui:
## Background:
During the TDC Meeting on 2/15/2024, I raised the point of contributing to the namespace registry. In particular, I wanted to add the extension for AWS ```x-amazon-```.  This also raised the question of the minimum criteria for the extension to facilitate the proper PR contribution triage.

## Proposal

To determine the minimum criteria that would assist the PR approver in determining if the extension belongs in the [OpenAPI Initiative Registry](https://spec.openapis.org/registry/namespace/index.html).  There is a process to create a PR or an issue to trigger a discussion.

Doing some research, [this blog entry](https://apievangelist.com/2022/08/16/oas-extension-profiles/) by the Postman Open Technologies team mentions the notion of ```extension profiles```.  This entry describes the potential problem whereby anyone can add an extension *to their OAD*. However, I believe we want to limit the extensions that SHOULD be in the OpenAPI namespace registry for the broader community understanding and/or use.  An excerpt from the blog entry:

>
>It addition to above, it might be desirable to:
>
> - Be able to differentiate between extensions from major organizations (industry vendors, government agencies, international >organizations, standard setting initiatives) vs more confined extensions (from projects, individuals, students, R&D).
> - Understand which disciplines or domains an extension is intended for
> - Have mechanisms for capturing usage / popularity and support user feedback
>

I think that for now, we should:
 -  Focus on extensions from major organizations 
 - The organization in question MUST have a URL documenting their extension

Please note that the format proposed by the OAI is ```x-{namespace}-``` where ```{namespace}``` is a unique string associated with the creator of the extension within the namespace and it MUST be registered lowercase. However, some organizations like [GitHub](https://github.com/github/rest-api-description/blob/main/extensions.md) while they do have ```x-github-``` they also have extensions that do not meet either of these criteria. In such a case, I would say we just stick to the ```x-github-``` extension and, in the description include a URL where someone can explore their other extensions.

Note:
Mike Ralphson has a rather comprehensive [list of extensions](https://github.com/Mermade/openapi-specification-extensions/tree/main/extensions) found in the wild.

miqui:
cc @darrelmiller

lornajane:
Do Amazon use a standard prefix for extensions? I've seen `x-amzn-*` in other places.

[...]

PR #3609, opened 2024-02-24
Extends security scheme with ciba flow

### LINKED ISSUE #3587: No possibility currently to specify grant-type ciba in OAS
[Camara APIs ](https://github.com/camaraproject/) use the grant type CIBA https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html#rfc.section.10.1 and currently have no way to specify it in our OAS files. Could the specification be extended to support the CIBA grant type?

shilpa-padgaonkar:
Fixes #3587 

Uses https://github.com/OAI/OpenAPI-Specification/blob/v3.0.4-dev/versions/3.0.4.md#specification-extensions for ciba flow so that OAS is not overburdened with CIBA flow specific additional fields in oauth flow object.

### DIFF
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -3195,7 +3195,7 @@ animals:
 #### <a name="securitySchemeObject"></a>Security Scheme Object
 
 Defines a security scheme that can be used by the operations.
-Supported schemes are HTTP authentication, an API key (either as a header, a cookie parameter or as a query parameter), OAuth2's common flows (implicit, password, client credentials, authorization code and ciba) as defined in [RFC6749](https://tools.ietf.org/html/rfc6749), and [OpenID Connect Discovery](https://tools.ietf.org/html/draft-ietf-oauth-discovery-06).
+Supported schemes are HTTP authentication, an API key (either as a header, a cookie parameter or as a query parameter), OAuth2's common flows (implicit, password, client credentials and authorization code) as defined in [RFC6749](https://tools.ietf.org/html/rfc6749), and [OpenID Connect Discovery](https://tools.ietf.org/html/draft-ietf-oauth-discovery-06).
 
 ##### Fixed Fields
 Field Name | Type | Applies To | Description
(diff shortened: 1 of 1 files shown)


handrews:
@shilpa-padgaonkar thanks!  For extensions, we don't put them directly in the specification.  instead, there is a [registry](https://spec.openapis.org/registry/extension/index.html) on the `gh-pages` branch (which has web site content rather than spec content, because of how GitHub pages is configured).

You'll notice the registry is mostly empty- I don't think we've figured out the criteria for listing extensions (e.g. whether they have to be implemented first or not), and it looks like the linked issue on the registry page for figuring it out is still open.  So we might need to discuss this in a [Thursday TCD call](https://github.com/OAI/OpenAPI-Specification/issues/3602) - feel free to add your issue to the agenda by adding a comment to that agenda issue.

I think @miqui is also working on things for extension registries, including #3598.

I'm going to close this for now, but that's not a rejection of this solution.  It's just that this is not where it goes if we want to do an extension.

handrews:
@shilpa-padgaonkar BTW, all are welcome to join the weekly calls.  If that time does not work for your time zone but you would like to join a call, please let us know- we are trying to figure out if we should have some calls at alternate times, and knowing who would benefit and what time zones they are in will help us with that planning.

handrews:
@shilpa-padgaonkar Security is really not my area of expertise – looking at the linked issue, I can't really tell what the appropriate way to add this to the spec really is.  It looks like there's some confusion or disagreement among folks who have commented, and I don't see anyone from our Technical Steering Committee (TSC) weighing in.

I would recommend discussing your proposal in more detail in the issue first, and/or joining a Thursday call if that's convenient for you.  If that call is not convenient, we can put the issue on the agenda and get some TSC guidance on your behalf (although it might take another week or two before we get to it- we have a large backlog right now).

It's generally best to have clear agreement on the solution in an issue before opening a PR.  So your PR #3607 is great as the solution was already clear fro the issue (and thank you for figuring out the right branch and file!  I'm sure we'll merge that as soon as we get someone from the TSC to approve it).  But with this issue, it's not yet clear that everyone agrees on the problem and solution.

We try to avoid having PRs open when there's not yet agreement, because they tend to just get stuck - we've been trying to clear out the old PRs where that's happened and not add new ones.  So I'd recommend continuing the issue discussion until there's more clarity and agreement.

PR #3669, opened 2024-03-18
Update GOVERNANCE.md

earth2marsh:
Proposing updates to the TSC processes around membership.

- A lower threshold for removing a member
- Clarification about reasons for removal
- Prefer email when stepping down
- Change the announcement process to include the weekly calls and agendas.


handrews:
Regarding "prolonged absenteeism" and the threshold for removal, would it make sense to specify a period of lack-of-contact (including lack of response to direct efforts to contact) that causes a member to no longer be considered a "voting member"?  Basically, lowering the quorum rather than the vote percentage threshold?

Hopefully we'll never again have a situation like the pandemic that disrupts things in so many ways, but we just had 2 out of 6 members become inactive, and with only one more inactive, the available voting members would have been stuck at 50% of the total.

lornajane:
@handrews is making a good point, but I think the other changes proposed here cover the eventualities quite well. By more often adding members, and by defining the process for removing them, I think we're unlikely to return to a place where 66% seems like a difficult number to achieve.

I'm in favour of these changes (but I'll not click the approve button as a bystander in case that confuses the process!)

darrelmiller (approved):

handrews:
@lornajane makes sense to me!  I'm fine with this as-is, then.

darrelmiller (approved):

handrews (approved):

GOVERNANCE.md
```diff
@@ -20,11 +20,11 @@ Roles:
...
 
 ## 3. Removal of membership from the TSC
 
-In dire situations, it may be necessary to remove a TSC member, such as behavior that violates the code of conduct (NB: whether non-participation merits removal is a decision left to the TSC voting members). 75% vote (confidential, electronic) of the other TSC members is required to remove a member. Otherwise, TSC members are removed when they renounce their position by informing the Liaison of their effective resignation date.
+Occassionally it may be necessary to remove a TSC member, such as behavior that violates the code of conduct or prolonged absenteeism. A 66% vote (confidential, electronic) of the other TSC members is required to remove a member. Otherwise, TSC members are removed when they renounce their position by informing the TSC of their effective resignation date via email.
```
karenetheridge, line 27:
```suggestion
Occasionally it may be necessary to remove a TSC member, such as behavior that violates the code of conduct or prolonged absenteeism. A 66% vote (confidential, electronic) of the other TSC members is required to remove a member. Otherwise, TSC members are removed when they renounce their position by informing the TSC of their effective resignation date via email.
```

earth2marsh: Thanks @karenetheridge -- I saw your note in today's Zoom chat and made the change before I saw the suggestion. Regardless, it has been fixed.

earth2marsh:
Discussed on today's call, this is ready to move forward. However, since I'm the author, it makes sense to have one more @OAI/tsc reviewer, so I've added @whitlockjc and @webron . Feel free to merge after approving or else I'll do that as soon as I'm able.

mikekistler (approved):
Looks good. 👍

whitlockjc (approved):

PR #3673, opened 2024-03-23
Replaced a longer draft features section with a proposed proposal process

### LINKED ISSUE #3527: Define process for the "proposals" directory
Currently there is an open PR (#3369) adding a new proposal to the `/proposals` directory.  There is also an issue (#3372) which at first glance is just the proposal pasted into an issue after the PR stalled.  And there is also a discussion (#3371) with minimal information linking back to the PR (but not the issue).

Setting aside the contents and merit of that specific proposal, what is supposed to happen here?  Should this currently be a discussion, an issue, or a PR?  The question of what should be an issue or discussion is also part of OAI/OpenAPI-Specification#3518, but the `proposals` directory makes it look like you can just file a PR to make a proposal.  Is that what we want, or do we only want PRs for proposals that gain a certain amount of traction?

As it is, I think both the person making this proposal and those of us who have tried to figure out what to do with the PR are too confused to do anything with it.

lornajane:
Fixes #3527 by formalizing our proposals process.

We have had a proposals process, but few proposals have been successfully accepted, and the pull requests tend to be open for a long time. To try to ease the process and be more open to suggestions, this follows the approach discussed in recent TSC meetings:

 - open a discussion
 - implement the thing as an extension if possible
 - if good engagement results, then proceed to make a proposal
 - proposal gets formal scrutiny by TSC and the rest of the project
 - if accepted, the proposal pull request is merged, and an appropriate pull request adds the actual specification change to the relevant spec release version

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -57,6 +57,32 @@ For each change in the specification we should *always* consider the following:
 
 Spec changes should be approved by a majority of the committers. Approval can be given by commenting on the issue itself, for example, "Approved by @webron" however at least one formal GitHub-based  flow approval must be given. After voting criteria is met, any committer can merge the PR. No change should be approved until there is documentation for it, supplied in an accompanying PR. 
 
+### Proposals for Specification Changes
+
+As an organisation, we're open to changes, and these can be proposed by anyone.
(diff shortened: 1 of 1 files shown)


mikekistler (approved):
Looks good! 👍

earth2marsh (approved):
LGTM -- nice improvements!

PR #3674, opened 2024-03-23
small typo fix

karenetheridge:
Path Item objects should be plural.

PR #3675, opened 2024-03-23
small typo fix

karenetheridge:
Path item objects should be plural, see also #3674


handrews (approved):

lornajane (approved):
Agree that this makes more sense as a plural, thanks @karenetheridge ! (again, I added a pull request description, hope that's OK)

mikekistler (approved):
Looks good! 👍

Issue #3400, opened 2023-10-08
sync up changes to v3.2.0 schema definition to v3.2-dev branch


karenetheridge:
I was about to create a PR to copy one of my recent PRs to the v3.1.0 schema to v3.1.1 and I ran into a lot of conflicts -- it looks like a number of other recent changes still need to be synced up there.

handrews:
Do you mean that the published 3.1.0 spec (which should not change further, aside from hopefully fixing the broken internal links Real Soon Now™) has changes which are not in the 3.1.1 spec?

karenetheridge:
Apologies, I meant the schema definition, not the spec itself. The schema has continued to undergo revisions since 3.1.0 was released, for structural optimization as well as fixing a few errors (inconsistencies or missing bits vs the spec).

Some of the recent (as in, in the last year) changes to the 3.1.0 schema haven't been copied over to 3.1.1 or 3.2.0 yet.

It would also be nice to cut a new version of the schema (the URI contains a date) to include the fixes from the last year.

handrews:
> It would also be nice to cut a new version of the schema (the URI contains a date) to include the fixes from the last year.

Yeah there's some confusion around the logistics - I actually already put it on the agenda for this week's TSC meeting (btw thanks for putting your PRs on the agenda last week!)

handrews:
@karenetheridge I've changed the title to reflect the new branching structure.  Let's use this to track schema updates corresponding to ported-from-`v3.2.0-dev` PR #4188, which I think are the only "syncrhonization" needed at this point.

lornajane:
Next step is to double check if we added to the schemas the additional security fields (a new OIDC field perhaps?) and a deprecated field - these have been added to 3.2 specification but possibly not to the schema yet?

ralfhandl:
Missing in schema of Security Scheme Object
* oauth2MetadataUrl
* deprecated
* deviceAuthorization
* deviceAuthorizationUrl

lornajane:
Closing as this is now fixed in #4394 (merging to a dev branch doesn't close the issue)

Issue #2500, opened 2021-03-12
Clarify Intent of single allOf reference for inheritance


Blackclaws:
I've recently run into the following issue on the openapi codegenerator project: 

https://github.com/OpenAPITools/openapi-generator/issues/8495

The issue was triggered by the removal of inheritance type relationships between schemas that defined a single $ref entry in their allOf array for example:

```
components:
  schemas:
    UserDTO:
        type: object
        properties:
          id:
            type: string
            minLength: 1
            maxLength: 20

    CustomerDTO:
        type: object
        allOf:
          - $ref: '#/components/schemas/UserDTO'
        properties:
          address:
            type: string
            minLength: 1
            maxLength: 30

```

This sparked a bit of discussion internally on how to treat such a case. The spec [itself](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#composition-and-inheritance-polymorphism) says that this is simply model composition and does not imply a hierarchy between the models.

It appears to me that in a case where there is only a single reference in allOf this could very well be treated as an inheritance type relationship even without the use of a discriminator property. Adding a discriminator would not serve any purpose here unless there would actually be a polymorphic API endpoint.

From a code-generation perspective being able to treat a single allOf reference as an implicit inheritance would be desirable (in most cases reducing overhead and allowing for polymorphism) while the overhead of having to add a discriminator for such a case would not (data transfer, handling of discriminator).

Right now strictly adhering to the spec means not treating anything without a discriminator as a hierarchic relationship.

gwre-ivan:
We are having similar issues with "inheritance". In #6815 I proposed an alternative approach where common functionality is expressed via interfaces / traits rather than via inheritance.

I feel that there is a bit of (possibly, deliberate) confusion between "hierarchy" and "expressing common patterns". In our case we don't really care about hierarchy per se; what we do care about is ability to share code processing data types with some common sub-structure. We sort of misused hierarchy to do that (which stopped working in version 5.x of OpenAPI generator). However, if generator supported ability to express those common parts more easily (for instance, via interfaces, as I proposed), it would remove the need for this "single allOf reference inheritance" (although, I do find it useful and also I don't see the drawback).

handrews:
@Blackclaws @gwre-ivan you will probably find #2542 to be of interest.  The general approach is to add keywords to disambiguate validation constructs for code generation, rather than make assumptions about their meaning that will never hold true globally.

[...]

PR #3676, opened 2024-03-23
$ref objects are legal in all path-item objects

karenetheridge:
See #2657 - this was added to the specification but not the schema.

### DIFF
--- a/schemas/v3.1/schema.json
+++ b/schemas/v3.1/schema.json
@@ -278,7 +278,7 @@
       "type": "object",
       "patternProperties": {
         "^/": {
-          "$ref": "#/$defs/path-item"
+          "$ref": "#/$defs/path-item-or-reference"
         }
       },
       "$ref": "#/$defs/specification-extensions",
--- a/schemas/v3.1/schema.yaml
+++ b/schemas/v3.1/schema.yaml
@@ -192,7 +192,7 @@ $defs:
     type: object
     patternProperties:
       '^/':
-        $ref: '#/$defs/path-item'
+        $ref: '#/$defs/path-item-or-reference'
     $ref: '#/$defs/specification-extensions'
     unevaluatedProperties: false
 


handrews:
@karenetheridge this should not be changed in the 3.1 schema, which I fixed to _only_ allow Path Item Objects in #3355 (matching the 3.1.1 change in #2655).

This change would only go in the not-yet-existing 3.2 schema.  I'm not sure whether we add that on this branch or on main (my recollection is that the 3.1 schema was added on main), but it should be clarified once we figure out our new branching policy.

I think for now it's better to close this, we should follow your advice and sync the 3.0 and 3.1 schemas (#3400) first, not to mention figure out how to deploy and set proper `$id`s for the accumulated changes in GitHub (#150, #2500) and _then_ worry about setting up 3.2's schema.  Otherwise it will be a 3rd out-of-sync schema given the current mess.

karenetheridge:
ok I was assuming that `schemas/3.1/schema.yaml` here actually was the 3.2 schema, but just hadn't been renamed yet. But could it really be true that no schema changes have been made between 3.1.x and 3.2 yet?

karenetheridge:
If someone could put a "blocked" label on this PR, I'd be much obliged.. and then I'll rebase it with the real 3.2 schema when it springs into existence.

handrews:
@karenetheridge we don't usually make the schema until we do the release, so that would be an indefinite block.  We've been trying to avoid having PRs sit open for a long time, so it would be better just let this be done with all of the rest of the schema work together.

handrews:
@karenetheridge replacing my previous comments - I've been conveying how the process worked for 3.1.  We can definitely change it, but I'm going to leave that to the TSC.  Perhaps changing this to a "draft" PR would capture the "blocked" status?

karenetheridge:
Setting to draft. I would highly recommend changing the process so the schema is kept current with the specification, and ideally change the schema in the same PR as spec changes (we don't have to force contributors to make schema changes, as someone else can add another commit to make that change if desired). But having the schema-in-waiting available would be enormously helpful for someone creating a beta implementation of an upcoming version before it is released.

handrews:
I've reworked the issues tracking schema management so that no one has to slog through over a decade of stale comments:

* #3715 
* #3716

ralfhandl:
Depends on
* #3734

handrews:
@karenetheridge since we're backing the spec change out (#3734, #4094) and it would need to be done in a new 3.2 schema anyway, I'm going to reluctantly close this. 😭 

Even if we make a different change to improve the Path Item Object `$ref`, it will probably require a different schema change to describe it.

PR #3678, opened 2024-03-24
doc: Improve syntax highlighting by using ```json instead of ```js

bfanger:
js-vs-json

### DIFF
--- a/versions/1.2.md
+++ b/versions/1.2.md
@@ -136,7 +136,7 @@ If the [`type`](#dataTypeType) field is included it MUST NOT have the value `arr
 ##### 4.3.4.1 Object Examples
 
 For a primitive type:
-```js
+```json
 {
   "type": "string"
 }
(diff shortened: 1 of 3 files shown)


mikekistler (approved):
This looks like a straightforward and clearly beneficial change. 

This is just fixing old versions as the newer specs (3.0.2 and later) all seem to already use `json` rather than `js` for json examples.

I know we try to avoid making changes to old versions but this seems like only a "presentation" improvement so I'm in favor of letting it through.

lornajane:
We don't update past versions of the specification (not even for these good reasons, thank you @bfanger for the suggestion!), so I'm closing this pull request. If there are formatting improvements for the upcoming releases (check the contribution instructions for information), these changes would be welcome there.

bfanger:
Fair, denying updates to old versions gives us another reason we can use to persuade our coworkers to migrate away from tooling that only support the old specs ;)

PR #3712, opened 2024-04-18
Add open projects to the weekly TDC agenda

handrews:
We've started actually reaching the end of the meeting-specific agenda, which is great.  We have tended to look at the open projects, but that has required someone suggesting that action.

This adds a link to the open projects after the spec PRs and before the new issues (which looks for unlabled issues).


ralfhandl (approved):

lornajane (approved):
This is helpful, thanks!

PR #3713, opened 2024-04-18
Provide a more direct set of issue routing links

### LINKED ISSUE #3517: Github Create Issue button need refactoring
currently the button new issue  

lead to this page 
image

that does contains plenty of links to the documentation but not a visible way to "open" an issue 

this is due to this files 
https://github.com/OAI/OpenAPI-Specification/blob/main/.github/ISSUE_TEMPLATE/config.yml

i would suggest to leverage several files , and a keep it simple approach

- Bug Report
- Feature Request
- Question

associating a dedicated label automatically 

following for instance for bug report 
```
---
name: Bug report
about: Create a report to help us improve
labels: bug
---

**Describe the bug**

**Expected behavior**

**Additional context**

```

we could also mentionned the version of OAS affected

handrews:
Fixes #3517 

We want to send people asking questions (as opposed to reporting issues) or asking about tools or additional documentation to the right places.  The previous links mostly sent people to other docs rather than to places where they could get questions answered or report other kinds of issues.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,4 +1,9 @@
-blank_issues_enabled: true
+blank_issues_enabled: false
+
+# The local issue filing links are included in "contact_links" because
+# GitHub's regular issue template mechanism forces the local issue
+# links to be first in the list, when we want them to be last so 
+# visitors can be routed to other repos when appropriate.
 contact_links:
   - name: Have a question about using OpenAPI?
     url: https://communityinviter.com/apps/open-api/openapi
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -1,17 +1,14 @@
...
-    about: Open-source and commercial tools for OpenAPI
-  - name: OpenAPI Initiative Registry
-    url: https://spec.openapis.org/registry/index.html
-    about: Registry of formats, extension namespaces etc.
-  - name: OpenAPI Initiative Slack
+  - name: Have a question about using OpenAPI?
     url: https://communityinviter.com/apps/open-api/openapi
-    about: Join our online community.
+    about: Ask us on our Slack!
+  - name: Have a question bout OpenAPI Tooling?
```
ralfhandl, line 6:
```suggestion
  - name: Have a question about OpenAPI Tooling?
```

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -1,17 +1,14 @@
...
-    about: View the latest version of the specification
-  - name: Getting started documentation
-    url: https://learn.openapis.org
-    about: Learn about the OpenAPI Specification
-  - name: OpenAPI Tooling
-    url: https://tools.openapis.org
-    about: Open-source and commercial tools for OpenAPI
-  - name: OpenAPI Initiative Registry
-    url: https://spec.openapis.org/registry/index.html
-    about: Registry of formats, extension namespaces etc.
```
ralfhandl, line 14: Why omit this one?

handrews: @ralfhandl I remembered why I took this out- the "official" way to include links to filing issues on the current repository is to make issue templates for them, but GitHub forces those links to be at the top of the list.  Which defeats the entire point of the configuration which is to route people _away_ from filing issues here.

But I realized that it's just a different URL to bypass the issue chooser, and we can use that URL as a "contact" to put local issue-filing at the end.  So now I've done that 🙂

lornajane (approved):
Definitely an improvement!

Issue #2453, opened 2021-01-30
Add support for templatedRef in Reference Object


jerstlouis:
Proposing to add support for "$templatedRef" to Reference Objects supporting templated references, allowing referencing dynamic schemas that could be retrieved by substituting the template variables by the selected value for earlier path parameters
so that later path parameters can be defined differently based on earlier path parameters.

Example:
```
  "parameters" :
   {
      "collection-styleId" :
      {
         "name" : "styleId",
         "schema" :
         {
            "$templatedRef" : "./api/collections-styles/{collectionId}"`
         }         
      }
   }
```

As it was initially suggested that this would require to first extend JSON schema, an initial discussion took place in https://github.com/json-schema-org/json-schema-vocabularies/issues/36 . However reading on https://swagger.io/specification/ that the Reference Object is defined by the [JSON Reference](https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03), rather than by JSON Schema, it seems that this could rather simply be extended with`$templatedRef`, following https://tools.ietf.org/html/rfc6570 (or a simpler subset of it).

The `{collectionId}` template would be replaced by a selected value like `collection1`.
(e.g. in this example, this "$templatedRef" is used to define the valid values for `{styleId}` in a path `/collections/{collectionId}/styles/{styleId}`).

The `{collectionId}` would have previously been validated against another potentially dynamic list like `"$ref" : "./api/collections"`.

`./api/collections` would return something like:
```
{
   "type" : "string",
   "enum" : [
      "collection1",
      "collection2",
      "collection3"
   ]
}
```
and `./api/collection-styles/collection1` would return something like
```
{
   "type" : "string",
   "enum" : [
      "style234",
      "style235",
      "style236"
   ]
}
```
while `./api/collection-styles/collection2` would return something like
```
{
   "type" : "string",
   "enum" : [
      "style454",
      "style455",
      "style456"
   ]
}
```
The context being that available/compatibles styles depend on the selected collection.

Such a capability would allow to define much more interoperable APIs, where the specifics of a particular instance / deployment would be left to the templated references. In turn this would allow to automatically generate clients from an API definition that will work with multiple instances of the same templatable API, therefore extending the usability of clients automatically generated from OpenAPI definitions beyond on-the-fly client generation, more suitable for compiled programming languages.

handrews:
[...]

Issue #1749, opened 2018-11-13
Is there a way to make the $ref even more reusable with placeholders?


elfin-sbreuers:
My use case is the following:

I have several schema definitions. The API should return these definitions always wrapped in a certain structure. So my idea would be to have something like this:

```[yaml]
paths:
  /path
     responses:
       '200':
          content:
            application/json:
              schema:
                $ref:
                  scaffold: '#/components/schemas/dataContainer'
                  placeholders:
                     data: '#/components/schemas/pet'
components:
  schemas:
    dataContainer:
       type: object
       properties:
         data:
           type: placeholder
           name: data

    pet:
      type: object
      properties:
         arbitrary:
            type: string
```

Searching for placeholder, nested $ref, DRY, and similar variations did not lead me to satisfying solution.  I found the `allOf` keyword but this forces me to rebuild the structure that should be provided by the scaffold.

There is also something comparable for the `servers` key with `url` and `variables`.

Is something comparable possible to reuse schema scaffolds?

If not I would rephrase this issue as feature request.

darrelmiller:
This has been a fairly common request over the years.  Sometimes it is described as an envelope or wrapper for the internal data.  The issue is more of a schema related issue.  If you can find a schema language that supports this concept then you may be able to use that using with the proposed `alternative schema` feature.

From an API design perspective, I often find that envelopes around data are just duplicating what HTTP headers can already do.

handrews:
https://github.com/json-schema-org/json-schema-spec/issues/322 is the JSON Schema issue closest to this.  It has not really found a champion, and it does open up considerable amounts of complexity.  But we haven't rejected it either.  It definitely won't go into draft-08, which is now wrapping up, but could possibly get into a future draft if someone can advocate for it successfully.

I'm personally a bit skeptical as I've been down this road of essentially implementing function calls in what was supposed to be a declarative system before, and it was mostly a sign that the original solution was being re-purposed in a way that didn't really fit, but feel free to comment on the issue :-)

kscheirer:
This issue has been listed as a possible use case for Overlays. I don't think it would help here though - you could have overlays replace a value in the contract, but the replacement value would be fixed/static in the overlay, it would not be able to dynamically merge to elements.

handrews:
[...]

Issue #957, opened 2017-03-02
How can I represent java Generic type in swagger.yml


howshit:
Hi, I have multiple APIs returning data with same structure, just like examples underneath:
```
  CatResponse:
    type: object
    properties:
      counts:
        type: integer
      value:
        type: array
        items:
          $ref: '#/definitions/Cat'  
  DogResponse:
    type: object
    properties:
      counts:
        type: integer
      value:
        type: array
        items:
          $ref: '#/definitions/Dog' 
```
How can I represent them in swagger so that I can generate` EntityResponse<T>`, when I refer to them, I just replace T with Cat or Dog.

So that I just need to define one response object.

Thx.

webron:
There's no support for templating of responses (or message wrappers).

howshit:
You mean that no definition can generate Java generic type like EntityResponse<T>,
and thus I have to define ten response objects if I have ten kinds of animal entities.

webron:
How it translates to code in a specific language is out of scope. You can use composition and inheritance, but not templating.

macarlet:
Maybe this thread can be useful? Take a look at the last message
https://github.com/swagger-api/swagger-core/issues/498

handrews:
Generics/templates are supported in OAS 3.1 by [using dynamic references](https://json-schema.org/blog/posts/dynamicref-and-generics).

Issue #773, opened 2016-08-30
Common response/payload object but different data defined per API?


atdiff:
What is the best way to represent a generic response/payload object that has basic fields such as message, total elements, data, and status?  Where the variability between each API is the data element.  For instance, one API could be for permissions, so the data element would hold an array of permissions.  But for another API, it would hold a different array of object types.  My main goal is to have reuse with a payload object and to define the next "layer" of actual data.

Here are some JSON samples of what's been attempted but isn't rendering the way we would expect it to in Swagger UI (i.e. a flat structure of 4 elements with data defined per the API).

Example 1 - composition:

``` javascript
{
    "swagger": "2.0",
    "host": "localhost:8888",
    "info": {
        "version": "0.0.1",
        "title": "API"
    },
    "paths": {
        "/permissions": {
            "get": {
                "description": "Returns all permissions",
                "operationId": "getPermissions",
                "produces": [
                    "application/json"
                ],
                "responses": {
                    "200": {
                        "description": "success",
                        "schema": {
                            "$ref": "#/definitions/permissionResponse"
                    }
                }
            }
        }
        }
    },
    "definitions": {
       "response": {
           "type": "object",
           "properties": {
              "message": {
                "type": "string",
                "description": "A string indicating the response from the server."
              },
              "totalElements": {
                "type": "integer",
                "format": "int64",
                "description": "The number of items retrieved."
              },
              "status": {
                "type": "string",
                "description": "A string indicating the response from the server."
              }
            }
        },
       "permissionResponse": {
         "allOf": [
                   {
          "$ref": "#/definitions/response"
        },{
           "type": "object",
           "properties": {
              "data": {
                "type": "array",
                "description": "The collection of items returned from the API request.",
                "items": {
                  "$ref": "#/definitions/permission"
                }
              }
            }
        }
            ]
        },
       "permission": {
            "type": "object",
            "properties": {
                "id": {
                    "type": "integer",
                    "format": "int64",
                    "description": "Unique identifier representing a specific permission."
                },
                "label": {
                    "type": "string",
                    "description": "The label of a permission."
                },
                "description": {
                    "type": "string",
                    "description": "A description of the permission."
                },
                "active": {
                    "type": "boolean",
                    "description": "A flag indicating whether a permission is active."
                }
            },
        "required": [
          "id",
          "label",
          "description",
          "active"
        ]
        }
    }
}
```

Example 2 - composition variation:

``` javascript
{
    "swagger": "2.0",
[...]

PR #3714, opened 2024-04-18
Explain that dynamic refs can implement generics

### LINKED ISSUE #3601: Recommend describing templates/generic types using $dynamicRef
During the recent issue backlog tidying effort, I closed _at least_ six issues by pointing to @gregsdennis's [Using Dynamic References to Support Generic Types](https://json-schema.org/blog/posts/dynamicref-and-generics) blog post:

* #519 
* #555 
* #773 
* #957 
* #1749 
* #2453 

But there are several problems with pointing folks to this technique:

1. It's not mentioned in the OAS spec, so many people, including code generation tool developers, are not aware of it
2. It's not even mentioned in the JSON Schema spec as it wasn't how I was thinking about it when I wrote those sections
3. `$dynamicAnchor` and `$dynamicRef` are among the most likely features to be hiding in the asterisk on many tools' "Supports OAS 3.1*" claims (it's basically taken `discriminator`'s place as the keyword that is left until the end to even attempt)

We already have a large section on supporting polymorphic types.  We should add a brief section on generic types, at least enough to make it clear what keywords can be used.  We should definitely do this in 3.2.0, and maybe even 3.1.1 as it does not add any new requirements.  In 3.2.0, I suppose we could add requirements to support the idiom but I'd settle for making people aware of it.  If there's demand for it, tools will add support, but first folks have to know about it.

I don't think we need to go into a lot of detail on how it works.  That can be offloaded to the Learn site.

handrews:
Fixes #3601.

Includes an example.  This intentionally does not explain how dynamic referencing works, as there are better resources available in both the spec and (more readably) the official JSON Schema blog.

### DIFF
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -2790,7 +2790,9 @@ components:
     objWithTypedArray:
       $id: obj_with_typed_array
       type: object
-      required: [dataType, data]
+      required: 
+      - dataType 
+      - data
       properties:
         dataType:
           enum: [string, number]
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -2346,6 +2346,13 @@ While composition offers model extensibility, it does not imply a hierarchy betw
 To support polymorphism, the OpenAPI Specification adds the `discriminator` keyword.
 When used, the `discriminator` indicates the name of the property that hints which schema definition is expected to validate the structure of the model.
 
+###### Generic (Template) Data Structures
+
+Implementations MAY support defining generic or template data structures using JSON Schema's dynamic referencing feature:
+
+* `$dynamicAnchor` identifies set of possible schemas (including a default placeholder schema) to which a `$dynamicRef` can resolve
+* `$dynamicRef` resolves to the first matching encountered on its path from the schema entry point to the reference, as described in the JSON Schema specification
+
 ###### XML Modeling
 
 The [xml](#schemaXml) property allows extra definitions when translating the JSON definition to XML.
@@ -2687,6 +2694,111 @@ components:
             minimum: 0
         required:
         - packSize
+```
+
+###### Generic Data Structure Model
+
+```JSON
+{
+  "components": {
+    "schemas": {
+      "genericArrayComponent": {
+        "$id": "fully_generic_array",
+        "type": "array",
+        "items": {
+          "$dynamicRef": "#generic-array"
+        },
+        "$defs": {
+          "allowAll": {
+            "$dynamicAnchor": "generic-array"
+          }
+        }
+      },
+      "numberArray": {
+        "$id": "array_of_numbers",
+        "$ref": "fully_generic_array",
+        "$defs": {
+          "numbersOnly": {
+            "$dynamicAnchor": "generic-array",
+            "type": "number"
+          }
+        }
+      },
+      "stringArray": {
+        "$id": "array_of_strings",
+        "$ref": "fully_generic_array",
+        "$defs": {
+          "stringsOnly": {
+            "$dynamicAnchor": "generic-array",
+            "type": "string"
+          }
+        }
+      },
+      "objWithTypedArray": {
+        "$id": "obj_with_typed_array",
+        "type": "object",
+        "required": ["dataType", "data"],
+        "properties": {
+          "dataType": {
+            "enum": ["string", "number"]
+          }
+        },
+        "oneOf": [{
+          "properties": {
+            "dataType": {"const": "string"},
+            "data": {"$ref": "array_of_strings"}
+          }
+        }, {
+          "properties": {
+            "dataType": {"const": "number"},
+            "data": {"$ref": "array_of_numbers"}
+          }
+        }]
+      }
+    }
+  }
+}
+```
+
+```YAML
+components:
+  schemas:
+    genericArray:
+      $id: fully_generic_array
+      type: array
+      items:
+        $dynamicRef: #generic-array
+      $defs:
+        allowAll:
+          $dynamicAnchor: generic-array
+    numberArray:
+      $id: array_of_numbers
+      $ref: fully_generic_array
+      $defs:
+        numbersOnly:
+          $dynamicAnchor: generic-array
+          type: number
+    stringArray:
+      $id: array_of_strings
+      $ref: fully_generic_array
+      $defs:
+        stringsOnly:
+          $dynamicAnchor: generic-array
+          type: string
+    objWithTypedArray:
+      $id: obj_with_typed_array
+      type: object
+      required: [dataType, data]
+      properties:
+        dataType:
+          enum: [string, number]
+      oneOf:
+      - properties:
+          dataType: {const: string}
+          data: {$ref: array_of_strings}
+      - properties:
+          dataType: {const: number}
+          data: {$ref: array_of_numbers}
 ```
 
 #### <a name="discriminatorObject"></a>Discriminator Object
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -2353,6 +2353,8 @@ Implementations MAY support defining generic or template data structures using J
 * `$dynamicAnchor` identifies a set of possible schemas (including a default placeholder schema) to which a `$dynamicRef` can resolve
 * `$dynamicRef` resolves to the first matching `$dynamicAnchor` encountered on its path from the schema entry point to the reference, as described in the JSON Schema specification
 
+An example is included in the "Schema Object Examples" section below, and further information can be found on the Learn OpenAPI site's ["Dynamic References"](https://learn.openapis.org/referencing/dynamic.html) page.
+
 ###### XML Modeling
 
 The [xml](#schemaXml) property allows extra definitions when translating the JSON definition to XML.
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -2350,7 +2350,7 @@ When used, the `discriminator` indicates the name of the property that hints whi
 
 Implementations MAY support defining generic or template data structures using JSON Schema's dynamic referencing feature:
 
-* `$dynamicAnchor` identifies set of possible schemas (including a default placeholder schema) to which a `$dynamicRef` can resolve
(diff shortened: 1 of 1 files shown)


vicspina:
Can you please take me off your list I don’t want no more emails from you
thank you

versions/3.1.1.md
```diff
@@ -2346,6 +2346,13 @@ While composition offers model extensibility, it does not imply a hierarchy betw
 To support polymorphism, the OpenAPI Specification adds the `discriminator` keyword.
 When used, the `discriminator` indicates the name of the property that hints which schema definition is expected to validate the structure of the model.
 
+###### Generic (Template) Data Structures
+
+Implementations MAY support defining generic or template data structures using JSON Schema's dynamic referencing feature:
+
+* `$dynamicAnchor` identifies set of possible schemas (including a default placeholder schema) to which a `$dynamicRef` can resolve
```
ralfhandl, line 2353:
```suggestion
* `$dynamicAnchor` identifies a set of possible schemas (including a default placeholder schema) to which a `$dynamicRef` can resolve
```

versions/3.1.1.md
```diff
@@ -2346,6 +2346,13 @@ While composition offers model extensibility, it does not imply a hierarchy betw
 To support polymorphism, the OpenAPI Specification adds the `discriminator` keyword.
 When used, the `discriminator` indicates the name of the property that hints which schema definition is expected to validate the structure of the model.
 
+###### Generic (Template) Data Structures
+
+Implementations MAY support defining generic or template data structures using JSON Schema's dynamic referencing feature:
```
ralfhandl, line 2351: Could you add a link to some "intro" material? I couldn't find a good starting point in https://json-schema.org/draft/2020-12/json-schema-core.

handrews: @ralfhandl it's funny, I actually wrote a fairly enormous explanatory section and then cut it as I didn't want to bloat the section.  It's not trivial to explain.

The best explanation of this is @gregsdennis [post on the JSON Schema blog](https://json-schema.org/blog/posts/dynamicref-and-generics).  I'm not sure if we want to have a link to that in the OAS document?  Perhaps I could add a page to the learn site mentioning this and other JSON Schema techniques (including linking from the learn site to the JSON Schema blog?)

It's absolutely true that the JSON Schema spec is not clear on this, in part because we didn't realize `$dynamicRef` and `$dynamicAnchor` would support this when we came up with it.  Also, I have regrets about the writing in that whole section... many regrets...

Does the learn site idea sound good?  If so I will open a PR for that page and we can wait until we agree on it to merge this.  If not, I can try to add back the explanatory material here and we can figure out how much is worth including.  Or any combination of these approaches.

versions/3.1.1.md
```diff
@@ -2689,6 +2696,111 @@ components:
...
+```
+
+```YAML
+components:
+  schemas:
+    genericArray:
+      $id: fully_generic_array
+      type: array
+      items:
+        $dynamicRef: #generic-array
```
gregsdennis, line 2770: This may need to be in quotes. The syntax highlighting makes it look like a comment.

versions/3.1.1.md
```diff
@@ -2346,6 +2346,13 @@ While composition offers model extensibility, it does not imply a hierarchy betw
 To support polymorphism, the OpenAPI Specification adds the `discriminator` keyword.
 When used, the `discriminator` indicates the name of the property that hints which schema definition is expected to validate the structure of the model.
 
+###### Generic (Template) Data Structures
+
+Implementations MAY support defining generic or template data structures using JSON Schema's dynamic referencing feature:
+
+* `$dynamicAnchor` identifies a set of possible schemas (including a default placeholder schema) to which a `$dynamicRef` can resolve
+* `$dynamicRef` resolves to the first matching encountered on its path from the schema entry point to the reference, as described in the JSON Schema specification
```
gregsdennis, line 2354:
```suggestion
* `$dynamicRef` resolves to the first matching `$dynamicAcnhor` encountered on its path from the schema entry point to the reference, as described in the JSON Schema specification
```

This has to be probably the most concise explanation of dynamic references that I've ever seen. 👍

versions/3.1.1.md
handrews: Oops, yeah that's a YAML comment, I'll fix, thanks!

versions/3.1.1.md
handrews: @gregsdennis it only took me three and a half years to figure out how to boil it down... 😅

handrews: Although you typo'd `$dynamicAcnhor` so I'll incorporate this separately as I don't know how to edit a suggestion.

ralfhandl: Fixed proposal

versions/3.1.1.md
handrews: @ralfhandl so it turns out I already wrote a Learn page on dynamic references that links to the JSON Schema blog posts on generics.  I've updated the PR to link to that Learn page.  And also note up front that there is an example in the Schema Object Examples section as otherwise it seems a little abrupt.  Does that work for you?

ralfhandl: I'm a bit torn on directly linking the Learn page without mentioning it as an "Informative reference" in the appendix.

On the other hand there is precedence for direct linking already in v3.1.0, for example the links to https://yaml.org/spec/1.2/spec.html.

ralfhandl: That means we only have to figure out where in the build process this gets added, and ideally also produce an appendix for "Informative references" without changing the spec sources.

handrews: @ralfhandl please feel free to file an issue for that and add it to the "Automation & Infrastructure" GitHub project.  I'd prefer that we not hold up this PR based on the references appendx, as the concerns are orthogonal.

handrews: Technically, only the markdown is normative, so our normative representation doesn't have that appendix at all.  Therefore it should not block anything.

ralfhandl: #3740

versions/3.1.1.md
```diff
@@ -2346,6 +2346,15 @@ While composition offers model extensibility, it does not imply a hierarchy betw
 To support polymorphism, the OpenAPI Specification adds the `discriminator` keyword.
 When used, the `discriminator` indicates the name of the property that hints which schema definition is expected to validate the structure of the model.
 
+###### Generic (Template) Data Structures
+
+Implementations MAY support defining generic or template data structures using JSON Schema's dynamic referencing feature:
```
ralfhandl, line 2351: Does Swagger UI support these already?

handrews: No, why should that matter?  Sswagger is the slowest-moving major tooling vendor out there.

handrews: Actually I'm just assuming "no" - the whole point of this change is to raise awareness that this works – please see the _six different issues requesting this feature_ linked in the issue for this PR.  I'll list them here for convenience- I closed them becauase it's technically possible, but got feedback that it's not obvious enough:

* #519 
* #555 
* #773 
* #957 
* #1749 
* #2453

handrews:
To add a little more context (with apologies for the rapid-fire comment spam):

* This feature is (as noted above) highly desired and has been requested repeatedly since 2015
* It is implemented in some places (I believe that @gregsdennis , who wrote the JSON Schema Blog's post about it, supports it in C#)
* `$dynamicRef` and `$dynamicAnchor` are kind of the `discriminator` of 3.1 – there was a long period where I noticed many OpenAPI tools claimed they supported everything, but with an asterisk saying "except `discriminator`... part of the reason for this is that these new keywords are moderately complex.  But part of it is that people don't understand that they can support a common programming paradigm that is otherwise unsupportable.  We will not see these implemented unless they are connected to real customer use cases, and that's what this PR does.

versions/3.1.1.md
ralfhandl: My experience is that what doesn't show up in Swagger UI isn't written into an OpenAPI description. Which shouldn't prevent us from advertising new features 😄

versions/3.1.1.md
```diff
@@ -2689,6 +2698,111 @@ components:
...
+      }
+    }
+  }
+}
+```
+
+```YAML
+components:
+  schemas:
+    genericArray:
```
ralfhandl, line 2768:
```suggestion
    genericArrayComponent:
```

versions/3.1.1.md
```diff
@@ -2689,6 +2698,111 @@ components:
...
+      $id: array_of_strings
+      $ref: fully_generic_array
+      $defs:
+        stringsOnly:
+          $dynamicAnchor: generic-array
+          type: string
+    objWithTypedArray:
+      $id: obj_with_typed_array
+      type: object
+      required: [dataType, data]
```
ralfhandl, line 2793:
```suggestion
      required: 
      - dataType 
      - data
```
I'd avoid mixing styles

versions/3.1.1.md
```diff
@@ -2689,6 +2698,111 @@ components:
...
+      properties:
+        dataType:
+          enum: [string, number]
+      oneOf:
+      - properties:
+          dataType: {const: string}
+          data: {$ref: array_of_strings}
+      - properties:
+          dataType: {const: number}
+          data: {$ref: array_of_numbers}
```
ralfhandl, line 2803:
```suggestion
      - properties:
          dataType:
            const: string
          data:
            $ref: array_of_strings
      - properties:
          dataType:
            const: number
          data:
            $ref: array_of_numbers
```

versions/3.1.1.md
handrews: @ralfhandl I'm glad you're not seeing it as a barrier :-)  But in all seriousness, there are other tooling vendors, and they do have actual customers/users.  I really don't think it's a good idea for the steering committee of a vendor-neutral project to publicly promote the idea that one company's single product is the gating factor for the entire ecosystem.  We've been working very hard for years now to separate "OpenAPI" from "Swagger", and comments like this do the opposite.

mikekistler (approved):
Looks good! 👍

lornajane (approved):

Issue #3716, opened 2024-04-18
Schema development: main or branch, in parallel or at release time?


handrews:
I'm splitting out the schema-related concerns from #3677 as we decided in a recent TDC meeting that the schema, spec, and gh-pages parts have different priorities (and probably involve different people).

* New schemas are created for each minor or major release
    * Patch releases are limited to changes that would not require a _separate_ schema for each patch release in a minor release line
    * Should new minor/major schemas be created and updated in parallel with spec changes, or deferred and created all at once just prior to the release, as part of the final review effort?
    * If created and updated in parallel, this should be integrated with the development branch process being sorted out in #3677, presumably continuing to be published to main in the current `schemas/`, just as #3677 recommends publishing the markdown to the current `versions/` location
* Schemas can currently _also_ be changed for bug-fixing or functional improvement at any time, and released independently from the release schedule (see also #3715)
    * This includes but is not limited to patch release-related bug fixes
    * This might require a different tagging strategy from whatever we do for spec releases

handrews:
So... given the proposals for branching in PR #4154 (background in issue #3677), I can see a few options for schemas:

## share the `X.Y-dev` branches, but release independently

This would consolidate the schemas, based on the 3.1.1 files, under `src/` (alongside the spec which is `src/oas.md`) on the `dev` branch, and we'd do the release-line-specific schema work on the appropriate `X.Y-dev` branch.

Since schema releases are independent of patch releases, instead of making `X.Y.Z-rel` branches to rename to the published file location, we'd make branches like `2024-10-22-rel`.

**PROS:**
* Same process and working branches as the spec
* We can merge changes across release lines
* This allows simultaneous spec and schema changes (which is a surprisingly controversial proposal, but I'm putting it under "PRO" based on personal preference)

**CONS:**
* The release branches are a bit unwieldy, and schemas might be published significantly more often (it's a bit hard to say, and also depends on how define the scope of schemas — see #4141 and #4147)

**MITIGATION:**
* We could probably automate more of the publishing process

## Keep the `schemas/` layout but do the work on `dev`

**PROS:**
* No renaming for release needed
* We can still batch up changes on `dev` and only merge them to `main` when we want them published

**CONS:**
* Schema dev is still divorced from the spec changes that drive it (which I consider a con)
* No merges, we'd have to do `git --format-patch | sed | git am -3` hackery still
* Different process to remember compared to spec work

**MITIGATION:**
[...]

PR #3718, opened 2024-04-19
 Clarify `openIdConnectUrl` #3630

m-mohr:
Clarification according to `openIdConnectUrl ` as discussed in #3630


versions/3.1.1.md
```diff
@@ -3198,7 +3198,7 @@ Field Name | Type | Applies To | Description
...
 <a name="securitySchemeBearerFormat"></a>bearerFormat | `string` | `http` (`"bearer"`) | A hint to the client to identify how the bearer token is formatted.  Bearer tokens are usually generated by an authorization server, so this information is primarily for documentation purposes.
 <a name="securitySchemeFlows"></a>flows | [OAuth Flows Object](#oauthFlowsObject) | `oauth2` | **REQUIRED**. An object containing configuration information for the flow types supported.
-<a name="securitySchemeOpenIdConnectUrl"></a>openIdConnectUrl | `string` | `openIdConnect` | **REQUIRED**. OpenId Connect URL to discover OAuth2 configuration values. This MUST be in the form of a URL. The OpenID Connect standard requires the use of TLS.
+<a name="securitySchemeOpenIdConnectUrl"></a>openIdConnectUrl | `string` | `openIdConnect` | **REQUIRED**. OpenID Connect URL to discover the [OpenID Connect Configuration Information](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig). This MUST be in the form of a URL that points to the well-known document and ends with `/.well-known/openid-configuration`. The OpenID Connect standard requires the use of TLS.
```
lornajane, line 3201:
```suggestion
<a name="securitySchemeOpenIdConnectUrl"></a>openIdConnectUrl | `string` | `openIdConnect` | **REQUIRED**. OpenID Connect URL to discover the [OpenID Connect Configuration Information](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig). This MUST be in the form of a URL. The OpenID Connect standard requires the use of TLS.
```

Could we remove the specifics of the OpenId implementation from this change, I don't think we should duplicate the contents of another spec.

lornajane (changes requested):

versions/3.1.1.md
m-mohr: Without it's unclear whether you need to provide the issuer or the well-known document. Changing/Removing this would pretty much render this PR useless.

handrews: @m-mohr the sentiment in the TDC call today was that the `.well-known` requirement is already present in the OpenIDConnect spec, and we normatively cite the spec.  So we already incorporate these, but the advantage of not repeating it is that if OpenIDConnect changes, we do not have to change.  We all agreed that the change to link to the OpenID Connect Configuration Information (instead of "OAuth2 configuration values) was a valuable improvement.  The link also takes you directly to the `.well-known` requirement so that should be clear.

AxelNennker: I think even without the .well-known part this PR is needed because of the reference  to https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig.
This part `OpenId Connect URL to discover OAuth2 configuration values` was always wrong. OIDC metadata and OAuth2 metadata are not the same.
We, Deutsche Telekom, have a keycloak service where each realm has its own metadata, so openIdConnectUrl cannot be at .well-known URL.
So, [this suggestion](https://github.com/OAI/OpenAPI-Specification/pull/3718#discussion_r1579821713) is not only removing something from another spec, but needed.
Sorry, I did not see this earlier.

m-mohr: But then you don't follow the OpenID Connect spec. It requires to use well known part?!

AxelNennker: The spec says `OpenID Providers supporting Discovery MUST make a JSON document available at the path formed by concatenating the string /.well-known/openid-configuration to the Issuer.`
The Issuer can be e.g. `https://keycloak.example.com/<realm>/` then the openIdConnectUrl would be 
`https://keycloak.example.com/realms/{realm}/.well-known/openid-configuration`.

I think OAI should not care too much, other that the value of openIdConnectUrl should be TLS and point to a JSON document that contains the OP's metadata. e.g.
https://mobileconnect.telekom.de/.well-known/openid-configuration 
https://accounts.login.idm.telekom.com/.well-known/openid-configuration

m-mohr: > We, Deutsche Telekom, have a keycloak service where each realm has its own metadata, so openIdConnectUrl cannot be at .well-known URL.

Please read more carefully, your use case is covered. It says "**ends** with `/.well-known/openid-configuration`", not the path must be exactly /.well-known/openid-configuration.

> So we already incorporate the requirement, but the advantage of not repeating it is that if OpenIDConnect changes.

So this means we can even further slim it down, because TLS is also required by the spec and it's a repetition.

```suggestion
<a name="securitySchemeOpenIdConnectUrl"></a>openIdConnectUrl | `string` | `openIdConnect` | **REQUIRED**. Well-known URL to discover the [OpenID Provider Metadata](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata).
```

handrews (approved):

versions/3.1.1.md
AxelNennker: I would keep the TLS requirement because I like openapi tools to insist on secure connections. Developers often ignore TLS requirements - I can only speculate why. Please keep the TLS requirement.

ralfhandl (approved):

ralfhandl:
@lornajane The text cited from the referenced spec was removed, instead the "well-known" is now part of the link text.

I think that matches the spirit of the requested change, please confirm.

versions/3.1.1.md
```diff
@@ -3198,7 +3198,7 @@ Field Name | Type | Applies To | Description
...
 <a name="securitySchemeBearerFormat"></a>bearerFormat | `string` | `http` (`"bearer"`) | A hint to the client to identify how the bearer token is formatted.  Bearer tokens are usually generated by an authorization server, so this information is primarily for documentation purposes.
 <a name="securitySchemeFlows"></a>flows | [OAuth Flows Object](#oauthFlowsObject) | `oauth2` | **REQUIRED**. An object containing configuration information for the flow types supported.
-<a name="securitySchemeOpenIdConnectUrl"></a>openIdConnectUrl | `string` | `openIdConnect` | **REQUIRED**. OpenId Connect URL to discover OAuth2 configuration values. This MUST be in the form of a URL. The OpenID Connect standard requires the use of TLS.
+<a name="securitySchemeOpenIdConnectUrl"></a>openIdConnectUrl | `string` | `openIdConnect` | **REQUIRED**. [Well-known URL](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig) to discover the [OpenID Provider Metadata](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata).
```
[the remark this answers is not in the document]
m-mohr: Another comment above asked to not cite content from the original spec, as such TLS should not be included here.

As such I limit the change to the casing changes:

```suggestion
<a name="securitySchemeOpenIdConnectUrl"></a>openIdConnectUrl | `string` | `openIdConnect` | **REQUIRED**. URL to the [OpenID provider metadata](https://openid.net/specs/openid-connect-discovery-1_0.html).
```

versions/3.1.1.md
m-mohr: see comment below

ralfhandl (approved):

lornajane (approved):
A good clarification for this existing field, thank you.

m-mohr:
How is this going to be made available in 3.2.0?

handrews:
@m-mohr someone needs to copy the fix over to the `versions/3.2.0.md` file on the `v3.2.0-dev` branch.  If you have time to do it, that would be most appreciated!  If not, we'll audit the branches at some point to reconcile.  (starting with 3.2, we will be moving to a [more normal use of git branches of a single spec file](https://github.com/OAI/OpenAPI-Specification/issues/3716).

PR #3719, opened 2024-04-19
Default encoding object (3.0.4)

handrews:
Addresses https://github.com/OAI/OpenAPI-Specification/issues/3710#issuecomment-2065222963 – If this is approved I will port it to 3.1.1 and 3.2.0.

If no Encoding Object is provided, the behavior is the same as an empty Encoding Object, as none of its fields are required.


versions/3.0.4.md
```diff
@@ -1373,7 +1373,7 @@ Field Name | Type | Description
...
-<a name="mediaTypeEncoding"></a>encoding | Map[`string`, [Encoding Object](#encodingObject)] | A map between a property name and its encoding information. The key, being the property name, MUST exist in the schema as a property. The encoding object SHALL only apply to `requestBody` objects when the media type is `multipart` or `application/x-www-form-urlencoded`.
+<a name="mediaTypeEncoding"></a>encoding | Map[`string`, [Encoding Object](#encodingObject)] | A map between a property name and its encoding information. The key, being the property name, MUST exist in the schema as a property. The encoding object SHALL only apply to `requestBody` objects when the media type is `multipart` or `application/x-www-form-urlencoded`.  If no Encoding Object is provided for a relevant parameter, the behavior is determined by the default values documented for the Encoding Object.
```
ralfhandl, line 1376: How do I document default values for an Encoding Object? Couldn't find corresponding specification text.

ralfhandl (approved):

versions/3.0.4.md
mikekistler, line 1376: I have some questions/concerns:

The statement
> The encoding object SHALL only apply to `requestBody` objects ...

is confusing to me, because the "encoding object" is the value part of the map entries of "encoding", which applies to properties of a schema.  Later in the spec I see this statement:
> An encoding attribute ... is only applicable to multipart and application/x-www-form-urlencoded request bodies.

Should "encoding object" actually be "encoding attribute" (or "encoding field") in the originally cited text?

Also confusing for me is the word "parameter" in the sentence:
> If no Encoding Object is provided for a relevant parameter ...

From the earlier text, I thought that encoding objects were provided for _properties_ of a schema -- not parameters. Maybe this is nit picking words but in OAS these are very different things. Secondarily, I'm not sure what "relevant" means in this context. When would a property be "irrelevant"? And if there are "irrelevant" properties, then shouldn't this qualification be added above:
> The key, being the _relevant_ property name,

I think the following suggested text would resolve my concerns:
```suggestion
<a name="mediaTypeEncoding"></a>encoding | Map[`string`, [Encoding Object](#encodingObject)] | A map between a property name and its encoding information. The key, being the property name, MUST exist in the schema as a property. The encoding attribute SHALL only apply to `requestBody` objects when the media type is `multipart` or `application/x-www-form-urlencoded`.  If no Encoding Object is provided for a property, the behavior is determined by the default values documented for the Encoding Object.
```

mikekistler (approved):
Looks good! 👍

handrews:
Two TSC members have approved, therefore I'm merging.

PR #3723, opened 2024-04-19
Clarify `spaceDelimited` with spaces (3.0.4)

### LINKED ISSUE #3550: Serialization of `spaceDelimited` query parameter with a list of strings.
The specification says that `spaceDelimited` should be used for array or object values. How should an space be escaped when serializing into this format?

## Example

Given the array `["Hello World", "Nice to see you"]` serialization of this value using the `spaceDelimited` style is not properly defined.

### Options
- `Hello%20World%20Nice%20to%20see%20you`: Looses information
- `Hello%20World Nice%20to%20see%20you`: Doesn't follow from the examples on the spec
- `Hello World%20Nice to see you`: Is not valid as a query parameter. If we URLEncode again then the separator ends up URLEncoded twice.
- `Hello%2520World%20Nice%2520to%2520see%2520you`: Encodes the array items twice. I think this is the most reasonable option but it is a bit weird.
- `Hello%20World%2520Nice%20to%20see%20you`: Ecodes the separator twice.

handrews:
Fixes #3550.  I will port this to 3.1.1 and 3.2.0 if it is accepted.

The OAS does not define how to avoid collisions between parameter values and style delimiters.  Mostly, this is straightforward, but the need to URL encode space characters introduces extra confusion.  Make it clear that managing this occurs outside of the use of style, and is the responsibility of users.

### DIFF
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -1058,6 +1058,7 @@ spaceDelimited | `array`, `object` | `query` | Space separated array values or o
 pipeDelimited | `array`, `object` | `query` | Pipe separated array values or object properties and values. This option replaces `collectionFormat` equal to `pipes` from OpenAPI 2.0.
 deepObject | `object` | `query` | Provides a simple way of rendering nested objects using form parameters.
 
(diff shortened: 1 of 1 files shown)


ralfhandl (approved):

mikekistler (approved):
Looks good! 👍

PR #3724, opened 2024-04-20
Clarify wording on when `encoding` is used (3.0.4)

### LINKED ISSUE #3421: Parameter schema specification via content
I'd like to clarify the purpose of the [Parameter Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#parameter-object) schema specification depending on the media type via content.

> For more complex scenarios, the [content](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#parameterContent) property can define the media type and schema of the parameter.

What is this complex scenario?
Having different media types defined for the body, the parameters schema should depend on the media type of the body?

```yaml
  parameters:
    DefinedParameter:
      in: query
      name: value
      required: true
      style: form
      explode: false
      content:
        "application/json":
          schema:
            type: integer
            default: 5
          example: 5
        "multipart/form-data":
          schema:
            type: object
            $ref: "…"
```

If the [Media Type Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#media-type-object) encoding property is not to be used when used with Parameter Objects, so the encoding of possibly different parameter types is set by the parameter independent of the schema.
If the encoding is free be used for Parameter Objects… for an object this will combine different encodings for the objects properties?

handrews:
Fixes #3421 

The previous wording could be read to only restrict what sort of request bodies could use an Encoding Object, while allowing that it could be used with other entities that use a Media Type Object.

This emphasizes that it is only relevant to Request Body Objects, and within that, only for specific media types.

***NOTE:** I am aware that the internal links are wrong.  I am assuming that we will fix all of those at once at some point, so I opted to be consistent for now.*

I will port this to 3.1.1 and 3.2.0 if it is accepted.


ralfhandl (approved):

versions/3.0.4.md
```diff
@@ -1373,7 +1373,7 @@ Field Name | Type | Description
...
-<a name="mediaTypeEncoding"></a>encoding | Map[`string`, [Encoding Object](#encodingObject)] | A map between a property name and its encoding information. The key, being the property name, MUST exist in the schema as a property. The encoding object SHALL only apply to `requestBody` objects when the media type is `multipart` or `application/x-www-form-urlencoded`.
+<a name="mediaTypeEncoding"></a>encoding | Map[`string`, [Encoding Object](#encodingObject)] | A map between a property name and its encoding information. The key, being the property name, MUST exist in the schema as a property. The encoding object SHALL only apply to [Request Body Objects](#requestBodyObject), and only when the media type is `multipart` or `application/x-www-form-urlencoded`.
```
mikekistler, line 1376:
```suggestion
<a name="mediaTypeEncoding"></a>encoding | Map[`string`, [Encoding Object](#encodingObject)] | A map between a property name and its encoding information. The key, being the property name, MUST exist in the schema as a property. The encoding attribute SHALL only apply to [Request Body Objects](#requestBodyObject), and only when the media type is `multipart` or `application/x-www-form-urlencoded`.
```

mikekistler: There is a similar statement later in the spec:
> This attribute is _only_ applicable to `multipart` and `application/x-www-form-urlencoded` request bodies.

Do you want to update the language there to be more consistent with this change?

mikekistler (approved):
Looks good. 👍 

I left a few minor suggestions for improvement, but I'll leave it to you to decide whether or not these are useful.

ralfhandl:
Two approvals, merged

PR #3727, opened 2024-04-20
Clarify how to model binary data in 3.1

### LINKED ISSUE #1544: Clarify: "type": "string", "format": "binary" in non-entity-body

### LINKED ISSUE #2476: Question: `Content-Encoding` HTTP header field and `contentEncoding`

### LINKED ISSUE #2868: conflicting order of operations when applying Content-Encoding and Content-Type headers
The specification describes how to handle requests/responses with both a Content-Type and Content-Encoding header:

" JSON Schema offers a contentEncoding keyword, which may be used to specify the Content-Encoding for the schema. ... The encoding specified by the contentEncoding keyword is independent of an encoding specified by the Content-Type header in the request or response or metadata of a multipart body – **when both are present, the encoding specified in the contentEncoding is applied first and then the encoding specified in the Content-Type header**."
--- https://spec.openapis.org/oas/v3.1.0#considerations-for-file-uploads

(emphasis mine)

However, this ordering is backwards with how an openapi validator would operate: the contentEncoding keyword is hidden inside the schema, at a lower level than the media-type property.

The way around this would be for the validator to peek inside the schema to see if there is a contentEncoding keyword at the top level, and use it to decode the content first before then using the media-type decoder. That's pretty ugly IMO; schemas ought to be opaque.  It would be better for this data to be brought up into the media-type object itself, or simply automatically decode the content according to the Content-Encoding header before applying the media-type in the Content-Type header (is it really a desired feature that the openapi document must specify the Content-Encoding that is used? what if multiple different encodings are supported, e.g. gzip compression and brotli compression - how would that be indicated?)

In practice, how is a desired content encoding indicated in a document in order to enable proper validation of the request/response body content?  I didn't find any hints on the [Swagger site](https://swagger.io/docs/specification/about/).

handrews:
* Fixes #2476 
* Fixes #2868
* Fixes #1544 for 3.1.x (See PR #3187 for the analogous 3.0.x fix).  If accepted, I will port this to 3.2.0.

_Note:  I wrote most of this quite some time ago and just tidied it up now... please review it carefully as I might have missed some unfinished aspect._

This reorganizes binary data-related guidance into a "Working With Binary Data" section, as has already been done in 3.0.4.

This includes more detailed guidance on when various approaches to binary data make sense (e.g. you cannot stuff raw binary into JSON no matter what you put in your Schema Object, and while you can base64-encode entire message bodies, it takes up a lot more space for no clear benefit).

Also noted that only `multipart` media types with named parts are supported, as they are modeled as an object.

### DIFF
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -174,7 +174,7 @@ Keyword | Raw | Encoded | Comments
 `contentMediaType` | `image/png` | `image/png` | can sometimes be omitted if redundant (see below)
 `contentEncoding`  | _omit_ | `base64`&nbsp;or&nbsp;`base64url` | other encodings are [allowed](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#section-8.3)
 
-Note that the encoding indicated by `contentEncoding`, which inflates the size of data in order to represent it as 7-bit ASCII text, is unrelated to HTTP's `Content-Encoding` header, which indicates whether and how a message body has been compressed and is applied after all content serialization described in this section has occurred.
+Note that the encoding indicated by `contentEncoding`, which inflates the size of data in order to represent it as 7-bit ASCII text, is unrelated to HTTP's `Content-Encoding` header, which indicates whether and how a message body has been compressed and is applied after all content serialization described in this section has occurred.  Since HTTP allows unencoded binary message bodies, there is no standardized HTTP header for indicating base64 or similar encoding of an entire message body.
 
 Using a `contentEncoding` of `base64url` ensures that URL encoding (as required in the query string and in message bodies of type `application/x-www-form-urlencoded`) does not need to further encode any part of the already-encoded binary data.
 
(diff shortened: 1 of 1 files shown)


handrews:
Paging @karenetheridge as it won't let me add you as a requested reviewer.

ralfhandl (approved):

handrews:
@ralfhandl after your approval I removed an example that contradicts the guidance in the revised section, and clarified that there is no HTTP header that can indicate base64-encoding of an entire message body (which was the problem with the removed example- it ended up senging a PNG as text with no in-message explanation of the encoding).

iox666999:
Nothing to preview

miqui (approved):

PR #3728, opened 2024-04-20
Non-string discriminator values have undefined behavior (3.0.4)

### LINKED ISSUE #3475: Clarify that the discriminating property MUST have only string values
The spec does not formally say that the property named in `propertyName` MUST only have string values.  This is more-or-less implied by how it is  used, but as seen in #3439, sometimes people expect other types to be mapped to strings, presumably somewhat like how they are in query parameters (which, as has been noted elsewhere, is also somewhat under-specified).

Alternatively, a minor release could define a mapping for integer, boolean, and null values if we really wanted to, but I doubt that we want to do that.

handrews:
* Fixes #3475 

The "undefined" wording was chosen to allow implementations that make a good-faith effort to accommodate non-string values to remain compliant.  However, new implementations are not expected to implement any sort of type coersion, and this guides API designers away from that expectation.

If accepted, I will port this to 3.1.1.  For 3.2.0 we might either want to strengthen this to outright forbid non-strings or define a mapping for at least certain non-strings (e.g. booleans).

### DIFF
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -2718,7 +2718,7 @@ Field Name | Type | Description
 <a name="propertyName"></a>propertyName | `string` | **REQUIRED**. The name of the property in the payload that will hold the discriminator value.
 <a name="discriminatorMapping"></a> mapping | Map[`string`, `string`] | An object to hold mappings between payload values and schema names or references.
 
-The discriminator object is legal only when using one of the composite keywords `oneOf`, `anyOf`, `allOf`.  Note that because the discrinating property's value is used as a component name and/or as the key in the `mapping` object, the behavior of any value that is not a string is undefined.
+The discriminator object is legal only when using one of the composite keywords `oneOf`, `anyOf`, `allOf`.  Note that because the discriminating property's value is used as a component name and/or as the key in the `mapping` object, the behavior of any value that is not a string is undefined.
 
 In OAS 3.0, a response payload MAY be described to be exactly one of any number of types:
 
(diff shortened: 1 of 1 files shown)


versions/3.0.4.md
```diff
@@ -2718,7 +2718,7 @@ Field Name | Type | Description
 <a name="propertyName"></a>propertyName | `string` | **REQUIRED**. The name of the property in the payload that will hold the discriminator value.
 <a name="discriminatorMapping"></a> mapping | Map[`string`, `string`] | An object to hold mappings between payload values and schema names or references.
 
-The discriminator object is legal only when using one of the composite keywords `oneOf`, `anyOf`, `allOf`.
+The discriminator object is legal only when using one of the composite keywords `oneOf`, `anyOf`, `allOf`.  Note that because the discrinating property's value is used as a component name and/or as the key in the `mapping` object, the behavior of any value that is not a string is undefined.
```
ralfhandl, line 2721:
```suggestion
The discriminator object is legal only when using one of the composite keywords `oneOf`, `anyOf`, `allOf`.  Note that because the discriminating property's value is used as a component name and/or as the key in the `mapping` object, the behavior of any value that is not a string is undefined.
```

ralfhandl (approved):

mikekistler (approved):
Looks good. 👍

handrews:
Since there are two TSC approvals here, I'm going to hit the merge button - the TSC emphasized again in today's TDC call that the 2nd TSC approver can merge, so I'm assuming I can merge with 2 TSC approvals.

PR #3731, opened 2024-04-20
Correctly escape example operationRef URLs

### LINKED ISSUE #1677: URL-encoding for operationRef

handrews:
* Fixes #1677 

It's ugly, but that's URL-encoding for you.  If accepted, I will port this to 3.1.1 and 3.2.0.

### DIFF
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -2124,7 +2124,7 @@ field in an [Operation Object](#operationObject)), references MAY also be made t
 links:
   UserRepositories:
     # returns array of '#/components/schemas/repository'
-    operationRef: '#/paths/~12.0~1repositories~1{username}/get'
+    operationRef: '#/paths/~12.0~1repositories~1%7Busername%7D/get'
     parameters:
       username: $response.body#/username
 ```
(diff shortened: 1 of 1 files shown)


ralfhandl (approved):

mikekistler (approved):
Looks ... ahh ... ugly ... but correct. 👍

karenetheridge:
I'm not sure that this is needed now that we have the change from #3455 ?

handrews:
@karenetheridge the schema is not normative and issue #1677 has TSC folks acknowledging that it needs URL-encoding.

So I guess we have to burn up Thursday meeting time on this after all?  This is what is so frustrating about trying to move forward.  We have such an immense backlog of minor things that they are all stepping on each other, and any attempt to resolve them runs into yet another contradiction or change in opinion.

handrews:
The normative spec clearly states that `operationRef` is:

> A relative or absolute URI reference to an OAS operation.

Not a template.  So I think this change is correct and the schema change was wrong.  If we want it to be an unencoded template we'd need to change the spec language.

PR #3735, opened 2024-04-21
Support streaming JSON formats such as json-seq

### LINKED ISSUE #3730: Support application/json-seq and similar JSON-based sequential formats
### Splitting this out of issue #1576...

...which was originally about binary streaming but later drifted into a discussion of JSON streaming.  Several real-world uses for JSON streaming were cited:

* https://github.com/OAI/OpenAPI-Specification/issues/1576#issuecomment-735024952 (Twitter API)
* https://github.com/OAI/OpenAPI-Specification/issues/1576#issuecomment-763257216 (someone's own company's API)
* https://github.com/OAI/OpenAPI-Specification/issues/1576#issuecomment-795882357 (Kubernetes API of some sort)
* https://github.com/OAI/OpenAPI-Specification/issues/1576#issuecomment-795882357 (request for "json-lines")

The binary streaming case just required a clarification and has been addressed in PR #3729 for 3.0.4, but adding proper support for a new media type will have to go into 3.2.0.

### Also Discussed in https://github.com/OAI/OpenAPI-Specification/discussions/2707

<div type='discussions-op-text'>

<sup>Originally posted by **Skeeve** September  9, 2021</sup>
Is there a way to describe a [json-lines](https://jsonlines.org/) with OpenAPI?

Besides the fact that there seems to be no mimetype for it yet, I'm wondering if it's possible to describe such a response.

In theory my response could be an array of objects, but I received the question whether or not I could deliver as json-lines, meaning: Just the objects, one per line.

Since I'm using OpenAPI to describe my API I'm puzzled as how to describe this response. I could simply define the response as being of type "string", but this is not very helpful for readers of my Api-spec.

P.S. I already asked at [stackoverflow](https://stackoverflow.com/questions/69123769/openapi-how-to-define-json-lines-response) and was pointed to here.</div>

### Further thoughts

There are now several formats for [JSON streaming](https://en.wikipedia.org/wiki/JSON_streaming)

* [RFC 7464](https://www.rfc-editor.org/rfc/rfc7464.html) "JavaScript Object Notation (JSON) Text Sequences" (`application/json-seq`)
    * [RFC 8124](https://www.rfc-editor.org/rfc/rfc8142.html) "GeoJSON Text Sequences" (`application/geo+json-seq`)
* [JSON Lines](https://jsonlines.org/) ([implementations](https://jsonlines.org/on_the_web/); [issue](https://github.com/wardi/jsonlines/issues/19) tracking proposal for `application/jsonl` media type [with recent activity](https://github.com/wardi/jsonlines/issues/19#issuecomment-1965152150)
* [NDJSON](https://github.com/ndjson/ndjson-spec) (currently uses `application/x-ndjson`; there might be an effort towards `application/ndjson` or merging with JSON lines

It should be clear from the lists of implementations, requests from several different folks in both issues and discussions, and the existence of significant derivative specs (GeoJSON is widely used within the geospatial data space) that this is a real use case with many applications.

AFAICT, the distinctions among the three formats are irrelevant to modeling their contents, and only involve the choice of delimiter, the allowability of blank lines/sequences (which are skipped during parsing regardless of the format), and the details of error handling.  JSON Lines and NDJSON might actually [merge](https://github.com/ndjson/ndjson-spec/issues/35#issuecomment-1285673417).

From a data modeling perspective, I think we could support these in OAS 3.2 by noting that:

* They can be modeled as if they were a JSON array
* Since JSON Schema implementations will not support these formats directly, tools will need to to do one of:
  * convert the data into an actual JSON array prior to schema validation (easy for a finite-length document)
  * individually apply the relevant subschema(s) to each JSON text in the sequence (better for an incrementally processed stream)

Tools that directly integrate and use JSON Schema implementations would need to handle the translation, but that's the main tooling impact.  We could write the requirements around any sequential JSON format rather than tying it to any specific media type, since there seem to be multiple more-or-less equivalent approaches (I'm probably missing some).

handrews:
* Fixes #3730 (which consolidated at least two other issues/discussions)

These formats can easily be modeled as if they were JSON arrays, with fairly minimal new tooling requirements.  As documented in the linked issue, there is ample evidence for their use, and the different formats do not require different modeling approaches.  This approach works for all known current streaming formats (which may or may not be literally streamed in the HTTP sense).

This seems like a nice, small-but-impactful enhancement for 3.2.0.

### DIFF
--- a/versions/3.2.0.md
+++ b/versions/3.2.0.md
@@ -2350,6 +2350,12 @@ There are two ways to define the value of a discriminator for an inheriting inst
 - Override the schema name by overriding the property with a new value. If a new value exists, this takes precedence over the schema name.
 As such, inline schema definitions, which do not have a given id, *cannot* be used in polymorphism.
 
+###### Streaming JSON
+
+Several media types, including but not limited to `application/json-seq` (defined in [RFC7464](https://datatracker.ietf.org/doc/html/rfc7464)), allow sending sequences of JSON documents separated by some sort of delimiter.  These formats can be modeled with [Schema Objects](#schemaObject) by treating the sequence as an array of JSON data.
+
+Since JSON Schema implementations do not support these formats directly, validating JSON streams requires either converting the stream to a single array document or manually applying the correct subschema(s) to each JSON document in the stream.  The latter approach is more suitable for indefinite-length streams where all documents (or all documents after a specific number of initial documents) are expected to be valid against the same subschema.
+
 ###### XML Modeling
 
 The [xml](#schemaXml) property allows extra definitions when translating the JSON definition to XML.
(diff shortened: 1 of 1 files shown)


ralfhandl (approved):

handrews:
Moving to draft for now as 3.2 PRs that aren't ports from the patch releases can wait until after we finish the patch releases.

miqui (approved):

handrews:
Closing this in preparation to re-submit once the repo is re-organized for 3.2

Issue #1502, opened 2018-03-06
Support for arbitrary query strings


pofallon:
Hello!  I'm trying to use the OpenAPI 3.0 spec to document an API that supports a subset of the [Resource Query Language (RQL)](https://github.com/persvr/rql#rql-rules).  However, because this query string format doesn't adhere to the typical `name=value` parameter structure, I'm struggling to express it in OpenAPI.  

For example, a valid query string might be:  `http://localhost:5000/books?select(title,author)&gt(published,2000-06-30T12:20:08Z)&sort(published)`.

I wouldn't expect OpenAPI to understand RQL, but if there were a way to specify that an endpoint may have a query string, and that the query string contents are beyond the scope of OpenAPI, that would suffice.  (I'm really thinking from the Swagger UI perspective -- so a user wanting to try the API can type in the entire RQL query string themselves.)

If anyone has any thoughts on other ways this can be accomplished with the existing 3.0 specification, that would be great too!

Thanks,
Paul

handrews:
Hmm.... is there (or could there be) a media type for RQL query strings?

Technically, the `name=value` parameter structure is not a URI thing, it's the `application/x-www-form-urlencoded` media type.  As far as URIs are concerned everything in between the `?` and the `#` is the query string with no mandatory structure.

OpenAPI could offer an `in` value of `queryContent` or something like that, and then use the `content` field for that parameter to express RQL as a media type.  Maybe?  Just brainstorming a bit here, there might be obvious problems with this approach as I've not thought it through.  I definitely have no idea what tooling challenges this might introduce (Hi @webron!)

(interestingly, this would also allow expressing relationships among query parameters as you could use `queryContent` with `application/x-www-form-urlencoded` and then apply `schema` to that to produce the `name=value` stuff - see also #256)

(also, I can't tell if #1054 is similar to this or not)

hkosova:
As @handrews mentioned, you probably need to use the [`content`](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#parameterContent) keyword for your query parameter. `content` uses the media type identifier to define the serialization format for the parameter schema.

You can find a similar example for SPARQL at https://github.com/OAI/Gluecon-Workshop/blob/master/Different/parameters.md#complex-parameters-in-v3

In your case, you can try something like this:

```yaml
    parameters:
    - name: complexQuery
      in: query
      content:
        application/rdf:
          schema:
            ???   # type: string? type: object?
[...]

PR #3738, opened 2024-04-22
Allow forbidding undefined query string params

### LINKED ISSUE #2511: How to disallow unknown query parameter?
The [operation object](https://spec.openapis.org/oas/v3.0.3#operation-object) allows folks to specify parameters, but how can I express that no unnamed parameters may be accepted?

We've had issues where clients are passing in query parameters thinking that they're accepted when actually they are not.

How do folks feel about introducing something like `additionalQueryParameters` for parameters?

```yaml
/foo:
  get:
    parameters:
      limit:
        in: query
      offset:
        in: query
    additionalQueryParameters: false
    ...
```

Then when folks do `GET /foo?start=xxx-xxx-xxx` they'll get an error instead of a silent failure.

Other links on the web:
- https://stackoverflow.com/questions/58187233/is-it-possible-to-disallow-unknown-query-parameters-in-an-openapi-v3-spec
- https://github.com/cdimascio/express-openapi-validator/issues/133
- https://github.com/opengeospatial/ogcapi-features/issues/128
- https://softwareengineering.stackexchange.com/a/311489

handrews:
* Fixes #2511


versions/3.2.0.md
```diff
@@ -752,6 +752,7 @@ Field Name | Type | Description
 <a name="pathItemTrace"></a>trace | [Operation Object](#operationObject) | A definition of a TRACE operation on this path.
 <a name="pathItemServers"></a>servers | [[Server Object](#serverObject)] | An alternative `server` array to service all operations in this path.
 <a name="pathItemParameters"></a>parameters | [[Parameter Object](#parameterObject) \| [Reference Object](#referenceObject)] | A list of parameters that are applicable for all the operations described under this path. These parameters can be overridden at the operation level, but cannot be removed there. The list MUST NOT include duplicated parameters. A unique parameter is defined by a combination of a [name](#parameterName) and [location](#parameterIn). The list can use the [Reference Object](#referenceObject) to link to parameters that are defined at the [OpenAPI Object's components/parameters](#componentsParameters). 
+<a name="pathItemUndefinedQueryParameters"></a>undefinedQueryParameters | `boolean` | Sets the default behavior of the same field at the [Operation Object](#operationUndefinedQueryParameters) level; see that field for the description of its effects.  Default value is `true`.
```
ralfhandl, line 755: Should this rather be `allowUndefinedQueryParameters` in line with `allowEmptyValue` and `allowReserved`?

mikekistler (commented):
I am wondering how this feature will be used in practice.

I don't think users of the API get any benefit from it. It does not "describe" any useful behavior of the API, unlike `additionalProperties` which can include a `description` that explains the meaning of additional properties. It specifies when the operation is _required_ to fail but not when it actually will fail, since undefined query parameters can still cause failures when `undefinedQueryParameters` is `true`, the default. So users must be prepared for failures either way.

I suppose this clarifies to implementers how they should treat undefined query parameters, but this assumes that the API description is written by someone other than the team that implements the API, which in my experience is rather rare.

I do understand that, from an API governance perspective, rejecting undefined query parameters is a best practice and failures to do so should be rooted out and corrected. But I don't think adding `undefinedQueryParameters` helps much on this front -- and could actually hinder it, since it gives implementers an excuse for not following this best practice -- "Well, we didn't document that undefined query parameters must fail, so we don't need to fix our service".

In short, I don't see much practical value in this new keyword and would not recommend it's use in any of the APIs I work with, but if others feel strongly that it is needed I wouldn't oppose it.

versions/3.2.0.md
handrews: @ralfhandl it would have to have a default of `true` in that case to retain minor release compatibility, but I'd be fine with making the naming more consistent.

handrews:
@mikekistler according to a recent discussion I had with some folks, it's actually increasingly common to _intentionally allow_ unexpected query parameters because of the tendency of various entities to tack various sorts of tracking parameters onto query strings.

It could be better documented here that _not forbidding_ undefined parameters is not a guarantee that they will be handled well or at all.  It's just that forbidding them ensures enforcement at all tooling levels instead of deferring to the server.

Also, you say this wouldn't be useful, but the issue has many people chiming in that they want it.

----

As a meta-topic, I want to note that this issue has been open for more than two years _without a single comment saying we shouldn't do it._  If the TSC doesn't want changes made, it should _close the issues and say so._  Not leave it open as if we might do something about it and then shut it down the instant someone takes action that looks non-controversial.

karenetheridge:
I'm unsure that a specific config is needed here given that it's possible to do already with style=form, explode=true, type=object, which will cause a single parameter will consume all values in the query string as a json object.

This functionality should be documented as an example in the specification, as it is quite useful but not immediately obvious to those unfamiliar with json schema (or those who have written an openapi validator 😛).

versions/3.2.0.md
handrews: I don't even know what I was thinking with my last comment, the default is `true` with the current name already.  I'll make the name change, although this is still blocked on the discussion in the issue.

handrews:
As noted in #2511, I now think solving #1502 is a more comprehensive solution to this and several other problems.  I agree with @karenetheridge that the workaround (which handles probably most cases) ought to be documented as well, but that can be handled separately for the patch releases and needn't wait to 3.2.

I'm going to close this to reduce clutter- we can revisit it if #1502 is rejected.

Issue #3737, opened 2024-04-22
pipeDelimited and spaceDelimited style examples can be unresolvable


charjr:
## What the Specification states

According to the style examples:

[`style`](#dataTypeFormat) | `explode` | `array` | `object`
----------- | ------ | -------- | -------
form | false | color=blue,black,brown | color=R,100,G,200,B,150
form | true | color=blue&color=black&color=brown | R=100&G=200&B=150
spaceDelimited | false | blue%20black%20brown | R%20100%20G%20200%20B%20150
pipeDelimited | false | blue\|black\|brown | R\|100\|G\|200\|B\|150

When I read the rows for `form` I observe that the parameter name is **explicitly used** within the examples: `color=blue,black,brown` for instance.

So naturally, when I do not see the name explicitly used in `spaceDelimited` or `pipeDelimited` examples my conclusion is that the names are not used.

## Problem with statement

Imagine the following example:

```yaml
  /pets:
    get:
      parameters:
        - name: type
          in: query
          style: pipeDelimited
          schema:
            type: array
            items: 
              type: string
        - name: personality
          in: query
          style: spaceDelimited
          schema:
            type: array
            items:
              type: string
```

A valid url according to the spec would be:
- `/pets?cat|shorthair|calico&confident%20greedy%20loud`
  - not too bad, we could identify each parameter based on the styles.
- `/pets?cat|shorthair|calico&loud`
  - Okay... no idea what style `loud` is... but if we can work out the first part then `loud` can be deduced by a process of elimination.
- `/pets?cat&loud`
  - hm.

Using the Swagger Editor with this spec it seems to think that the urls would come out as follows:
- `/pets?type=cat|shorthair|calico&personality=confident%20greedy%20loud`
- `/pets?type=cat|shorthair|calico&personality=loud`
- `/pets?type=cat&personality=loud`

All of which would immediately be identifiable based on the name. I realise the Swagger Editor is not the source of all knowledge, but the urls it generated matched what I would have expected.

Considering that `form`, `spaceDelimited` and `pipeDelimited`, are based off of 2.0's `collectionFormat` I was wondering if `spaceDelimited`, and `pipeDelimited` were meant to have the name explicitly used in the same way as `form`?

handrews:
@charjr My initial thought was that the `style` table only shows parameter values, but some of them clearly incorporate the name.  I belive that the Swagger editor is correct _and_ in compliance with the spec as written, but the table is somewhat misleading.

[...]

PR #3741, opened 2024-04-25
fix style=form examples (3.1.1)

karenetheridge:
These should have the parameter name in the resulting string (see issue #3737).


karenetheridge:
Note that we'll also want to fix the examples in https://learn.openapis.org/specification/parameters.html.

ralfhandl (approved):

handrews (approved):

mikekistler (approved):
Looks good! 👍

PR #3742, opened 2024-04-25
fix style=form examples (3.0.4)

karenetheridge:
These should have the parameter name in the resulting string (see issue #3737).

### DIFF
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -1080,8 +1080,8 @@ form | false | color= | color=blue | color=blue,black,brown | color=R,100,G,200,
 form | true | color= | color=blue | color=blue&color=black&color=brown | R=100&G=200&B=150
 simple | false | n/a | blue | blue,black,brown | R,100,G,200,B,150
 simple | true | n/a | blue | blue,black,brown | R=100,G=200,B=150
-spaceDelimited | false | n/a | n/a | blue%20black%20brown | R%20100%20G%20200%20B%20150
-pipeDelimited | false | n/a | n/a | blue\|black\|brown | R\|100\|G\|200\|B\|150
+spaceDelimited | false | n/a | n/a | color=blue%20black%20brown | color=R%20100%20G%20200%20B%20150
+pipeDelimited | false | n/a | n/a | color=blue\|black\|brown | color=R\|100\|G\|200\|B\|150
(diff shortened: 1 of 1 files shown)


karenetheridge:
This will need me to resolve conflicts resolved after merging #3744.

ralfhandl (approved):

handrews (dismissed):

versions/3.0.4.md
```diff
@@ -1080,8 +1080,8 @@ form | false | color= | color=blue | color=blue,black,brown | color=R,100,G,200,
 form | true | color= | color=blue | color=blue&color=black&color=brown | R=100&G=200&B=150
```
mikekistler, line 1080: Should this have the name too?
```suggestion
form | true | color= | color=blue | color=blue&color=black&color=brown | color=R=100&G=200&B=150
```

mikekistler: Also, the lines above, which I cannot comment on, seems to be missing parameter name in the "object" column.

[`style`](#dataTypeFormat) | `explode` | `empty` | `string` | `array` | `object`
----------- | ------ | -------- | -------- | -------- | -------
matrix | true | ;color | ;color=blue | ;color=blue;color=black;color=brown | ;color=R=100;G=200;B=150

mikekistler (commented):
Are there other fixes needed here?

versions/3.0.4.md
ralfhandl: I think objects are "exploded" by using their members as query parameter names, see [Parameter Serialization Control](https://learn.openapis.org/specification/parameters.html#parameter-serialization-control), so the current text is what I would expect.

ralfhandl:
@karenetheridge please resolve conflict

handrews:
@ralfhandl @mikekistler @miqui @karenetheridge I resolved the conflicts but that invalidated the approval - can someone re-approve and merge?

handrews (approved):

handrews:
Oh weird, it let me approve it?  We might want to look into the permissions here, but for now I will merge since there were already 2 TSC approvals and the 3.1.1 and 3.2.0 versions were merged.

ralfhandl:
@handrews you are not the PR author, so you can approve.
We technically require only one approval for merging, and you are a maintainer, so you can approve and then directly merge as long as no other maintainer has requested changes.

"Two TSC approvals" is currently only a convention, not technically enforced.

PR #3399, opened 2023-10-08
switch the order of these styles in the tables

karenetheridge:
This is a more natural grouping of similar types, making the data much easier to read.

(as in #3398)

PR #3744, opened 2024-04-25
switch the order of these styles in the tables (3.0.4)

karenetheridge:
This is a more natural grouping of similar types, making the data much easier to read.

(This is the same as #3398, #3399)


ralfhandl (approved):

lornajane (approved):
Makes sense to me!

PR #3748, opened 2024-04-25
Correctly escape example operationRef URLs (3.2.0 port of #3731)

### LINKED ISSUE #1677: URL-encoding for operationRef

### LINKED ISSUE #2738: JSON Reference mention in 3.1.0
Hi all,

There is a single occurrence of `JSON reference` words withing the 3.1.0 spec:

> Note that in the use of operationRef, the escaped forward-slash is necessary when using JSON references.

https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#operationref-examples

This might be confusing, as it's not clear what JSON references really means here. I assume it is refereing to https://datatracker.ietf.org/doc/html/draft-pbryan-zyp-json-ref-03, but I'm not really sure.

IMHO it would be more appropriate to have following sentence:

>  Note that in the use of operationRef, the escaped forward-slash is necessary when using JSON Pointer in fragment identifier part of the URI

...and have JSON Pointer link to [JSON Pointer specification](https://datatracker.ietf.org/doc/html/rfc6901).

handrews:
* Ports #3731 to 3.2.0
* Fixes #1677 
* Fixes #2738

### DIFF
--- a/versions/3.2.0.md
+++ b/versions/3.2.0.md
@@ -2124,7 +2124,7 @@ field in an [Operation Object](#operationObject)), references MAY also be made t
 links:
   UserRepositories:
     # returns array of '#/components/schemas/repository'
-    operationRef: '#/paths/~12.0~1repositories~1{username}/get'
+    operationRef: '#/paths/~12.0~1repositories~1%7Busername%7D/get'
     parameters:
       username: $response.body#/username
 ```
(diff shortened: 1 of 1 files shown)


lornajane (approved):
Thanks!

PR #3747, opened 2024-04-25
Correctly escape example operationRef URLs (3.1.1 port of #3731)

handrews:
* Fixes issue #1677 
* Fixes issue #2738

PR #3749, opened 2024-04-25
operationRef is a URI-reference

handrews:
Restore these "format" values to the schemas, which were removed when they were removed from the Server Object's "url" field, which is a template and not a URI-reference.

* Partially reverts #3455 (but leaves the Server Object changes in place)
* Companion to #3731, #3747, and #3748


ralfhandl (approved):

mikekistler (approved):
Looks good! 👍

PR #3752, opened 2024-04-25
Non-string discriminator values (3.2.0 port of #3728)

handrews:
The "undefined" wording was chosen to allow implementations that make a good-faith effort to accommodate non-string values to remain compliant.  However, new implementations are not expected to implement any sort of type coersion, and this guides API designers away from that expectation.

### DIFF
--- a/versions/3.2.0.md
+++ b/versions/3.2.0.md
@@ -2707,7 +2707,7 @@ Field Name | Type | Description
 
 This object MAY be extended with [Specification Extensions](#specificationExtensions).
 
-The discriminator object is legal only when using one of the composite keywords `oneOf`, `anyOf`, `allOf`.
+The discriminator object is legal only when using one of the composite keywords `oneOf`, `anyOf`, `allOf`.  Note that because the discrinating property's value is used as a component name and/or as the key in the `mapping` object, the behavior of any value that is not a string is undefined.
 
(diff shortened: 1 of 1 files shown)


versions/3.2.0.md
```diff
@@ -2707,7 +2707,7 @@ Field Name | Type | Description
 
 This object MAY be extended with [Specification Extensions](#specificationExtensions).
 
-The discriminator object is legal only when using one of the composite keywords `oneOf`, `anyOf`, `allOf`.
+The discriminator object is legal only when using one of the composite keywords `oneOf`, `anyOf`, `allOf`.  Note that because the discrinating property's value is used as a component name and/or as the key in the `mapping` object, the behavior of any value that is not a string is undefined.
```
ralfhandl, line 2710:
```suggestion
The discriminator object is legal only when using one of the composite keywords `oneOf`, `anyOf`, `allOf`.  Note that because the discriminating property's value is used as a component name and/or as the key in the `mapping` object, the behavior of any value that is not a string is undefined.
```

ralfhandl (approved):

ralfhandl:
Thanks for keeping the three work-in-progress branches in sync!

lornajane (approved):
Thanks @handrews :)

PR #3756, opened 2024-04-26
Clarify wording on encoding field (3.1.1 port of #3724)

### LINKED ISSUE #3421: Parameter schema specification via content

handrews:
* Ports #3724 from 3.0.4 to 3.1.1
* Fixes #3421

The previous wording could be read to only restrict what sort of request bodies could use an Encoding Object, while allowing that it could be used with other entities that use a Media Type Object.

This emphasizes that it is only relevant to Request Body Objects, and within that, only for specific media types.


ralfhandl (approved):

lornajane (approved):
Port looks good

Issue #2159, opened 2020-02-25
Base URI of external files? (more $id and $anchor fun with $ref resolution)


handrews:
I never got an answer to this question from #2092 (`$id` scope resolution):

>  If I put a non-Schema Object (like a Response Object) in its own file, what is the base URI for Reference Objects in that file? In particular, is `#/components/whatever` expected to work (because it is resolved relative to the primary OAS file) or fail (because Response Objects don't have a components property, and the file's base URI is distinct from the primary OAS document file)?

Regarding `$id` and particularly `$anchor`, this also impacts how something like `{"$ref": "#foo"}` is resolved.  In the main file you might have:

```YAML:
components:
  schemas:
    fooSchema:
      $anchor: foo
      type: object
      properties: {...}
    barSchema:
      type: object
      properties:
        someFoo: {$ref: "#foo"}
```

In this example, the `someFoo` property of the `barSchema` schema references the `fooSchema` schema using the fragment created by `$anchor`.  That fragment is attached to the base URI for the overall OAS file, so the same question about `#/components/schemas/fooSchema` from a separate file exists for `#foo` from a separate file.  

@OAI/tsc

darrelmiller:
What does document mean?  That is the question.

MikeRalphson:
For `v3.0` we got away without rewriting https://github.com/OAI/OpenAPI-Specification/blob/master/guidelines/v2.0/REUSE.md - I think for `v3.1` we will need to provide guidance, even if most of it is links back to the relevant parts of the JSON Schema spec.

handrews:
@MikeRalphson I agree, and I can work on that.  It will be a separate PR.  Should it also be a separate issue?  I'd say yes if there needs to be discussion on it.

handrews:
@darrelmiller was I supposed to wait on another PR closing for this?  I remember that coming up on the call but for some reason I didn't make a note of it.

darrelmiller:
@handrews In the TSC call today we (or some of us) believe that this was resolved in the changes that were made in the relative references text in 3.1.  To summarize, relative references are resolved either based on $Id or the location of the resource.
Also, Mike's PR related to option paths clarifies that an OpenAPI document is "composite thing" and not necessarily a physical file.  
We are going to merge the changes into rc0 and then re-review this issue holistically.

handrews:
[...]

PR #3758, opened 2024-04-26
Clarify how Schema Objects require full-document parsing (3.1.1)

### LINKED ISSUE #3545: Improve confusing "Relative References in URIs" / "Relative References in URLs" section titles and wording
These sections are quite hard to parse.  The URIs section applies to all forms of `$ref` and `operationRef`.  The URLs section applies to URLs in the Server (and by extension, Paths) Objects.  Some other places are unclear (see #1701, for example).

handrews:
This is the next step in un-blocking OASComply.  Unlike the previous OASComply-related PR, this PR ***does not*** apply to v3.0.4.  What, if any, wording about document parsing needs to go in 3.0.4 is something we can sort out later.

The implied requirement might be surprising to some, but is an unambiguous result of normatively citing JSON Schema 2020-12.  Resolution of references to `$id` values that do not match the document's location (e.g. are resolved as URIs rather than URLs, including by noticing them in already-parsed content) was affirmed multiple times by TSC decisions during the work on OAS 3.1:

* https://github.com/OAI/OpenAPI-Specification/pull/1977#issuecomment-571665656
* https://github.com/OAI/OpenAPI-Specification/issues/2092#issuecomment-585881781
* https://github.com/OAI/OpenAPI-Specification/issues/2159#issuecomment-601283402

Unfortunately, it seems like we never did get around to adding explicit wording highlighting this.  Which does not change the fact that a.) it was decided and re-affirmed, and b.) normatively citing JSON Schema 2020-12 mandates this behavior as the only possible way to satisfy its requirements.

There might be further changes related to this in the "Resolving Relative References in URIs", but I'll deal with that when I fix #3545, which covers both that section and "Resolving Relative References in URLs".

I also expect we'll want to add further guidance for both implementors and description authors on the Learn site.  I tried to keep the text here as minimal as possible while heading off the misunderstandings we've seen as folks implement 3.1.

------
_Commit message:_

JSON Schema draft 2020-12 includes numerous keywords that require parsing the entire document prior to deeming a reference unresolvable. This makes that more clear and outlines several approaches.

The practice of embedding OpenAPI fragments in other formats is deemed to have implementation-defined (non-interoperable) behavior, as the potential complications that might arise are not predictable.

### DIFF
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -154,7 +154,7 @@ Implementations MAY support complete-document parsing in any of the following wa
 * Detecting a document containing a referenceable Object at its root based on the expected type of the reference
 * Allowing users to configure the type of documents that might be loaded due to a reference to a non-root Object
 
-Embedding fragments of OpenAPI content in other formats can produce unexpected interactions among the enclosing format, the OAS, and JSON Schema.  While implementations MAY choose to support such arrangements, the resulting behavior is _implementation defined_.  This includes parsing fragments of OpenAPI out of a complete OpenAPI document while treating the rest of the document as arbitrary JSON or YAML.
+Embedding fragments of OpenAPI content in other formats can produce unexpected interactions among the enclosing format, the OAS, and JSON Schema.  While implementations MAY choose to support such arrangements, the resulting behavior is _implementation defined_.  This includes parsing fragments of OpenAPI out of a complete OpenAPI document while treating the rest of the document as arbitrary JSON or YAML.  Since parsing isolated fragments in this way will fail to notice keywords that might impact meaning or behavior, including in ways that contradict this specification's requirements, it is RECOMMENDED that OpenAPI Description authors avoid depending on such behavior.
 
 #### <a name="structuralInteroperability"></a>Structural Interoperability
 
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -154,7 +154,16 @@ Implementations MAY support complete-document parsing in any of the following wa
 * Detecting a document containing a referenceable Object at its root based on the expected type of the reference
 * Allowing users to configure the type of documents that might be loaded due to a reference to a non-root Object
 
-Embedding fragments of OpenAPI content in other formats can produce unexpected interactions among the enclosing format, the OAS, and JSON Schema.  While implementations MAY choose to support such arrangements, the resulting behavior is _implementation defined_.  This includes parsing fragments of OpenAPI out of a complete OpenAPI document while treating the rest of the document as arbitrary JSON or YAML.  Since parsing isolated fragments in this way will fail to notice keywords that might impact meaning or behavior, including in ways that contradict this specification's requirements, it is RECOMMENDED that OpenAPI Description authors avoid depending on such behavior.
+Implementations that parse referenced fragments of OpenAPI content without regard for the content of the rest of the containing document will miss keywords that change the meaning and behavior of the reference target.
+In particular, failing to take into account keywords that change the base URI introduces security risks by causing references to resolve to unintended URIs, with unpredictable results.
+While some implementations support this sort of parsing due to the requirements of past versions of this specification, in version 3.1, the result of parsing fragments in isolation is _undefined_ and likely to contradict the requirements of this specification.
+
+While it is possible to structure certain OpenAPI Descriptions to ensure that they will behave correctly when references are parsed as isolated fragments, depending on this is NOT RECOMMENDED.
+This specification does not explicitly enumerate the conditions under which such behavior is safe, and provides no guarantee for continued safety in any future versions of the OAS.
+
+A special case of parsing fragments of OAS content would be if such fragments are embedded in another format, referred to as an _embedding format_ with respect to the OAS.
+Note that the OAS itself is an embedding format with respect to JSON Schema, which is embedded as Schema Objects.
+It is the responsibility of an embedding format to define how to parse embedded content, and OAS implementations that do not document support for an embedding format cannot be expected to parse embedded OAS content correctly.
 
 #### <a name="structuralInteroperability"></a>Structural Interoperability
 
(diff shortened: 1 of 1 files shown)


miqui (approved):
Nice job ! @handrews

miqui:
This PR IMHO is very clear. I am very much tempted to merge this before the TDC call. @lornajane  , @ralfhandl thoughts?

handrews:
@miqui this one has some pretty deep impacts so I'd rather leave it open for longer.

ralfhandl (approved):

mikekistler (approved):
Looks good! 👍

miqui:
Discussed during TDC 5/2/2024 - It was decided we need 2 more TSC approvals.

versions/3.1.1.md
```diff
@@ -154,7 +154,7 @@ Implementations MAY support complete-document parsing in any of the following wa
...
 
-Embedding fragments of OpenAPI content in other formats can produce unexpected interactions among the enclosing format, the OAS, and JSON Schema.  While implementations MAY choose to support such arrangements, the resulting behavior is _implementation defined_.  This includes parsing fragments of OpenAPI out of a complete OpenAPI document while treating the rest of the document as arbitrary JSON or YAML.
+Embedding fragments of OpenAPI content in other formats can produce unexpected interactions among the enclosing format, the OAS, and JSON Schema.  While implementations MAY choose to support such arrangements, the resulting behavior is _implementation defined_.  This includes parsing fragments of OpenAPI out of a complete OpenAPI document while treating the rest of the document as arbitrary JSON or YAML.  Since parsing isolated fragments in this way will fail to notice keywords that might impact meaning or behavior, including in ways that contradict this specification's requirements, it is RECOMMENDED that OpenAPI Description authors avoid depending on such behavior.
```
ralfhandl, line 157: I have troubles parsing the added sentence, can we simplify it?

On a more technical note: can we put new sentences on new lines, one line per sentence? That makes reviewing changes easier, and it makes no difference in the rendered MD and HTML.

Simplification proposal:
```suggestion
Embedding fragments of OpenAPI content in other formats can produce unexpected interactions among the enclosing format, the OAS, and JSON Schema.  While implementations MAY choose to support such arrangements, the resulting behavior is _implementation defined_.  This includes parsing fragments of OpenAPI out of a complete OpenAPI document while treating the rest of the document as arbitrary JSON or YAML.
Since parsing isolated fragments in this way may fail to notice keywords that might impact meaning or behavior, it is RECOMMENDED that OpenAPI Description authors avoid depending on such behavior.
```

handrews:
@ralfhandl I've pushed a new commit with the following commit message:

-----

[Rework the guidance around fragmentary parsing.](https://github.com/OAI/OpenAPI-Specification/pull/3758/commits/ffe04346fac9921ab4e1147065b03d6a11b19c85)

This goes into more detail and uses "undefined" instead of
"implementation-defined" as the behavior is likely to be
incorrect (rather than just a different interpretationof an
ambiguous requirement), and may result in security concerns
as well.

-----

I don't know that it's a simplification, but I think it's more explicit and should be easier to follow.  Please do continue to suggest simplified wording - it's something I struggle with sometimes.

ralfhandl (approved):
Good explanation!

versions/3.1.1.md
```diff
@@ -140,6 +140,31 @@ An OpenAPI Description (OAD) MAY be made up of a single document or be divided i
...
+
+* Detecting OpenAPI or JSON Schema documents using media types
+* Detecting OpenAPI documents through the root `openapi` property
+* Detecting JSON Schema documents through detecting keywords or otherwise successfully parsing the document in accordance with the JSON Schema specification
+* Detecting a document containing a referenceable Object at its root based on the expected type of the reference
+* Allowing users to configure the type of documents that might be loaded due to a reference to a non-root Object
+
+Implementations that parse referenced fragments of OpenAPI content without regard for the content of the rest of the containing document will miss keywords that change the meaning and behavior of the reference target.
+In particular, failing to take into account keywords that change the base URI introduces security risks by causing references to resolve to unintended URIs, with unpredictable results.
+While some implementations support this sort of parsing due to the requirements of past versions of this specification, in version 3.1, the result of parsing fragments in isolation is _undefined_ and likely to contradict the requirements of this specification.
```
notEthan, line 159: I may be missing some distinction, but this seems like the same behavior that's described as "implementation defined" in https://github.com/OAI/OpenAPI-Specification/pull/3732/files#diff-b92507e7acda65ae00a02236c555cefc68b6fca4661077b84c2fb9ab150e5e17R151

This section particularly addresses schema objects but the other PR seems to encompass schema objects too.

handrews: @notEthan eh... (wobbles hand back and forth in a wishy-washy way).  There's an overlap, but it's not quite the same.  This is where we get into the "it's too hard to even explain" that I mentioned in the undefined/implementation-defined PR, but this ones not one of the worst so here's an explanation:

If you have a document that consists of an empty JSON object, `{}`, that document is syntactically valid as several different Objects.  Parsing that document from different reference contexts is still _full-document_ parsing, so it does not fall into this undefined behavior.  But it _is_ parsing using multiple conflicting contexts, so it falls into the implementation-defined behavior of the other PR.

It's a pretty fine hair to split, but it is deterministic.  The whole-document case is implementation-defined because at least one implementor I've spoken with did not think it required a fix in the spec.  I'm trying to invalidate as few existing implementations as possible, which is tricky to balance with the scenarios that result in outright wrong behavior.

handrews: Neither of these changes get into the deeper question of whether documents are parsed as JSON/YAML once or each time, or whether, once parsed as JSON/YAML, the resulting structures are further parsed as OAS Objects once or each time.  This starts getting into whether various parsing steps are cached, and what is done when a cached Object doesn't agree with a new parsing context.  There are many ways this might or might not be detected, and many strategies one could pick to handle it - new each time, first wins, last wins, any conflict is an error, etc.

notEthan: >  There's an overlap, but it's not quite the same.

I'm happy to accept that, and appreciate the explanation.

handrews: I figured it might help folks to see what "complicated to explain" looks like :-)

RomanHotsiy:
Looks good 👎

PR #3760, opened 2024-04-26
Clarify constraints on Parameter Object fields (3.0.4)

### LINKED ISSUE #1542: Requirement of schema or content in parameter object is difficult to notice
One of the paragraphs within the "Parameter Object" section of the current OpenAPI v3 [specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#parameterObject) contains this statement:

> A parameter MUST contain either a schema property, or a content property, but not both.

The `schema` or `content` fields in the tables of fields in the "Parameter Object" section do not contain the **REQUIRED** keyword. That is okay, because these fields are not strictly required. But at the same time, the entries in these tables do not mention a combined requirement of one of those fields.

The statement mentioned at the beginning is easy to overlook. I think that the conditional requirement of any fields should also be mentioned in the table, because it is the first and often the only place where a person would normally look for the field's definition.

handrews:
Fixes #1542 

I'm actually uncertain if `example` and `examples` (or, for that matter, `allowReserved`) are really only allowed with `schema`, but that _seems_ to be what's implied by the existing text?  I'm happy to rearrange this if it's not quite right.

As usual, I will forward-port if accepted.

-----

_Commit Message:_

This makes the relational requirements of `content` vs `schema` and associated fields more clear.  The mutual exclusivity requirement is moved up to the top of the Fixed Fields section, and each table table is given a subsection clarifying its meaning.  This makes it easier to follow the multi-table format, which is unusual within the OAS.

The commentary on `example` and `examples` is moved to the section on `schema` and related parameters as it appears to only apply in that case.

### DIFF
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -1016,6 +1016,14 @@ There are four possible parameter locations specified by the `in` field:
 
 
 ##### Fixed Fields
+
+The rules for serialization of the parameter are specified in one of two ways.
+Parameter Objects MUST include either a `content` field or a `schema` field, but not both.
+
+###### Common Fixed Fields
+
+These fields MAY be used with either `content` or `schema`.
+
 Field Name | Type | Description
 ---|:---:|---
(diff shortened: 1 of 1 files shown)


ralfhandl (approved):

mikekistler (approved):
Looks good. 👍 

I was not involved in the 3.0 spec development but I think these changes present the two alternatives for parameter description in a clearer way.

miqui (approved):

PR #3765, opened 2024-05-01
Backport case insensitivity note for security schemes to 3.0.4

lornajane:
We added a note about case insensitivity in v3.1.1 (see #2644 ), add it to 3.0.4 as well.


handrews:
Are we considering this a clarification because it was already implied by the spec but not clear?  Or would this require tools to change to support case-insensitivity?

lornajane:
Yes, I consider that it would be assumed to be case-insensitive, but that adding a note to say so was useful. I'm also happy to not backport this if it seems at all risky!

handrews:
@lornajane that works for me!

handrews (approved):

ralfhandl (approved):

versions/3.0.4.md
```diff
@@ -3208,7 +3208,7 @@ Field Name | Type | Applies To | Description
...
 <a name="securitySchemeName"></a>name | `string` | `apiKey` | **REQUIRED**. The name of the header, query or cookie parameter to be used.
 <a name="securitySchemeIn"></a>in | `string` | `apiKey` | **REQUIRED**. The location of the API key. Valid values are `"query"`, `"header"` or `"cookie"`.
-<a name="securitySchemeScheme"></a>scheme | `string` | `http` | **REQUIRED**. The name of the HTTP Authorization scheme to be used in the [Authorization header as defined in RFC7235](https://tools.ietf.org/html/rfc7235#section-5.1).  The values used SHOULD be registered in the [IANA Authentication Scheme registry](https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml).
+<a name="securitySchemeScheme"></a>scheme | `string` | `http` | **REQUIRED**. The name of the HTTP Authorization scheme to be used in the [Authorization header as defined in RFC7235](https://tools.ietf.org/html/rfc7235#section-5.1).  The values used SHOULD be registered in the [IANA Authentication Scheme registry](https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml).  The value is case-insensitive.
```
mikekistler, line 3211: Is this name the "token" that is referred to in [Section 2.1 of RFC 7235](https://datatracker.ietf.org/doc/html/rfc7235#section-2.1)?
> It uses a case-insensitive token as a means to identify the authentication scheme

If so, perhaps a link to that section would make it clear that this new text is simply a clarification.

ralfhandl, line 3211:
```suggestion
<a name="securitySchemeScheme"></a>scheme | `string` | `http` | **REQUIRED**. The name of the HTTP Authorization scheme to be used in the [Authorization header as defined in RFC7235](https://tools.ietf.org/html/rfc7235#section-5.1).  The values used SHOULD be registered in the [IANA Authentication Scheme registry](https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml).  The value is case-insensitive, as defined in [RFC7235](https://datatracker.ietf.org/doc/html/rfc7235#section-2.1).
```

lornajane:
I'm not sure what to do here, if we make additional changes, rather than just backporting - do I then need to followup forward-port the additions to the 3.1.1 spec where the original note about case sensitivity was added?

Personally, I think this can merge without the additions to bring the two specs into sync - but I'd need another approval for that. Opinions and reviews welcome!

PR #3769, opened 2024-05-01
Clarify how to model streaming binary data (3.1.1 port of #3729)

### LINKED ISSUE #1576: Support for Streams in Responses

handrews:
* Ports #3729, plus an additional subsection heading as the section was getting complex
* Fixes #1576 

Also make the translation from 3.0 binary modeling a subsection, as otherwise the line about streaming either feels randomly inserted into the middle of the section, or feels disconnected when placed after the conversion table.

### DIFF
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -196,6 +196,9 @@ The `contentMediaType` keyword is redundant if the media type is already set:
 
 If the Schema Object will be processed by a non-OAS-aware JSON Schema implementation, it may be useful to include `contentMediaType` even if it is redundant.  However, if `contentMediaType` contradicts a relevant Media Type Object or Encoding Object, then `contentMediaType` SHALL be ignored.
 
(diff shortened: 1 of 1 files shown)


ralfhandl (approved):

miqui (approved):

mikekistler (approved):
Looks good! 👍

PR #3770, opened 2024-05-01
Clarify how to model binary data (3.2.0 port of #3727 and #3729

### LINKED ISSUE #1544: Clarify: "type": "string", "format": "binary" in non-entity-body

### LINKED ISSUE #1576: Support for Streams in Responses

### LINKED ISSUE #2476: Question: `Content-Encoding` HTTP header field and `contentEncoding`

### LINKED ISSUE #2868: conflicting order of operations when applying Content-Encoding and Content-Type headers

handrews:
* Ports two PRs as separate commits, because one is tiny and would conflict with the much larger one
    * #3727
    * #3729
* Fixes #1544
* Fixes #1576 
* Fixes #2476 
* Fixes #2868 

-----
***First commit message***

This reorganizes binary data-related guidance into a
"Working With Binary Data" section, as has already been
done in 3.0.4.

This includes more detailed guidance on when various
approaches to binary data make sense (e.g. you cannot
stuff raw binary into JSON no matter what you put in your
Schema Object, and while you can base64-encode entire
message bodies, it takes up a lot more space for no clear
benefit).  An example that contradicts this guidance has
been removed.

The lack of and HTTP header for base64 encoding is now
explicitly called out.

Also note that only `multipart` media types with named parts
are supported, as they are modeled as an object.

-----
***Second commit message***

Clarify how to model streaming binary data

### DIFF
--- a/versions/3.2.0.md
+++ b/versions/3.2.0.md
@@ -160,6 +160,40 @@ The formats defined by the OAS are:
 `number` | `double` | |
 `string` | `password` | A hint to UIs to obscure input.
 
+#### <a name="binaryData"></a>Working With Binary Data
+
+The OAS can describe either _raw_ or _encoded_ binary data.
+
+* **raw binary** is used where unencoded binary data is allowed, such as when sending a binary payload as the entire HTTP message body, or as part of a `multipart/*` payload that allows binary parts
+* **encoded binary** is used where binary data is embedded in a text-only format such as `application/json` or `application/x-www-form-urlencoded` (either as a message body or in the URL query string).
+
+In the following table showing how to use Schema Object keywords for binary data, we use `image/png` as an example binary media type.  Any binary media type, including `application/octet-stream`, is sufficient to indicate binary content.
+
+Keyword | Raw | Encoded | Comments
+------- | --- | ------- | --------
+`type`  | _omit_ | `string` | raw binary is [outside of `type`](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#section-4.2.3)
+`contentMediaType` | `image/png` | `image/png` | can sometimes be omitted if redundant (see below)
+`contentEncoding`  | _omit_ | `base64`&nbsp;or&nbsp;`base64url` | other encodings are [allowed](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#section-8.3)
+
+Note that the encoding indicated by `contentEncoding`, which inflates the size of data in order to represent it as 7-bit ASCII text, is unrelated to HTTP's `Content-Encoding` header, which indicates whether and how a message body has been compressed and is applied after all content serialization described in this section has occurred.  Since HTTP allows unencoded binary message bodies, there is no standardized HTTP header for indicating base64 or similar encoding of an entire message body.
+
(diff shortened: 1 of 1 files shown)


ralfhandl (approved):

miqui (approved):

mikekistler (approved):
Looks good! 👍

PR #3779, opened 2024-05-03
Define "undefined" and "implementation-defined" (3.0.4)

### LINKED ISSUE #3775: Codify the meaning of "undefined" and "implementation-defined"
We already use these terms a bit in the spec, but with PRs like these:

* #3732 
* #3772 
* #3758 

we are relying more on such descriptions to define areas of interoperability and expected compliance.  A new subsection under the _Definitions_ section seems like the best approach.

handrews:
* Fixes #3775

Undefined means "don't use this even if it seems to work, as there are cases where it won't."  Implementation-defined means "this isn't portable or interoperable, but it is still correct."

### DIFF
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -100,7 +100,7 @@ Status codes SHOULD be selected from the available status codes registered in th
 
 This specification deems certain situations to have either _undefined_ or _implementation-defined_ behavior.
 
-Behavior described as _undefined_ is likely, at least in some circumstances, to contradict the specification.
+Behavior described as _undefined_ is likely, at least in some circumstances, to result in outcomes that contradict the specification.
 This description is used when detecting the contradiction is impossible or impractical.
 Implementations MAY support undefined scenarios for historical reasons, including ambiguous text in prior versions of the specification.
 This support might produce correct outcomes in many cases, but relying on it is NOT RECOMMENDED as there is no guarantee that it will work across all tools or with future specification versions, even if those versions are otherwise strictly compatible with this one.
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -96,6 +96,19 @@ Some examples of possible media type definitions:
 The HTTP Status Codes are used to indicate the status of the executed operation. 
 Status codes SHOULD be selected from the available status codes registered in the [IANA Status Code Registry](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml).
 
+##### <a name="undefinedAndImplementationDefinedBehavior"></a>Undefined and Implementation-Defined Behavior
+
+This specification deems certain situations to have either _undefined_ or _implementation-defined_ behavior.
+
+Behavior described as _undefined_ is likely, at least in some circumstances, to contradict the specification.
+This description is used when detecting the contradiction is impossible or impractical.
+Implementations MAY support undefined scenarios for historical reasons, including ambiguous text in prior versions of the specification.
+This support might produce correct outcomes in many cases, but relying on it is NOT RECOMMENDED as there is no guarantee that it will work across all tools or with future specification versions, even if those versions are otherwise strictly compatible with this one.
+
+Behavior described as _implementation-defined_ allows implementations to choose which of several different-but-compliant approaches to a requirement to implement.
+This documents ambiguous requirements that API description authors are RECOMMENDED to avoid in order to maximize interoperability.
+Unlike undefined behavior, it is safe to rely on implementation-defined behavior if _and only if_ it can be guaranteed that all relevant tools support the same behavior.
+
 ## Specification
 
 ### Versions
@@ -147,7 +160,7 @@ Models are defined using the [Schema Object](#schemaObject), which is an extende
 <a name="dataTypeFormat"></a>Primitives have an optional modifier property: `format`.
 OAS uses several known formats to define in fine detail the data type being used.
 However, to support documentation needs, the `format` property is an open `string`-valued property, and can have any value.
-Formats such as `"email"`, `"uuid"`, and so on, MAY be used even though undefined by this specification.
+Formats such as `"email"`, `"uuid"`, and so on, MAY be used even though they are not defined by this specification.
 Types that are not accompanied by a `format` property follow the type definition in the JSON Schema. Tools that do not recognize a specific `format` MAY default back to the `type` alone, as if the `format` is not specified.
 
 The formats defined by the OAS are:


versions/3.0.4.md
```diff
@@ -96,6 +96,19 @@ Some examples of possible media type definitions:
...
+This specification deems certain situations to have either _undefined_ or _implementation-defined_ behavior.
+
+Behavior described as _undefined_ is likely, at least in some circumstances, to contradict the specification.
+This description is used when detecting the contradiction is impossible or impractical.
+Implementations MAY support undefined scenarios for historical reasons, including ambiguous text in prior versions of the specification.
+This support might produce correct outcomes in many cases, but relying on it is NOT RECOMMENDED as there is no guarantee that it will work across all tools or with future specification versions, even if those versions are otherwise strictly compatible with this one.
+
+Behavior described as _implementation-defined_ allows implementations to choose which of several different-but-compliant approaches to a requirement to implement.
+This documents ambiguous requirements that API description authors are RECOMMENDED to avoid in order to maximize interoperability.
+Unlike undefined behavior, it is safe to rely on implementation-defined behavior if _and only if_ it can be guaranteed that all relevant tools support the same behavior.
```
ralfhandl, line 110: Well, that guarantee may be hard to come by 😁

ralfhandl (dismissed):

versions/3.0.4.md
```diff
@@ -96,6 +96,19 @@ Some examples of possible media type definitions:
 The HTTP Status Codes are used to indicate the status of the executed operation. 
 Status codes SHOULD be selected from the available status codes registered in the [IANA Status Code Registry](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml).
 
+##### <a name="undefinedAndImplementationDefinedBehavior"></a>Undefined and Implementation-Defined Behavior
+
+This specification deems certain situations to have either _undefined_ or _implementation-defined_ behavior.
+
+Behavior described as _undefined_ is likely, at least in some circumstances, to contradict the specification.
+This description is used when detecting the contradiction is impossible or impractical.
+Implementations MAY support undefined scenarios for historical reasons, including ambiguous text in prior versions of the specification.
```
notEthan, line 105: This is confusing to me, both my general understanding of how the term 'undefined behavior' is used and as defined here. In particular the description "likely to contradict the specification" - my understanding has been that undefined behavior, either explicitly stated as undefined or by a lack of explicit specification, has no expectation for the outcome of that behavior. That would mean there's nothing specified for behavior to contradict.

And the statement "Implementations MAY support undefined scenarios" seems at odds with that behavior being likely to contradict specification. Combining the two, to my interpretation, results in "Implementations MAY have behavior that probably contradicts the specification" or so.

notEthan: Having since been reading related issues that led to this (on structural interoperability), this is making more sense - it seems not the undefined behavior itself (treating references' resolved objects as one type when their structural placement indicates another type) that will contradict the spec, but the natural result of doing that in some circumstances (not detecting schema `$id`s, not having the correct base URI).

Wording that distinguishes the undefined-but-allowed behavior from the possible specified-disallowed result might be clearer, though I'm not sure quite how that would be worded.

handrews: Thanks, @notEthan , I'll try to think about better wording.

> Wording that distinguishes the undefined-but-allowed behavior from the possible specified-disallowed result might be clearer, though I'm not sure quite how that would be worded.

That's specifically what I'm trying to avoid doing.  Or, rather, I have laid out the specifics before and had people refuse to read it for being too long or complicated.  One person complained that the topics are so obscure that "only four or five people" could probably even understand it.

It's much easier to say "just don't do anything like this" than try to explain all of the different things that have to happen for it to not work.

Allowing for implementations that do it anyway is just a concession to reality:  Many tools do not implement 3.1 Schema Object reference parsing correctly, and are likely to ignore a directive to rip out their incorrect support.  And given that it was not obvious to most how the specifications require different behavior in 3.1, I hesitate to penalize them.

If they have something that works for them and their customers, the most likely outcome is that they keep doing it.  But changing the wording in the spec makes it clear that if there is a problem, it's not a spec problem (aside from past lack of clarity).  It's outside the spec, and we know it won't work sometimes.

It's a difficult situation to articulate, really.

notEthan: Maybe just "Behavior described as _undefined_ is likely, at least in some circumstances, to result in outcomes that contradict the specification." or so. Undefined behavior itself isn't contradicting the spec but its result likely will - if that isn't too much indirection for one sentence.

I don't know, the wording that goes into spec authorship is hard (much respect for all you have put into various specs my work relies on).

handrews:
@notEthan @kevinswiber @ralfhandl in light of @notEthan 's comments, I'm wondering if "undefined" is really the right word to use.  I was using it thinking of the following from [RFC 7231 §4.3.1 GET](https://www.rfc-editor.org/rfc/rfc7231#section-4.3.1)

>    A payload within a GET request message has no defined semantics;
   sending a payload body on a GET request might cause some existing
   implementations to reject the request.


Here is the analogous text from [RFC 9110 §9.3.1 GET](https://www.rfc-editor.org/rfc/rfc9110.html#section-9.3.1):

> Although request message framing is independent of the method used, content received in a GET request has no generally defined semantics, cannot alter the meaning or target of the request, and might lead some implementations to reject the request and close the connection because of its potential as a request smuggling attack ([Section 11.2](https://www.rfc-editor.org/rfc/rfc9112#section-11.2) of [[HTTP/1.1](https://www.rfc-editor.org/rfc/rfc9110.html#HTTP11)]). A client SHOULD NOT generate content in a GET request unless it is made directly to an origin server that has previously indicated, in or out of band, that such a request has a purpose and will be adequately supported. An origin server SHOULD NOT rely on private agreements to receive content, since participants in HTTP communication are often unaware of intermediaries along the request chain.

What both of these have in common with what I'm trying to do is that they don't require an error, even though they note that an error might occur.  Basically, they boil down to "if you want to do this and it seems to work, we're not stopping you, but it quite likely won't work."  9110 gets a bit more detailed in its recommendation, including allowing that a server might indicate that it accepts a GET body.  But it also notes that intermediaries might not be aware of such agreements, indicating that you  might not be able to get your GET body through to the server even if it is willing to accept it.

That seems to fit with my general desire to say "you can do this if you really want to, and maybe a tool documents that it works in some cases, but there might be reasons it still won't work because you don't control everything here."

Notably, RFC 2616 did not include anything about GET request bodies/content, so I assume that it was added after people did unexpected things with it, and it was necessary to discourage without invalidating.  Which is very similar to some situations in OAS.

versions/3.0.4.md
handrews: Indeed... it mirrors a few other spec passages, like the one in the JSON RFC that says you can use something other than UTF-8 only in a "closed system."

But it's not that uncommon to have internal APIs where your customers are mandated to use a particular client or technology stack.  It's not something I'd go out of my way to accommodate, but there _are_ scenarios where you can be pretty certain that you won't hit a different interpretation of an implementation-defined requirement.

If nothing else, some people don't do anything with OpenAPI but generate docs.

versions/3.0.4.md
```diff
@@ -96,6 +96,19 @@ Some examples of possible media type definitions:
 The HTTP Status Codes are used to indicate the status of the executed operation. 
 Status codes SHOULD be selected from the available status codes registered in the [IANA Status Code Registry](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml).
 
+##### <a name="undefinedAndImplementationDefinedBehavior"></a>Undefined and Implementation-Defined Behavior
+
+This specification deems certain situations to have either _undefined_ or _implementation-defined_ behavior.
+
+Behavior described as _undefined_ is likely, at least in some circumstances, to contradict the specification.
```
ralfhandl, line 103: Failing to come up with a better phrasing I suggest to use @notEthan's proposal

```suggestion
Behavior described as _undefined_ is likely, at least in some circumstances, to result in outcomes that contradict the specification.
```

ralfhandl (approved):

lornajane (approved):
I think this is pretty clear. Thanks everyone for the input and the revisions!

PR #3782, opened 2024-05-06
Use pinned version of MAINTAINERS.md per spec version

### LINKED ISSUE #3693: ReSpec HTML generation needs to use correct revision of MAINTAINERS.md
As seen in PR #3688, `scripts/md2html` fails to use the correct revision of the MAINTAINERS.md file for the 3.0 and 3.1 specifications.  Specifically, `build.sh` [uses a specific revision for the 2.0 spec](https://github.com/OAI/OpenAPI-Specification/blob/main/scripts/md2html/build.sh#L13), but [not for the later specs](https://github.com/OAI/OpenAPI-Specification/blob/main/scripts/md2html/build.sh#L23).  Also, the [collection of emeritus members](https://github.com/OAI/OpenAPI-Specification/blob/main/scripts/md2html/md2html.js#L124) in `md2html.js` seems to no longer work with the current file format.

ralfhandl:
Fixes #3693

Quick fix for respec build script:
- use pinned version of `MAINTAINERS.md` for all 3.x specs

TODO:
- [x] use nearest version of `MAINTAINERS.md` for each 3.x spec
- [x] create static content for missing version of `MAINTAINERS.md` for 2.0 spec

### DIFF
--- a/scripts/md2html/build.sh
+++ b/scripts/md2html/build.sh
@@ -10,7 +10,8 @@ mkdir -p deploy/js
 
 cd scripts/md2html
 mkdir -p history
-git show c740e950d:MAINTAINERS.md > history/MAINTAINERS_v2.0.md
+git show c740e95:MAINTAINERS.md > history/MAINTAINERS_v2.0.md
+git show e9fe5bc:MAINTAINERS.md > history/MAINTAINERS_v3.x.md
 cp -p js/* ../../deploy/js 2> /dev/null
 cp -p markdown/* ../../deploy/ 2> /dev/null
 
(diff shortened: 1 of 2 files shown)


scripts/md2html/build.sh
```diff
@@ -10,17 +10,28 @@ mkdir -p deploy/js
 
 cd scripts/md2html
 mkdir -p history
-git show c740e950d:MAINTAINERS.md > history/MAINTAINERS_v2.0.md
+cat > history/MAINTAINERS_v2.0.md <<EOF
+## Active
+* Jeremy Whitlock [@whitlockjc](https://github.com/whitlockjc)
+* Marsh Gardiner [@earth2marsh](https://github.com/earth2marsh)
```
ralfhandl, line 16: @whitlockjc @earth2marsh please confirm, my OpenAPI history knowledge is spotty.

earth2marsh: @webron remains an active maintainer, too! otherwise lgtm (assuming I understood the question)

ralfhandl: Yes, @webron is an active maintainer in all versions of `MAINTAINERS.md`.

The question is whether I listed all active editors of version 2.0, and whether there were any changes between publication of 2.0 and 3.0.0.

miqui (approved):

lornajane (approved):
Thanks for diving into this @ralfhandl (I think the historical data is correct)

Issue #3740, opened 2024-04-24
Add appendix "Informative references" to published spec documents


ralfhandl:
Check existing non-RFC links whether they are normative or informative, for example links to the YAML specification, or links to the JSON Schema specifications.

Add appendix "Informative references" for selected external links that we deem non-normative and worth mentioning.

See [IESG Statement: Normative and Informative References
](https://datatracker.ietf.org/doc/statement-iesg-iesg-statement-normative-and-informative-references-20060419/)

Candidates (default: normative)
- ABNF
- CommonMark
- HTML 4.01
- IANA registries
- JSON Schema
- JSON Schema Validation
- OpenAPI learn - informative
- OpenAPI registries - normative as it is a defined extension point
- OpenID Connect Core
- OpenID Connect Discovery
- SPDX
- WhatWG URL
- XML Namespaces
- YAML

Preview: https://ralfhandl.github.io/OpenAPI-Specification/oas/latest.html#informative-references

handrews:
For those wondering why this is in "Automation & Infrastructure", the "Normative References" appendix seems to be added by the HTML generation script- it is not present in our (allegedly normative) Markdown document.  Which strikes me as odd- are not normative references a normative part of the spec?

ralfhandl:
Just noted this inconsistency:

![image](https://github.com/OAI/OpenAPI-Specification/assets/951576/2cc70cef-de55-495e-b216-6ab15115f9bf)

The Markdown spec source already has an "Appendix A" with the version history, which is rendered in HTML as chapter 5, and the normative references are added as second Appendix A.

handrews:
ReSpec has syntax for informative references.  We don't consistently use its syntax for references, and we possibly should.  That might be more predictable than relying on inferences and heuristics.

lornajane:
Discussion concludes that the OpenAPI Learn is the only informative link from the collection

karenetheridge:
I think we might to also add the RFC describing URIs (3986?) because although `"format": "uri"` is managed by JSON Schema, we do also have URI references (via `$ref`) in OpenAPI itself. Although, this could be considered covered by "an openapi schema must be valid according to the JSON Schema specification here".

karenetheridge:
Yes, sorry, I called it out because I didn't see it in the candidates list at the top of this issue.

PR #3708, opened 2024-04-15
Spec links to https://path.id - v3.2.0

Bellangelo:
Closes https://github.com/OAI/OpenAPI-Specification/issues/3679

v3.1.1 PR: https://github.com/OAI/OpenAPI-Specification/pull/3707
v3.0.4 PR: https://github.com/OAI/OpenAPI-Specification/pull/3706

Issue #3785, opened 2024-05-06
Style guide for the specification


handrews:
As I write PRs, I keep finding little things that I'm not sure how to do.  Some examples:

* What are the rules about which headings go in the TOC and which don't?
* It seems clear that RFCs are to be written without a space between the "RFC" and the number; are there other conventions to follow?
* RFC referencdes often go to a specific section without noting that section in the text, is that intentional or should we be more explicit in the readable text?  I think some _do_ mention section numbers or titles, but I have no idea what's ideal
* Section title capitalization - it seems like we go full title-case, but it would be a good idea to document it
* We should document that we use **REQUIRED** but not **OPTIONAL** as came up in a recent PR.  Unless we want to follow issue #2470 and start using **OPTIONAL** - we should either do that or close #2470

There are other things with less impact (e.g. for bulleted lists do we treat bullet points as sentences or not?), but these are the ones that I hit the most.

lornajane:
* If I had to guess, the TOC hasn't been well-maintained and we should probably regenerate that.
* I don't really want to go back and update the RFC references but good practice is probably to cite the section as well in the link test.
* Title case is fine and we need to improve our markdown automation anyway. Let's document it for now (I know how you feel about our not-contributing file, but maybe let's add a section for style and then we can clean up around it)
* I do not feel it would improve anything for our users if we start explicitly putting OPTIONAL everywhere so I'd like the "just document that we don't do it that way" option, please!

handrews:
Thanks @lornajane – yeah I don't think we need to go update every RFC citation, but in some areas where I'm already doing PRs that are thick with them, knowing how to handle references to different sections, some of which happen multiple times, would be really helpful.

I'm totally fine with documenting things in the not-contributing file, I've just realized that I'm over capacity and can't take on rewriting that right now even when I'm the person pushing for changes / documented policies :-/  At some point, if no one else steps up, I'll get back to it, but that's multiple months out right now.

handrews:
Some pre-release checklist items

- [ ] re-generate TOC
    - [ ] consistent section titling
    - [ ] consistent linking
    - [ ] update links for GitHub changes done on main
- [ ] #3576 (authoritative URL - might need to change specification texgt)
- [ ] make RFC links consistent
    - [ ] rfc-editor.org vs datatracker.ietf.org
    - [ ] different anchor styles (auto-id vs section numbers vs ???)
    - [ ] link to sections (always, sometimes, never?) (#3863)
[...]

PR #3786, opened 2024-05-06
Add appendix "Informative references" to published spec documents

Bellangelo:
Closes https://github.com/OAI/OpenAPI-Specification/issues/3740

This PR parses every link that is not local and is not an formative RFC using the format that the Respec expects. See https://respec.org/docs/#:~:text=ReSpec%20uses%20the%20context%20of,reference%20is%20treated%20as%20normative.

It achieves it by using the [localBiblio](https://respec.org/docs/#localBiblio) feature of the Respec. Although, its usage is discouraged I couldn't find any other native way of adding a Informative list. Also, it is only discouraged to force you to apply your references to their free database.

You can check its result here: https://codebeautify.org/htmlviewer/y247136c8

## A few notes
1. `path.id` is appearing as a link is fixed in another [PR](https://github.com/OAI/OpenAPI-Specification/pull/3708)
2. The first entry in the list ( `1.2` ) is basically the referenced version in the YAML spec. We could change the url's text to contain more data so it would be more user friendly. For example, `YAML-Version-12`.
3. You might have noticed that all references have their spaces replaced with dashes. This is because the Respec does not allow spaces in the references.
4. After this PR is merged we might need to change how we handle the formative links since we will have 2 places where we handle links.

### DIFF
--- a/scripts/md2html/md2html.js
+++ b/scripts/md2html/md2html.js
@@ -47,7 +47,6 @@ const md = require('markdown-it')({
     }
 });
 
-function preface(title,options) {
 const localBiblio = {};
 const baseURL = 'https://github.com/OAI/OpenAPI-Specification/';
 
@@ -81,6 +80,7 @@ md.renderer.rules.link_open = function(tokens, idx, options, env, self) {
     return self.renderToken(tokens, idx, options);
 };
 
+function preface(title, options, localBublio) {
     const respec = {
         specStatus: "base",
         editors: maintainers,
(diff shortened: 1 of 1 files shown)


handrews:
@Bellangelo thanks, this is a really great step forward!  It also highlights some of the concerns I raised in #3785 just an hour or two before you submitted this, which is that we're really inconsistent with linking citations.  Which explains why the rendered result thinks "GET" and "HEAD" are informative references (they're really just subsection links to normative reference RFC 7231).  There are also some non-RFC references, most notably the JSON Schema drafts, which are normative.

I'm not quite sure what the next steps are here, but this really helps us figure out how to handle citations – we need to balance clarity in the text with what tools like this can extract.  I suspect we'll need to make changes in both the spec and this PR to get something correct in the end.  Let's see what others have to say about this PR and issue #3785 so far before trying to sort that out.

ralfhandl:
@Bellangelo Sorry, the issue description was too fluffy, and your PR literally fixed it by adding every external link to the "Informative References", including links to the editors' Github user pages, or "deep links" to RFC sections for HTTP verbs, ...

Resharpened the issue description to require conscious decision by the editors which links to mention in which appendix, and which continue to not mention there.

PR #3788, opened 2024-05-07
Fix 3.1.1 .gitignore

ralfhandl:
Sync .gitignore with main

- [x] contains `package-lock.json` in this branch: is that a good idea?
  - no

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,5 +7,4 @@ atlassian-ide-plugin.xml
 node_modules/
 deploy/
 history
-package-lock.json
 Gemfile.lock
--- a/.gitignore
+++ b/.gitignore
@@ -5,5 +5,7 @@
 target
 atlassian-ide-plugin.xml
 node_modules/
+deploy/
+history
 package-lock.json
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -5,5 +5,7 @@
 target
 atlassian-ide-plugin.xml
 node_modules/
+deploy/
+history
 package-lock.json
```
ralfhandl, line 10: Remove this line?

handrews: Hmm... I'd say yes, because we've checked it in on main.  But we have not checked it in on the branches.  I'm not sure if that's because it's not needed on the branches or because we just didn't think to do it, though.

But we did decided to check in `package-lock.json` because that is currently deemed the best practice, so either this line is irrlevant (b/c it doesn't need to be on the branch) or wrong (because it does, and we would check it in).

ralfhandl: Removed it to avoid accidentally merging this into main

handrews (approved):

PR #3790, opened 2024-05-08
[new] scripts/update-schema-id

TakahikoKawasaki:
This script updates the schema IDs in the following YAML files:

- schemas/v{major}.{minor}/schema.yaml
- schemas/v{major}.{minor}/schema-base.yaml

and generates the following JSON files from the updated YAML files:

- schemas/v{major}.{minor}/schema.json
- schemas/v{major}.{minor}/schema-base.json

### DIFF
--- a/scripts/update-schema-id
+++ b/scripts/update-schema-id
@@ -26,7 +26,7 @@
 # EXAMPLE
 # -------
 #
-#   ./update-schema-id -v 3.1
+#   ./scripts/update-schema-id -v 3.1
 #
 #   If the above command line is executed on May 7, 2024, the following
 #   external commands are invoked by the script.
--- a/scripts/update-schema-id
+++ b/scripts/update-schema-id
@@ -1,0 +1,245 @@
+#!/usr/bin/env ruby
+#
+# OVERVIEW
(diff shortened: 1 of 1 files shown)


scripts/update-schema-id
```diff
@@ -0,0 +1,245 @@
...
+      @quiet = true
+    end
+  end
+
+  private
+
+  def error_exit(message)
+    $stderr.puts message
+    exit false
+  end
```
notEthan, line 135: built-in `abort` (method of Kernel) does the same as this `error_exit` method

TakahikoKawasaki: @notEthan Thank you! I have replaced `error_exit` with `abort`.

lornajane:
Thanks for getting us started with this change. We don't currently have any Ruby code in our repository, and it's not a skill present in our current maintainer cohort either. Could we discuss what this script does, and whether we can use a tech stack that is already in use?

TakahikoKawasaki:
Okay. A shell script version is here --> PR #3814

handrews:
Replaced by #3814

PR #3795, opened 2024-05-09
Explain dynamic refs for generics (3.2.0 port of #3714)

### LINKED ISSUE #3601: Recommend describing templates/generic types using $dynamicRef

handrews:
* Ports #3714 
* Fixes #3601 

Includes an example.  This intentionally does not explain how dynamic referencing works, as there are better resources available in both the spec and (more readably) the official JSON Schema blog.

### DIFF
--- a/versions/3.2.0.md
+++ b/versions/3.2.0.md
@@ -2398,6 +2398,15 @@ There are two ways to define the value of a discriminator for an inheriting inst
 - Override the schema name by overriding the property with a new value. If a new value exists, this takes precedence over the schema name.
 As such, inline schema definitions, which do not have a given id, *cannot* be used in polymorphism.
 
+###### Generic (Template) Data Structures
+
+Implementations MAY support defining generic or template data structures using JSON Schema's dynamic referencing feature:
+
+* `$dynamicAnchor` identifies a set of possible schemas (including a default placeholder schema) to which a `$dynamicRef` can resolve
+* `$dynamicRef` resolves to the first matching `$dynamicAnchor` encountered on its path from the schema entry point to the reference, as described in the JSON Schema specification
+
+An example is included in the "Schema Object Examples" section below, and further information can be found on the Learn OpenAPI site's ["Dynamic References"](https://learn.openapis.org/referencing/dynamic.html) page.
+
 ###### XML Modeling
 
 The [xml](#schemaXml) property allows extra definitions when translating the JSON definition to XML.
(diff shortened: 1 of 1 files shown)


lornajane (dismissed):
An example always helps :) Thanks!

versions/3.2.0.md
```diff
@@ -2741,6 +2750,111 @@ components:
...
+      }
+    }
+  }
+}
+```
+
+```YAML
+components:
+  schemas:
+    genericArray:
```
ralfhandl, line 2820:
```suggestion
    genericArrayComponent:
```

versions/3.2.0.md
```diff
@@ -2741,6 +2750,111 @@ components:
...
+      $id: array_of_strings
+      $ref: fully_generic_array
+      $defs:
+        stringsOnly:
+          $dynamicAnchor: generic-array
+          type: string
+    objWithTypedArray:
+      $id: obj_with_typed_array
+      type: object
+      required: [dataType, data]
```
ralfhandl, line 2845:
```suggestion
      required: 
      - dataType 
      - data
```

versions/3.2.0.md
```diff
@@ -2741,6 +2750,111 @@ components:
...
+      properties:
+        dataType:
+          enum: [string, number]
```
ralfhandl, line 2848:
```suggestion
          enum:
          - string
          - number
```

versions/3.2.0.md
```diff
@@ -2741,6 +2750,111 @@ components:
...
+      oneOf:
+      - properties:
+          dataType: {const: string}
+          data: {$ref: array_of_strings}
```
ralfhandl, line 2852:
```suggestion
          dataType:
            const: string
          data:
            $ref: array_of_strings
```

versions/3.2.0.md
```diff
@@ -2741,6 +2750,111 @@ components:
...
+      - properties:
+          dataType: {const: number}
+          data: {$ref: array_of_numbers}
```
ralfhandl, line 2855:
```suggestion
          dataType:
            const: number
          data:
            $ref: array_of_numbers
```

ralfhandl (changes requested):
- Align YAML with JSON example
- Use consistent YAML formatting identical to #3714

handrews:
@ralfhandl ah, I see what happened here - I auto-ported the fix with `git format-patch` and `git am`, but spliced off some (but not all) of the during-review commits.  I will rebase and re-port this so that we get the exact same fixes.

handrews:
@ralfhandl I've marked your suggestions as resolved as the rebased force-push I just did _should_ contain all of your corrections from the previous PR (as well as some from other folks).

ralfhandl (approved):

handrews:
This had two approvals and the minor fix cleared them - one re-approval from the person who asked for the minor fix is sufficient, so I'll merge this (it's a port, after all).

PR #3797, opened 2024-05-09
Clarify constraints on Param Obj fields (3.2.0 port of #3760)

### LINKED ISSUE #1542: Requirement of schema or content in parameter object is difficult to notice

handrews:
* Ports #3760 
* Fixes #1542 

This makes the relational requirements of `content` vs `schema` and associated fields more clear.  The mutual exclusivity requirement is moved up to the top of the Fixed Fields section, and each table table is given a subsection clarifying its meaning.  This makes it easier to follow the multi-table format, which is unusual within the OAS.

The commentary on `example` and `examples` is moved to the section on `schema` and related parameters as it appears to only apply in that case.

### DIFF
--- a/versions/3.2.0.md
+++ b/versions/3.2.0.md
@@ -1086,7 +1086,7 @@ Field Name | Type | Description
 
 ###### Fixed Fields and considerations for use with `content`
 
-For more complex scenarios, the [`content`](#parameterContent) property can define the media type and schema of the parameter, as well as give examples of it use.
+For more complex scenarios, the [`content`](#parameterContent) property can define the media type and schema of the parameter, as well as give examples of its use.
 
 Field Name | Type | Description
 ---|:---:|---
--- a/versions/3.2.0.md
+++ b/versions/3.2.0.md
@@ -1053,6 +1053,14 @@ There are four possible parameter locations specified by the `in` field:
 
 
 ##### Fixed Fields
+
+The rules for serialization of the parameter are specified in one of two ways.
+Parameter Objects MUST include either a `content` field or a `schema` field, but not both.
+
+###### Common Fixed Fields
+
+These fields MAY be used with either `content` or `schema`.
+
 Field Name | Type | Description
 ---|:---:|---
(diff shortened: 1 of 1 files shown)


versions/3.2.0.md
```diff
@@ -1074,10 +1084,9 @@ Field Name | Type | Description
...
 <a name="parameterExamples"></a>examples | Map[ `string`, [Example Object](#exampleObject) \| [Reference Object](#referenceObject)] | Examples of the parameter's potential value. Each example SHOULD contain a value in the correct format as specified in the parameter encoding. The `examples` field is mutually exclusive of the `example` field. Furthermore, if referencing a `schema` that contains an example, the `examples` value SHALL _override_ the example provided by the schema.
 
-For more complex scenarios, the [`content`](#parameterContent) property can define the media type and schema of the parameter.
-A parameter MUST contain either a `schema` property, or a `content` property, but not both.
-When `example` or `examples` are provided in conjunction with the `schema` object, the example MUST follow the prescribed serialization strategy for the parameter.
+###### Fixed Fields and considerations for use with `content`
 
+For more complex scenarios, the [`content`](#parameterContent) property can define the media type and schema of the parameter, as well as give examples of it use.
```
lornajane, line 1089:
```suggestion
For more complex scenarios, the [`content`](#parameterContent) property can define the media type and schema of the parameter, as well as give examples of its use.
```

lornajane (approved):
Applied the typo fix from the sister PR while I was here - thanks @handrews for this clarification!

ralfhandl (approved):

PR #3799, opened 2024-05-12
Encapsulate page layout

Bellangelo:
Partially addresses https://github.com/OAI/OpenAPI-Specification/issues/1823.

This PR has the same changes as the https://github.com/OAI/OpenAPI-Specification/pull/1824 plus the requested changes that were requested here: https://github.com/OAI/OpenAPI-Specification/pull/1824#issuecomment-460532960.

Following the same logic as the original PR, it only fixes the alternative schema and after it is merged we will need to make the same changes to other collections.

### DIFF
--- a/_includes/alternative-schema-entry.md
+++ b/_includes/alternative-schema-entry.md
@@ -1,0 +1,11 @@
+# <a href="..">{{ page.collection }}</a>
+
+## {{ page.slug }} - {{ page.description }}
+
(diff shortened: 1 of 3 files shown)


handrews (approved):
Looks good, let's deploy it and see what happens- I suspect the alternative schemas registry was chosen to test it because that feature never got out of the proposal stage.

Issue #3792, opened 2024-05-08
Behavior of `allowEmptyValue: false`


handrews:
TL;DR: I have no idea how to define compliance criteria for `allowEmptyValue: false`, which is the default behavior, with setting `true` being NOT RECOMMENDED.

&lt;rant&gt;
The clarification added in 3.0.2 just says using it is NOT RECOMMENDED, which would, AFAICT, leave the default `false` behavior in place.  Looking at the [issue regarding that clarification](https://github.com/OAI/OpenAPI-Specification/issues/1573#issuecomment-389029180), there is a lot there that never made it into the spec.  Some of that says that setting it to `true` allows the undefined value behavior already defined by RFC6570.

However, `style`, `explode`, and `allowReseved` already control the RFC6570 behavior.  Are we by default forbidding empty values?  Does that mean RFC6570's undefined values (which includes `null`, empty arrays, empty objects, and objects where all properties are `null`, but ***NOT*** the empty string)?  Does it impact the empty string as well?  If `schema` allows these does the default `allowEmptyValue: false` behavior contradict that?  How are implementations expected to support this?

If using `content`, does `allowEmptyValue` have meaning?  What if the `content` is legitimately a zero-length string?

This seems like a field with a default value that blocks common cases in a very non-obvious way, requiring complex code to work around the normal behavior of Schema Objects, Media Type Objects, and RFC6570.  Do we really expect implementations to enforce this?  If we are recommending against its use, how do we expect users to support empty values, which can easily happen by accident (particularly empty arrays)?
&lt;/rant&gt;

lornajane:
We discussed in TSC and decided that since the use isn't recommended, we might try to continue as if this field does not exist.

handrews:
After thinking on this more, the only interpretation I can come up with that makes sense to me is:

* `allowEmptyValue: true` was intended to allow empty values _regardless of whether the schema allows them or not._
* this would make `allowEmptyValue: false` (the default) be a no-op:  if the schema allows an empty value, then it is still allowed

It just doesn't make sense to me that the default would contradict how schemas work everywhere else.  So it must be that the non-default value is the one that contradicts the schema.  This would also parallel how `nullable` works by contradicting the established `type`.

In the absence of anyone knowing how this was supposed to work, that's what I'm going to explain in the text.

ralfhandl:
[...]

PR #3803, opened 2024-05-14
Partial alternative attempt at allowEmptyValue clarification

handrews:
Alternative to:

* #3792 

This attempts to clarify based on objections to the previous attempt to clarify.  However, it is incomplete as the concept of "empty value" is not well-defined, which is why the following is embedded in the PR text and (obviously) not intended to be merged:

***TODO: What values are empty?  RFC6570 defines `null`, `[]`, `{}`, and objects with only `null` properties (e.g. `{"foo": null, "bar": null}` to be "undefined" and serialize as empty values, but it does not consider the empty string to be undefined even though it also serializes as an empty value.  Also, how does this interact with `content`, which is not RFC6570-based and may result in empty values through any number of different circumstances.***


handrews:
@ralfhandl @LasneF this is not about changing anything.  This keyword does... _something_?  It is not at all clear what.

Whether this PR or #3798 is accepted should not be about what we want, it should be about what was intended.  If necessary, that can be balanced with what a _representative set of multiple tooling vendors_ have done with it.

But this is a patch release.  It's not supposed to change anything.  If @ralfhandl 's previous assertion that this keyword, when `false` (the default) disallows empty values is correct, then yes, there is a MUST NOT here.  Not because I'm adding it, but because it was always there.  That's what we need to figure out.  That's why I submitted this despite not thinking it's a good approach, and indeed not being able to even properly define the implications.

This is the sort of ambiguity our tool developers and end users struggle with.  And if we can't figure it out, how can we expect them to do so?

Issue #3449, opened 2023-11-13
Webhook :  wrong example no path defined


LasneF:
according to the documentation web hooks are defined like this 

<html>
<body>

webhooks | Map[string, Path Item Object \| Reference Object] ] | The incoming webhooks that MAY be received as part of this API and that the API consumer MAY choose to implement. Closely related to the callbacks feature, this section describes requests initiated other than by an API call, for example by an out of band registration. The key name is a unique string to refer to each webhook, while the (optionally referenced) Path Item Object describes a request that may be initiated by the API provider and the expected responses. An example is available.
-- | -- | --

but in the official sample  presents i think a really corner case , i am not even sure it is valid 

as here we have webhook , a string here newPet , but so far no path object  , just a post operation

whereas i think we should have a path mentionning that  would go to /pet/ meaning that the implementer of the webhook must implements /pet/ with the describe payload

here the sample looks minimalistic not even sure it works does the implementer need to implements just post on the root 

to me the sample does not looks here

the comment in the files 
https://github.com/OAI/OpenAPI-Specification/blob/main/examples/v3.1/webhook-example.yaml

saying  ' This is a Path Item Object ' is unclear as it is describing only a verb not a path 

``
openapi: 3.1.0
info:
  title: Webhook Example
  version: 1.0.0

webhooks:
  // Each webhook needs a name
  newPet:
    // This is a Path Item Object, the only difference is that the request is initiated by the API provider
    post:
      requestBody:
        description: Information about a new pet in the system
        content:
          application/json:
            schema:
              $ref: "#/components/schemas/Pet"
      responses:
        "200":
          description: Return a 200 status to indicate that the data was received successfully

``

lornajane:
The specification and example are correct, describing the request that the _server_ can send to a client (rather than the more usual client sending a request to the server) using the same structure as the existing path items. The OpenAPI description does not specify where the webhook should be sent to, because it is common for clients to specify the location where they want to receive the webhook when registering for it, and that operation is out-of-band from the OpenAPI description.

Let us know if you have more questions on this, I'm not clear if there's something we can help more with, or if this explanation will be what you need.

lornajane:
@LasneF please ignore the bot, we added some new automation to help manage our repositories and it malfunctioned here. Issue is still open, and your comments are still welcome!

baywet:
[...]

PR #3805, opened 2024-05-14
Define "undefined" and "implementation-defined" (3.1.1 port of #3449)

### LINKED ISSUE #3775: Codify the meaning of "undefined" and "implementation-defined"

handrews:
* Port of #3779 
* Fixes #3775 

Undefined means "don't use this even if it seems to work, as there are cases where it won't."  Implementation-defined means "this isn't portable or interoperable, but it is still correct."

### DIFF
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -103,7 +103,7 @@ Status codes SHOULD be selected from the available status codes registered in th
 
 This specification deems certain situations to have either _undefined_ or _implementation-defined_ behavior.
 
-Behavior described as _undefined_ is likely, at least in some circumstances, to contradict the specification.
+Behavior described as _undefined_ is likely, at least in some circumstances, to result in outcomes that contradict the specification.
 This description is used when detecting the contradiction is impossible or impractical.
 Implementations MAY support undefined scenarios for historical reasons, including ambiguous text in prior versions of the specification.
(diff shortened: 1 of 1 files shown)


versions/3.1.1.md
```diff
@@ -99,6 +99,19 @@ Some examples of possible media type definitions:
 The HTTP Status Codes are used to indicate the status of the executed operation. 
 Status codes SHOULD be selected from the available status codes registered in the [IANA Status Code Registry](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml).
 
+##### <a name="undefinedAndImplementationDefinedBehavior"></a>Undefined and Implementation-Defined Behavior
+
+This specification deems certain situations to have either _undefined_ or _implementation-defined_ behavior.
+
+Behavior described as _undefined_ is likely, at least in some circumstances, to contradict the specification.
```
ralfhandl, line 106:
```suggestion
Behavior described as _undefined_ is likely, at least in some circumstances, to result in outcomes that contradict the specification.
```
Latest change missing

ralfhandl (changes requested):

versions/3.1.1.md
handrews: I've pushed a fix porting the other commit.

ralfhandl (approved):

mikekistler (approved):
Looks good. 👍

PR #3807, opened 2024-05-15
DEVELOPMENT.md: Criteria for supporting other specifications

handrews:
_[Apparently I wrote this a while back and forgot to submit it?  If this doesn't match what people remember of that conversation, I'm happy to drop it.]_

From a recent TDC call: changes related to supporting other specifications need to show that those specifications are actually being used.


mikekistler (approved):
I don't think I was present when this was discussed previously, but this seems like a positive change. 👍

miqui (approved):

Issue #3800, opened 2024-05-13
RFC 7231 has been obsoleted by RFC 9110


dret:
Looking at https://spec.openapis.org/oas/latest.html it still references RFC 7231 when referring to HTTP. But that specification was obsoleted by RFC 9110 in June 2022. I am not confident enough to make the change myself, but if somebody can point me at some "here's how to change the OpenAPI spec" primer, I'd be more than happy to give it a try.

handrews:
> if somebody can point me at some "here's how to change the OpenAPI spec" primer, I'd be more than happy to give it a try.

We're kind of buried in a backlog of issues trying to define and document this:

* https://github.com/orgs/OAI/projects/5

dret:
>   * #1778 <https://github.com/OAI/OpenAPI-Specification/issues/1778> ***@***.*** <https://github.com/dret> if you have any idea whether we really need to mention that 3986 obsoletes 1738 since we cite 1866 which cites 1738, that would be really helpful as I've thought myself in circles on it and no one else here seems to know)

for that specific case my opinion is: you cannot make IETF's publishing strategy more coherent than it is, and there always are some odd corner cases. your case is special because 1866 never was obsoleted by an IETF spec and that may cause some ripples. but why are you referencing that one and not the latest HTML spec which is for example also what's referenced in the IANA media type registry:

https://www.iana.org/assignments/media-types/media-types.xhtml

my general guideline is: if i reference an RFC and it shows up as "obsoleted", then i'll reference the "obsoleted by" RFC instead. but that's of course assuming everything is nicely cross-referenced...

handrews:
> not the latest HTML spec

I didn't make this reference, so I don't know, but I assume because it's actually in WHATWG's URL "Living Standard" which doesn't even acknowledge the existence of relative URI-references, among many other failures to cover 3986, which it makes an utterly laughable claim to "obsolete".  And then the other refuses to define relative references because, (and I [quote](https://github.com/whatwg/url/issues/531#issuecomment-714305738)) there is a "lack of browser-related use cases" 🤦 

While I work on the parameter serialization areas, I have debated mentioning WHATWG.  But they are aggressively browser-only, and I feel like that is confusing in the context of an API specification.

Plus, WHATWG specs are pseudocode tied to browser implementations.  They're completely unreadable for anyone else IMNSHO.

[...]

Issue #3677, opened 2024-03-23
Define and document branch strategy for the spec, both development and publishing


handrews:
This issue consolidates the various interrelated branch / merge / publish / etc. we need to sort out, as discussed in the Moonwalk 19 March and TDC 21 March meetings.  This needs to cover:

* Specification development
* Specification release / publishing
* Schema development
* Schema release / publishing
* Any additional publishing currently handled through GitHub pages (the `gh-pages` branch) such as registries

_[EDIT: Schema (#3716) and gh-pages (#3717) have been split out into their own issues.]_

I've included notes from the meeting below, followed by my own thoughts.

-----

@lornajane's notes from the [21 March 2024 TDC meeting](https://github.com/OAI/OpenAPI-Specification/issues/3633):

We also talked about our git versioning/branching/development workflow and it's clear that we are in need of a change. A fruitful discussion seems to have happened this week in the Moonwalk SIG meeting. A high-level summary of what I understood would be:
 - there is one file in the repository called `oas.md` and this is the canonical spec version (so I guess on `main` it's the latest stable release? We didn't talk about that)
 - we maintain one branch for 3.0, where `oas.md` is under active development and will become the 3.0.4 release. The branch will be tagged at release time. The same pattern for a 3.1 branch where the next tag would be 3.1.1 and a 3.2 branch where we'll tag 3.2.0.
 - when we tag a new version, we also copy the `oas.md` into the `main` branch under the `versions` structure that we have now.

The advantages of this structure are that we can diff the different versions and revisions of the same file in a sane way. There are fewer branches to keep track of so things should be easier to reason about. And it gives a simpler workflow for adding versions (hello 4.0) as we move forward.

My questions: where are we having the discussion about this that isn't in an agenda comment thread? And how do we apply changes such as tooling updates to the minor version branches without affecting the `oas.md` file? /cc @miqui since I know you're putting something together on this.

As a followup to getting the branches sorted, at some point we can change our use of github pages - currently it uses a branch, which was a very early implementation. We think it would be more useful to use a subfolder now.

-----

My thoughts:

* `main` is where officially published / publishable things live; the `versions` and `schemas` directories are just fine as they are, and the tagging can happen as it does now if we want to keep doing that
* I'd recommending initializing a new `dev` branch _purely to serve as the base for this new process._  This should include the work-in-progress development directory / files that aren't published, and therefore don't appear on main, initialized from `v3.0.4-dev`
[...]

PR #3809, opened 2024-05-15
Preliminary review - conformance to RFC9110

ioggstream:
## This PR

- preliminary review to comply with HTTP specifications RFC9110

## Notes

- payload / body -> content

### DIFF
--- a/versions/3.2.0.md
+++ b/versions/3.2.0.md
@@ -41,7 +41,7 @@ An OpenAPI definition can then be used by documentation generation tools to disp
 		- [Operation Object](#operationObject)
 		- [External Documentation Object](#externalDocumentationObject)
 		- [Parameter Object](#parameterObject)
-		- [Request Body Object](#requestBodyObject)
+		- [request content Object](#requestBodyObject)
 		- [Media Type Object](#mediaTypeObject)
 		- [Encoding Object](#encodingObject)
 		- [Responses Object](#responsesObject)
@@ -164,8 +164,8 @@ The formats defined by the OAS are:
 
 The OAS can describe either _raw_ or _encoded_ binary data.
 
-* **raw binary** is used where unencoded binary data is allowed, such as when sending a binary payload as the entire HTTP message body, or as part of a `multipart/*` payload that allows binary parts
-* **encoded binary** is used where binary data is embedded in a text-only format such as `application/json` or `application/x-www-form-urlencoded` (either as a message body or in the URL query string).
+* **raw binary** is used where unencoded binary data is allowed, such as when sending a binary payload as the entire HTTP message content, or as part of a `multipart/*` payload that allows binary parts
+* **encoded binary** is used where binary data is embedded in a text-only format such as `application/json` or `application/x-www-form-urlencoded` (either as a message content or in the URL query string).
 
 In the following table showing how to use Schema Object keywords for binary data, we use `image/png` as an example binary media type.  Any binary media type, including `application/octet-stream`, is sufficient to indicate binary content.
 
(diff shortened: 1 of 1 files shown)


versions/3.2.0.md
```diff
@@ -891,7 +891,7 @@ Field Name | Type | Description
...
+<a name="operationRequestBody"></a>requestBody | [request content Object](#requestBodyObject) \| [Reference Object](#referenceObject) | The request content applicable for this operation.  The `requestBody` is fully supported in HTTP methods where the HTTP 1.1 specification [RFC7231](https://tools.ietf.org/html/rfc7231#section-4.3.1) has explicitly defined semantics for request contents.  In other cases where the HTTP spec is vague (such as [GET](https://tools.ietf.org/html/rfc7231#section-4.3.1), [HEAD](https://tools.ietf.org/html/rfc7231#section-4.3.2) and [DELETE](https://tools.ietf.org/html/rfc7231#section-4.3.5)), `requestBody` is permitted but does not have well-defined semantics and SHOULD be avoided if possible.
```
ioggstream, line 894: reference https://datatracker.ietf.org/doc/html/rfc9110

versions/3.2.0.md
```diff
@@ -41,7 +41,7 @@ An OpenAPI definition can then be used by documentation generation tools to disp
 		- [Operation Object](#operationObject)
 		- [External Documentation Object](#externalDocumentationObject)
 		- [Parameter Object](#parameterObject)
-		- [Request Body Object](#requestBodyObject)
+		- [request content Object](#requestBodyObject)
```
ioggstream, line 44: Here and elsewere, retain Request Body

versions/3.2.0.md
```diff
@@ -1292,23 +1292,23 @@ content:
...
-<a name="requestBodyContent"></a>content | Map[`string`, [Media Type Object](#mediaTypeObject)] | **REQUIRED**. The content of the request body. The key is a media type or [media type range](https://tools.ietf.org/html/rfc7231#appendix-D) and the value describes it.  For requests that match multiple keys, only the most specific key is applicable. e.g. text/plain overrides text/*
-<a name="requestBodyRequired"></a>required | `boolean` | Determines if the request body is required in the request. Defaults to `false`.
+<a name="requestBodyDescription"></a>description | `string` | A brief description of the request content. This could contain examples of use.  [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation.
+<a name="requestBodyContent"></a>content | Map[`string`, [Media Type Object](#mediaTypeObject)] | **REQUIRED**. The content of the request content. The key is a media type or [media type range](https://tools.ietf.org/html/rfc7231#appendix-D) and the value describes it.  For requests that match multiple keys, only the most specific key is applicable. e.g. text/plain overrides text/*
```
ioggstream, line 1303:
```suggestion
<a name="requestBodyContent"></a>content | Map[`string`, [Media Type Object](#mediaTypeObject)] | **REQUIRED**. The request content. The key is a media type or [media type range](https://tools.ietf.org/html/rfc7231#appendix-D) and the value describes it.  For requests that match multiple keys, only the most specific key is applicable. e.g. text/plain overrides text/*
```

handrews:
@ioggstream thanks for this!  We are rearranging how our branches and directories are laid out which is about to make this PR invalid, so I'm closing it (and other 3.2.0 draft PRs).  However, issue #3800 is tracking this, and you're welcome to re-submit once the branching/directory changes described in #3677 are finalized and implemented.

Otherwise, issue #3800  is in the 3.2.0 milestone already so we'll definitely look at it, and it will have a link to this PR for us to consider when deciding how to handle it.  As you guessed with your later comments, we won't change "Request Body Object" to "Content Body Object" in a minor release, but we should consider the general vocabulary alignment.

Issue #3717, opened 2024-04-18
Should we-organize our GitHub pages to a separate repo (by way of a directory on main)?


handrews:
Splitting this out from #3766, to which it is still somewhat related as we should consider the impact on how many PRs to how many different branches are required for a change that spans the spec and the registries.

Originally, gh-pages sites needed to be on a `gh-pages` branch that (as ours does) typically has a completely different layout than `main`.  That is no longer required, and it is much more common to deploy from a directory on `main`.  Or at least from a directory on a branch that is no more different from `main` than any other working branch.

Bellangelo:
I totally agree with this one. This might be a personal preference but I usually find it confusing when each branch is a different project completely. I can start with this one after this is closed https://github.com/OAI/OpenAPI-Specification/issues/1823

frankkilcommins:
We should consider having a dedicated repo for [spec.openapis.org](http://spec.openapis.org/)? As we have now moved to a multi-specification project (and likely more coming - Overlays), nesting the website housing the specs under one of the specification repos is a little confusing.

Each separate specification repo that wants to publish can issue PRs to the repo housing the HTML versions and other information relevant to spec.openapis.org. We could then have the same process in place for both `OpenAPI-Specification` and `Arazzo-Specification` repos.

Bellangelo:
@frankkilcommins Should each spec has its own domain maybe? For example, `arazzo.openapis.org`, `overlays.openapis.org` etc. Or a single repo for everyone is the best option?

handrews:
@Bellangelo We want a unified brand across the different specs, so I think we'll want to stick with spec.openapis.org.

Bellangelo:
@handrews Fair enough. I see that @darrelmiller invited me to the organization, so now I have access to create a repo. Should I create a `spec.openapis.org` repo and move there the website?

handrews:
@Bellangelo I was surprised about repository creation privileges but it looks like a recent change that GitHub made, giving members more abilities.  Let's make sure @darrelmiller or another admin is OK with the separate-repository choice.

There might need to be a bit of work to figure out how the publishing of specs and schemas changes.  Not so much the technical logistics (which I'm sure you can sort out) but the policy side of things.

lornajane:
We are generally in favour. But we will need some process, governance and maintenance setup to enable this change. A good discussion in TDC, but we are not ready to make the change immediately.

lornajane:
[...]

PR #3810, opened 2024-05-16
Correctly handle appendicies in md2html

handrews:
We have an Appendix A that was getting treated as a numbered section, with a second Appendix A generated for references.

This now correctly detects sections with "Appendix" in the name, and assigns letters appropriately.  The generated references section is now Appendix B (and will always be the last appendix).

### DIFF
--- a/scripts/md2html/md2html.js
+++ b/scripts/md2html/md2html.js
@@ -293,18 +293,22 @@ for (let l in lines) {
         //if (delta<0) delta = -1;
         if (Math.abs(delta)>1) console.warn(delta,line);
         let prefix = '';
+        let newSection = '<section>';
+        if (line.includes('Appendix')) {
+            newSection = '<section class="appendix">';
+        }
 
         // heading level delta is either 0 or is +1/-1, or we're in respec mode
         /* respec insists on <section>...</section> breaks around headings */
 
         if (delta === 0) {
-            prefix = '</section><section>';
+            prefix = '</section>'+newSection;
         }
         else if (delta > 0) {
-            prefix = '<section>'.repeat(delta);
+            prefix = newSection.repeat(delta);
         }
         else {
-            prefix = '</section>'+('</section>').repeat(Math.abs(delta))+'<section>';
+            prefix = '</section>'+('</section>').repeat(Math.abs(delta))+newSection;
         }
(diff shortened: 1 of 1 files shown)


ralfhandl:
Nice!

This results in 

image

which is almost what I'd expect: can we snip away the `Appendix A: `?

handrews:
@ralfhandl let me see if I can figure out enough JavaScript for that 😅

handrews:
@ralfhandl it looks a bit weird to me without the "Appendix" because then the word "Appendix" doesn't appear anywhere.

oas-appendix

Ideally we would keep "Appendix A: " in the main text and leave it out of the TOC- that is what IETF RFCs do.  But I don't know how to make that happen and don't have more time to spend on this.  If you feel strongly about taking it out we can just go with bare letters, but out of the two not-great options I'd rather leave the redundancy in (it's not worse than what we current have) and hope someone can figure out a better solution later.

ralfhandl (approved):

lornajane:
I'm sorry, I'm confused about how to preview the changes. I don't see Appendix B at all but I know we do some extra rendering. Am I just missing the docs on this?

handrews:
@lornajane 
> Am I just missing the docs on this?

Bold of you to assume we have docs!  🤪 

I was lazy and did not publish a rendering, as you have to bounce between different branches to do it (at least until #3717 is fixed) but i should have.  Here are three relevant samples (edit the URLs in the obvious ways to see other versions if you want):

* [2.0](https://handrews.github.io/renderings/oas/deploy/oas/v2.0.html#references) only has the auto-generated "References" Appendix
* [3.0.0](https://handrews.github.io/renderings/oas/deploy/oas/v3.0.0.html#appendix-a-revision-history) has a manually created "Change History" and auto-generated "References"
* [3.0.4](https://handrews.github.io/renderings/oas/deploy/oas/v3.0.4.html#appendix-a-revision-history) (rendered from the branch for #3818) has two manually-written appendices plus the auto-generated "References"

darrelmiller (approved):

mikekistler (approved):
Looks good! 👍

Montiwa11 (commented):
[mage]()

Issue #3739, opened 2024-04-23
Encoding Object `contentType` field: "comma-separated list of the two types"?


handrews:
Docs for the `contentType` field of the Encoding Object in 3.1.0 read (***emphasis added***):

> The Content-Type for encoding a specific property. Default value depends on the property type: for object - `application/json`; for array – the default is defined based on the inner type; for all other cases the default is `application/octet-stream`. The value can be a specific media type (e.g. `application/json`), a wildcard media type (e.g. `image/*`), ***or a comma-separated list of the two types***

In 3.0.3 it reads (***emphasis added*** again):

> The Content-Type for encoding a specific property. Default value depends on the property type: for `string` with `format` being `binary` – `application/octet-stream`; for other primitive types – `text/plain`; for object - `application/json`; for array – the default is defined based on the inner type. The value can be a specific media type (e.g. `application/json`), a wildcard media type (e.g. `image/*`), ***or a comma-separated list of the two types***.

Two types?  Just exactly two?  Was this meant to be "two or more types"?  The 3.0.3. wording of "***the*** two types" feels a bit different but no more clear to me.

karenetheridge:
Side note: when I was looking at this area of the spec a few weeks ago while adding some multipart support in my implementation, I noticed that the schema specifies this field as a media-type, which isn't valid if there are multiple comma-separated values.

Also, "inner type" in that paragraph needs clarifications. An example would help too :)

handrews:
@OAI/tsc review request:  Does anyone know what this was supposed to mean?

darrelmiller:
Only allowing two media types makes no sense to me, especially considering we claim to allow wildcards.  Considering the JSON schema mentions media-range my guess is that we were expecting to support this https://www.rfc-editor.org/rfc/rfc9110.html#section-12.5.1 media-range here.

mikekistler:
I think what was meant here is:
> The value can be a specific media type (e.g. `application/json`), a wildcard media type (e.g. image/*), or a comma-separated list of values that can be either a specific media type or a wildcard media type.

handrews:
@mikekistler I'm 99.999% sure that's the same as saying a list of media-ranges (as is taken by the `Accept` header).  However we describe it for the Responses Object, we should probably do the same here.

mikekistler:
My reading of the BNF:
```
media-range = ( "*/*" / ( type "/*" ) / ( type "/" subtype ) ) *( OWS
[...]

PR #3813, opened 2024-05-16
Encoding Obj contentType takes a media range (3.0.4)

handrews:
Fixes:
* #3739 

The "the two types" appears to have come from an example involving two types, and the TSC agrees that there was no intention to limit it to two.  The syntax described also matches media type ranges, which is what is used for keys under `content` keywords that take objects of Media Type Objects.  As that is analogous to the use of `contentType` in the Encoding Object, this change aligns the wording.

The guidance on default values was difficult to format in a readable way inside of the fixed fields table, so this change also moves it to its own table.

### DIFF
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -1601,13 +1601,24 @@ A single encoding definition applied to a single schema property.
 ##### Fixed Fields
 Field Name | Type | Description
 ---|:---:|---
-<a name="encodingContentType"></a>contentType | `string` | The Content-Type for encoding a specific property. Default value depends on the property type: for `string` with `format` being `binary` – `application/octet-stream`; for other primitive types – `text/plain`; for `object` - `application/json`; for `array` – the default is defined based on the inner type. The value can be a specific media type (e.g. `application/json`), a wildcard media type (e.g. `image/*`), or a comma-separated list of the two types.
+<a name="encodingContentType"></a>contentType | `string` | The Content-Type for encoding a specific property. The value is a media type or [media type range](https://tools.ietf.org/html/rfc7231#appendix-D). Default value depends on the property type as shown in the table below.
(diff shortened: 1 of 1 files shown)


versions/3.0.4.md
```diff
@@ -1601,14 +1601,25 @@ A single encoding definition applied to a single schema property.
 ##### Fixed Fields
 Field Name | Type | Description
 ---|:---:|---
-<a name="encodingContentType"></a>contentType | `string` | The Content-Type for encoding a specific property. Default value depends on the property type: for `string` with `format` being `binary` – `application/octet-stream`; for other primitive types – `text/plain`; for `object` - `application/json`; for `array` – the default is defined based on the inner type. The value can be a specific media type (e.g. `application/json`), a wildcard media type (e.g. `image/*`), or a comma-separated list of the two types.
+<a name="encodingContentType"></a>contentType | `string` | The Content-Type for encoding a specific property. The value is a media type or [media type range](https://tools.ietf.org/html/rfc7231#appendix-D). Default value depends on the property type as shown in the table below.
```
ralfhandl, line 1604:
```suggestion
<a name="encodingContentType"></a>contentType | `string` | The Content-Type for encoding a specific property. The value is a comma-separated list of [media type ranges](https://tools.ietf.org/html/rfc7231#appendix-D). Default value depends on the property type as shown in the table below.
```
Note: rule `media-range` is only a single thingy, and it includes non-wildcard `type "/" subtype` as an alternative.

We can't take away the "list of" in a patch version.

ralfhandl (changes requested):

versions/3.0.4.md
handrews: @ralfhandl there is no intention to take away anything.  This is what we thought made sense in a TDC call, so either it was supposed to be "list of media-ranges" or something else.

I think perhaps what we want is list of media-ranges-without-parameters, because that's what the previous awkward wording (aside from the "the two types" part) described.

handrews:
Converting to draft because I did some deeper digging and this is murkier than expected.

handrews:
OK on second thought I'm closing this and starting over.  @ralfhandl thanks for catching the un-intended removal of list-ness, sadly there's more complex history here as well.

PR #3814, opened 2024-05-17
[new] scripts/update-schema-id.sh

TakahikoKawasaki:
This script updates the schema IDs in the following YAML files:

- schemas/v{major}.{minor}/schema.yaml
- schemas/v{major}.{minor}/schema-base.yaml

and generates the following JSON files from the updated YAML files:

- schemas/v{major}.{minor}/schema.json
- schemas/v{major}.{minor}/schema-base.json


ralfhandl (changes requested):
Ran the script with `-v 3.1` and it worked fine.

Running with `-v 3.0` showed problems:
- schema.yaml was not updated
- schema-base.yaml was created empty
- schema-base.json was created with content `null`

ralfhandl:
Hi @TakahikoKawasaki,

Thanks for providing the script. 

The script works fine for the v3.1 case with separate `schema` and `schema-base`, and fail for the v3.0 case where there is only `schema`.

Would you mind adding this distinction to the scripts?

Thanks in advance!

TakahikoKawasaki:
@ralfhandl 
- Updated the script to support the version 3.0.
- Versions older than 3.0 are rejected on the assumption that maintenance of 2.x and older versions has stopped.

If it is necessary to support versions older than 3.0, please let me know.

miqui:
@ralfhandl script so far looks good to me, but I'll let you run with it.

ralfhandl (approved):
Works for both v3.1 and v3.0

miqui (approved):

karenetheridge:
I see you're using `ex` (which is a neutered form of `vim`) to do the substitutions -- I'm wondering why you made this choice, vs. the more usual choice of `sed` which provides access to a richer set of regular expressions?

TakahikoKawasaki:
@karenetheridge Sorry, I did not know the `-i` option of `sed`. The `ex` command line can be replaced with `sed` if it is desired.

lornajane:
I'm closing this pull request since it's no longer needed since the recent changes to the schema publishing process.

PR #3801, opened 2024-05-14
Link for case insensitivity of auth scheme

ralfhandl:
and correct expansion of "auth".

Will forward-port this to 3.1.1 and 3.2.0 once accepted.

PR #3817, opened 2024-05-17
3.2.0 upport of #3801 and #2644: case insensitivity of auth scheme

ralfhandl:
Upport of
- #3801 
- #2644


miqui (approved):
looks good to me

Issue #1551, opened 2018-04-19
Encoding object has style and contentType


jwalton:
Let's say I have:

```yaml
requestBody:
  content:
    application/x-www-form-urlencoded:
      schema:
        type: object
        properties:
          address:
            type: object
            properties: {}
      encoding:
        address:
            contentType: "application/json" # This is the default
            style: form # This is the default
            explode: true # This is the default
```

Since encoding has both a `contentType` and a `style`, what do I do with the `address` property?  Do I decode it as an exploded form object?  Do I extract the value of 'address' from the message body and run it through `JSON.parse()`?  Do I decode it as an exploded form, but run the value of each key through `JSON.parse()`?

Or does contentType here only apply to `multipart` bodies?

alexadrake:
I am also running into this question; is there any resolution?

simplesagar:
Also running into the same

handrews:
OK, digging through this more, it was actually fixed in 3.1.0 with the following new text added to `style` (which, along with `explode` (but not `allowReserved`), was extended to support `multipart/form-data` as well):

> If a value is explicitly defined, then the value of [contentType](https://spec.openapis.org/oas/v3.1.0#encodingContentType) (implicit or explicit) SHALL be ignored.

So that at least explains how to override `contentType` (default or otherwise).  The implication is that if it's not explicitly defined then `contentType` wins... I guess?  The `style` text still references query params for default values, so it's still not entirely clear, nor is it entirely clear what is supposed to happen in 3.0, but that's at least a clue.

Since `allowReserved` was not extended to `form-data`, then there needs to be a way to _not_ use RFC6570 value expansion (with its automatic percent-encoding), so I'm going to interpret this to mean that despite the wording around defaults, part of this is changing the behavior when no explicit `style` is defined.

This is going to take more investigation and thought to fully unravel.

handrews:
@OAI/tsc review request:  It's unclear how to clarify 3.0.4 here.  Even 3.1.1 is a bit murky, but the only thing I can think of for 3.0.4 is to more-or-less backport the 3.1.1 behavior.  Or maybe deem the area implementation-defined (where implementations can choose which of _two_ options to support) while making the 3.1.1-compatible behavior RECOMMENDED?

ralfhandl:
We can only recommend the 3.1.1 behavior in a patch version, so let's do that.

handrews:
[...]

Issue #1322, opened 2017-08-16
Example of deepObject + explode=false


hkosova:
How are objects serialized when using `style: deepObject` and `explode: false`? [Style Examples](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#style-examples) contain a `deepObject` example only for `explode: true`.

(Related - OAI/learn.openapis.org#100)

lornajane:
We discussed that we probably don't need exploded (or not) alternatives for deepObject; @handrews will add a clarification note but we won't add another entry for it with explode=false

handrews:
Fix added to PR #3859.  @lornajane I know you said don't add more rows but it was just bothering the heck out of me that it's unclear in the one place people glance at the most (rather than poring over the text).  If you want me to remove them I will, but I as I look over the voluminous history of issues and complaints in this area, I think failing to address it in the table is problematic (and likewise for the other incomplete entires).

handrews:
PR merged for 3.0.4 and ported to 3.1.1 via PR #3921!

Issue #1288, opened 2017-07-22
More and complete examples for serialization styles


slinkydeveloper:
OAS 3 include more complex serialization styles than Swagger 2.0 . Can you include in the spec more complete examples for every in/serialization style/explode/type combination on **how a request is supposed to be done**? Also where parameters need to be encoded or not. For example:

**Path parameters**

| style  | exploded | example path | example uri request | example array uri request | example object uri request |
|--------|----------|--------------|---------------------|---------------------------|----------------------------|
| matrix | true     | /{color}     | /;color=blue        | /;color=blue,black,brown  | /;color=R,100,G,200,B,150  |

**Cookie parameters**

| style | exploded | example cookie header | example array cookie header                | example object cookie header |
|-------|----------|-----------------------|--------------------------------------------|------------------------------|
| form  | true     | Cookie: color=blue    | Cookie: color=blue; color=red; color=green | Cookie: R=100; G=100; B=200  |

**Query parameters**

| style | exploded | example uri      | example array uri                | example object uri                      |
|-------|----------|------------------|----------------------------------|-----------------------------------------|
| form  | false    | /path?color=blue | /path?color=blue%2Cblack%2Cbrown | /path?color=R%2C100%2CG%2C200%2CB%2C150 |

Of course this examples can be included outside the spec markdown, but they can be really useful for tool implementers. Thank you

MikeRalphson:
A good case for a revised Wiki (see #1249)

darrelmiller:
@slinkydeveloper I attempted to make the serialization rules a subset of RFC 6570's rules.  https://tools.ietf.org/html/rfc6570#section-3.2.7  The idea was that you should could use an existing URI Template library to do the serialization for you.  Just construct a template based on the operation definitions and "ta-da" no work to do.  Let me know if there are places I failed to do that.

slinkydeveloper:
Okay i'll give a try thank you so much!

hkosova:
I've published serialization examples at https://swagger.io/docs/specification/serialization/. Hope it helps!

slinkydeveloper:
That's awesome thank you so much!

In the project I'm working on, I've created a "unit test generator" (specific for my project) based on an oas 3 spec. This generator actually takes all operations declared in an oas 3 spec and generates a specific test to validate the correct parsing of parameters on server side. With some small changes this can be a complete server libraries/frameworks compatibility test tool.
[...]

PR #3818, opened 2024-05-17
Appendix on RFC6570-derived behavior

handrews:
Addresses:
* #2940 
* #1551 

This is the first of several appendixes to be added to clarify the most obscure details of Parameter Object and Encoding Object serialization.

***Why an appendix?*** Because these are really fiddly details, and technically they all follow from the existing requirements.

***Why not put this on the Learn site?*** Because this is part of a larger set of serialization issues that took me several weeks to sort out, despite being very comfortable researching and reconciling RFCs in general, and being very familiar with the relevant RFCs in particular.  It's too high of a bar to relegate to secondary documentation.  It needs clarity in the spec, IMNSHO.

This clarifies the correspondence between OAS fields and RFC6570 operators, and acknowledges that some field values and combinations do not have analogues.  It provides further guidance for how to use RFC6570 implementations to support these configurations.

This includes a SHOULD directive regarding using RFC6570 expansion with the non-RFC6570 styles, as the use of "explode" and "allowReserved" does not otherwise make any sense.  It perhaps could be a MUST.

This also includes an acknowledgement of ambiguity in reconciling `contentType` and `style` in the Encoding Object, make the behavior


versions/3.0.4.md
```diff
@@ -3495,3 +3497,53 @@ Version   | Date       | Notes
...
+
+This example is equivalent to RFC6570's `{?foo*,bar}`, and ***NOT*** `{?foo*}{&bar}`.
+The `&` operator has no equivalent in the Parameter Object.
+
+### Non-RFC6570 Field Values and Combinations
+
+Configurations with no direct RFC6570 equivalent SHOULD also be handled according to RFC6570.
+Implementations MAY pass individual names and values to RFC6570 regular or reserved expansion (based on `allowReserved`), and then assemble the results into a string with the appropriate delimiters.
+
+This includes the styles `pipeDelimited`, `spaceDelimited`, and `deepObject`, as well as the combination of the `form` style with `allowReserved: true`.
```
ralfhandl, line 3549:
```suggestion
This includes the styles `pipeDelimited`, `spaceDelimited`, and `deepObject`, as well as the combination of style `form` with `allowReserved: true`.
```

ralfhandl (commented):
+1, one phrasing suggestion

ralfhandl (dismissed):

handrews:
@darrelmiller after thinking about your comment in https://github.com/OAI/OpenAPI-Specification/issues/1288#issuecomment-342539281 I'm wondering if maybe I'm not giving the right guidance on the combination of `style: form, allowReserved: true`.

In this PR, I basically suggest using single-variable templates like `"{+foo}"` and assembling the query string with the results.  Alternatively, you could assemble a template using a mixture of RFC6570's `?`, `&`, and `+` operators, and pass that whole thing through to an RFC6570s process.  Which you could also do for `pipeDelimited`, `spaceDelimited`, and `deepObject`.

I'm going to move this to a draft PR because I think this needs some consideration.

versions/3.0.4.md
```diff
@@ -3495,3 +3497,53 @@ Version   | Date       | Notes
...
+
+field | value | operator
+----- | ----- | --------
+style | simple | _n/a_
+style | matrix | `;`
+style | label | `.`
+style | form | `?`
+allowReserved | `false` | _n/a_
+allowReserved | `true` | `+`
+explode | `false` | _n/a_
```
darrelmiller, line 3524: RFC 6570 refers to explode as a modifier because it is a suffix added to the variable as opposed to an operator which appears as a prefix.

versions/3.0.4.md
```diff
@@ -3495,3 +3497,53 @@ Version   | Date       | Notes
...
+  schema:
+    type: object
+  explode: true
+- name: bar
+  in: query
+  schema:
+    type: string
+```
+
+This example is equivalent to RFC6570's `{?foo*,bar}`, and ***NOT*** `{?foo*}{&bar}`.
```
darrelmiller, line 3541: > {?foo*}{&bar}
Is this even a valid URI Template?  If foo is empty/undefined then the resulting href would be invalid.

versions/3.0.4.md
handrews: Ah, yes.  I was just going by the characters, not the placement, but you're right that it should be more clear.

versions/3.0.4.md
handrews: Yeah that's kind of my point.  I mean, it's a valid URI Tempate in the sense that it's syntactically legal (I think?) it just might not produce a valid URI.  The URI Template spec is pretty intentionally blind to the context outside of the variables.

But perhaps I could note that that's why you need to do it the `{?foo*,bar}` way.

handrews:
Based on @darrelmiller 's comments and trying to think this through further, I have significantly expanded this and changed the recommendation regarding how to construct query strings with configurations that don't quite match RFC6570.

I've also rewritten the `allowReserved` section and incorporated that aspect in the appendix.  I had planned to address that in another appendix on percent-encoding, but this allows readers to find all query string-related RFC6570 content in one place, and will let the percent-encoding appendix focus on the confusing interoperability landscape there.

Finally, the appendix now has several examples showing both easy and difficult cases.

versions/3.0.4.md
handrews: I think I've addressed this in the latest, much-expanded version.

versions/3.0.4.md
handrews: I changed the headings to make things more clear in the latest reworking of this PR.

versions/3.0.4.md
```diff
@@ -3523,3 +3527,215 @@ Version   | Date       | Notes
...
+Configurations with no direct RFC6570 equivalent SHOULD also be handled according to RFC6570.
+Implementations MAY create a properly delimited URI Template with variables for individual names and values using RFC6570 regular or reserved expansion (based on `allowReserved`).
+
+This includes:
+  * the styles `pipeDelimited`, `spaceDelimited`, and `deepObject`, which have no equivalents at all
+  * the combination of the style `form` with `allowReserved: true`, which is not allowed because only one prefix operator can be used at a time
+  * any parameter name that is not a legal RFC6570 variable name (alphanumerics, `_`, and percent-encoded characters)
+
+The Parameter Object's `name` field has a much more permissive syntax than [RFC6570 variable name syntax](https://www.rfc-editor.org/rfc/rfc6570#section-2.3).
+A parameter name that includes characters outside of the allowed RFC6570 variable character set MUST be percent-encoded before it can be used in a URI Template, and MUST set `allowReserved: true` to avoid double percent-encoding.
```
darrelmiller, line 3590: I don't think it is necessary to set allowReserved to true.
```csharp
            var template = new UriTemplate("http://example.org/{?foo%2Fid}");
            template.SetParameter("foo%2Fid", "bar/");
            var uriString = template.Resolve();
            Assert.Equal("http://example.org/?foo%2Fid=bar%2F", uriString);
```
I tried this using my Tavis.UriTemplates library and it passed without having to use allowReserved.

versions/3.0.4.md
```diff
@@ -3523,3 +3527,215 @@ Version   | Date       | Notes
...
+So we need to restructure the data to fit a manually constructed URI Template that passes all of the pieces through the right sort of expansion.
+
+Here is one such template.
+The exact variable names are unimportant, but in this example "f" and "w" are meant to suggest "formulas" and "words", while "n" and "v" suggest "name" and "value":
+
+```urlencoded
+?{+fn0}={+fv0}&{+fn1}={+fv1}&{+fn2}={+fv2}&{wn}={wv0} {wv1} {wv2}
+```
+
+We'll also need to pre-process the names and values for `formaulas` because while `/` and most other reserved characters are allowed in the query string by RFC3986, `[`, `]`, and `#` [are not](https://datatracker.ietf.org/doc/html/rfc3986#appendix-A), and `&`, `=`, and `+` all have [special behavior](https://www.rfc-editor.org/rfc/rfc1866#section-8.2.1) in the `application/x-www-form-urlencoded` format, which is what we are using in the query string.
```
darrelmiller, line 3674:
```suggestion
We'll also need to pre-process the names and values for `formulas` because while `/` and most other reserved characters are allowed in the query string by RFC3986, `[`, `]`, and `#` [are not](https://datatracker.ietf.org/doc/html/rfc3986#appendix-A), and `&`, `=`, and `+` all have [special behavior](https://www.rfc-editor.org/rfc/rfc1866#section-8.2.1) in the `application/x-www-form-urlencoded` format, which is what we are using in the query string.
```

versions/3.0.4.md
```diff
@@ -3523,3 +3527,215 @@ Version   | Date       | Notes
...
+```
+
+We can't combine the `?` and `+` RFC6570 prefixes, and there's no way with RFC6570 to replace the `,` separator with a space character.
+So we need to restructure the data to fit a manually constructed URI Template that passes all of the pieces through the right sort of expansion.
+
+Here is one such template.
+The exact variable names are unimportant, but in this example "f" and "w" are meant to suggest "formulas" and "words", while "n" and "v" suggest "name" and "value":
+
+```urlencoded
+?{+fn0}={+fv0}&{+fn1}={+fv1}&{+fn2}={+fv2}&{wn}={wv0} {wv1} {wv2}
```
darrelmiller, line 3671: It isn't clear to me why we need the + for the formula names.  Is this intended to cover the deep object scenario where someone uses  [ ]  characters in the names.

versions/3.0.4.md
```diff
@@ -3523,3 +3527,215 @@ Version   | Date       | Notes
...
+```YAML
+wn: words
+wv0: hello
+wv1: world
+```
+
+Manually constructed URI Template:
+
+```urlencoded
+?{wn}={wv0} {wv1}
```
darrelmiller, line 3734: I'm not sure I understand the advantage of that manually constructed URI template over

> ?words={wv0} {wv1}

versions/3.0.4.md
```diff
@@ -3523,3 +3527,215 @@ Version   | Date       | Notes
...
+
+Expanding our manually assembled template with our restructured data yields the following query string:
+
+```urlencoded
+?a=x%2By&b=x/y&c=x%5Ey&words=math%20is%20fun
+
+```
+The `/` and the pre-percent-encoded `%2B` have been left alone, but the disallowed `^` character (inside a value) and space characters (in the template but outside of the expanded variables) were percent-encoded.
+
+#### Undefined Values and Manual URI Template Construction
```
darrelmiller, line 3703: This is a really interesting exploration of doing "late generation" of URI templates based on the parameter values to work around the features in OpenAPI that are not supported by 6570.

versions/3.0.4.md
handrews: I now want to rewrite the example to use forMAULas so it's more dangerous... 🐻

versions/3.0.4.md
handrews: It's a good question and relates to your other two comment about [not needing to set `allowReserved` for names with percent encoding](https://github.com/OAI/OpenAPI-Specification/pull/3818#discussion_r1616048516) and [not needing the `+` on the formula names](https://github.com/OAI/OpenAPI-Specification/pull/3818#pullrequestreview-2080933650).

My original thought was "simulate the normal behavior of parameter names in `form` by putting them through the same process as the values" which led to "if you need pre-percent encoded values in names then you'll have to set `allowReserved: true` to get them.

But since that seems not to be true (two Python URI Template implementations both agree with you on that), then there's no need to use `allowReserved` and therefore no need to make the names template variables since they have to be pre-encoded anyway.  It's simpler to just copy them as part of the literal portion of the template.  I'll make an update taking all of this into account.

versions/3.0.4.md
handrews: It is not, it was about [names that aren't valid template variable names](https://github.com/OAI/OpenAPI-Specification/pull/3818#discussion_r1616235641).  Which would include the `deepObject`-style names and, yeah, that could be interesting, but... UGH.  `deepObject`.  idek know what to do about that.  See #1322 - I can't entirely figure out how it's even supposed to work.  I think I'm going to ignore it for now as I'll need to fix several `deepObject` things as it is.  I've been putting it off to the last thing.

versions/3.0.4.md
handrews: Yes, I tried two Python implementations and they agree, sse also [this comment about the larger context](https://github.com/OAI/OpenAPI-Specification/pull/3818#discussion_r1616235641)

versions/3.0.4.md
```diff
@@ -1068,11 +1068,13 @@ Field Name | Type | Description
...
+<a name="parameterAllowReserved"></a>allowReserved | `boolean` | When this is true, parameter values are serialized  using reserved expansion, as defined by [RFC6570](https://datatracker.ietf.org/doc/html/rfc6570#autoid-20), which allows [RFC3986's reserved character set](https://datatracker.ietf.org/doc/html/rfc3986#autoid-13), as well as percent-encoded triples, to pass through unchanged, while still percent-encoding all other disallowed characters (including `%` outside of percent-encoded triples). Applications are still responsible for percent-encoding reserved characters that are not allowed in the query string, or have a meaning other than their literal value (e.g. `+` as an escape for the space character in `application/x-www-form-urlencoded` query strings). This property only applies to parameters with an `in` value of `query`. The default value is `false`.
```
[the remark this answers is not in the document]
handrews: The upshot is that if you use `allowReserved: true`, you still need to manually percent-encode the following reserved characters:

* `#` because it is the delimiter between the query and fragment
* `[` and `]` because while they are reserved, they are forbidden in the query by RFC3986 (yes, `deepObject` examples are wrong, see #3859 for more on this)
* `+` because `form-urlencoded` treats it as a space
* `&` and `=` because `form-urlencoded` uses them as delimiters

You can leave any of the others unencoded, and you MUST leave a `+` that represents a space unencoded.  Which maybe I did not show in the example and should have... hm....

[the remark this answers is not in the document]
handrews: @darrelmiller I'm open to suggestion on how to better state this.  At one point I had a short paragraph after thist table going into a bit more detail, and then moved it all to the appendix.  Perhaps a short paragraph is warranted, though.

handrews:
@darrelmiller I've added [another commit](https://github.com/OAI/OpenAPI-Specification/pull/3818/commits/ed7a6d8dc79d262b97d8359c4d9735f597eab541) that hopefully addresses all of your concerns.

versions/3.0.4.md
```diff
@@ -1068,11 +1068,13 @@ Field Name | Type | Description
...
+<a name="parameterAllowReserved"></a>allowReserved | `boolean` | When this is true, parameter values are serialized  using reserved expansion, as defined by [RFC6570](https://datatracker.ietf.org/doc/html/rfc6570#autoid-20), which allows [RFC3986's reserved character set](https://datatracker.ietf.org/doc/html/rfc3986#autoid-13), as well as percent-encoded triples, to pass through unchanged, while still percent-encoding all other disallowed characters (including `%` outside of percent-encoded triples). Applications are still responsible for percent-encoding reserved characters that are [not allowed in the query string](https://datatracker.ietf.org/doc/html/rfc3986#autoid-24) (`[`, `]`, `#`), or have a special meaning in `application/x-www-form-urlencoded` (`-`, `&`, `+`); see Appendices [C](#usingRFC6570Implementations) and [E](#percentEncodingAndFormMediaTypes) for details. This property only applies to parameters with an `in` value of `query`. The default value is `false`.
```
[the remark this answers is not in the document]
handrews: @ralfhandl these are inconsistent all over and I'd rather deal with them all at once (like numerous other document-wide issues).

handrews:
@ralfhandl I've started a pre-release checklist and put the RFC links on it:

* https://github.com/OAI/OpenAPI-Specification/issues/3785#issuecomment-2136168834

I'm not going to change them in this PR because it would just be added noise as we don't know the target yet (and I do not want to hold this PR up for that decision- we need to do that all at once, like fixing the section links).

versions/3.0.4.md
```diff
@@ -3523,3 +3527,244 @@ Version   | Date       | Notes
 1.2       | 2014-03-14 | Initial release of the formal document.
 1.1       | 2012-08-22 | Release of Swagger 1.1
 1.0       | 2011-08-10 | First release of the Swagger Specification
+
+## <a name="usingRFC6570Implementations"></a>Appendix C: Using RFC6570 Implementations
+
+Serialization is defined in terms of RFC6570 URI Templates in two scenarios:
```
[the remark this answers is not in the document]
handrews: I see you filed an issue to track this, and I added it to the pre-release checklist, so let's handle it that way and not do anything right now.  There are places all over the spec (not just my recent PRs) where this needs sorting out.

versions/3.0.4.md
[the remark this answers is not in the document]
ralfhandl: Do we have an issue for these?

ralfhandl (approved):

darrelmiller (approved):

handrews:
This now has 2 tsc approvals, one of which was cleared by the merge conflict resolution which was trivial.  So I'm going to merge this.

rafalkrupinski:
```YAML
parameters:
- name: foo
  in: query
  schema:
    type: object
  explode: true
- name: bar
  in: query
  schema:
    type: string
```
> This example is equivalent to RFC6570's `{?foo*,bar}`, and **_NOT_** `{?foo*}{&bar}`, which is problematic because if `foo` is not defined, the result will be an invalid URI.

Can you clarify what's the difference between the two URL patterns, and why the example schema is equivalent to one vs the other?

The conclusion doesn't sound right to me. Looking at the OAD fragment, I don't see why missing `foo` should result in an invalid URL.

handrews:
@rafalkrupinski I think this is something that folks can figure out by putting the data into an RFC6570 implementation and checking out the results.  Alternatively, I think having more examples on learn.openapis.org would be great, and you could add this request to:

* OAI/learn.openapis.org#100


In this case, given a path of `/some/path`, if `foo` is missing and `bar` is `"hello"`, then `{?foo*,bar}` expands to `/some/path?bar=hello`, while `{?foo*}{&bar}` expands to `/some/path&bar=hello`, which is not something that will be parsed as a query string because there's no `?` delimiter.

rafalkrupinski:
@handrews 
OK, so an implementation of the RFC `{?foo*}{&bar}` would behave like this, I still see no reason why an implementation of the OAD should. Is this a backdoor backporting of RFC6570 compatibility from OAS3.1?

I mean it's not obvious _that_ RFC should be used to interpret path/query parameters definition to higher degree than using the right separator tokens in the right places.*
Then it's not obvious _how_ to interpret given OAD parameters as a URL template. I don't even recall a requirement that query parameters should be processed in order, which would be implied by `{?foo*}{&bar}`

* Then again what "the right places" mean? To me it always meant '?' between path and query parameters, '&' between query parameters. To someone else it means `{?foo*}{&bar}`...  A subtle difference.

handrews:
@rafalkrupinski The point of that appendix is given in this sentence:

> Implementations of this specification MAY use an implementation of RFC6570 to perform variable expansion, however, some caveats apply.

There is no SHOULD or SHOULD NOT, just a MAY.

The authoritative requirements are given in the spec.  Sometimes they are given in terms of RFC6570, and the intent was that some aspects of this part of OpenAPI could be handled by RFC6570 implementations.  Since it's not trivial to figure out exactly what does and does not match RFC6570, this appendix goes through the more confusing areas and says _how_ you can use an RFC6570 implementation, _if_ you want to.

Nothing in this appendix changes any of the text from the spec proper.  It all flows from that text, but you have to have spent a *lot* of time with the various specifications to work it all out.

In the case of the fragment you mention, there are two different ways to specify a URL query string parameter in RFC6570:  using the `?` prefix or the `&` prefix.  Only one of them produces the _same behavior_ that the OAS _already requires._  That is `{?foo*,bar}`.  But that's not obvious because some people assume they should convert each Parameter Object to a distinct URI Template entry, which would give the `{?foo*}{&bar}` version, which does *NOT* do the same thing that the OAS already requires.

This is an appendix because it's extra information that is essential enough to go into the spec document (because we got tons of issues opened on all of these things over the years) but not _strictly_ necessary to implement the spec (because the spec proper says the same thing, if you really, really, really know how to put it all together).

So to go back to your question:  There is no SHOULD, there is a MAY.  Since we defined the OAS's behavior by referencing RFC6570, there are times when you MAY safely rely on an RFC6570 implementation _if_ you want to.  This appendix helps avoid common errors in doing so, without changing the actual OAS requirements.

rafalkrupinski:
@handrews 

I think I get it

> This example is equivalent to RFC6570's {?foo*,bar}, and NOT {?foo*}{&bar}, which is problematic because if foo is not defined, the result will be an invalid URI. The & prefix operator has no equivalent in the Parameter Object.

The `which is problematic` part refers to `{?foo*}{&bar}` and not the entire `This example is equivalent to RFC6570's {?foo*,bar}, and NOT {?foo*}{&bar}` part.

To me it reads like the entire first part is problematic, which implies that there's a problem with how would you normally interpret the OAD parameters as RFC6570. Let me add brackets to clarify.

- Your intended meaning:
This example is equivalent to RFC6570's {?foo*,bar}, **[[** and NOT {?foo*}{&bar}, which is problematic because if foo is not defined, the result will be an invalid URI **]]**.

- My interpretation:
This example is equivalent to RFC6570's {?foo*,bar}, **[[** and NOT {?foo*}{&bar} **]]**, which is problematic because if foo is not defined, the result will be an invalid URI.

If this is correct, perhaps wording like `the latter is problematic...` would help avoid confusion?

handrews:
@rafalkrupinski good point

@ralfhandl would you be willing to add this to PR #4001?

ralfhandl:
> wording like `the latter is problematic...` would help avoid confusion

Good point, done: https://github.com/OAI/OpenAPI-Specification/pull/4001/commits/439e8eb9da4194e600ad830aeb1740287d08fe14

PR #3819, opened 2024-05-19
Clarify entry document terminology (3.0.4)

handrews:
@ralfhandl caught that our phrasing here seemed to imply only one document with an OpenAPI Object, which was emphatically not the intention.  also added "entry document" as shorthand for "entry OpenAPI document" because I know I've used it that way and it reads better in some contexts.


handrews:
messed something up, will re-submit momentarily

PR #3820, opened 2024-05-19
Clarify entry/complete document terminology (3.0.4)

handrews:
@ralfhandl caught that our phrasing here seemed to imply only one document with an OpenAPI Object, which was emphatically not the intention.

Clarifies that there can be multiple complete OpenAPI documents, only one of which is an entry OpenAPI document.  The term "complete OpenAPI document" is useful in other guidance (to be tidied up in a later PR as it is currently only mentioned in 3.1.1).

Also added "entry document" as shorthand for "entry OpenAPI document" because I know I've used it that way and it reads better in some contexts.

### DIFF
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -146,7 +146,10 @@ In order to preserve the ability to round-trip between YAML and JSON formats, YA
 
 ### <a name="documentStructure"></a>OpenAPI Description Structure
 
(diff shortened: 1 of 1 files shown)


miqui (dismissed):
I pre-approve this. I deem the language here to be appropriate.

ralfhandl (dismissed):

handrews:
Conflicts merged - it needs at least one re-approval to merge because pushing the conflict resolution unset both existing approvals.  There is no change to the actual change here- there was just another change to different words on the same line.

ralfhandl (approved):

miqui (approved):

Issue #3776, opened 2024-05-02
Use URI references for Security Requirements in 3.2


handrews:
Resolving component names used in _referenced_ (as opposed to _entry_) documents is ambiguous, particularly in 3.1 where we advertise the ability to have a `components`-only document to use as a shared component library.  Historically, component names were resolved from the entry document, as reference targets were extracted from the document in which they were found without regard for the contents of the rest of the document.

In 3.1, whole-document parsing is required to properly implement the Schema Object, and there is a reasonable intuition that component names should be resolved within the document in which they appear.  This is at odds with the historical behavior, and one could argue that for Security Schemes in particular, it makes sense to treat them as part of the "deployment" aspect of things, which is arguably more relevant to the entry document.

For the Discriminator Object, the ambiguity can avoided by using the `mapping` keyword with unambiguous URI-references (meaning URI-references that are not syntactically valid as component names).  There is no similar workaround for the Security Requirement Object.

Adding a URI-reference mechanism for Security Requirements, whether by allowing `$ref` somewhere, or just allowing URI-references as keys in place of the current component names, will make it possible to write unambiguous security requirements.

handrews:
There is a similar issue with tags.  While a tag doesn't _need_ to connect to a Tag Object, it is similarly ambiguos as to which Tag Objects are "in scope", and any choice made makes it impossible to connect to whichever Tag Objects (or Security Scheme Objects) are deemed "out of scope."

I'm going to write a proposal on this to consider a couple of options holistically.

kevinswiber:
@handrews Security Requirement Objects could definitely use this. I'm actually okay with tags working this way. We could be more explicit that Tag Objects will be resolved in the order that all documents are resolved in the description and may be overwritten upon duplicate declarations. Linting can check for duplicates.

handrews:
I'm going to take tags out of this, because they have challenges that Security Requirements do not, and (as @kevinswiber notes) are not as much of a problem.  See issue #3853 (Consolidated `$ref`-to-Some Object feature request) to track further thoughts on URI-references to Tag Objects (among others).

Issue #1972, opened 2019-07-12
How do I refer to external `securitySchema` object.


yuji38kwmt:
I want to split the following file.

```yaml
openapi: "3.0.0"
info:
  version: 1.0.0
  title: Swagger Petstore
  license:
    name: MIT
servers:
  - url: http://petstore.swagger.io/v1
paths:
  /pets/{petId}:
    get:
      summary: Get Pet
      operationId: showPetById
      security:
      - Everyone: []      
      tags:
        - animals
      parameters:
        - name: petId
          in: path
          required: true
          description: The id of the pet to retrieve
          schema:
            type: string
      responses:
        '200':
          description: Expected response to a valid request
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Pet"
components:
  securitySchemes:
    Everyone:
      description: everyone
      type: "apiKey"
      name: "Authorization"
      in: "header"
  schemas:
    Pet:
      required:
        - id
        - name
      properties:
        id:
          type: integer
          format: int64
        name:
          type: string
        tag:
          type: string

```

I write  `components` object to external file.

`petstore-parent.yaml`

```yaml
openapi: "3.0.0"
info:
  version: 1.0.0
  title: Swagger Petstore
  license:
    name: MIT
servers:
  - url: http://petstore.swagger.io/v1
paths:
  /pets/{petId}:
    get:
      summary: Get Pet
      operationId: showPetById
      security:
      - Everyone: []      
      tags:
        - animals
      parameters:
        - name: petId
          in: path
          required: true
          description: The id of the pet to retrieve
          schema:
            type: string
      responses:
        '200':
          description: Expected response to a valid request
          content:
            application/json:
              schema:
                $ref: "./petstore-child.yaml#/components/schemas/Pet"
```

`petstore-child.yaml`

```
components:
  securitySchemes:
    Everyone:
      description: everyone
      type: "apiKey"
      name: "Authorization"
      in: "header"
  schemas:
    Pet:
      required:
        - id
        - name
      properties:
        id:
          type: integer
          format: int64
        name:
          type: string
        tag:
          type: string
```

But I do not know that I refer to external `securitySchema` object.

```yaml
    get:
      summary: Get Pet
      operationId: showPetById
      security:
      - Everyone: []      
```

How do I refer to external `securitySchema` object.

handrews:
[...]

PR #3821, opened 2024-05-19
Allow URI-references for Security Requirements (3.2.0)

handrews:
_**NOTE:** This is a draft because I was going to make it a proposal, but its **much** shorter to just show the PR than explain it some other way.  See PR #3823 for more context._

Fixes:
* #3776 
* #1972

This allows Security Requirement Objects to reference Security Scheme Objects by URI instead of implicit component name. Without this ability, it is difficult to share Security Schemes in a way that is consistent with re-usable component documents.

This approach provides parity with how the Discriminator Object's `mapping` field works.

### DIFF
--- a/versions/3.2.0.md
+++ b/versions/3.2.0.md
@@ -3558,7 +3558,9 @@ flows:
 #### <a name="securityRequirementObject"></a>Security Requirement Object
 
 Lists the required security schemes to execute this operation.
-The name used for each property MUST correspond to a security scheme declared in the [Security Schemes](#componentsSecuritySchemes) under the [Components Object](#componentsObject).
+The name used for each property MUST either correspond to a security scheme declared in the [Security Schemes](#componentsSecuritySchemes) under the [Components Object](#componentsObject), or be the URI of a Security Scheme Object.
+Property names that match the syntax of a component name under the Components Object MUST be treated as a component name.
+To reference a Security Scheme with a single-segment relative URI reference (e.g. `foo`), use the `.` path segment (e.g. `./foo`).
 
 Security Requirement Objects that contain multiple schemes require that all schemes MUST be satisfied for a request to be authorized.
 This enables support for scenarios where multiple query parameters or HTTP headers are required to convey security information.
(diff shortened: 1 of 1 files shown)


versions/3.2.0.md
```diff
@@ -3558,7 +3558,9 @@ flows:
 #### <a name="securityRequirementObject"></a>Security Requirement Object
 
 Lists the required security schemes to execute this operation.
-The name used for each property MUST correspond to a security scheme declared in the [Security Schemes](#componentsSecuritySchemes) under the [Components Object](#componentsObject).
+The name used for each property MUST either correspond to a security scheme declared in the [Security Schemes](#componentsSecuritySchemes) under the [Components Object](#componentsObject), or be the URI of a Security Scheme Object.
+Property names that match the syntax of a component name under the Components Object MUST be treated as a component name.
```
ralfhandl, line 3562:
```suggestion
Property names identical to a component name under the Components Object MUST be treated as a component name.
```

ralfhandl (commented):
+1 with minor nit

versions/3.2.0.md
handrews: @ralfhandl this word choice was intentional to ensure that any given string is unambiguously resolved as either a component or a URI, and cannot change its resolution based on whether a component exists (that's a problem with `mapping` today, which we actually considered fixing in a PR before deciding that we would not make breaking changes, even small ones, in 3.2 and closed the PR).

The next line shows how to use a URI that would, in the most normalized form, be interpeted as a component name.

This is also something of a a compatibility thing to me (although I'll grant this point is debatable)- keys that always and only resolved as component names in 3.1 should continue to always and only resolve that way.  Only previously illegal keys should have the new behavior.

ralfhandl: Ah, so any property name matching `^[a-zA-Z0-9\.\-_]+$` MUST be treated as a security scheme name, and processing MUST fail (or use a default/fallback security scheme) if there is nothing inside `components.securitySchemes` with this name?

handrews: @ralfhandl correct.  If there is default/fallback behavior then we definitely need to keep the syntax partitioned so that still works (it shouldn't switch to trying to treat the component name as a path segment, because if it did and by some chance there happened to be a file of that name in the same directory, then when you updated to 3.2 it would start referencing that file instead of falling back).

ralfhandl (approved):

miqui (approved):
@handrews looks good

ralfhandl:
Looking forward to the next PR 😄

Issue #3591, opened 2024-02-18
How to find discriminator targets


floitsch:
I just implemented a JSON-Schema validator and would like to add OpenAPI support to it.

I'm now stuck at adding discriminators with the "allOf" variant.
According to the spec a "discriminator MAY be added to a parent schema definition, and all schemas comprising the parent schema in an allOf construct may be used as an alternate schema."

However, I can't find the definition of "parent". The term "parent" in JSON Schema seems to mean something unrelated in JSON Schema, and I didn't see anything in the OpenAPI spec. I'm assuming it is the "parent" in the inheritance hierarchy, but that's not something that is well defined in JSON Schema (afaik).

Also: I can't just do this dynamically (waiting to encounter an discriminator), as I need to have the mappings targets statically. After, all, I might get the generic schema as "type" and then use the discriminator to validate the more specialized one.

There are a few cases that are obvious:
``` yaml
    Dog:
      allOf:
      - $ref: '#/components/schemas/Pet'
...
```

According to https://github.com/OAI/OpenAPI-Specification/issues/2165#issuecomment-592957945 an `allOf` of an `allOf` is also a valid target.

``` yaml
components:
  schemas:
    Pet:
      type: object
      required:
        - petType
      properties:
        petType:
          type: string
      discriminator:
        propertyName: petType
    Reptile:
      allOf:
        - $ref: '#/components/schemas/Pet'
    Lizard:
      allOf:
        - $ref: '#/components/schemas/Reptile'
        - type: object
          properties:
            lovesRocks:
              type: boolean
```

What about a reference in an `if`?
``` yaml
components:
  schemas:
    Pet:
      type: object
      required:
        - petType
      properties:
        petType:
          type: string
      discriminator:
        propertyName: petType
    Reptile:
      if: true  # Or some complex schema.
      then:
        allOf:
          - $ref: '#/components/schemas/Pet'
```
I'm guessing that one isn't valid, but my gut instinct would have said the same for the `allOf` of the `allOf`.

There are also some interesting cases:
If I have a schema in a different schema-resource (for example 'https://gigantic-server.com/schemas/Monster') that references my discriminator schema with "allOf", then clearly I can't find that schema. However, what if that same schema-resource is loaded through other means?
Does that schema then count as a target for the mapping?

Roughly:
``` yaml
# Let's give this the URI "URI-DISCRIMINATOR"
Pet:
  type: object
  required:
    - petType
  properties:
    petType:
      type: string
[...]

Issue #3424, opened 2023-10-30
Clarification on `discriminator` usage on `allOf` schemas.


iglosiggio:
Hi, the spec says:
> The discriminator object is legal only when using one of the composite keywords `oneOf`, `anyOf`, `allOf`.

What's the meaning behind something like this?
```yaml
components:
  schemas:
    Cat:
      allOf:
      - $ref: '#/components/schemas/Pet'
      - $ref: '#/components/schemas/Furry'
      - $ref: '#/components/schemas/Cute'
      discriminator:
        propertyName: "petType"
    # ... descriptions of other schemas
```

I don't see an obvious intepretation for an schema that has both `discriminator` and `allOf`.

karenetheridge:
`allOf` (or `anyOf` or `oneOf`) are expected, but they don't have to be at the same level -- there is one example in the spec at the same level and one where they are in separate definitions. The schema that requires them together is a mistake - see https://github.com/OAI/OpenAPI-Specification/pull/3405.

handrews:
I think the `oneOf` or `anyOf` need to be in the same schema object as `discriminator`, but the `allOf` use case is different and is always in a different schema object.  I think.

LasneF:
is there a use case of having discriminator and allof ? 

to me discriminator should only be use for the oneOf  the usage of it inside an anyOf beeing quite alredy limited/corner case

discriminator could even be dropped 
pushing that  schema with a oneOf must be completly exclusive by nature (ie a content cannot match 2 schema)

iglosiggio:
> The schema that requires them together is a mistake - see #3405.

Sorry but what does _"The discriminator object is legal only when using one of the composite keywords"_ means then?

handrews:
@LasneF OAS 4.0 a.k.a. Moonwalk is [definitely separating concerns more thoroughly](https://www.openapis.org/blog/2023/12/06/openapi-moonwalk-2024), including better separation between content schema formats and OAS "proper".  There aren't any specifics on what that means yet, but it will be addressed in the coming year (note that there is a separate OAI/Moonwalk repository for that project).

Doing any such thing would be too big of a change for a 3.1.x, or even a 3.2, although of course in 3.1 JSON Schema is at least theoretically extensible with 3rd-party keywords.

handrews:
PRs merged!  closing.

Issue #2520, opened 2021-04-05
Discriminator: how properly detect valid reference


Logioniz:
Suppose we have such OpenAPI document which served by location http://example.com/openapi/desc

```yaml
servers:
    - url: 'http://example.com/'
components:
    schemas:
        Pet:
          type: object
          required:
          - petType
          properties:
            petType:
              type: string
          discriminator:
            propertyName: petType
            mapping:
              dog: Dog
        Cat:
          allOf:
          - $ref: '#/components/schemas/Pet'
          - type: object
            # all other properties specific to a `Cat`
            properties:
              name:
                type: string
        Dog:
          allOf:
          - $ref: '#/components/schemas/Pet'
          - type: object
            # all other properties specific to a `Dog`
            properties:
              bark:
                type: string
        Lizard:
          allOf:
          - $ref: '#/components/schemas/Pet'
          - type: object
            # all other properties specific to a `Lizard`
            properties:
              lovesRocks:
                type: boolean

```

it is not clear what mean Dog in the mapping subscheme.
Variants:
1. http://example.com/openapi/Dog (accroding to base uri http://example.com/openapi/desc)
2. #/components/schemas/Dog

If i have subschema in another url (http://example.com/openapi/Dog) thеn Dog is relative link.
If i have subschema in components thеn Dog is #/components/schemas/Dog (which is also a relative link, but different)

How to correctly determine that Dog is a relative link or subschema in #/components/schema?

Sorry for bad english.

Logioniz:
Any recommendations?
Not sure how to properly handle this kind of situation.

jdesrosiers:
My understanding is that the values of `mapping` are _not_ a reference. The value is used to construct references such as `#/components/schemas/Dog`.

jdesrosiers:
(Note: I deleted my previous wrong response)

I believe is example has a bug. The values of `mapping` are URIs that must match the URI of one of the schemas being discriminated. I don't think OpenAPI has a concept of an identifier the way JSON Schema does. In that case, `Dog` doesn't have anything to resolve against and that is an invalid reference. If OpenAPI does have a concept of an identifier, then it would resolve to `http://example.com/openapi/Dog`, which doesn't exist. Constructing `#/components/schemas/Dog` should only happen when auto mapping, not with `mapping`.

The discriminator in the example should be,

```yaml
discriminator:
  propertyName: petType
  mapping:
    dog: '#/components/Dog'
```

[...]

PR #3822, opened 2024-05-19
Clarify discriminator "*Of" and "mapping" usage (3.0.4)

handrews:
Fixes:
* #2520 
* #3424 
* #3591 

This moves some guidance up to the fixed fields section where it is more obvious, makes it clear that `allOf` is used differently from `anyOf`/`oneOf`, and explicitly designates other configurations as having undefined behavior.  It also reworks the somewhat cryptic statement about "inline" schemas, and provides guidance around ambiguous `mapping` values.

### DIFF
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -2752,7 +2752,7 @@ However, it is RECOMMENDED to search all named schemas in the Components Object 
 The behavior of any configuration of `oneOf`, `anyOf`, `allOf` and `discriminator` that is not described above is undefined.
 
 ##### Options for Mapping Values to Schemas
-The value of the property named in `propertyName` is is used as the name of the associated schema under the [Components Object](#componentsObject), _unless_ a `mapping` is present for that value.
+The value of the property named in `propertyName` is used as the name of the associated schema under the [Components Object](#componentsObject), _unless_ a `mapping` is present for that value.
 The `mapping` entry maps a specific property value to either a different schema component name, or to a schema identified by a URI.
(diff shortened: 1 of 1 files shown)


ralfhandl (dismissed):
Very nice!

versions/3.0.4.md
```diff
@@ -2734,17 +2734,34 @@ components:
...
+It is implementation-defined as to whether all named [Schema Objects](#schemaObject) under the [Components Object](#componentsObject), or only those that are otherwise directly referenced are searched for `allOf` references to the parent schema.
+However, it is RECOMMENDED to search all named schemas in the Components Object because it is common with the `allOf` usage for other parts of the API to only directly reference the parent schema.
+
+The behavior of any configuration of `oneOf`, `anyOf`, `allOf` and `discriminator` that is not described above is undefined.
 
-The discriminator object is legal only when using one of the composite keywords `oneOf`, `anyOf`, `allOf`.  Note that because the discriminating property's value is used as a component name and/or as the key in the `mapping` object, the behavior of any value that is not a string is undefined.
+##### Options for Mapping Values to Schemas
+The value of the property named in `propertyName` is is used as the name of the associated schema under the [Components Object](#componentsObject), _unless_ a `mapping` is present for that value.
```
mikekistler, line 2755:
```suggestion
The value of the property named in `propertyName` is used as the name of the associated schema under the [Components Object](#componentsObject), _unless_ a `mapping` is present for that value.
```

mikekistler (approved):
One little nit, otherwise Looks good! 👍

ralfhandl (approved):

PR #3825, opened 2024-05-20
Jekyll does not load variables in front-matter

Bellangelo:
Jekyll does not support variables in the front-matter. After the merge of this [PR](https://github.com/OAI/OpenAPI-Specification/pull/3799) this is how it is shown:
image

Based on this https://stackoverflow.com/questions/10218779/custom-variables-in-jekyll-front-matter we can pass variables to the included file.
Documentation: https://jekyllrb.com/docs/includes/

### DIFF
--- a/_includes/alternative-schema-entry.md
+++ b/_includes/alternative-schema-entry.md
@@ -2,7 +2,7 @@
 
 ## {{ page.slug }} - {{ page.description }}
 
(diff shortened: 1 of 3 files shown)


handrews (approved):
Thanks- My apologies for not looking at the right page post-merge to make sure it worked!

PR #3826, opened 2024-05-20
Capture allows the usage of curly braces

Bellangelo:
In this [PR](https://github.com/OAI/OpenAPI-Specification/pull/3825) we tried to bypass the issue that Jekyll does not support variables in the front-matter. What we did instead was to break the syntax, once more: https://github.com/OAI/OpenAPI-Specification/actions/runs/9163554322/job/25192911220.

The documentation even proposes the syntax but i missed it originally ( face-palm ): https://jekyllrb.com/docs/includes/#passing-parameter-variables-to-includes

You can check the rendered results here: https://bellangelo.github.io/OpenAPI-Specification/registry/alternative-schema/xmlSchema.html

### DIFF
--- a/registries/_alternative-schema/jsonSchema.md
+++ b/registries/_alternative-schema/jsonSchema.md
@@ -5,4 +5,8 @@ description: JSON Schema
 layout: default
 ---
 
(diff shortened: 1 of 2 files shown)


handrews (approved):
Thanks for the explanation - Jekyll always makes my head hurt, and its just as much on me for missing these things while looking over the changes.  But this is a good low-impact page to experiment with!

PR #3212, opened 2023-03-23
[gh-pages] reskin registry site and prep for release

MikeRalphson:
As discussed on the TDC call 16th March 2023, drop the 'draft' status from the registry and hide the extensions registry (not the extensions namespace registry).

As discussed with @oai/TSC this PR also re-skins the registry site with the same theme as the documentation site for consistency, visual appeal, accessiblity and the presence of a ToC and search bar.

The preview can be seen [here](https://mikeralphson.github.io/OpenAPI-Specification/registry/index.html)

PR #3828, opened 2024-05-20
Encapsulate page layout for extensions

Bellangelo:

This [commit](https://github.com/OAI/OpenAPI-Specification/commit/ccbf6a835d78d1bd44b9c21f5820f61437c30849) re-enables the page building of the extension which was disabled in https://github.com/OAI/OpenAPI-Specification/pull/3212.

I have no idea why it was disabled but in production it produces a 404 link. You can check it here https://spec.openapis.org/registry/extension/index.html by clicking on the `x-twitter` link.

A rendered result can be seen here: https://bellangelo.github.io/OpenAPI-Specification/registry/extension/x-twitter.html

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -20,8 +20,7 @@ collections:
   extension:
     slug: extension
     name: Specification Extension Registry
-    output: false
-    hidden: true
+    output: true
     permalink: /registry/:collection/:title
   draft-feature:
     slug: draft-feature
--- a/_includes/extension-entry.md
+++ b/_includes/extension-entry.md
@@ -1,0 +1,16 @@
+# <a href="..">{{ page.collection }}</a>
+
+## {{ page.slug }} - {{ page.description }}
+
+{{ include.summary }}
+
+### Schema
+
+```yaml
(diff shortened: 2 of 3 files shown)


handrews:
@OAI/tsc based on #3212 there was a conscious decision to hide this registry- anyone recall why this was done and whether we should bring it back or just delete it?

ralfhandl:
As far as I can recall the discussion, the extension registry is practically not used and its only entry may already be obsolete. Better keep it hidden, and also hide the menu entry for it in the left side bar:
image

Bellangelo:
> As far as I can recall the discussion, the extension registry is practically not used and its only entry may already be obsolete. Better keep it hidden, and also hide the menu entry for it in the left side bar:

Thanks for your input @ralfhandl, should we delete then all together? I don't see why we should keep it if it hidden. If anyone wants to "re-enable" it later on, he can just look at the git history.

lornajane (approved):
Thanks for finding and rescuing this

handrews (approved):

PR #3830, opened 2024-05-20
Encapsulate page layout for formats - Part 1

Bellangelo:

A rendered result can be seen here: https://bellangelo.github.io/OpenAPI-Specification/registry/format/index.html

This is the part 1 of a series of PRs to address the formats.

### DIFF
--- a/_includes/format-entry.md
+++ b/_includes/format-entry.md
@@ -2,6 +2,8 @@
 
 ## {{ page.slug }} - {{ page.description }}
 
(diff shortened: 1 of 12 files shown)


handrews (approved):
Looks great!  I reviewed all of the renderings and noticed that this fixed a bug in `decimal` where it was showing the issue number (again, as a non-link) instead of remarks.  So this is already proving its value by fixing copy-paste bugs!

PR #3831, opened 2024-05-21
Encapsulate page layout for formats - Part 2

Bellangelo:

A rendered result can be seen here: https://bellangelo.github.io/OpenAPI-Specification/registry/format/index.html

This is the part 2 of a series of PRs to address the formats.

Part 1: https://github.com/OAI/OpenAPI-Specification/pull/3830

### DIFF
--- a/_includes/format-entry.md
+++ b/_includes/format-entry.md
@@ -2,7 +2,7 @@
 
 ## {{ page.slug }} - {{ page.description }}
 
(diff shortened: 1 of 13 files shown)


handrews (approved):
Looks great and (based on comparing each rendered page) fixes another bug - a "numberstring" that was missed before.  🎉

PR #3837, opened 2024-05-22
feat(spec): codify Encoding in specific requestBody mediatypes

jeremyfiel:
the specification indicates
> The encoding object SHALL only apply to requestBody objects when the media type is multipart or application/x-www-form-urlencoded.

This PR codifies that behavior in the JSON Schema to enable validation.

Per the [Development.md](https://github.com/OAI/OpenAPI-Specification/blob/main/DEVELOPMENT.md#changing-the-schemas), this change was already merged to the Specification on [this PR](https://github.com/OAI/OpenAPI-Specification/pull/3724)

### DIFF
--- a/schemas/v3.0/schema.yaml
+++ b/schemas/v3.0/schema.yaml
@@ -405,15 +405,35 @@ definitions:
           oneOf:
             - $ref: '#/definitions/Example'
             - $ref: '#/definitions/Reference'
-      encoding:
-        type: object
-        additionalProperties:
-          $ref: '#/definitions/Encoding'
     patternProperties:
       '^x-': {}
     additionalProperties: false
     allOf:
       - $ref: '#/definitions/ExampleXORExamples'
+
+    MediaTypeWithEncoding:
+      type: object
+      properties:
+        schema:
+          oneOf:
+            - $ref: '#/definitions/Schema'
+            - $ref: '#/definitions/Reference'
+        example: {}
+        examples:
+          type: object
+          additionalProperties:
+            oneOf:
+              - $ref: '#/definitions/Example'
+              - $ref: '#/definitions/Reference'
+        encoding:
+          type: object
+          additionalProperties:
+            $ref: '#/definitions/Encoding'
+      patternProperties:
+        '^x-': {}
+      additionalProperties: false
+      allOf:
+        - $ref: '#/definitions/ExampleXORExamples'
 
   Example:
     type: object
@@ -763,6 +783,9 @@ definitions:
         type: object
         additionalProperties:
           $ref: '#/definitions/MediaType'
+        patternProperties:
+          '^application/x-www-form-urlencoded$|^multipart\/.+$':
+            $ref: '#/definitions/MediaTypeWithEncoding'
       required:
         type: boolean
         default: false
--- a/schemas/v3.0/schema.yaml
+++ b/schemas/v3.0/schema.yaml
@@ -784,7 +784,7 @@ definitions:
         additionalProperties:
           $ref: '#/definitions/MediaType'
         patternProperties:
-          '\*\/\*|^application/x-www-form-urlencoded$|^multipart\/.+$':
+          '^\*\/\*$|^application/x-www-form-urlencoded$|^multipart\/.+$':
             $ref: '#/definitions/MediaTypeWithEncoding'
       required:
         type: boolean
--- a/schemas/v3.0/schema.yaml
+++ b/schemas/v3.0/schema.yaml
@@ -784,7 +784,7 @@ definitions:
         additionalProperties:
           $ref: '#/definitions/MediaType'
         patternProperties:
-          '^application/x-www-form-urlencoded$|^multipart\/.+$':
+          '\*\/\*|^application/x-www-form-urlencoded$|^multipart\/.+$':
             $ref: '#/definitions/MediaTypeWithEncoding'
       required:
         type: boolean
--- a/schemas/v3.1/schema.yaml
+++ b/schemas/v3.1/schema.yaml
@@ -480,7 +480,7 @@ $defs:
     propertyNames:
       format: media-range
 
-  media-type:
+  media-type-common:
     $comment: https://spec.openapis.org/oas/v3.1.0#media-type-object
     type: object
     properties:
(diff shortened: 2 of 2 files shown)


handrews:
Also paging @karenetheridge

schemas/v3.1/schema.yaml
```diff
@@ -474,10 +474,24 @@ $defs:
...
+    type: object
+    properties:
+      schema:
+        $dynamicRef: '#meta'
+    allOf:
+      - $ref: '#/$defs/specification-extensions'
+      - $ref: '#/$defs/examples'
+    unevaluatedProperties: false
+
+  media-type-with-encoding:
```
handrews, line 494: I would factor out everything from `media-type` _except_ `unevaluatedProperties` into a `media-type-common`, and then `$ref: "#/$defs/media-type-common"` from both `media-type` and `media-type-with-encoding`.  Both would then have `unevaluatedProperties: false` and `media-type-with-encoding` would have a `properties` keyword adding the `encoding` property, but no duplication of `type` or the other properties or `allOf`s.

schemas/v3.0/schema.yaml
```diff
@@ -763,6 +783,9 @@ definitions:
         type: object
         additionalProperties:
           $ref: '#/definitions/MediaType'
+        patternProperties:
+          '^application/x-www-form-urlencoded$|^multipart\/.+$':
+            $ref: '#/definitions/MediaTypeWithEncoding'
```
handrews, line 788: Nice!

handrews (commented):
I've suggested a change although if you feel strongly about it it's not a hill I care to die on :-)

schemas/v3.1/schema.yaml
jeremyfiel: do you mean like this? 

```yaml
media-type-common:
   $comment: https://spec.openapis.org/oas/v3.1.0#media-type-object
    type: object
    properties:
      schema:
        $dynamicRef: '#meta'
    allOf:
      - $ref: '#/$defs/specification-extensions'
      - $ref: '#/$defs/examples'

 media-type:
   $comment: https://spec.openapis.org/oas/v3.1.0#media-type-object
   $ref: '#/$defs/media-type-common'
   unevaluatedProperties: false 

media-type-with-encoding:
   $comment: https://spec.openapis.org/oas/v3.1.0#media-type-object
    $ref: '#/$defs/media-type-common'
    type: object
    properties:
      encoding:
        $ref: '#/$defs/encoding-object'
    unevaluatedProperties: false
```

jeremyfiel: I did consider something like this just to avoid updating media-type in two places.

handrews: Yup, like that!  If we have similar places where this kind of `$comment` gets replicated, then definitely replicate it like you are now.  Otherwise I'd be willing to bet that you can just put it in the common schema.  Although I'm not sure if anything is relying on them - you're not supposed to programmatically rely on `$comment` in general.

darrelmiller:
Minor concern that this would prevent `*/*` as a media type.  Generally supportive of this addition.

handrews:
@OAI/tsc I think what is needed here is a policy on whether field combinations that are said to be _ignored_ or to _not apply_ can be forbidden by the schema or not.  That would help us decide a lot of schema design issues, rather than having to re-think each proposed restriction.  There are quite a few places in the spec with this kind of language, and we should treat them consistently in the schemas.

I'd be fine with either way the TSC wants to decide this - if it is decided that ignored/non-applicable things can be forbidden in the schema, then I'm fine with this PR being merged.

schemas/v3.0/schema.yaml
```diff
@@ -763,6 +783,9 @@ definitions:
         type: object
         additionalProperties:
           $ref: '#/definitions/MediaType'
+        patternProperties:
+          '\*\/\*|^application/x-www-form-urlencoded$|^multipart\/.+$':
```
notEthan, line 787: Should `*/*` have `^`/`$` anchors?
Should `application/*` be matched, as that would apply to `application/x-www-form-urlencoded`?

jeremyfiel: what is the purpose of the `/s` anchors?  Why would it be needed? 

`application/*` would match any `application` media type and that's _not_ what we want here., although that would enable quite a few different media types to allow `encoding` schema.   Matching `*/*` was only for backwards compatibility per Darrell's comment.

mikekistler:
@handrews on this point:
> @OAI/tsc I think what is needed here is a policy on whether field combinations that are said to be ignored or to not apply can be forbidden by the schema or not.

My opinion is that if the spec says a field combination **must** be ignored or not apply, then we can make these forbidden by the schema.

mikekistler:
Following up here ... what I see in the v3.0.3 spec is:
> The encoding object SHALL only apply to requestBody objects when the media type is multipart or application/x-www-form-urlencoded.

Since the spec uses "SHALL" and not "MUST", and my prior comment, I think my view is that we should _NOT_ make this change to the schema, at least for v3.0. Do you agree @jeremyfiel ?

ralfhandl:
> Since the spec uses "SHALL" and not "MUST"

Note that [RFC2119 does not differentiate between these two words](https://www.rfc-editor.org/rfc/rfc2119):

> 1. MUST   This word, or the terms "REQUIRED" or "SHALL", mean that the
     definition is an absolute requirement of the specification.

We cannot base any decisions on this stylistic choice.

mikekistler (approved):
Looks good. 👍

jeremyfiel:
Thanks Ralf and Mike for reviewing and your comments.

handrews:
@mikekistler MUST be ignored is not the same as forbidden.

I don't think we can forbid anything in the schema on the grounds that it is to be ignored.  Those are two different things.

Again, if we want to supply _two_ schemas, one that _only_ constrains what is required by the OAS, and one that has _additional_ constraints that help avoid pitfalls, I think that would be a great idea.

But I think it is incorrect to over-constrain the schema if it is the only schema.  (I realize this is a bit different from what I said earlier, but there are a whole bunch of "MUST be ignored" in the OAS, and if we start changing the schema based on that, that is a huge amount of change, and I think an incorrect change).

handrews (changes requested):
Regrettably, I don't think we can forbid things that are mandated to be ignored.  Not if we only have one schema.  I would support a more "helpful" schema that added more constraints, though.

handrews:
Based on today's TDC call, I'm going to close this for the same reasons as #4069 and #4070.  Technically, the `encoding` field is allowed _but ignored_ in these circumstances, so it cannot be forbidden in the validation schema.  But as @lornajane noted when closing those PRs, there is likely to be a home for this in some sort of separate linting schema or schemas:

> Please keep an eye on the project for more discussion about linting schemas in addition to our validation schemas.

Issue #2100, opened 2020-01-10
Figuring out $ref and $id in OAS v3.1 in general


philsturgeon:
@darrelmiller 

> Is it expected that two distinct OAS Schema Objects would be able to refer to each other using canonical Ids?

yes

> If the above example is possible, then is it also expected that $ref objects in other parts of the OAS document can use the canonical Id?

yes

> This example uses the standard OAS way of referencing a component. However, it ignores the fact that $Id has been defined. I am guessing that this is still fine.

More or less.  Technically, if the OpenAPI document were considered a JSON Schema, then the behavior is undefined.  But particularly for landing directly on the schema object with an `$id`, I suspect in practice most implementations will support it.  It was (or ambiguously appeared to be) required in the last few drafts.

Also, I think it would be fine if OAS had its own rules for things directly under `/components/schemas`.  Although I guess that would actually be more complicated for tooling vendors.

_Originally posted by @handrews in https://github.com/OAI/OpenAPI-Specification/pull/1977#issuecomment-571437344_

philsturgeon:
Duplicate of https://github.com/OAI/OpenAPI-Specification/issues/2092

PR #3838, opened 2024-05-22
Clarify "URI" and "URL" meaning and usage (3.1.1)

handrews:
***NOTE 1:** This is not as long as it looks!  After the first block, it's just a word here and a word there to make things consistent throughout the spec.*

***NOTE 2:** This is not at all relevant to 3.0.4*

***NOTE 3:** This does not address links in CommonMark text, which is being discussed in issue #1701*

Fixes:
* #3545 

I went through many iterations on this, some of which went into much more detail.  In the end I decided to do just a few things:

* State an overall criteria for URLs vs URIs rather than just mentioning a few fields or trying to include a complete list:
    * API interaction involves URLs (by definition, really)
    * Connecting pieces of the API description involves identifier-not-necessarily-locator URIs (because JSON Schema requires it, and the text _mostly_ uses "URI" including for the Reference Object and `operationRef`)
* Make sure all fields with URI behavior are described as URIs and not URLs
    * This includes using less location-oriented phrasing in several places
    * The Discriminator Object's `mapping` field gets a URI description in PR #3822, so it's not in this PR 
* Add minimal commentary on URI behavior
    * Reference the "Document Parsing" section which mentions configuration options for identifier-base loading
    * Briefly mention key use cases as this can feel like a puzzling or overly-academic requirement
    * It's worth noting that several tools actually allow "faking" the document location already

-----

_For those who want the full backstory (everyone else can skip the rest of this)_

The most likely point of controversy would be removing "location of" from "identifies the location of the value being referenced" in the Reference Object.  The "identifies the location" phrasing was copied over from the (not cited in 3.1) JSON Reference draft, and indicates location-based (URL) behavior.  However, the new wording added in 3.1 calls the Reference Object's `$ref` a URI rather than a URL, and includes the Reference Object in the list in 3.1.0's URI section.

The history of this apparent contradiction is rather muddled:

* Jan 2020: I assert that [reference objects can be treated as URIs rather than URLs](https://github.com/OAI/OpenAPI-Specification/issues/1887#issuecomment-576487839) and no one seems to object
* Feb 2020: Later in the same issue, the [TSC states](https://github.com/OAI/OpenAPI-Specification/issues/1887#issuecomment-584435639) that "identifies the location" means URL behavior in 3.0, but in 3.1 that wording is no longer used
* March 2020: The TSC endorses URI (`$id`-based) behavior in the Schema Object, location behavior elsewhere, and pledges to ["re-review this issue holistically"](https://github.com/OAI/OpenAPI-Specification/issues/2159#issuecomment-601283402)
* Jan. 2021: A TSC member introduces the distinction between [document authoring and api operation](https://github.com/OAI/OpenAPI-Specification/pull/2437), which I actually didn't notice until just now, although I did notice the [discussion involving two TSC Members and a longtime contributor](https://github.com/OAI/OpenAPI-Specification/pull/2437#discussion_r554491968) stating that `externalValue` is a URI and evaluated as an identifier rather than a locator; however, this is also where the "identifies the location" wording was pulled in from the no-longer-cited JSON Reference spec
* There are several issues (#2092, #2099, #2100) that might have been tracking the "holistic review" that never got resolved in a clear way - some were just closed for reasons that aren't clear, or are pointed to from elsewhere as tracking the solution but not actually saying anything about it.  I really cannot figure out quite what happened.

My take on this is that the general URI/URL split, and the clear presence of the Reference Object in the URI section, means that it has URI behavior, and the "identifies the location" wording is something that never got taken out after that was resolved.  So I'm taking it out now, because different parts of the spec treat it differently, and URI seems more consistent with more parts of the spec.

### DIFF
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -248,7 +248,6 @@ OAS < 3.1 | OAS 3.1 | Comments
 ### <a name="richText"></a>Rich Text Formatting
 Throughout the specification `description` fields are noted as supporting CommonMark markdown formatting.
 Where OpenAPI tooling renders rich text it MUST support, at a minimum, markdown syntax as described by [CommonMark 0.27](https://spec.commonmark.org/0.27/). Tooling MAY choose to ignore some CommonMark features to address security concerns. 
-Hyperlinks in CommonMark text are considered to be [#relativeReferencesURI](API description URIs).
 
 ### <a name="relativeReferencesURI"></a>Relative References in API Description URIs
 
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -262,7 +262,7 @@ Unless specified otherwise, all properties that are URIs MAY be relative referen
 
 Relative references in [`Schema Objects`](#schemaObject), including any that appear as `$id` values, use the nearest parent `$id` as a Base URI, as described by [JSON Schema Specification Draft 2020-12](https://tools.ietf.org/html/draft-bhutton-json-schema-00#section-8.2).
 
-Relative URI references in other Objects, and in Schema Objects where no parent schema contains an `$id`, MUST resolved using the referring document's base URI, which is determined in accordance with [RFC3986 §5.1.2 – 5.1.4](https://tools.ietf.org/html/rfc3986#section-5.1.2).
+Relative URI references in other Objects, and in Schema Objects where no parent schema contains an `$id`, MUST be resolved using the referring document's base URI, which is determined in accordance with [RFC3986 §5.1.2 – 5.1.4](https://tools.ietf.org/html/rfc3986#section-5.1.2).
 In practice, this is usually the retrieval URI of the document, which MAY be determined based on either its current actual location or a user-supplied expected location.
 
 If a URI contains a fragment identifier, then the fragment should be resolved per the fragment resolution mechanism of the referenced document.  If the representation of the referenced document is JSON or YAML, then the fragment identifier SHOULD be interpreted as a JSON-Pointer as per [RFC6901](https://tools.ietf.org/html/rfc6901).
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -248,18 +248,28 @@ OAS < 3.1 | OAS 3.1 | Comments
 ### <a name="richText"></a>Rich Text Formatting
 Throughout the specification `description` fields are noted as supporting CommonMark markdown formatting.
 Where OpenAPI tooling renders rich text it MUST support, at a minimum, markdown syntax as described by [CommonMark 0.27](https://spec.commonmark.org/0.27/). Tooling MAY choose to ignore some CommonMark features to address security concerns. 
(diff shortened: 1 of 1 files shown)


versions/3.1.1.md
```diff
@@ -248,18 +248,28 @@ OAS < 3.1 | OAS 3.1 | Comments
 ### <a name="richText"></a>Rich Text Formatting
 Throughout the specification `description` fields are noted as supporting CommonMark markdown formatting.
 Where OpenAPI tooling renders rich text it MUST support, at a minimum, markdown syntax as described by [CommonMark 0.27](https://spec.commonmark.org/0.27/). Tooling MAY choose to ignore some CommonMark features to address security concerns. 
+Hyperlinks in CommonMark text are considered to be [#relativeReferencesURI](API description URIs).
```
ralfhandl, line 251: This link looks wrong. Should it be something like
```suggestion
Hyperlinks in CommonMark text are considered to be [relative references](#relativeReferencesURI).
```

versions/3.1.1.md
```diff
@@ -248,18 +248,28 @@ OAS < 3.1 | OAS 3.1 | Comments
...
 
 Unless specified otherwise, all properties that are URIs MAY be relative references as defined by [RFC3986](https://tools.ietf.org/html/rfc3986#section-4.2).
 
-Relative references, including those in [`Reference Objects`](#referenceObject), [`PathItem Object`](#pathItemObject) `$ref` fields, [`Link Object`](#linkObject) `operationRef` fields and [`Example Object`](#exampleObject) `externalValue` fields, are resolved using the referring document as the Base URI according to [RFC3986](https://tools.ietf.org/html/rfc3986#section-5.2).
+Relative references in [`Schema Objects`](#schemaObject), including any that appear as `$id` values, use the nearest parent `$id` as a Base URI, as described by [JSON Schema Specification Draft 2020-12](https://tools.ietf.org/html/draft-bhutton-json-schema-00#section-8.2).
+
+Relative URI references in other Objects, and in Schema Objects where no parent schema contains an `$id`, MUST resolved using the referring document's base URI, which is determined in accordance with [RFC3986 §5.1.2 – 5.1.4](https://tools.ietf.org/html/rfc3986#section-5.1.2).
```
ralfhandl, line 265:
```suggestion
Relative URI references in other Objects, and in Schema Objects where no parent schema contains an `$id`, MUST be resolved using the referring document's base URI, which is determined in accordance with [RFC3986 §5.1.2 – 5.1.4](https://tools.ietf.org/html/rfc3986#section-5.1.2).
```

ralfhandl (changes requested):
Minor nits

versions/3.1.1.md
handrews: @ralfhandl thanks for bringing this up, I had to think about it a bit because it's not clear to me how this interacts with rendering the Markdown.  I'm going to take this part out and continue the discussion in issue #1701, which I will take off of the "fixed" list for this PR.

ralfhandl (approved):

lornajane (approved):
Thanks!

PR #3839, opened 2024-05-22
Clarify entry/complete document terminology (3.1.1 of #3820)

handrews:
Ports:
* #3820 

_[Note that there is no port to 3.2.0 yet because I'm going to batch up all of the document structure / referencing commits into one PR – with separate commits – for convenience]_

Clarifies that there can be multiple complete OpenAPI documents, only one of which is an entry OpenAPI document.


ralfhandl (approved):

lornajane (approved):
Thanks for the port!

Issue #2819, opened 2021-12-06
should Cookie headers be ignored by 'parameters'?


karenetheridge:
In https://spec.openapis.org/oas/v3.1.0#fixed-fields-9: "If in is "header" and the name field is "Accept", "Content-Type" or "Authorization", the parameter definition SHALL be ignored."  Should this also include "Cookie" and/or "Set-Cookie", since these are handled by the 'cookie' parameter type?

Similarly for the header object at https://spec.openapis.org/oas/v3.1.0#header-object (and I noticed that "Accept", "Content-Type" or "Authorization" are not excluded there, and probably should be).

handrews:
I think for the Header Object the "follows the structure of the Parameter Object" is sufficient, because the forbidden names are in the field definition.  But we should say something about Cookie.  I'm going with its behavior being undefined (for compatibility purposes - we can't make it illegal).

handrews:
I think it's possible to describe `Set-Cookie` in a Header Object as it is a response header and its use is discussed in issue #1237, so I'm just going to make a note about `Cookie`.

handrews:
PR merged for 3.0.4 and ported to 3.1.1 via PR #3921!
This is addressed by the new Appendix D.

PR #3841, opened 2024-05-23
Warnings for header and cookie serialization (3.0.4)

handrews:
***Rendering with all currently open 3.0.4 changes (as of the timestamp of this edit - use the dropdown on the word "edited" in the bar for this comment) can be [found here](https://handrews.github.io/renderings/oas/deploy/oas/v3.0.4.html#appendix-d-serializing-headers-and-cookies).***

Fixes:
* #1528 
* #2819 
* #2940 (the cookie part)
* plus other things no one noticed

_This is Appendix D because other open PRs add B and C, I'll sort it all out when resolving merge conflicts._

This makes serializing cookie paramters and most header parameters with `schema` and `style` NOT RECOMMENDED.
It is not clear that any `schema`-based serialization for cookies will produce a correct value (although the reason is sufficiently obscure that many implementations might ignore it and produce cookie-compliant output anyway).
With headers, there are numerous pitfalls and only the simplest scenarios will work properly, although perhaps the warning here could be reworded to emphasize the safe scenarios more clearly.

The details are relegated to an appendix, because truly, most people will not want to know.  But recommending against syntactically legal configurations really does need to be explained in the spec.

Also, don't use

- in: header
  name: Cookie

Because... yeah.

### DIFF
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -1044,7 +1044,6 @@ There are four possible parameter locations specified by the `in` field:
 * header - Custom headers that are expected as part of the request. Note that [RFC7230](https://tools.ietf.org/html/rfc7230#page-22) states header names are case insensitive.
 * cookie - Used to pass a specific cookie value to the API.
 
-
 ##### Fixed Fields
 
 The rules for serialization of the parameter are specified in one of two ways.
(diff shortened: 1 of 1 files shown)


ralfhandl (commented):
+1, minor nit

ralfhandl (approved):

miqui (approved):

miqui (approved):

PR #3846, opened 2024-05-23
Clarify discriminator + oneOf/anyOf/allOf usage (3.1.1 modified port of #3822)

handrews:
This is not a direct port of #3822 (although it is close), for two reasons:

1.  Full-document parsing (#3758) and 3.1.1-specific implicit component resolution (PR #3823) means that the 3.0.4 guidance on finding `allOf`-using schemas is not needed in 3.1.1
2. @jdesrosiers made some improvements and bug-fixes to the Discriminator Object language in 3.1.1 (PR #2618) but not 3.0.4 (because when he submitted the PR, we hadn't decided to do a 3.0.4 yet)

I have tried to merge the improvements as best I can.  There were two notable changes to the improved text:

* A paragraph at the beginning was dropped because it's points are now all addressed elsewhere:
    * The definition of "name" is superseded by the definition of the term "component name" in PR #3823, which we've been using consistently in several PRs now; I added an example in PR #3823 to ensure that it is as clear as what @jdesrosiers wrote in the removed paragraph
    * The bit about implicit schemas not being considered (which might not have been changed in the previous 3.1.1 PR) moved down into text that I rewrote substantially in PR #3822 
* The MUST wording on requiring the `propertyName` property sadly has to be weakened to a "SHOULD but the behavior is otherwise undefined" due to our tightened compatibility requirements for patch releases - this is how we have handled other problematic ambiguities – in practice, this is essentially as strong since disregarding the SHOULD is clearly not interoperable (see PR #3805 for the meaning of "undefined" and "implementation-defined"

I also revived some of @jdesrosiers 's ideas that were taken out for compatibility reasons and included a co-author credit; see the commit message below.

Once this PR is accepted, I will backport @jdesrosiers 's changes, as merged here, to 3.0.4 and include a co-author credit (unless you'd rather do it yourself, Jason).

Fixes:
* #2520 
* #3424 
* #3591 

------
Commit message:

This moves some guidance up to the fixed fields section where
it is more obvious, and explicitly designates other configurations
as having undefined behavior.

It also creates subsections to organize the different topics, pulls
key guidance out of the examples and up into those sections,
and provides clarification on the ambiguity of names and URIs.

Finally, it incorporates ideas from @jdesrosiers regarding
the ambiguous `mapping` syntax submitted in a prior PR, but
does so in a way that meets our compatibility requirements
for patch releases.
For the same compatibility reasons, the MUST wording for
requiring the named discriminator property in the schema
was (regrettably) weakened to a "SHOULD but otherwise undefined",
as we have done for other problematic ambiguities.

Co-authored-by: Jason Desrosiers <jdesrosi@gmail.com>

### DIFF
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -193,8 +193,6 @@ If the same JSON/YAML object is parsed multiple times and the respective context
 
 #### <a name="resolvingImplicitConnections"></a>Resolving Implicit Connections
 
-***TODO: In another PR***
-
 ### <a name="dataTypes"></a>Data Types
 
 Data types in the OAS are based on the types supported by the [JSON Schema Specification Draft 2020-12](https://tools.ietf.org/html/draft-bhutton-json-schema-00#section-4.2.1).
@@ -2710,7 +2708,7 @@ components:
         ]
       },
       "Cat": {
-        "description": "A representation of a cat. Note that `Cat` will be used as the discriminator value.",
+        "description": "A representation of a cat. Note that `Cat` will be used as the discriminating value.",
         "allOf": [
           {
             "$ref": "#/components/schemas/Pet"
@@ -2737,7 +2735,7 @@ components:
         ]
       },
       "Dog": {
-        "description": "A representation of a dog. Note that `Dog` will be used as the discriminator value.",
+        "description": "A representation of a dog. Note that `Dog` will be used as the discriminating value.",
         "allOf": [
           {
             "$ref": "#/components/schemas/Pet"
@@ -2779,7 +2777,7 @@ components:
       required:
       - name
       - petType
-    Cat:  # "Cat" will be used as the discriminator value
+    Cat:  # "Cat" will be used as the discriminating value
       description: A representation of a cat
       allOf:
       - $ref: '#/components/schemas/Pet'
@@ -2795,7 +2793,7 @@ components:
             - aggressive
         required:
         - huntingSkill
-    Dog:  # "Dog" will be used as the discriminator value
+    Dog:  # "Dog" will be used as the discriminating value
       description: A representation of a dog
       allOf:
       - $ref: '#/components/schemas/Pet'
@@ -2933,7 +2931,7 @@ The Discriminator Object does this by implicitly or explicitly associating the p
 ##### Fixed Fields
 Field Name | Type | Description
 ---|:---:|---
-<a name="propertyName"></a>propertyName | `string` | **REQUIRED**. The name of the property in the payload that will hold the discriminator value. This property SHOULD be required in the payload schema, as the behavior when the property is absent is undefined.
+<a name="propertyName"></a>propertyName | `string` | **REQUIRED**. The name of the property in the payload that will hold the discriminating value. This property SHOULD be required in the payload schema, as the behavior when the property is absent is undefined.
 <a name="discriminatorMapping"></a> mapping | Map[`string`, `string`] | An object to hold mappings between payload values and schema names or URI references.
 
 This object MAY be extended with [Specification Extensions](#specificationExtensions).
@@ -3008,7 +3006,7 @@ MyResponseType:
       monster: https://gigantic-server.com/schemas/Monster/schema.json
 ```
 
-Here the discriminator property _value_ of `dog` will map to the schema `#/components/schemas/Dog`, rather than the default (implicit) value of `#/components/schemas/dog`.  If the discriminator property _value_ does not match an implicit or explicit mapping, no schema can be determined and validation SHOULD fail.
+Here the discriminating value of `dog` will map to the schema `#/components/schemas/Dog`, rather than the default (implicit) value of `#/components/schemas/dog`.  If the discriminating value does not match an implicit or explicit mapping, no schema can be determined and validation SHOULD fail.
 
 When used in conjunction with the `anyOf` construct, the use of the discriminator can avoid ambiguity for serializers/deserializers where multiple schemas may satisfy a single payload.
 
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -191,6 +191,10 @@ When parsing an OAD, JSON or YAML objects are parsed into specific Objects (such
 
 If the same JSON/YAML object is parsed multiple times and the respective contexts require it to be parsed as _different_ Object types, the resulting behavior is _implementation defined_, and MAY be treated as an error if detected.  An example would be referencing an empty Schema Object under `#/components/schemas` where a Path Item Object is expected, as an empty object is valid for both types.  For maximum interoperability, it is RECOMMENDED that OpenAPI Description authors avoid such scenarios.
 
+#### <a name="resolvingImplicitConnections"></a>Resolving Implicit Connections
+
+***TODO: In another PR***
+
 ### <a name="dataTypes"></a>Data Types
 
 Data types in the OAS are based on the types supported by the [JSON Schema Specification Draft 2020-12](https://tools.ietf.org/html/draft-bhutton-json-schema-00#section-4.2.1).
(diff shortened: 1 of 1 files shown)


versions/3.1.1.md
```diff
@@ -191,6 +191,10 @@ When parsing an OAD, JSON or YAML objects are parsed into specific Objects (such
 
 If the same JSON/YAML object is parsed multiple times and the respective contexts require it to be parsed as _different_ Object types, the resulting behavior is _implementation defined_, and MAY be treated as an error if detected.  An example would be referencing an empty Schema Object under `#/components/schemas` where a Path Item Object is expected, as an empty object is valid for both types.  For maximum interoperability, it is RECOMMENDED that OpenAPI Description authors avoid such scenarios.
 
+#### <a name="resolvingImplicitConnections"></a>Resolving Implicit Connections
+
+***TODO: In another PR***
+
```
ralfhandl, line 197: Could you please remove this?

```suggestion
```

versions/3.1.1.md
```diff
@@ -2922,24 +2926,38 @@ components:
...
-
-`discriminator` uses a schema's "name" to automatically map a property value to
-a schema. The schema's "name" is the property name used when declaring the
-schema as a component in an OpenAPI document. For example, the name of the
-schema at `#/components/schemas/Cat` is "Cat". Therefore, when using
-`discriminator`, _inline_ schemas will not be considered because they don't have
-a "name".
+When request bodies or response payloads may be one of a number of different schemas, a Discriminator Object gives a hint about the expected schema of the document.
+This hint can be used to aid in serialization, deserialization, and validation.
+The Discriminator Object does this by implicitly or explicitly associating the possible values of a named property with alternative schemas.
```
ralfhandl, line 2931: I like the 3.0.4 text better, and can't spot a semantic difference between the two variants.

```suggestion
When request bodies or response payloads may be one of a number of different schemas, a Discriminator Object can be used to aid in serialization, deserialization, and validation by implicitly or explicitly associating the possible values of a named property with alternative schemas.
```

versions/3.1.1.md
```diff
@@ -2990,13 +3008,11 @@ MyResponseType:
       monster: https://gigantic-server.com/schemas/Monster/schema.json
 ```
 
-Here the discriminator _value_ of `dog` will map to the schema `#/components/schemas/Dog`, rather than the default (implicit) `#/components/schemas/dog`.  If the discriminator _value_ does not match an implicit or explicit mapping, no schema can be determined and validation SHOULD fail. Mapping keys MUST be string values, but tooling MAY convert response values to strings for comparison.
+Here the discriminator property _value_ of `dog` will map to the schema `#/components/schemas/Dog`, rather than the default (implicit) value of `#/components/schemas/dog`.  If the discriminator property _value_ does not match an implicit or explicit mapping, no schema can be determined and validation SHOULD fail.
```
ralfhandl, line 3011: Given that the definition of `propertyName` also uses  "discriminator value", I'd remove the "property" here, same as in 3.0.4:

```suggestion
Here the discriminator _value_ of `dog` will map to the schema `#/components/schemas/Dog`, rather than the default (implicit) value of `#/components/schemas/dog`.  If the discriminator _value_ does not match an implicit or explicit mapping, no schema can be determined and validation SHOULD fail.
```

ralfhandl (commented):
+1, minor nits

jdesrosiers:
> unless you'd rather do it yourself, Jason

No thanks, you go ahead. But, I appreciate you asking and the recognition. Thanks for getting this work over the finish line.

versions/3.1.1.md
handrews: @ralfhandl it will get removed when I resolve conflicts with #3823, but in the meantime the PR markdown check fails if I don't have this here.

versions/3.1.1.md
handrews: @ralfhandl we previously approved the additional wording (submitted by another contributor) in 3.1.1 and I'd rather not take it out on one person's stylistic preference.  If you don't think this is the correct way to merge them, then we can change it, but I was trying not to delete someone else's work here.

versions/3.1.1.md
handrews: I'd rather fix the definiton of `propertyName` as I found this wording very confusing.  To me, the discriminator value is the value of the `discriminator` field, which is how we'd talk about any other field.

versions/3.1.1.md
handrews: I will convert this to a draft so it doesn't accidentally get merged with the placeholder.

versions/3.1.1.md
ralfhandl: Whatever works better for you, as long as it is consistent between 3.0.4 and 3.1.1.

handrews: @ralfhandl

> as long as it is consistent between 3.0.4 and 3.1.1.

See my previous comment:

> Once this PR is accepted, I will backport jdesrosiers 's changes, as merged here, to 3.0.4 and include a co-author credit (unless you'd rather do it yourself, Jason).

which is why I don't want to do it piecemeal- I'm mentally tracking about 30 issues and being systematic about how I synchronize them across the various releases.  I made a change on 3.0.4, Jason made a change on 3.1.1.  This is me forward-porting my 3.0.4 changes, merging them, and then I'll backport the merge result to ensure Jason's changes make it to 3.0.4.

handrews: OK I have [switched to "discriminating value"](https://github.com/OAI/OpenAPI-Specification/pull/3846/commits/dd44da96dbca2aa92d4c0efe87448accfd213ed0) across the whole PR, as I agree with @ralfhandl that what i had was confusing, but I also find "discriminator property" to be confusing as it means "value of the `discriminator` property in the OAD" to me, rather than value of the property nmaed by `#/.../discriminator/propertyName`.

@ralfhandl If that doesn't feel right, I would also be open to "discriminating property".  The important part for me is distinguishing between "discriminator" (which to me is the `discriminator` keyword and Discriminator Object) and "discriminating" (which to me implies the instance property and value).

versions/3.1.1.md
handrews: @ralfhandl I have taken the TODO out but left the header stub in.  Of the three ways I've tried to handle the PR checks failing when a link target doesn't exist across different PRs, that's the only one that has not provoked a complaint in review.

miqui (approved):
Looks good

versions/3.1.1.md
```diff
@@ -191,6 +191,8 @@ When parsing an OAD, JSON or YAML objects are parsed into specific Objects (such
 
 If the same JSON/YAML object is parsed multiple times and the respective contexts require it to be parsed as _different_ Object types, the resulting behavior is _implementation defined_, and MAY be treated as an error if detected.  An example would be referencing an empty Schema Object under `#/components/schemas` where a Path Item Object is expected, as an empty object is valid for both types.  For maximum interoperability, it is RECOMMENDED that OpenAPI Description authors avoid such scenarios.
 
+#### <a name="resolvingImplicitConnections"></a>Resolving Implicit Connections
+
```
mikekistler, line 195: This looks like an empty H4 ... is that what was intended? It looks very odd.

mikekistler (approved):
Looks good. 👍 

I commented on an empty H4. If that was not intended it should be fixed but everything else looked fine.

versions/3.1.1.md
handrews: It's to keep the markdown validator from complaining about broken links.  That is being added in another PR.  I have like 6 different PRs that cross-reference each other and there's no good way to do this.  Of the three ways I've tried, this was the only one that @ralfhandl didn't object to.

It will all get sorted as we merge and resolve conflicts.

ralfhandl (approved):

ralfhandl:
@handrews Do you want to merge now?

Issue #3626, opened 2024-03-08
Synchronize 3.0.4, 3.1.1, and 3.2.0


handrews:
At the moment, we are working towards releasing three releases, off three branches (`v3.0.4-dev`, `v3.1.1-dev`, `v3.2.0-dev`).

In order to release consistent documents (please let's not face-plant with our first releases in three years), we need to synchronize these three branches (and also apply the GitHub markdown link fixes that were made to main, but _not yet to pending release branches_).

This is made difficult by our policy of using separate files on each branch, [making it impossible to manage such things with github in a normal way](https://github.com/OAI/OpenAPI-Specification/issues/3602#issuecomment-1959986977).  But there are some less-normal ways git can help us as seen in the scripts linked below.

I previously said to a few people that I'd sort this out, but I am overloaded with other OpenAPI tasks and do not expect to be able to do this anytime soon.  I did learn that we have two scripts designed to help out with this, although they aren't really robust in the face of the current rather muddled situation.  But they help and could probably be improved:

* [`scripts/fwdport.sh`](https://github.com/OAI/OpenAPI-Specification/blob/main/scripts/fwdport.sh)
* [`scripts/fwdabort.sh`](https://github.com/OAI/OpenAPI-Specification/blob/main/scripts/fwdabort.sh)

What I'd expect from PRs here is that:

* Each ported commit is still a separate commit (although merge commits need not be ported - it's fine to merge or rebase things differently)
* Commits should remain credited to their original authors, or use the GitHub-recognized Co-Author line (GitHub has documentation about this somewhere)
* It's fine to make one PR for many commits since they've been reviewed before, and GitHub allows examining each commit separately as needed

The main trick is figuring out which changes are applicable or not (for example, fixes to the reference to RFC 3339 in 3.0.4 do **not** need to be forward-ported, because from 3.1.0 onwards the OAS doesn't have that reference - it's handled by JSON Schema compatiblity).  A related thing to watch out for is that sometimes the same change is applied to different branches with differently-titled commits.

We really need to fix this before we can move forward with finalizing these releases.  Once we do this, we can decide whether we will continue parallel releases or not (in a recent Thursday call, we tentatively agreed that 3.0.4 would be the last 3.0.y barring a really compelling need; we did not make a decision on whether we would continue 3.1.y patch releases).

lornajane:
[...]

PR #3850, opened 2024-05-24
Typo in discriminator example

ralfhandl:
Based on the proposed change to 3.1.1 in #3846 the 3.0.4 text seems to be wrong.

### DIFF
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -2826,7 +2826,7 @@ MyResponseType:
       monster: https://gigantic-server.com/schemas/Monster/schema.json
 ```
 
-Here the discriminator _value_ of `dog` will map to the schema `#/components/schemas/Dog`, rather than the default (implicit) value of `Dog`.  If the discriminator _value_ does not match an implicit or explicit mapping, no schema can be determined and validation SHOULD fail.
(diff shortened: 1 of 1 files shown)


handrews:
@ralfhandl there are several things that need backporting from PR #2618, this is only one of them- could you close this as it will be easier for me to backport if I don't have to split pieces out.  I've added #2618 to https://github.com/OAI/OpenAPI-Specification/issues/3626#issuecomment-2016888941 so we won't forget it.

handrews (changes requested):
As noted in other comments, I'd prefer to port PRs as PRs and not individual lines.

ralfhandl:
@handrews Don't find this typo explicitly mentioned in https://github.com/OAI/OpenAPI-Specification/issues/3626#issuecomment-2016888941, so why not just fix it by merging this trivial PR instead of risking it falling through the cracks?

Note: from my perspective 3.0.4 is the most relevant of the four specs we are currently working on.

handrews:
This is now addressed on [this line](https://github.com/OAI/OpenAPI-Specification/pull/3880/files#diff-89e34372e8c32d43f8e3912385b9238745456d34646ebfbf91b3b0a263f9120dR2837) of the full merge resolution backport:
* #3880

PR #3855, opened 2024-05-25
Clarify resolving implicit connections (3.1.1 - ALTERNATIVE to #3823)

handrews:
This clarifies how to handle resolving implicit (non-URI-based) connections in multi-document OpenAPI Descriptions.

While the behavior is implementation-defined overall, this RECOMMENDS a single approach based on how things behaved going back to the 2.0 referencing model.  This allows Security Schemes and Tags to (like the top-level Server Objects) define a deployment-specific interface for referenced documents to access.

This entry document interface approach makes less sense for the Discriminator Object, but it can use the URI syntax of `mapping` to keep things within the local document.

This also aligns the search for matching `operationId`s with 3.1's full-document parsing requirements.

Note that the term "complete OpenAPI document" has been defined in another change pending approval on the 3.0.4 branch.


handrews:
Oh hey it turns out hitting enter while renaming the title posts the PR.

CLOSING as this was not ready.

PR #3856, opened 2024-05-25
Alternative: Clarify resolving implicit connections (3.1.1- alternative to #3823)

handrews:
***This is an alternative approach to the one in PR #3823***

The other PR pushes users more towards document-oriented resolution, which is what I personally prefer.  But it only works if the documents are self-contained, and the way people organize multi-document OADs based on 2.0 conventions does not work with it.  So the other PR provides two different approaches depending on how the OAD is structured.  That's a bit complicated.

This PR advises one approach: resolve implicit connections from the entry document, and use URI-based alternatives to resolve locally.

This works because the ones where you are most likely to want local connections both have URI-based alternatives, while the ones without URI-based alternatives arguably make more sense associated with the API deployment, which is handled by the entry document:

* local connections:
    * Discriminator Objects are used with hierarchies that are likely to be defined together, and `mapping` supports URIs
    * Link Objects need to worry about `operationId` collisions across multiple documents, and `operationRef` is a URI-based alternative that does that
* deployment interface:
    * Security Requirement Objects linking to deployment-specific Security Schemes (using the component name as interface)
    * Operation Objects tagging based on deployment-specific Tag Object definitions, as the organization of operations is probably more of a deployment-specific concern (and tags that don't match tag objects still work as tags, they just don't come with ordering information or metadata)

I think we would want to still provide the missing URI-based alternatives in 3.2 because sometimes you will want to package up a set of Tag Objects or Security Schemes+Requirements.  We in fact have issues open for both of those things.

But this is a more straightforward option, even if I'd prefer the self-contained document approach.

-----

This clarifies how to handle resolving implicit (non-URI-based) connections in multi-document OpenAPI Descriptions.

While the behavior is implementation-defined overall, this RECOMMENDS a single approach based on how things behaved going back to the 2.0 referencing model.  This allows Security Schemes and Tags to (like the top-level Server Objects) define a deployment-specific interface for referenced documents to access.

This entry document interface approach makes less sense for the Discriminator Object, but it can use the URI syntax of `mapping` to keep things within the local document.

This also aligns the search for matching `operationId`s with 3.1's full-document parsing requirements.

Note that the term "complete OpenAPI document" has been defined in another change pending approval on the 3.0.4 branch.

### DIFF
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -3832,9 +3832,38 @@ security:
 
 See [Resolving Implicit Connections](#resolvingImplicitConnections) for more information.
 
-Entry document `openapi.yaml`:
+First, our entry document is where parsing begins. It defines the `MySecurity` security scheme to be JWT-based, and it defines on Path Item as a reference to a component in another document:
 
-```YAML
+```HTTP
+GET /api/description/openapi HTTP/1.1
+Host: www.example.com
+Accept: application/openapi+json
+```
+
+```json
+"components": {
+  "securitySchemes": {
+    "MySecurity": {
+      "type": "http",
+      "scheme": "bearer",
+      "bearerFormat": "JWT"
+    }
+  }
+},
+"paths": {
+  "/foo": {
+    "$ref": "other#/components/pathItems/Foo"
+  }
+}
+```
+
+```HTTP
+GET /api/description/openapi HTTP/1.1
+Host: www.example.com
+Accept: application/openapi+yaml
+```
+
+```yaml
 components:
   securitySchemes:
     MySecurity:
@@ -3843,12 +3872,44 @@ components:
       bearerFormat: JWT
 paths:
   /foo:
-    $ref: "other.yaml#/components/pathItems/Foo"
+    $ref: "other#/components/pathItems/Foo"
 ```
 
-Referenced document `other.yaml`:
+Next, we have our referenced document, `other`, that we presumably request in the same format we requested for the entry document.  But the fact that we don't use file extensions gives the client the flexibilty to choose on a resource-by-resource basis, assuming both representations are available:
 
-```YAML
+```HTTP
+GET /api/description/other HTTP/1.1
+Host: www.example.com
+Accept: application/openapi+json
+```
+
+```json
+"components": {
+  "securitySchemes": {
+    "MySecurity": {
+      "type": "http",
+      "scheme": "basic"
+    }
+  },
+  "pathItems": {
+    "Foo": {
+      "get": {
+        "security": [
+         "MySecurity": []
+        ]
+      }
+    }
+  }
+}
+```
+
+```HTTP
+GET /api/description/other HTTP/1.1
+Host: www.example.com
+Accept: application/openapi+yaml
+```
+
+```yaml
 components:
   securitySchemes:
     MySecurity:
@@ -3861,7 +3922,7 @@ components:
         - MySecurity: []
 ```
 
-In this example, it is implementation-defined whether the Security Requirement for "MySecurity" in `other.yaml` resolves to `other.yaml#/components/securitySchemes/MySecurity` or the RECOMMENDED resolved location of `openapi.yaml#/components/securitySchemes/MySecurity`.
+In this `other` document, the reference path item has a Security Requirement for the Security Scheme "MySecurity".  But there is a Security Scheme by that name in the `other` document as well.  As discussed in [Resolving Implicit Connections](#resolvingImplicitConnections), which "MySecurity" gets used is [implementation-defined](#undefinedAndImplementationDefinedBehavior).  However, as also documented in that section, it is RECOMMENDED that tools resolve component names from the [entry document](#documentStructure).  As with all implementation-defined behavior, it is important to check tool documentation to determine which behavior is supported.
 
 ### <a name="specificationExtensions"></a>Specification Extensions
 
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -193,7 +193,8 @@ If the same JSON/YAML object is parsed multiple times and the respective context
 
 #### <a name="resolvingImplicitConnections"></a>Resolving Implicit Connections
 
-Several features of this specification require resolving a non-URI-based connection to some other part of the OpenAPI Description (OAD).
+Several features of this specification require resolution of non-URI-based connections to some other part of the OpenAPI Description (OAD).
+``
 
 These connections are easily resolved in single-document OADs, but the resolution process in multi-document OADs has never been spelled out, and is therefore _implementation-defined_, within the constraints described in this section.
 In some cases, an unambiguous URI-based alternative is available, and OAD authors are RECOMMENDED to always use the alternative:
@@ -210,10 +211,10 @@ A fifth implicit connection, which involves appending the templated URL paths of
 It is RECOMMENDED to consider all Operation Objects from all parsed documents when resolving any Link Object `operationId`.
 This requires ensuring that all referenced documents have been parsed prior to determining an `operationId` to be unresolvable.
 
-The implicit connections in the Security Requirement Object and Discriminator Object rely on the _component name_, which is the property name holding the component in the appropriate typed sub-object of the Components Object.
+The implicit connections in the Security Requirement Object and Discriminator Object rely on the _component name_, which is the property name holding the component in the appropriately typed sub-object of the Components Object.
 For example, the component name of the Schema Object at `#/components/schemas/Foo` is `Foo`.
-The implicit connection of tags in the Operation Object use the `name` field of Tag Objects, which (like the Components Object) are found under the root OpenAPI Object.
-This means that resolving component names and tag names both depend on starting from the correct OpenAPI Object.
+The implicit connection of `tags` in the Operation Object use the `name` field of Tag Objects, which (like the Components Object) are found under the root OpenAPI Object.
+This means resolving component names and tag names both depend on starting from the correct OpenAPI Object.
 
 For resolving component and tag name connections from a referenced (non-entry) document, it is RECOMMENDED that tools resolve from the entry document, rather than the current document.
 This allows Security Scheme Objects and Tag Objects to be defined with the API's deployment information (the top-level Server Objects), and treated as an interface for referenced documents to access.
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -220,7 +220,7 @@ This allows Security Scheme Objects and Tag Objects to be defined with the API's
 
 The interface approach can also work for Discriminator Objects and Schema Objects, but it is also possible to keep the Discriminator Object's behavior within a single document using the relative URI-reference syntax of `mapping`.
 
-There are not currently URI-based alternatives for the Security Requirement Object or for the Operation Object's `tags` field.
+There currently are no URI-based alternatives for the Security Requirement Object or for the Operation Object's `tags` field.
 These limitations are expected to be addressed in a future release.
 
 Note that no aspect of implicit connection resolution changes how [URIs are resolved](#relativeReferencesURI), or restricts their possible targets.
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -3832,7 +3832,7 @@ security:
 
 See [Resolving Implicit Connections](#resolvingImplicitConnections) for more information.
 
-First, our entry document is where parsing begins. It defines the `MySecurity` security scheme to be JWT-based, and it defines on Path Item as a reference to a component in another document:
+First, our entry document is where parsing begins. It defines the `MySecurity` security scheme to be JWT-based, and it defines a Path Item as a reference to a component in another document:
 
 ```HTTP
 GET /api/description/openapi HTTP/1.1
@@ -3875,7 +3875,7 @@ paths:
     $ref: "other#/components/pathItems/Foo"
 ```
 
-Next, we have our referenced document, `other`, that we presumably request in the same format we requested for the entry document.  But the fact that we don't use file extensions gives the client the flexibilty to choose on a resource-by-resource basis, assuming both representations are available:
+Next, we have our referenced document, `other`. The fact that we don't use file extensions gives the client the flexibility to choose an acceptable format on a resource-by-resource basis, assuming both representations are available:
 
 ```HTTP
 GET /api/description/other HTTP/1.1
@@ -3922,7 +3922,7 @@ components:
         - MySecurity: []
 ```
 
-In this `other` document, the reference path item has a Security Requirement for the Security Scheme "MySecurity".  But there is a Security Scheme by that name in the `other` document as well.  As discussed in [Resolving Implicit Connections](#resolvingImplicitConnections), which "MySecurity" gets used is [implementation-defined](#undefinedAndImplementationDefinedBehavior).  However, as also documented in that section, it is RECOMMENDED that tools resolve component names from the [entry document](#documentStructure).  As with all implementation-defined behavior, it is important to check tool documentation to determine which behavior is supported.
+In the `other` document, the referenced path item has a Security Requirement for a Security Scheme, `MySecurity`. The same Security Scheme exists in the original entry document. As outlined in [Resolving Implicit Connections](#resolvingImplicitConnections), `MySecurity` is resolved with an [implementation-defined behavior](#undefinedAndImplementationDefinedBehavior). However, documented in that section, it is RECOMMENDED that tools resolve component names from the [entry document](#documentStructure). As with all implementation-defined behavior, it is important to check tool documentation to determine which behavior is supported.
 
 ### <a name="specificationExtensions"></a>Specification Extensions
 
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -225,6 +225,9 @@ The interface approach can also work for Discriminator Objects and Schema Object
 There are no URI-based alternatives for the Security Requirement Object or for the Operation Object's `tags` field.
 These limitations are expected to be addressed in a future release.
 
+See [Security Requirement in a Referenced Document](#security-requirement-in-a-referenced-document) for an example of the possible resolutions, including which one is recommended by this section.
+The behavior for Discrimator Object non-URI mappings and for the Operation Object's `tags` field operate on the same principles.
+
 Note that no aspect of implicit connection resolution changes how [URIs are resolved](#relativeReferencesURI), or restricts their possible targets.
 
 ### <a name="dataTypes"></a>Data Types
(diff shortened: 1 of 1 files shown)


versions/3.1.1.md
```diff
@@ -185,12 +185,46 @@ It is the responsibility of an embedding format to define how to parse embedded
...
+For example, the component name of the Schema Object at `#/components/schemas/Foo` is `Foo`.
+The implicit connection of tags in the Operation Object use the `name` field of Tag Objects, which (like the Components Object) are found under the root OpenAPI Object.
+This means that resolving component names and tag names both depend on starting from the correct OpenAPI Object.
+
+For resolving component and tag name connections from a referenced (non-entry) document, it is RECOMMENDED that tools resolve from the entry document, rather than the current document.
+This allows Security Scheme Objects and Tag Objects to be defined with the API's deployment information (the top-level Server Objects), and treated as an interface for referenced documents to access.
+
+The interface approach can also work for Discriminator Objects and Schema Objects, but it is also possible to keep the Discriminator Object's behavior within a single document using the relative URI-reference syntax of `mapping`.
+
+There are not currently URI-based alternatives for the Security Requirement Object or for the Operation Object's `tags` field.
```
ralfhandl, line 223:
```suggestion
There currently are no URI-based alternatives for the Security Requirement Object or for the Operation Object's `tags` field.
```

ralfhandl (commented):
+1, with formulation proposal

ralfhandl:
I prefer this alternative over #3823.

ralfhandl (dismissed):

lornajane:
A good discussion about this in TSC this week. Examples might be findable in https://github.com/OAI/oascomply/tree/main/reports to illustrate this point.

versions/3.1.1.md
```diff
@@ -185,12 +185,46 @@ It is the responsibility of an embedding format to define how to parse embedded
...
-* As a full OpenAPI Description document (an [OpenAPI Object](#oasObject) taking up an entire document)
+* As a complete OpenAPI Description document
 * As the Object type implied by its parent Object within the document
 * As a reference target, with the Object type matching the reference source's context
 
 If the same JSON/YAML object is parsed multiple times and the respective contexts require it to be parsed as _different_ Object types, the resulting behavior is _implementation defined_, and MAY be treated as an error if detected.  An example would be referencing an empty Schema Object under `#/components/schemas` where a Path Item Object is expected, as an empty object is valid for both types.  For maximum interoperability, it is RECOMMENDED that OpenAPI Description authors avoid such scenarios.
 
+#### <a name="resolvingImplicitConnections"></a>Resolving Implicit Connections
+
+Several features of this specification require resolving a non-URI-based connection to some other part of the OpenAPI Description (OAD).
```
jeremyfiel, line 196:
```suggestion
Several features of this specification require resolution of non-URI-based connections to some other part of the OpenAPI Description (OAD).
``

**Review by jeremyfiel — COMMENTED** [review]
I've updated the text to clarify understanding and use cohesive terminology, throughout.

File: versions/3.1.1.md [hunk 3]
```diff
@@ -185,12 +185,46 @@ It is the responsibility of an embedding format to define how to parse embedded
...
+For example, the component name of the Schema Object at `#/components/schemas/Foo` is `Foo`.
+The implicit connection of tags in the Operation Object use the `name` field of Tag Objects, which (like the Components Object) are found under the root OpenAPI Object.
+This means that resolving component names and tag names both depend on starting from the correct OpenAPI Object.
+
+For resolving component and tag name connections from a referenced (non-entry) document, it is RECOMMENDED that tools resolve from the entry document, rather than the current document.
+This allows Security Scheme Objects and Tag Objects to be defined with the API's deployment information (the top-level Server Objects), and treated as an interface for referenced documents to access.
+
+The interface approach can also work for Discriminator Objects and Schema Objects, but it is also possible to keep the Discriminator Object's behavior within a single document using the relative URI-reference syntax of `mapping`.
+
+There currently are no URI-based alternatives for the Security Requirement Object or for the Operation Object's `tags` field.
```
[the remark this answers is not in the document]
handrews: Hmm... I wonder how to best make the intent more clear here.

In this sentence, the structure is making a distinction between the Security Requirement Object (as a whole, or more precisely all top-level keys in the Object) and the specific `tags` field in the Operation Object.  Your rewrite reads to me like both the Operation Object and Security Requirement Object have `tags` fields, which is not the case.

versions/3.1.1.md
```diff
@@ -185,12 +185,46 @@ It is the responsibility of an embedding format to define how to parse embedded
...
+These limitations are expected to be addressed in a future release.
+
+Note that no aspect of implicit connection resolution changes how [URIs are resolved](#relativeReferencesURI), or restricts their possible targets.
```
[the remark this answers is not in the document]
handrews: I count 14 instances of "Note that" in the spec and two instances of "Note:", one of which is a very specific usage (an italicized note inside of a field definition set off by a blank line, which is not done anywhere else in the spec AFAIK).  I'd rather stick with "Note that" (and maybe put this in the [style guide](https://github.com/OAI/OpenAPI-Specification/issues/3785)?

handrews (commented):
@jeremyfiel I appreciate you going over this for consistency!  We will need an effort like this on a larger scale before publishing, and I hope we can build up a [style guide](https://github.com/OAI/OpenAPI-Specification/issues/3785), which perhaps you could help with?

I've replied to a few where I interpret grammatical rules differently, or see other precedents in the existing text, or just think the sentence was unclear to start with which led you to fix the grammar in the "wrong" direction.  But I appreciate this feedback and am merging several other comments.

versions/3.1.1.md
```diff
@@ -185,12 +185,46 @@ It is the responsibility of an embedding format to define how to parse embedded
...
+In some cases, an unambiguous URI-based alternative is available, and OAD authors are RECOMMENDED to always use the alternative:
+
+Source | Target | Alternative
+------ | ------ | -----------
+[Security Requirement Object](#securityRequirementObject) `{name}` |  [Security Scheme Object](#securitySchemeObject) name under the [Components Object](#componentsObject) | _n/a_
+[Discriminator Object](#discriminatorObject) `mapping` _(implicit, or explicit name syntax)_ | [Schema Object](#schemaObject) name under the Components Object | `mapping` _(explicit URI syntax)_
+[Operation Object](#operationObject) `tags` | [Tag Object](#tagObject) `name` (in the Components Object) | _n/a_
+[Link Object](#linkObject) `operationId` | [Path Item Object](#pathItemObject) `operationId` | `operationRef`
+
+A fifth implicit connection, which involves appending the templated URL paths of the [Paths Object](#pathsObject) to the appropriate [Server Object](#serverObject)'s `url` field, is unambiguous because only the entry document's Paths Object contributes URLs to the described API.
```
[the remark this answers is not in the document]
handrews: I ended up taking your suggestion of splitting the sentence and removed all commas and "which" in the first sentence as I think that is correct and then we don't need to debate complex sentence structure.

handrews:
@jeremyfiel I've updated several areas that you commented on with changes that I think address your concerns in slightly different ways.  There are a few I have not (yet) changed, pending your replies.  I've also rebased the branch to resolve merge conflicts - while I had to force-push that, my latest changes are a separate commit if you want to look at them specifically.

ralfhandl (dismissed):

ralfhandl:
> a Shatner, comma

Amazing, the things I learn here 😁

versions/3.1.1.md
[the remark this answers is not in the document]
jeremyfiel: i'm ok with _that_  :D

versions/3.1.1.md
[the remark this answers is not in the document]
jeremyfiel: you're right, that does seem to infer both of them have the `tags`.   I added a `the` in there.. Not sure if that is sufficient.

versions/3.1.1.md
```diff
@@ -185,12 +185,46 @@ It is the responsibility of an embedding format to define how to parse embedded
...
+
+It is RECOMMENDED to consider all Operation Objects from all parsed documents when resolving any Link Object `operationId`.
+This requires ensuring that all referenced documents have been parsed prior to determining an `operationId` to be unresolvable.
+
+The implicit connections in the Security Requirement Object and Discriminator Object rely on the _component name_, which is the property name holding the component in the appropriate typed sub-object of the Components Object.
+For example, the component name of the Schema Object at `#/components/schemas/Foo` is `Foo`.
+The implicit connection of tags in the Operation Object use the `name` field of Tag Objects, which (like the Components Object) are found under the root OpenAPI Object.
+This means that resolving component names and tag names both depend on starting from the correct OpenAPI Object.
+
+For resolving component and tag name connections from a referenced (non-entry) document, it is RECOMMENDED that tools resolve from the entry document, rather than the current document.
```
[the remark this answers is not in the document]
jeremyfiel: i'm ok with _that_ in this one.  I can't point to any hard, fast grammar rule for it's use or lack there of.

versions/3.1.1.md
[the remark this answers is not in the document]
handrews: I think it needs the repeated "for."  I did [remove the "currently"](https://github.com/OAI/OpenAPI-Specification/pull/3856/files#diff-b92507e7acda65ae00a02236c555cefc68b6fca4661077b84c2fb9ab150e5e17R225-R226), though.

versions/3.1.1.md
```diff
@@ -185,12 +185,46 @@ It is the responsibility of an embedding format to define how to parse embedded
...
+------ | ------ | -----------
+[Security Requirement Object](#securityRequirementObject) `{name}` |  [Security Scheme Object](#securitySchemeObject) name under the [Components Object](#componentsObject) | _n/a_
+[Discriminator Object](#discriminatorObject) `mapping` _(implicit, or explicit name syntax)_ | [Schema Object](#schemaObject) name under the Components Object | `mapping` _(explicit URI syntax)_
+[Operation Object](#operationObject) `tags` | [Tag Object](#tagObject) `name` (in the Components Object) | _n/a_
+[Link Object](#linkObject) `operationId` | [Path Item Object](#pathItemObject) `operationId` | `operationRef`
+
+A fifth implicit connection, which involves appending the templated URL paths of the [Paths Object](#pathsObject) to the appropriate [Server Object](#serverObject)'s `url` field, is unambiguous because only the entry document's Paths Object contributes URLs to the described API.
+
+It is RECOMMENDED to consider all Operation Objects from all parsed documents when resolving any Link Object `operationId`.
+This requires ensuring that all referenced documents have been parsed prior to determining an `operationId` to be unresolvable.
```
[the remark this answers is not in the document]
jeremyfiel: as a native speaker, I don't know enough rules to argue for or against it.  haha

I think this may sound better

handrews:
I have now added an example of resolving a Security Requirement in a referenced document.  I am hoping that is sufficient and we don't need examples of everything.

versions/3.1.1.md
```diff
@@ -185,14 +185,51 @@ It is the responsibility of an embedding format to define how to parse embedded
...
+* As a complete OpenAPI Description document
 * As the Object type implied by its parent Object within the document
 * As a reference target, with the Object type matching the reference source's context
 
 If the same JSON/YAML object is parsed multiple times and the respective contexts require it to be parsed as _different_ Object types, the resulting behavior is _implementation defined_, and MAY be treated as an error if detected.  An example would be referencing an empty Schema Object under `#/components/schemas` where a Path Item Object is expected, as an empty object is valid for both types.  For maximum interoperability, it is RECOMMENDED that OpenAPI Description authors avoid such scenarios.
 
 #### <a name="resolvingImplicitConnections"></a>Resolving Implicit Connections
 
+Several features of this specification require resolution of non-URI-based connections to some other part of the OpenAPI Description (OAD).
+``
```
ralfhandl, line 197:
```suggestion
```

versions/3.1.1.md
```diff
@@ -3792,6 +3829,41 @@ security:
     - read:pets
 ```
 
+###### Security Requirement in a Referenced Document
```
ralfhandl, line 3832: Why rely on the implicit anchor generated by Github here?

versions/3.1.1.md
```diff
@@ -3792,6 +3829,41 @@ security:
...
+components:
+  securitySchemes:
+    MySecurity:
+      type: http
+      scheme: basic
+  pathItems:
+    Foo:
+      get:
+        security:
+        - MySecurity: []
```
ralfhandl, line 3862:
```suggestion
          MySecurity: []
```
Pattern field, not array item

ralfhandl (changes requested):
+1, minor nits

lornajane (commented):
I think the example works well, but must we provide that in JSON and YAML?

versions/3.1.1.md
handrews: @ralfhandl I thought it was because you did that in one of the suggestions I committed to the another PR!  I hadn't realized we could just do that until I saw _someone_ do it.  Maybe that wasn't you?  The plan is to switch them all over before releasing anyway because the other ones won't work anymore.

versions/3.1.1.md
handrews: @ralfhandl array of objects with pattern field, not object with pattern field :-)
Screen Shot 2024-06-11 at 8 01 14 AM

handrews:
@lornajane we usually do but sometimes do not.  I suppose I can add it.

handrews:
@lornajane I remember why I didn't want to add the JSON  It annoys me every time I see a `$ref` with a file extension but I know other people are baffled by omitting them.  To me, showing a YAML example with a `$ref: other.yaml` and a JSON example with `"$ref": "other.json"` is a horrible anti-pattern as the whole point of supporting either format is to have the references work the same either way.  So to make myself less frustrated with it I decided to just do one.

I really want either both references to be to just plain `other`, or there to only be one format.  Doing both with a rewritten reference for the file extension... if folks insist I might be able to do it when I'm more awake.  But I just think it's a really bad anti-pattern and I don't know what to do about that. ¯\\_(ツ)_/¯

Maybe I could frame it as HTTP content negotiatoin?  This is the version you get with `Accept: application/openapi+json` and this is the version you get with `Accept: application/openapi+yaml`?  That might make it palatable to the super-literal-URL-with-local-files faction?  I realize I'm probably being unreasonably dogmatic... but it's also kind-of the entire point of the URI vs URL sections so maybe we should leverage that?

ralfhandl:
> both references to be to just plain `other`

I like that!

versions/3.1.1.md
```diff
@@ -3843,12 +3872,44 @@ components:
...
+    "MySecurity": {
+      "type": "http",
+      "scheme": "basic"
+    }
+  },
+  "pathItems": {
+    "Foo": {
+      "get": {
+        "security": [
+         "MySecurity": []
```
ralfhandl, line 3898:
```suggestion
          "MySecurity": []
```

ralfhandl (commented):
Very nice!

ralfhandl (approved):

versions/3.1.1.md
```diff
@@ -3792,6 +3828,102 @@ security:
     - read:pets
 ```
 
+###### Security Requirement in a Referenced Document
+
+See [Resolving Implicit Connections](#resolvingImplicitConnections) for more information.
+
+First, our entry document is where parsing begins. It defines the `MySecurity` security scheme to be JWT-based, and it defines on Path Item as a reference to a component in another document:
```
jeremyfiel, line 3835:
```suggestion
First, our entry document is where parsing begins. It defines the `MySecurity` security scheme to be JWT-based, and it defines a Path Item as a reference to a component in another document:
```

versions/3.1.1.md
```diff
@@ -3792,6 +3828,102 @@ security:
...
+    MySecurity:
+      type: http
+      scheme: bearer
+      bearerFormat: JWT
+paths:
+  /foo:
+    $ref: "other#/components/pathItems/Foo"
+```
+
+Next, we have our referenced document, `other`, that we presumably request in the same format we requested for the entry document.  But the fact that we don't use file extensions gives the client the flexibilty to choose on a resource-by-resource basis, assuming both representations are available:
```
jeremyfiel, line 3878:
```suggestion
Next, we have our referenced document, `other`. The fact that we don't use file extensions gives the client the flexibility to choose an acceptable format on a resource-by-resource basis, assuming both representations are available:
```

versions/3.1.1.md
```diff
@@ -3792,6 +3828,102 @@ security:
...
+      type: http
+      scheme: basic
+  pathItems:
+    Foo:
+      get:
+        security:
+        - MySecurity: []
+```
+
+In this `other` document, the reference path item has a Security Requirement for the Security Scheme "MySecurity".  But there is a Security Scheme by that name in the `other` document as well.  As discussed in [Resolving Implicit Connections](#resolvingImplicitConnections), which "MySecurity" gets used is [implementation-defined](#undefinedAndImplementationDefinedBehavior).  However, as also documented in that section, it is RECOMMENDED that tools resolve component names from the [entry document](#documentStructure).  As with all implementation-defined behavior, it is important to check tool documentation to determine which behavior is supported.
```
jeremyfiel, line 3925:
```suggestion
In the `other` document, the referenced path item has a Security Requirement for a Security Scheme, `MySecurity`. The same Security Scheme exists in the original entry document. As outlined in [Resolving Implicit Connections](#resolvingImplicitConnections), `MySecurity` is resolved with an [implementation-defined behavior](#undefinedAndImplementationDefinedBehavior). However, documented in that section, it is RECOMMENDED that tools resolve component names from the [entry document](#documentStructure). As with all implementation-defined behavior, it is important to check tool documentation to determine which behavior is supported.
```

jeremyfiel (changes requested):
couple nits on the language. 

I'm not a huge proponent of the conversational text, especially for a spec.  But I won't rewrite the whole thing. :)

handrews:
@jeremyfiel 

> I'm not a huge proponent of the conversational text, especially for a spec. But I won't rewrite the whole thing. :)

That's an editing thing rather than a writing thing.  With these PRs the point is to get the information in the spec.

In an ideal world someone like you who has a good eye for editing would go over everything and do wordsmithing.  I do not care if my writing style is retained, as long as we _first_ get the content in that I can provide (and no one else is currently both willing _and_ able to do), and _then_ worry about the style.

And preferably [make a style guide](https://github.com/OAI/OpenAPI-Specification/issues/3785)!  It would make it so much easier to add text if we had one.

But for just getting the content in- whoever does the work gets to determine how it's written up.  And if someone wants to do an editing pass before publishing- I'm all for it.

handrews:
@jeremyfiel also, I tend to write more conversationally for examples, and less so for definitions.  My preference would be to integrate the spec and the Learn site more closely and move most examples to the Learn site, which would have a different tone.  But for now, people expect to find the examples in the spec, and (up to a point) we need to put some there.  I struggle with where to draw the line on what goes in or doesn't – for now it's been "apparently I'm the one writing the PRs so the line is where I say it is", but that's not a viable policy long-term and leads to questions like in #3874.

versions/3.1.1.md
```diff
@@ -3792,6 +3828,102 @@ security:
...
+    MySecurity:
+      type: http
+      scheme: bearer
+      bearerFormat: JWT
+paths:
+  /foo:
+    $ref: "other#/components/pathItems/Foo"
+```
+
+Next, we have our referenced document, `other`. The fact that we don't use file extensions gives the client the flexibility to choose an acceptable format on a resource-by-resource basis, assuming both representations are available:
```
miqui, line 3878: @handrews This is good.

miqui (approved):

ralfhandl (approved):

ralfhandl:
Not merging, waiting for @jeremyfiel to approve the latest changes

jeremyfiel (approved):

ralfhandl:
@lornajane Please confirm whether the text now has sufficient examples.

lornajane (approved):
Thanks for getting into the weeds to get this under control!

Issue #3331, opened 2023-07-19
Clarification on support for application/json encoded object properties in application/x-www-form-urlencoded messages


LGraber:
I have been trying to read the spec and understand this to understand if some tools are "broken" or if the spec is just ... unclear (or I am reading it wrong). I need to support an application/x-www-form-urlencoded message where some properties are application/json encoded objects. From what I can read in the spec, this should be the expected, default serialization. The section on Encoding Object says: "The Content-Type for encoding a specific property. Default value depends on the property type: for object - application/json; ". I will say that I am very confused by that line when reading the section on "explode":

> When this is true, property values of type array or object generate separate parameters for each value of the array, or key-value-pair of the map. For other types of properties this property has no effect. When [style](https://swagger.io/specification/#encoding-style) is form, the default value is true. For all other styles, the default value is false. This property SHALL be ignored if the request body media type is not application/x-www-form-urlencoded or multipart/form-data. If a value is explicitly defined, then the value of [contentType](https://swagger.io/specification/#encoding-content-type) (implicit or explicit) SHALL be ignored.

That doesn't make sense to me since if I want my property to simply be a json string, then I need to set it to explode: false but then it ignores contentType so ... I guess I have to assume it is application/json but then what would be the point of this parameter. 

So, could someone tell me how to make the system accept a application/x-www-form-urlencoded message where one of the properties is a json encoded object. I attached a sample spec I created which I tested in Postman and SwaggerUI and they behave differently. I also tried with a codegen tool to see what they did and it was different also. I want to know what is "right" and then maybe I can fix something :)

Thanks
[NestedOpenAPISpec.yml.txt](https://github.com/OAI/OpenAPI-Specification/files/12100426/NestedOpenAPISpec.yml.txt)

LGraber:
To add, Postman always does comma delimted for the explode:false and explode:false, contentType:application/json. SwaggerUI on the other hand, also does json for the property value. It feels like I should be able to have either by using contentType attribute and it seems like the default should be application/json and not csv??

handrews:
@LGraber 

I agree that this is not very clear.  My reading is that this is the key part:

> If a value is explicitly defined, then the value of `contentType` (implicit or explicit) SHALL be ignored.

This tells me that the text around defaults is a bit incomplete.  An _explicit_ `explode: false` causes the `contentType` to be ignored, but omitting `explode` _entirely_ allows `contentType` to take effect.

[...]

Issue #2282, opened 2020-07-05
Questions about schemas for multipart and application/x-www-form-urlencoded request bodies


hkosova:
1\. The [Special Considerations for multipart Content](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#special-considerations-for-multipart-content) section says:

> In contrast to 2.0, a `schema` is REQUIRED to define the input parameters to the operation when using `multipart` content.

Please consider including this requirement in the description of [Media Type Object.`schema`](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#user-content-mediatypeschema), where it can be easily found.

2\. Is `schema` also required for `application/x-www-form-urlencoded` request bodies?

3\. All `multipart/*` and `application/x-www-form-urlencoded` examples in the spec use schemas with `type: object`, where schema `properties` define individual form fields. Does this mean that `type: object` is the only allowed schema `type` for these media types? If yes, please consider mentioning this explicitly somewhere. If not, how would other schema `types` work here?

handrews:
@hkosova I reworked a lot of that section in PR #2200 for OAS 3.1.  I'm not sure I addressed these questions but you might want to take a look at the changes as things might show up differently now.

As for my personal thoughts off the top of my head:

2.  I think this is a question of defaulting.  If a missing schema is the same as an empty schema (the usual defaulting approach JSON Schema) which allows everything, then omitting it means the request body might have any content.  A `false` schema in OAS 3.1 (or `{"not": {}}` in OAS 3.0 or earlier) would then be needed to forbid a request body.    I'd personally recommend doing that- if you want to constrain the set of valid request bodies, you need a schema.  Otherwise it's an empty schema by default so it could be anything.

3.  I believe it's simply that no other type makes sense.  I'd avoid saying that `type: object` is the only allowed type, as someone could theoretically figure out something else that makes sense.  Maybe.  What I'd prefer here is to talk about how `multipart/*` and `application/x-www-form-urlencoded` media types map into the JSON Schema data model.  Once you do that, the relevant `type` value or values should be obvious.  If you write a thing that can only be mapped into an object, then of course `type: object` is the only useful type.  It's not a special case, it just flows naturally.  I'd be happy to help work on wording for this if there is interest.

handrews:
oops wrong issue

handrews:
@OAI/tsc review request:  I'm not quite sure what to do with this.  

[...]

PR #3857, opened 2024-05-25
Encoding Object content and header clarifications (3.0.4)

handrews:
***Rendering with all currently open 3.0.4 changes (as of the timestamp of this edit - use the dropdown on the word "edited" in the bar for this comment) can be [found here](https://handrews.github.io/renderings/oas/deploy/oas/v3.0.4.html#encoding-object).***

Fixes:
* #1551 (by RECOMMENDING how to handle apparent `contentType` and `style`/`explode`/`allowReserved` conflicts)
* #2282 (by clarifying how things are correlated between schema, encoding, and data)
* #3331 (similar to 1551)
* #3739 (by rephrasing how the comma-separated-list value of `contentType` works)
* other un-filed confusions such as the conditions under which `headers` can be used

The Encoding Object's `contentType` field takes a comma-separated list of either regular or wildcared media types.  These are the "two types" mentioned in the previous wording – "two" here did *not* refer to a limit on the number of entries in the list.

These are not exactly media-type or media-range values, as both of those include parameters.

This change also moves the hard-to-follow list of default values out of the individual field cell and into its own table.

It takes `Content-Disposition` out of the header field's cell and instead explains limitations on header usage, and explains how `Content-Disposition` is used for encoding.

This explanation includes a suggestion on how other `multipart` formats could be used with an Encoding Object, since their unnamed parts otherwise cannot be supported.

Finally, it clarifies the interaction between `contentType` and the three fields imported from the Parameter Object, by aligning the recommended (but not, for compatibility reasons, required) behavior with guidance added in 3.1.0.


versions/3.0.4.md
```diff
@@ -1609,19 +1609,51 @@ An `encoding` attribute is introduced to give you control over the serialization
...
-<a name="encodingHeaders"></a>headers | Map[`string`, [Header Object](#headerObject) \| [Reference Object](#referenceObject)] | A map allowing additional information to be provided as headers, for example `Content-Disposition`.  `Content-Type` is described separately and SHALL be ignored in this section. This property SHALL be ignored if the request body media type is not a `multipart`.
+<a name="encodingContentType"></a>contentType | `string` | The `Content-Type` for encoding a specific property. The value is a comma-separated list, each element of which is either a specific media type (e.g. .`image/png`) or a wildcard media type (e.g. `image/*`). Default value depends on the property type as shown in the table below.
```
ralfhandl, line 1617:
```suggestion
<a name="encodingContentType"></a>contentType | `string` | The `Content-Type` for encoding a specific property. The value is a comma-separated list, each element of which is either a specific media type (e.g. `image/png`) or a wildcard media type (e.g. `image/*`). Default value depends on the property type as shown in the table below.
```

versions/3.0.4.md
```diff
@@ -1609,19 +1609,51 @@ An `encoding` attribute is introduced to give you control over the serialization
...
+Property Type | Property Format | Default `contentType`
+------------- | --------------- | ---------------------
+`string` | `binary` | `application/octet-stream`
+`string`, `number`, `integer`, or `boolean` | _n/a_ | `text/plain`
+`object` | _n/a_ | `application/json`
+`array` | _n/a_ | according to the `type` and `format` of the `items` schema
+
+###### Encoding `multipart` Media Types
+
+The `form-data` disposition and its `name` parameter that is used to correlate parts with encodings are mandatory for `multipart/form-data` ([RFC7578 §4.2](https://www.rfc-editor.org/rfc/rfc7578.html#section-4.2)).
```
ralfhandl, line 1636: I like having links directly to the relevant sections of referenced documents. Unfortunately this seems to be not supported by [ReSpec](https://respec.org/docs/#references-0), and links like these are not (automatically) added to the References appendix.

versions/3.0.4.md
```diff
@@ -1609,19 +1609,51 @@ An `encoding` attribute is introduced to give you control over the serialization
...
```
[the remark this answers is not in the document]
handrews: `§` means section, `¶` means paragraph.  Maybe this is a US vs German thing?  But in our rendered HTML of the spec, we use `§` by every section header as a link to that section.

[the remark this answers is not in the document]
ralfhandl: Yep, `§` does have different meanings in English and German - learned something today 😄.

I'd still feel more comfortable with using the unambiguous word "section".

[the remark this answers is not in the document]
handrews: @ralfhandl please take that up in [issue 3785 Style guide / release checklist for the specification](https://github.com/OAI/OpenAPI-Specification/issues/3785).  I've used § all over the place and I don't want to introduce inconsistencies in this one PR.  We can go over all of these sorts of stylistic nits at once before we ship.  It's too much noise to go back and forth on them in one random PR (much less all of them simultaneously).

versions/3.0.4.md
ralfhandl: Opened #3863 to track this

versions/3.0.4.md
```diff
@@ -1609,19 +1609,51 @@ An `encoding` attribute is introduced to give you control over the serialization
...
+
+Note also that `Content-Transfer-Encoding` is deprecated for `multipart/form-data` ([RFC7578 §4.7](https://www.rfc-editor.org/rfc/rfc7578.html#section-4.7)) where binary data is supported, as it is in HTTP.
+Using `format: byte` for a multipart field is equivalent to setting `Content-Transfer-Encoding: base64`.
+If `format: byte` is used along with setting a different `Content-Transfer-Encoding` value with the `headers` field, the result is undefined.
+
+###### Encoding the `application/x-www-form-urlencoded` Media Type
+
+The role of `contentType` with `application/x-www-form-urlencoded` request bodies was not described in detail in version 3.0.3 and earlier of this specification.
+To match the intent of these fields and be compatible with version 3.1 of this specification, it is RECOMMENDED that whenever any of `style`, `explode`, or `allowReserved` are present with an explicit values, that the value of `contentType`, whether it is explicitly defined or has the default value, be ignored.
```
notEthan, line 1652: /with an explicit values/with an explicit value/

versions/3.0.4.md
```diff
@@ -1609,19 +1609,51 @@ An `encoding` attribute is introduced to give you control over the serialization
...
+Likewise, if all three of those fields are absent, it is RECOMMENDED that they be ignored, and that encoding be based on `contentType` alone.
```
notEthan, line 1653: I'm not sure what it means to ignore these fields when they are not present. Ignore any default value, maybe?

handrews: Yeah, that's what it means, I'll improve the wording.  Thanks!  This is a weird thing to even have to explain, and has to do with how the Encoding Object was kind-of spliced together from ideas that weren't designed together.

versions/3.0.4.md
handrews: Should be fixed now- thanks!

versions/3.0.4.md
handrews: OK I ended up reworking this section a bit to make all of the default / explicit interactions more clear.

@notEthan please let me know what you think of this version!

notEthan (approved):
looks good

ralfhandl (dismissed):

handrews:
@notEthan @ralfhandl I'm afraid I did a little re-arranging after resolving conflicts because it got less clear.  This now splits the fields like they are in the Parameter Object and the Header Object (now that both of those have their own tables structured the same way, the Encoding Object became a weird exception to the pattern).

This will conflict with the percent-encoding PR, but it would have anyway.  I'll resolve conflicts for whichever isn't merged first.

versions/3.0.4.md
```diff
@@ -1621,19 +1621,67 @@ An `encoding` attribute is introduced to give you control over the serialization
...
+<a name="encodingHeaders"></a>headers | Map[`string`, [Header Object](#headerObject) \| [Reference Object](#referenceObject)] | A map allowing additional information to be provided as headers. `Content-Type` is described separately and SHALL be ignored in this section. This property SHALL be ignored if the request body media type is not a `multipart`.
+
+This object MAY be extended with [Specification Extensions](#specificationExtensions).
+
+The default values for `contentType` are as follows:
+
+Property Type | Property Format | Default `contentType`
+------------- | --------------- | ---------------------
+`string` | `binary` | `application/octet-stream`
+`string`, `number`, `integer`, or `boolean` | _n/a_ | `text/plain`
```
ralfhandl, line 1643: Do we have to consider `type: null` here?

versions/3.0.4.md
```diff
@@ -1621,19 +1621,67 @@ An `encoding` attribute is introduced to give you control over the serialization
...
+
+* All three keywords are to be entirely ignored, rather than treated as having their default values
+* Encoding is to be based on `contentType` alone, whether it is present with an explicit value or absent and treated as having its default value
+
+This makes the presence of at least one of `style`, `explode`, or `allowReserved` with an explicit value equivalent to using `schema` with `in: query` Parameter Objects, and their absence the equivalent of using `content`, but with the media type specified in `contentType` rather than through a Media Type Object.
+
+###### Encoding `multipart` Media Types
+
+The `form-data` disposition and its `name` parameter are mandatory for `multipart/form-data` ([RFC7578 §4.2](https://www.rfc-editor.org/rfc/rfc7578.html#section-4.2)).
+Array properties are handled by applying the same `name` to multiple parts, as is recommended by [RFC7578 §4.2](https://www.rfc-editor.org/rfc/rfc7578.html#section-4.3) for supplying multiple values per form field.
```
ralfhandl, line 1673:
```suggestion
Array properties are handled by applying the same `name` to multiple parts, as is recommended by [RFC7578 §4.3](https://www.rfc-editor.org/rfc/rfc7578.html#section-4.3) for supplying multiple values per form field.
```

ralfhandl (commented):
Minor nits

versions/3.0.4.md
handrews: This is 3.0.4 so it's `nullable: true` and I wasn't sure if that should be handled differently as it's not a `type` keyword thing in this version.

We could change this from the value of keyword `type` to the JSON data model type (string, number, boolean, null) rather than the `type` + `nullable` keyword.

In any event, I suspect in 3.0 it is supposed to be determined by the non-`null` type.  Hmm... we'll have to figure that out for 3.1 as no one has ever asked what happens for a dual (or more) typed field.

ralfhandl (dismissed):

versions/3.0.4.md
handrews: @ralfhandl the most recent force-push adds a commit that adds some discussion of `null` and reformats the table a bit.

ralfhandl (approved):

Issue #3734, opened 2024-04-21
Did we break 3.2 compatibility by removing the Path Item $ref special case?


handrews:
It grieves me to even bring this up, as no one is a fan of the special-case behavior of `$ref` with siblings in the Path Item Object.  But I feel that it's extremely important to restore trust by making _all valid 3.1 OADs_ also be _valid 3.2 OADs_.  The change in PR #2657 would invalidate certain 3.1 OADs.

We did _deprecate_ the special case in 3.1.1 with PR #2656, but we don't really have a published deprecation policy that would allow us to break compatibility.

I hesitate to suggest any sort of creative workaround here, as _any_ exception to compatibility will just remind everyone of the 3.1 experience.

ralfhandl:
I don't think that PR #2657 technically invalidates 3.1 OADs: the value of a path still is a JSON object that MAY contain `$ref` and any other fields next to it, so any 3.1 OAD should pass structure validation with 3.2 rules.

The semantic change is that
- in 3.1
  > In case a Path Item Object field appears both in the defined object and the referenced object, the behavior is undefined.

  Which means that the behavior is undefined even for `summary` and `description` 😮 
- in 3.2 an object containing `$ref` is a [Reference Object](https://spec.openapis.org/oas/v3.1.0#reference-object) which
  - explicitly allows `summary` and `description` and defines how to interpret them
  - *requires* to ignore additional properties, and thereby acknowledges that they may exist:
    > any properties added SHALL be ignored.

If we worry about implementations that don't ignore these properties and have some specific behavior not defined by the spec, we could relax that sentence to
> This object SHOULD NOT be extended with additional properties and any properties added SHOULD be ignored.

Implementations ignoring additional properties in Reference Objects are still compliant, and implementations with special behavior for additional properties in former Path Item Objects with `$ref` that now must be interpreted as Reference Objects also are.

As are future implementations that ignore both the SHOULD NOT and the SHOULD 🤷🏻‍♂️.

handrews:
@ralfhandl I definitely see your point here, and it's worth considering that approach.  However, I'm concerned about limiting compatibility to whether or not past OADs are _syntactically_ invalidated.

[...]

PR #3860, opened 2024-05-27
Path Item $ref change warning (3.1.1)

handrews:
Fixes:
* #3734 

Previously we deprecated the special $ref behavior entirely, but doing so violates the compatibility guidelines that we have re-established for ourselves within 3.x.  This wording allows changes in 3.2, but emphasizes convergence rather than discouraging the usage.

***NOTE:** I'm not sure what we should do for this in 3.2, but options include treating `summary` and `description` like in the Reference Object but retaining the special behavior for the other entries, and/or allowing the same overwriting for some or all other Path Item entries (although I am reluctant to open that can of worms with `parameters`).  We could also just formalize the behavior that is currently undefined in some way (e.g. make it a clear error - undefined behavior can be changed to anything without breaking compatibility, but defined behavior cannot be removed).*


versions/3.1.1.md
```diff
@@ -829,7 +829,7 @@ The path itself is still exposed to the documentation viewer but they will not k
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="pathItemRef"></a>$ref | `string` | Allows for a referenced definition of this path item. The referenced structure MUST be in the form of a [Path Item Object](#pathItemObject).  In case a Path Item Object field appears both in the defined object and the referenced object, the behavior is undefined. See the rules for resolving [Relative References](#relativeReferencesURI). <br><br>**Deprecated:** Usage of the `$ref` property has been deprecated when accompanied with properties other than `summary` and `description`.
+<a name="pathItemRef"></a>$ref | `string` | Allows for a referenced definition of this path item. The referenced structure MUST be in the form of a [Path Item Object](#pathItemObject).  In case a Path Item Object field appears both in the defined object and the referenced object, the behavior is undefined. See the rules for resolving [Relative References](#relativeReferencesURI). <br><br>_**Note:** The behavior of `$ref` with adjacent properties is likely to change in future versions of this specification to bring it into closer alignment with the behavior of the Reference Object._
```
ralfhandl, line 832:
```suggestion
<a name="pathItemRef"></a>$ref | `string` | Allows for a referenced definition of this path item. The referenced structure MUST be in the form of a [Path Item Object](#pathItemObject).  In case a Path Item Object field appears both in the defined object and the referenced object, the behavior is undefined. See the rules for resolving [Relative References](#relativeReferencesURI). <br><br>_**Note:** The behavior of `$ref` with adjacent properties is likely to change in future versions of this specification to bring it into closer alignment with the behavior of the [Reference Object](#referenceObject)]._
```

ralfhandl (commented):
+1, please add link

ralfhandl (approved):

lornajane (approved):

ralfhandl (approved):

PR #3861, opened 2024-05-27
Various minor editorial improvements (3.0.4)

handrews:
Fixes:

* #1423 
* #1867 
* #2034 (final resolution to link to the Learn site rather than adding a full-scale example up front)
* #2106 
* #2160 
* #2785

Issue #3349, opened 2023-08-15
Example contains deprecated property


energister:
Schema Object example [contains](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#model-with-example) `example` property that [has been deprecated](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#fixed-fields-20).

LasneF:
@handrews  ; Most of the job has been done 

there is one remaiing 
https://github.com/OAI/OpenAPI-Specification/blob/v3.1.1-dev/versions/3.1.1.md?plain=1#L1991 

but still we could close it ... or fix the last one

handrews:
@LasneF thanks for checking!  Those two are actually the `example` field in the Media Type Object, which supports both `example` and `examples`, which are distinct from the JSON Schema `examples` or the OAS legacy Schema Object `example` 🤦 

The whole mess has been explained better with PR #3895, so I'm going to close this as it should be as clear as it gets at this point.

I did search 3.1.1.md for `/examples?"?:/` and didn't find anything that looked incorrect to me.

Issue #2503, opened 2021-03-16
Spelling: WebHook -> Webhook


dret:
3.1.0 talks about "WebHook" (confusingly in the section on callbacks, but that's a different issue). the correct spelling is "Webhook" (https://en.wikipedia.org/wiki/Webhook). i couldn't quite figure out what to create a PR against, but if somebody points me in the right direction, i'd happily do that.

MikeRalphson:
@dret we now have a `v3.1.1-dev` branch containing a `versions/3.1.1.md` file which would be the target of this PR.

I think the "WebHook" (sic) wording pre-dates the introduction of the `webhooks` object (containing callbacks which are registered out-of-band) - so it might be better to find rewording for this sentence as you noted, but a first PR just correcting the capitalisation would also be a fine first step.

handrews:
PR merged!

PR #3862, opened 2024-05-27
Minor webhooks/callbacks and schema `examples` improvements (3.1.1)

handrews:
Fixes:
* #2503 
* #3349 

These are the minor editorial issues that do not apply to 3.0.4 and therefore do not fit in PR #3861 (which will be forward-ported as usual, but separately to ensure no one misses the 3.1.1-specific improvements).

Hopefully I did the right thing with webhooks vs callbacks here.

### DIFF
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -2587,10 +2587,12 @@ additionalProperties:
   "required": [
     "name"
   ],
-  "example": {
-    "name": "Puma",
-    "id": 1
-  }
+  "examples": [
+    {
+      "name": "Puma",
+      "id": 1
+    }
+  ]
 }
 ```
 
@@ -2604,8 +2606,8 @@ properties:
     type: string
 required:
 - name
-example:
-  name: Puma
+examples:
+- name: Puma
   id: 1
 ```
 
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -2021,7 +2021,7 @@ $response.header.Location | https://example.org/subscription/1
 
 ##### Callback Object Examples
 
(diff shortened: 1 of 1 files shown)


versions/3.1.1.md
```diff
@@ -2021,7 +2021,7 @@ $response.header.Location | https://example.org/subscription/1
 
 ##### Callback Object Examples
 
-The following example uses the user provided `queryUrl` query string parameter to define the callback URL.  This is an example of how to use a callback object to describe a WebHook callback that goes with the subscription operation to enable registering for the WebHook.
+The following example uses the user provided `queryUrl` query string parameter to define the callback URL.  This is similar to a [`webhooks`](#oasWebhooks), but differs in that the callback only occurs because of the initial request that sent the `queryUrl`.
```
ralfhandl, line 2024:
```suggestion
The following example uses the user provided `queryUrl` query string parameter to define the callback URL.  This is similar to a [`webhooks`](#oasWebhooks) callback, and differs in that the callback only occurs because of the initial request that sent the `queryUrl`.
```

ralfhandl (dismissed):
+1, formulation suggestion

versions/3.1.1.md
handrews: I can't get behind this one.  There is an error to fix here which is that it should have been "webhook" (singular) rather than "webhooks".  I really don't want to conflate webhooks and callbacks here- that's why I reworded it in the first place.

As for `and` vs `but`, the point here is to emphasize the different (`but`) rather than the similarity (`and`).

I'll fix the plural, but I'd rather not do the others unless there's some aspect of the definitions of "webhook" and "callback" that I'm missing (_which is very possible so please do lmk if you think so)_.

miqui (approved):

Issue #3761, opened 2024-04-27
"body parameter" example in 3.x does not fit how Request Body Objects work


handrews:
In the Request Body Examples section of both [3.0](https://spec.openapis.org/oas/v3.0.3#request-body-examples) and [3.1](https://spec.openapis.org/oas/v3.1.0#request-body-examples), the second example is as follows (JSON omitted for brevity):

-----

A body parameter that is an array of string values:

```YAML
description: user to add to the system
required: true
content:
  text/plain:
    schema:
      type: array
      items:
        type: string
```

-----

This is identical to an example in the 2.0 spec, where the request body was treated as a parameter.  I'm not entirely sure how it was supposed to be serialized even there, as the 3.x `style` and `explode` fields don't exist, but I definitely don't see how to serialize it in 3.0.  There is no JSON Schema data model defined for `text/plain`, and nothing in the Request Body Object talks about how to serialize a list of values into plain text (Is it a comma-separated list?  Is it one value per line?)

What was this example trying to show, and how can we show it more clearly?

handrews:
@OAI/tsc review request: see above

karenetheridge:
I expect some git spelunking (in an old swagger repo, perhaps)? might be needed to derive the original intent here. I agree it makes no sense at all - text/plain's serializer/deserializer is the identity function (and indeed my implementation has this hardcoded for 'text/*')!

darrelmiller:
I do not believe this is a helpful example.  I do not believe we have ever talked about how a JSON model should be serialized in text/plain payloads.

mikekistler:
I agree that this is not a helpful example and we should make it more useful.

But I think there is a larger question here which is: how to interpret -- or whether to ignore -- `schema` for various media types. Does it make any sense for `text/plain` content to have any schema other than `type: string`? Likewise, for `image/<anything>` does it make any sense to have any schema other than `type: string, format: binary`?

In other words, I think that certain content types _imply_ the schema -- maybe even _impose_ a schema.  I realize this is a much larger topic than just fixing an example, but shouldn't we address this somehow?

handrews:
@mikekistler I have had similar thoughts that there's a general topic here.  I wasn't going to bring it up yet, but since you did (and I'm glad to hear someone else have this sort of thought), I have started a discussion on a registry for data modeling:

* #3771 

[...]

PR #3864, opened 2024-05-28
Improve media type serialization examples (3.0.4)

handrews:
_I know I said I was done with PRs until the current ones and/or some blocking issues are resolved, but I realized I forgot to post this one._

Fixes:
* #3761 
* Some other issue asking about `form-urlencoded` bodies and application/json values that I cannot for the life of me find right now

* Replace the outdated "model" terminology with "schema"
* Remove the outdated `text/plain` array example, which does not correlate with current OAS requirements
* Rather than replacing the `text/plain` example direclty, enhance the example of serializing `application/json` content in `application/x-www-form-urlencoeded` request bodies.

### DIFF
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -1297,7 +1297,7 @@ This object MAY be extended with [Specification Extensions](#specificationExtens
 
 ##### Request Body Examples
 
-A request body with a referenced model definition.
+A request body with a referenced schema definition.
 ```json
 {
   "description": "user to add to the system",
@@ -1372,35 +1372,6 @@ content:
         summary: User example in other format
         externalValue: http://foo.bar/examples/user-example.whatever
 ```
-
-A body parameter that is an array of string values:
-```json
-{
-  "description": "user to add to the system",
-  "content": {
-    "text/plain": {
-      "schema": {
-        "type": "array",
-        "items": {
-          "type": "string"
-        }
-      }
-    }
-  }
-}
-```
-
-```yaml
-description: user to add to the system
-required: true
-content:
-  text/plain:
-    schema:
-      type: array
-      items:
-        type: string
-```
-
 
 #### <a name="mediaTypeObject"></a>Media Type Object
 Each Media Type Object provides schema and examples for the media type identified by its key.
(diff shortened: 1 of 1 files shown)


versions/3.0.4.md
```diff
@@ -1557,9 +1528,35 @@ requestBody:
...
+```
+
+Assuming the most compact representation of the JSON value (with unnecessary whitespace removed), we would expect to see the following request body, where space characters have been replaced with `+` and `"`, `{`, and `}` have been percent-encoded to `%22`, `%7B`, and `%7D`, respectively:
 
-When passing complex objects in the `application/x-www-form-urlencoded` content type, the default serialization strategy of such properties is described in the [`Encoding Object`](#encodingObject)'s [`style`](#encodingStyle) property as `form`.
+```urlencoded
+id=f81d4fae-7dec-11d0-a765-00a0c91e6bf6&address=%7B%22streetAddress%22:%22123+Example+Dr.%22,%22city%22:%22Somewhere%22,%22state%22:%22CA%22,%22zip%22:99999%7D
+```
+
+Note that the `id` keyword is treated a `text/plain` per the [Encoding Object](#encodingObject)'s default behavior, and is serialized as-is.
```
ralfhandl, line 1552:
```suggestion
Note that the `id` keyword is treated as `text/plain` per the [Encoding Object](#encodingObject)'s default behavior, and is serialized as-is.
```

ralfhandl (commented):
+1, one typo

versions/3.0.4.md
```diff
@@ -1557,9 +1528,35 @@ requestBody:
...
+When passing complex objects in the `application/x-www-form-urlencoded` content type, the default serialization strategy of such properties is described in the [Encoding Object](#encodingObject)'s [`style`](#encodingStyle) property as `form`.
+
+With this example, given an `id` of `f81d4fae-7dec-11d0-a765-00a0c91e6bf6` and a US-style address as follows:
+
+```json
+{
+  "streetAddress": "123 Example Dr.",
+  "city": "Somewhere",
+  "state": "CA",
+  "zip": 99999
```
notEthan, line 1542: This is extremely minor/beside the point, but a zip should be a string as they can start with 0 (though, maybe not beside the point if wanting to show encoding of a number)

handrews: Let's go with string- technically ZIP can also be something like 99999+1234, so I'll use that so it's not ambiguous.  Good catch!  It's always better if examples don't confuse with extraneous details.

notEthan (approved):

miqui (approved):

handrews:
Two TSC approvals - merging.

PR #3875, opened 2024-06-03
Proposal: Bundled streams w/ self-identification

handrews:
_[**NOTE**: Please prioritize the open 3.0.4 and 3.1.1 PRs]_

This adds a proposal for bundling both OAS (including full 3.1 support) and Arazzo using YAML native streams or any of several JSON document stream formats.  It includes self-identification of documents to ensure that all references within the stream can be resolved without looking outside of the stream (except for External Documents).

Also paging @frankkilcommins for the Arazzo aspect (GitHub won't let me make you a reviewer, for who knows what reason).


proposals/2024-06-01-Self-Identification-and-Bundling.md
```diff
@@ -0,0 +1,123 @@
...
+
+### Most tools do not fully support OAS 3.1
+
+Many referencing preprocessors only work in OAS 3.0 or earlier, because they violate the [full-document, JSON Schema keyword-aware](https://github.com/OAI/OpenAPI-Specification/pull/3758).  Supporting referencing as a preprocessor requires handling not only the [many different `$ref` variations](https://github.com/OAI/Arazzo-Specification/issues/181#issuecomment-2085586524), but also the `$id`, `$anchor`, `$dynamicAnchor`, and `$dynamicRef` keywords.  While it is theoretically possible to preprocess `$dynamicRef` (aside from circular references), it can cause an [exponential growth in document size](https://arxiv.org/pdf/2307.10034).  Dynamic referencing is an important technique for [modeling generic data types](https://github.com/OAI/OpenAPI-Specification/pull/3714).
```
ralfhandl, line 38:
```suggestion
Many referencing preprocessors only work in OAS 3.0 or earlier, because they violate the [full-document, JSON Schema keyword-aware parsing requirement](https://github.com/OAI/OpenAPI-Specification/pull/3758).  Supporting referencing as a preprocessor requires handling not only the [many different `$ref` variations](https://github.com/OAI/Arazzo-Specification/issues/181#issuecomment-2085586524), but also the `$id`, `$anchor`, `$dynamicAnchor`, and `$dynamicRef` keywords.  While it is theoretically possible to preprocess `$dynamicRef` (aside from circular references), it can cause an [exponential growth in document size](https://arxiv.org/pdf/2307.10034).  Dynamic referencing is an important technique for [modeling generic data types](https://github.com/OAI/OpenAPI-Specification/pull/3714).
```

proposals/2024-06-01-Self-Identification-and-Bundling.md
```diff
@@ -0,0 +1,123 @@
...
+Building on the JSON Schema bundling model will ensure that, as much as possible:
+
+* New behavior will be identical or at least analogous to behavior that is already required, making it easier to support
+* The mental model will parallel one that is already successful
+
+## Proposed solution
+
+These challenges can be solved by combining two already-existing concepts:
+
+1. A simplified analog of JSON Schema's `$id` that appears in exactly one place: the `self` field in root OpenAPI Objects / Arazzo Objects
```
ralfhandl, line 59:
```suggestion
1. A simplified analog of JSON Schema's `$id` that appears in exactly one place: a new `self` field in root OpenAPI Objects / Arazzo Objects
```

proposals/2024-06-01-Self-Identification-and-Bundling.md
```diff
@@ -0,0 +1,123 @@
...
+
+...
+
+\#\# OpenAPI Description Structure
+
+...
+
+\#\#\# Bundling Documents as YAML or JSON Streams
+
+Multiple OpenAPI Description documents MAY be bundled in a YAML stream ([RFC 9512 §3.2](https://www.rfc-editor.org/rfc/rfc9512.html#name-yaml-streams)) or a JSON streaming format such as JSON Text Sequences [RFC 7464](https://www.rfc-editor.org/rfc/rfc7464), [JSON Lines](https://jsonlines.org/), or [NDJSON](https://github.com/ndjson/ndjson-spec).  Documents bundled in this way MUST set their own URI using the `self` field in the [OpenAPI Object](#oasObject).  If a document in the stream has `self` set to a relative URI-reference, it MUST be resolved relative to the location of the entire stream.  However, it is strongly RECOMMENDED to set `self` to an absolute-URI for use within multi-document streams.
```
ralfhandl, line 88:
```suggestion
Multiple OpenAPI Description documents MAY be bundled in a YAML stream ([RFC 9512 §3.2](https://www.rfc-editor.org/rfc/rfc9512.html#name-yaml-streams)) or a JSON streaming format such as JSON Text Sequences [RFC 7464](https://www.rfc-editor.org/rfc/rfc7464), [JSON Lines](https://jsonlines.org/), or [NDJSON](https://github.com/ndjson/ndjson-spec).  Documents bundled in this way MUST set their own URI using the `self` field in the [OpenAPI Object](#oasObject).  If a document in the stream has `self` set to a relative URI-reference, it MUST be resolved relative to the location of the entire stream.  However, it is strongly RECOMMENDED to set `self` to an absolute URI for use within multi-document streams.
```

proposals/2024-06-01-Self-Identification-and-Bundling.md
```diff
@@ -0,0 +1,123 @@
...
+
+As [OAS 3.1.1 clarifies](https://github.com/OAI/OpenAPI-Specification/pull/3758), it is already mandatory to separate location and identity for Schema Object support.  Currently, associating a URI other than the current URL with a document to meet this requirement has to be done externally.  Many tools effectively support this in the form of allowing the retrieval URL to be set manually, without verifying that the document actually lives at the given URL.
+
+The various streaming formats do not state how to resolve links among the parts, as noted in [RFC 9512 YAML Media Type §3.2](https://www.rfc-editor.org/rfc/rfc9512.html#name-yaml-streams), which makes an explicit analogy to `application/json-seq` for this behavior.  
+
+A `self` field that is a relative URI-reference would be resolved against the document location just as all Reference Object and similar URIs are resolved in OAS 3.1.  A relative reference within a stream would resolve against the URL of the entire stream; however it is probably better to either RECOMMEND or require (MUST) resolving `self` to an absolute-URI when bundling into a stream for maximum predictability.
```
ralfhandl, line 72:
```suggestion
A `self` field that is a relative URI-reference would be resolved against the document location just as all Reference Object and similar URIs are resolved in OAS 3.1.  A relative reference within a stream would resolve against the URL of the entire stream; however it is probably better to either RECOMMEND or require (MUST) resolving `self` to an absolute URI when bundling into a stream for maximum predictability.
```

proposals/2024-06-01-Self-Identification-and-Bundling.md
```diff
@@ -0,0 +1,123 @@
...
+
+\#\#\# Fixed Fields
+
+Field Name | Type | Description
+---|:---|:---
+self | `URI-reference` (without a fragment) | Sets the URI of this document, which also serves as its base URI in accordance with [RFC 3986 §5.1.1](https://www.rfc-editor.org/rfc/rfc3986#section-5.1.1); the value MUST NOT be the empty string and MUST NOT contain a fragment
+
+## Backwards compatibility
+
+OAS 3.2 and Arazzo 1.1 documents that do not use the `self` field will be have exactly the same as OAS 3.1 and Arazzo 1.0 documents.  The change in minor version is sufficient to manage the compatibility issues, as no software that only supports up to 3.1/1.0 should attempt to parse 3.2/1.1 documents.
```
ralfhandl, line 106:
```suggestion
OAS 3.2 and Arazzo 1.1 documents that do not use the `self` field will behave exactly the same as OAS 3.1 and Arazzo 1.0 documents.  The change in minor version is sufficient to manage the compatibility issues, as no software that only supports up to 3.1/1.0 should attempt to parse 3.2/1.1 documents.
```

ralfhandl (commented):
+1, minor nits

ralfhandl (approved):

proposals/2024-06-01-Self-Identification-and-Bundling.md
```diff
@@ -0,0 +1,123 @@
...
+|---- | ---------------- | ---------- |
+|2024-06-01 | @handrews | Initial submission
+
+## Introduction
+
+Poor support for external references has fractured the OAS tooling landscape, with many tools requiring multi-document OpenAPI Descriptions (OADs) to be combined into a single document.  Arazzo requires resolving sources and runtime expressions from multiple OADs, each of which might consist of multiple documents.  There is no way to combine all of the OAD and Arazzo documents involved into a single document, but an alternate solution would be a bundle similar to [what we recommend on our blog for bundling Schema Objects](https://www.openapis.org/blog/2021/08/23/json-schema-bundling-finally-formalised).  This would require a similar mechanism to JSON Schema's `$id` for OAS and Arazzo documents and the components within them.  It would also provide an alternative to current multi-to-single-document OAD tools, most (possibly all) of which do not fully support OAS 3.1, and allow for _lossless bundling of identifiable components_, which is increasingly needed by industry standards groups publishing API "building blocks" for use across many APIs by many different providers.
```
mikekistler, line 24: > There is no way to combine all of the OAD and Arazzo documents involved into a single document

I'd like to understand why this is the case. What about just OAD documents?

kevinswiber: Objects in descriptions are scoped to the document in which they originate. We don't really have semantics for merging multiple documents. In some cases, it would involve renaming unique identifiers like map keys. Tools that have done this create an inconsistent experience, and to reverse the process requires a source map of some kind. Maintaining document identity in a multi-document scenario obviates the need for that.

handrews: @mikekistler basically what @kevinswiber said, but I'll add two things:

I chatted with someone on some Slack a while ago who was having a lot of problems because his "merge to a single document" tools that he needed in order to get something (AWS gateway, maybe?) to accept the OAD was messing things up with what it though were "safe" transformations.  As Kevin pointed out, there are a lot of choices that tools doing this have to make, and those can be surprising and sometimes breaking.  When you start with a single document, split it, and then use the same toolchain to re-combine the pieces, that works quite well because the split and combine tool are the same and make the same assumptions.  But mixing toolchains doesn't work well here.

Another way to look at it is to look at how [JSON Schema bundling works](https://www.openapis.org/blog/2021/08/23/json-schema-bundling-finally-formalised).  It relies heavily on `"$id"` to preserve not just the referencing behavior but the literal reference values.  This way there isn't any "rewriting" of the documents, and no need to merge them.  They are simply placed in the standard location (`"$defs"`) and used in a way that the name under `"$defs"` is irrelevant.

As briefly mentioned in the call this morning, if we want to support `"$id"` (or an equivalent named differently to reduce confusion with JSON Schema) in every Object type in the OAS and Arazzo, we could absolutely implement single-document bundling.  But nested `"$id"`s make it much harder to determine the base URI, which is something that tooling vendors seem to struggle with (or are just indifferent to) as it is.  Using a document stream minimizes the changes and implementation work involved.

I'm trying to get the maximum improvement for the minimum effort with this proposal.

proposals/2024-06-01-Self-Identification-and-Bundling.md
```diff
@@ -0,0 +1,123 @@
...
+
+The author of this proposal is not aware of any reference preprocessing tool that fully supports OAS 3.1, although the last comprehensive survey on this was done in 2022.
+
+### Arazzo not supported
+
+All current tools depend on it being possible to structure OADs as single JSON or YAML documents.  This is not possible with Arazzo, as it coordinates multiple OADs without being part of any of them.  We do not yet know how much of a challenge this will be for Arazzo, but history suggests that the ecosystem will be healthier if a clear solution is endorsed early on.
```
kevinswiber, line 44: Could we make composite keys using an identifier for the root OAD?

proposals/2024-06-01-Self-Identification-and-Bundling.md
```diff
@@ -0,0 +1,123 @@
...
+
+...
+
+\#\# OpenAPI Object
+
+\#\#\# Fixed Fields
+
+Field Name | Type | Description
+---|:---|:---
+self | `URI-reference` (without a fragment) | Sets the URI of this document, which also serves as its base URI in accordance with [RFC 3986 §5.1.1](https://www.rfc-editor.org/rfc/rfc3986#section-5.1.1); the value MUST NOT be the empty string and MUST NOT contain a fragment
```
kevinswiber, line 102: For interleaving multiple OAD streams, we could also include an optional field to identify the bundle, perhaps using the entry document ID.  Ideally, it would allow multiple bundles to prevent duplicating documents in certain scenarios, but I'm usually the odd one out for being cool with arrays.

proposals/2024-06-01-Self-Identification-and-Bundling.md
handrews: @kevinswiber Arazzo identifying targets in OADs is not a problem, if that's what you mean.  I'm just kind-of assuming that since multi-document support has been a challenge (or just not viewed as cost-effective) for our community, publishing a spec that depends on reading and using multiple documents seems a touch risky.

proposals/2024-06-01-Self-Identification-and-Bundling.md
handrews: @kevinswiber I think this would be a good further direction, but I worry about making the initial thing too complicated.  Really, the only thing you need to solve for having multiple OADs in a stream is how to identify each entry document.  This proposal just says "the first document in the stream is the entry document", which means there can only be one clear OAD in the stream.  But there's no reason you couldn't get a stream, and then select (by URI) each entry document you want to try to parse from the stream.  As long as all needed documents are in the stream, it would work fine.

handrews:
Per yesterday's TDC call, I am going to split this into two proposals, as the `self` field met with wide acceptance, while there are still questions and concerns about bundling.  Ironically, these two were only together because I though the bundling piece would be enthusiastically embraced, and would be the only way to sell folks on the `self` field 🙃

PR #3824, opened 2024-05-19
Clarify Link Obj parameters as best we can (3.0.4)

handrews:
Fixes:
* #1143 
* #3276 

This acknowledges the ambiguity in the key and value syntax of the Link Object's `parameter` field, and provides a bit of guidance on how to implement it.  Sadly it is not possible to fully solve in a point release.

Issue #3276, opened 2023-05-11
Clarify parsing ambiguity of "parameters" in Link Object


handrews:
The question came up in today's call about fields that could take either a literal value or a runtime expression.  The `parameters` field in the Link Object allows this.  Probably we should clarify that anything that parses as a runtime expression SHOULD (MUST?) be handled that way, and that it's therefore not possible to specify a literal value that looks like a runtime expression (since creating an escaping syntax would require a minor version increment).

cc @webron @mkistler

handrews:
There's a further ambiguity: the property names under `parameters` are either parameter names or qualified with the parameter's `in` field, e.g. `path.id` vs `query.id`.  Parameter names are strings without any restrictions other than that imposed by their locations, and `.` is a valid character in at least the path and query parameter names.

In this case, there's an easy solution which is that if you (for some bizarre reason) have a parameter named `path.id`, then to reference it you would always qualifiy it (`path.path.id`, or I guess if you're particularly mind-bending ,`query.path.id`).

handrews:
PRs merged – closing!

Issue #1143, opened 2017-06-03
Problems in parameters field type for Link Object


timburks:
Found in Version 3.0.0-rc2

<img width="943" alt="screen shot 2017-06-03 at 9 27 21 am" src="https://cloud.githubusercontent.com/assets/405/26755311/9e88ca16-483f-11e7-8a70-d3981cbbeda7.png">

I guess this was intended to be `Map[string, Any|{expression}]`.

But would `Map[string,{expression}]` suffice?

darrelmiller:
The `Any` was included to allow linked operations to have certain parameters be provided with constant values.  The runtime expression syntax doesn't allow for expressing constant values.

It does the beg the question, how does tooling distinguish between a mistyped expression and literal string.  My initial thought would be to check the first character for a $ and then attempt to interpret it as an expression.  This does mean that strings that start with "$" would not be valid literal values.  Not exactly ideal.

handrews:
PRs merged – closing!

PR #3884, opened 2024-06-04
Clarify Link Obj parameters as best we can (3.2.0 port of #3824)

handrews:
Ports:
* #3824

Fixes:
* #1143 
* #3276 

This acknowledges the ambiguity in the key and value syntax of the Link Object's `parameter` field, and provides a bit of guidance on how to implement it.  Sadly it is not possible to fully solve in a point release.


lornajane (approved):
Thanks again!

PR #3906, opened 2024-06-13
Resolving implicit connections (3.0.4 port of #3856)

handrews:
* Ports #3856 

Since we chose the simpler of the two ways to handle this, we can just directly backport the guidance instead of needing to rework it for 3.0.  🎉 

_[**EDIT:** Except changing "URIs" to "URLs" in the relative resolution link, which I missed but have now added as a separate commit.]_

### DIFF
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -3802,7 +3802,7 @@ security:
 
 See [Resolving Implicit Connections](#resolvingImplicitConnections) for more information.
 
-First, our entry document is where parsing begins. It defines the `MySecurity` security scheme to be JWT-based, and it defines on Path Item as a reference to a component in another document:
+First, our entry document is where parsing begins. It defines the `MySecurity` security scheme to be JWT-based, and it defines a Path Item as a reference to a component in another document:
 
 ```HTTP
 GET /api/description/openapi HTTP/1.1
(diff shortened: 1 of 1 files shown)


versions/3.0.4.md
```diff
@@ -166,14 +166,50 @@ It is RECOMMENDED that the entry OpenAPI document be named: `openapi.json` or `o
...
+
+The interface approach can also work for Discriminator Objects and Schema Objects, but it is also possible to keep the Discriminator Object's behavior within a single document using the relative URI-reference syntax of `mapping`.
+
+There are no URI-based alternatives for the Security Requirement Object or for the Operation Object's `tags` field.
+These limitations are expected to be addressed in a future release.
+
+See [Security Requirement in a Referenced Document](#security-requirement-in-a-referenced-document) for an example of the possible resolutions, including which one is recommended by this section.
+The behavior for Discrimator Object non-URI mappings and for the Operation Object's `tags` field operate on the same principles.
+
+Note that no aspect of implicit connection resolution changes how [URLs are resolved](#relativeReferences), or restricts their possible targets.
```
ralfhandl, line 211: Why "URLs" here and "URIs" in v3.1.1?

handrews: Because the title of the section here is "Relative References in URLs".  Why would I say "URI" and link to a section that says "URLs"?

Also, everything in 3.0 is specified to behave as a URL.  If the text is inconsistent on this (which does not surprise me - "basae URI" should still be "base URI" and the XML Object's `namespace` field should be "URI", but probably nothing else) that is not a problem to solve in this PR.  Feel free to open a separate PR for that.

miqui (approved):
@ralfhandl Looks good to me.  Thoughts?

ralfhandl (approved):

Issue #3877, opened 2024-06-04
Clarify default dialect for referenced JSON Schema documents


handrews:
We need to clarify that `jsonSchemaDialect` (whether defaulted or explicitly set) governs the dialect throughout the document in which it appears, regardless of how the schema object was reached via referencing.

However, we need to also determine how to handle fragmentary documents, including standalone JSON Schema documents (which are Schema Object fragmentary documents from OAS's point of view).  The most logically consistent view is to say that it's always document-scoped, and if your document does not have an OpenAPI Object at the root, if you want to change the schema behavior you need to use `$id` and `$schema` to do it.

But it's definitely going to be implementation-defined for fragmentary documents except if schemas have `$schema` (and `$id` if the schema is not a document root, as `$id` is necessary for `$schema` to be usable in a non-document-root schema).

It might be good to wait until we know which of PR #3823 or PR #3855 get merged, as they are working with similar concepts.

handrews:
PR merged for 3.1.1!

PR #3907, opened 2024-06-13
Clarify discriminator non-impact on validation

### LINKED ISSUE #3901: OAS Dialect does not require the OAS vocabulary
[Schema Object#Properties](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#properties) states that 

> The OpenAPI Schema Object dialect is defined as requiring the OAS base vocabulary ...

yet the dialect schema does not require that vocab:

https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.1/dialect/base.schema.json#L16C57-L16C62

handrews:
* Fixes #3901 (`discriminator`-specific aspects)

Be very explicit that discriminator MUST NOT change the validation outcome, and explain the implication for the "allOf" use case.

### DIFF
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -3026,6 +3026,8 @@ When request bodies or response payloads may be one of a number of different sch
 This hint can be used to aid in serialization, deserialization, and validation.
 The Discriminator Object does this by implicitly or explicitly associating the possible values of a named property with alternative schemas.
 
+Note that `discriminator` MUST NOT change the validation outcome of the schema.
+
 ##### Fixed Fields
 Field Name | Type | Description
 ---|:---:|---
@@ -3034,10 +3036,15 @@ Field Name | Type | Description
 
 ##### Conditions for Using the Discriminator Object
 The Discriminator Object is legal only when using one of the composite keywords `oneOf`, `anyOf`, `allOf`.
+
 In both the `oneOf` and `anyOf` use cases, where those keywords are adjacent to `discriminator`, all possible schemas MUST be listed explicitly.
+
 To avoid redundancy, the discriminator MAY be added to a parent schema definition, and all schemas building on the parent schema via an `allOf` construct may be used as an alternate schema.
 It is implementation-defined as to whether all named [Schema Objects](#schemaObject) under the [Components Object](#componentsObject), or only those that are otherwise directly referenced are searched for `allOf` references to the parent schema.
 However, it is RECOMMENDED to search all named schemas in the Components Object because it is common with the `allOf` usage for other parts of the API to only directly reference the parent schema.
+
+The `allOf` form of `discriminator` is _only_ useful for non-validation use cases; validation with the parent schema with this form of `discriminator` _does not_ perform a search for child schemas or use them in validation in any way.
+This is because `discriminator` cannot change the validation outcome, and no standard JSON Schema keyword connects the parent schema to the child schemas.
 
 The behavior of any configuration of `oneOf`, `anyOf`, `allOf` and `discriminator` that is not described above is undefined.
 


jam01:
Would you consider changing [Schema Object # Properties](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#properties)

to go from 

> The OpenAPI Schema Object [dialect](https://tools.ietf.org/html/draft-bhutton-json-schema-00#section-4.3.3) is defined as requiring the [OAS base vocabulary](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#fixed-fields-20), in addition to the vocabularies as specified in the JSON Schema draft 2020-12 [general purpose meta-schema](https://tools.ietf.org/html/draft-bhutton-json-schema-00#section-8).

to 

> The OpenAPI Schema Object declares the OAS base vocabulary as optional, in addition to the vocabularies as specified in the JSON Schema draft 2020-12 general purpose meta-schema. When supported an implementation can optimize validation performance, but does not change the outcome as optional vocabularies do not directly impact validation.

Or something along those lines...?

handrews:
@jam01 I'll handle that in the fix to issue #3877 which is about dialects specifically.

handrews:
@jam01 although I probably won't go into that level of detail- more likely I will link to the JSON Schema documentation for `$vocabulary`.

versions/3.0.4.md
```diff
@@ -3034,11 +3036,16 @@ Field Name | Type | Description
 
 ##### Conditions for Using the Discriminator Object
 The Discriminator Object is legal only when using one of the composite keywords `oneOf`, `anyOf`, `allOf`.
```
ralfhandl, line 3038: Does this mean that the `Pet` example right above is illegal because it uses `discriminator` and lacks

```yaml
      anyOf:
```

???

versions/3.0.4.md
```diff
@@ -3026,6 +3026,8 @@ When request bodies or response payloads may be one of a number of different sch
 This hint can be used to aid in serialization, deserialization, and validation.
 The Discriminator Object does this by implicitly or explicitly associating the possible values of a named property with alternative schemas.
 
+Note that `discriminator` MUST NOT change the validation outcome of the schema.
```
ralfhandl, line 3029: This is surprising: if `discriminator` is a hint for picking the "right" schema out of an `anyOf` array, the result may differ if a validator uses the hint and successfully validates (only) against the hinted alternative, or if it sequentially tries all alternatives and stops on the first success, which may not be the one `discriminator` hints at.

Do we indirectly say here that validators MUST ignore `discriminator`?

mikekistler: Does JSON Schema proscribe an order in which the schemas of an `anyOf` are checked? I did not think so but I'm certainly no expert here.

ralfhandl: [Draft 04](https://json-schema.org/draft-04/draft-fge-json-schema-validation-00#rfc.section.5.5.4) only says

> An instance validates successfully against this keyword if it validates successfully against at least one schema defined by this keyword's value.

So my implementation would loop over the schemas in array order and stop at the first success. Why waste energy on the remaining schemas.

[2020-12](https://json-schema.org/draft/2020-12/json-schema-core#section-10.2.1.2) requires that

> all subschemas MUST be examined

As this PR is for 3.0.4 I would go with my interpretation of Draft-04.

mikekistler: Okay, so I don't think there is a requirement for validators to ignore the `discriminator`. If JSON Schema allows validators to choose the order of validation, then a validator is free to choose to start with the schema associated with the discriminator, or not.

versions/3.0.4.md
handrews: No, that's explained in the next few paragraphs.  Also, this is not part of the change in this PR.

versions/3.0.4.md
handrews: @ralfhandl 

> Do we indirectly say here that validators MUST ignore `discriminator`?

No, if you know what sort of static analysis of the entire schema that you need to do to calculate whether an `anyOf` branch is safe to skip _and_ you don't support annotations (which is sad because annotations are the easiest way to extend JSON Schema), you can technically use it to short-circuit.  I think @karenetheridge 's implementation does work out whether one of the annotation-consuming keywords like `unevaluatedProperties` is involved, for example, although I think she just uses it to decide whether it's necessary to collect property annotations for certain subschemas.

handrews: @ralfhandl 

> That implementaiton is incorrect for draft 2020-12. It's fine up through draft-07, but not after that.

I'm having a slow morning and have been forward-porting things to 3.1.1, so I was in a 3.1 mindset.  But this is on 3.0.4 🤦 

So in 3.0.4 it's entirely reasonable to short-circuit `anyOf` because draft-04 did not define how to collect annotations (much less have keywords relying on that sort of thing).

But that _still_ doesn't change the validation outcome.  I was very specific in my word choice of "validation outcome" as there are many other things about schema evaluation than just the validation outcome.

versions/3.0.4.md
ralfhandl: Found it, now that I know it is there 😄

ralfhandl (approved):

mikekistler (approved):
Looks good. 👍

PR #3908, opened 2024-06-13
Use document/description/definition consistently (3.0.4)

handrews:
This tries to be consistent about using "describe" rather than "define" and using "description" for the entire potentially-multi-document thing, and "document" (or sometimes "description document") to refer to individual documents in the description.

### DIFF
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -10,7 +10,7 @@ This document is licensed under [The Apache License, Version 2.0](https://www.ap
 
 The OpenAPI Specification (OAS) defines a standard, language-agnostic interface to RESTful APIs which allows both humans and computers to discover and understand the capabilities of the service without access to source code, documentation, or through network traffic inspection. When properly defined, a consumer can understand and interact with the remote service with a minimal amount of implementation logic.
 
-An OpenAPI definition can then be used by documentation generation tools to display the API, code generation tools to generate servers and clients in various programming languages, testing tools, and many other use cases.
+An OpenAPI description can then be used by documentation generation tools to display the API, code generation tools to generate servers and clients in various programming languages, testing tools, and many other use cases.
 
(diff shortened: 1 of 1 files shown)


ralfhandl (approved):

versions/3.0.4.md
```diff
@@ -126,11 +126,11 @@ The OpenAPI Specification is versioned using a `major`.`minor`.`patch` versionin
 
 Occasionally, non-backwards compatible changes may be made in `minor` versions of the OAS where impact is believed to be low relative to the benefit provided.
 
-An OpenAPI document compatible with OAS 3.\*.\* contains a required [`openapi`](#oasVersion) field which designates the version of the OAS that it uses.
+An OpenAPI description document compatible with OAS 3.\*.\* contains a required [`openapi`](#oasVersion) field which designates the version of the OAS that it uses.
```
mikekistler, line 129: Do we really need to add the "description" qualifier here? Is there some other kind of "OpenAPI document" that we are trying to distinguish from?

mikekistler:
There are still a couple instances of "OpenAPI definition" in the file:
- [Line 13](https://github.com/OAI/OpenAPI-Specification/blob/58052838d769277aa10af66d6d79b805f73d484d/versions/3.0.4.md?plain=1#L13)
- [Line 2278](https://github.com/OAI/OpenAPI-Specification/blob/58052838d769277aa10af66d6d79b805f73d484d/versions/3.0.4.md?plain=1#L2278)

Should those be changed to "OpenAPI description"?

handrews:
@mikekistler 

> Should those be changed to "OpenAPI description"?

Probably yes!  Good catch- I'll take a look and update.

ralfhandl (approved):

versions/3.0.4.md
lornajane: I think it does help in this context, so I think we should leave this as-is

lornajane (approved):
Thanks @handrews . Words matter!

PR #3909, opened 2024-06-13
More guidance on schema dialects.

handrews:
Fixes:
* #3877 
* #3901 (vocabulary / dialect aspects)

This makes it more clear that jsonSchemaDialect, like $schema, is per document, and provides guidance for incomplete OAS documents and using $schema in standalone schema documents.

It also clarifies the nature of "requiring" the OAS extension vocabulary.

### DIFF
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -2583,8 +2583,8 @@ The following properties are taken from the JSON Schema specification but their 
 
 In addition to the JSON Schema properties comprising the OAS dialect, the Schema Object supports keywords from any other vocabularies, or entirely arbitrary properties.
 
-Validators MAY choose to ignore keywords defined by the OpenAPI Specification's base vocabulary, due to its inclusion in the OAS dialect with a [`$vocabulary`](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#section-8.1.2) value of `false`.
-It is comprised of the following keywords:
+JSON Schema implementations MAY choose to treat keywords defined by the OpenAPI Specification's base vocabulary as [unknown keywords](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#section-4.3.1), due to its inclusion in the OAS dialect with a [`$vocabulary`](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#section-8.1.2) value of `false`.
+The OAS base vocabulary is comprised of the following keywords:
 
 ##### <a name="baseVocabulary"></a>Fixed Fields
 
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -2583,7 +2583,8 @@ The following properties are taken from the JSON Schema specification but their 
 
 In addition to the JSON Schema properties comprising the OAS dialect, the Schema Object supports keywords from any other vocabularies, or entirely arbitrary properties.
 
-The OpenAPI Specification's base vocabulary is comprised of the following keywords:
+The OpenAPI Specification's base vocabulary is safe for standard validators to ignore, as indicated by its inclusion in the OAS dialect's [`$vocabulary`](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#section-8.1.2) keyword with a value of `false`.
+It is comprised of the following keywords:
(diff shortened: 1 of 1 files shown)


versions/3.1.1.md
```diff
@@ -2623,11 +2624,12 @@ The [XML Object](#xmlObject) contains additional information about the available
...
+To allow use of a different default `$schema` value for all Schema Objects contained within an OAS document, a `jsonSchemaDialect` value may be set within the <a href="#oasObject">OpenAPI Object</a>. If this default is not set, then the OAS dialect schema id MUST be used for these Schema Objects. The value of `$schema` within a resource root Schema Object always overrides any default.
 
-When a Schema Object is referenced from an external resource which is not an OAS document (e.g. a bare JSON Schema resource), then the value of the `$schema` keyword for schemas within that resource MUST follow [JSON Schema rules](https://tools.ietf.org/html/draft-bhutton-json-schema-00#section-8.1.1).
+For standalone JSON Schema documents that do not set `$schema`, or for Schema Objects in OpenAPI description documents that are _not_ [complete documents](#documentStructure), the dialect SHOULD be assumed to be the OAS dialect.
```
ralfhandl, line 2631: Just for my education: does this statement now define the desired implementation behavior in the OAS context?

And will this become a MUST in Moonwalk?

ralfhandl (approved):

versions/3.1.1.md
handrews: These two lines aren't directly related, because I can't figure out what the old line meant.  The way we use git is so messed up that much of the past change history is obscured in a giant squashed mega-commit at the point when we released 3.1, so I can't figure out who wrote that text or why.  It was not part of the changes that I made for JSON Schema support in 3.1.

I have no idea why it would matter that you're coming _from_ some other JSON Schema document.

So it's possible that I'm changing something by removing that.  I'm just not sure what.  Maybe that you can't expect JSON Schema to know to use the OpenAPI dialect if the implementation doesn't know about OpenAPI?  Which is true, but the "JSON Schema rules" are that the behavior is implementation-defined, which means it can be anything and be valid according to those rules.  I'd be OK with leaving it in since AFAICT it's a no-op.  If you don't use `$schema` with regular JSON Schema then you potentially get random behavior anyway, so don't do that.

The SHOULD is just an interpretation of the `jsonSchemaDialect` guidance for what to do when `jsonSchemaDialect` is not set, but taking into account the fact that people write incomplete documents.  In this view, I'm treating a standalone JSON Schema document used by an OpenAPI Description as an incomplete OpenAPI Description document.

The SHOULD here has nothing to do with references _from_ JSON Schemas.

versions/3.1.1.md
```diff
@@ -2583,7 +2583,8 @@ The following properties are taken from the JSON Schema specification but their
 
 In addition to the JSON Schema properties comprising the OAS dialect, the Schema Object supports keywords from any other vocabularies, or entirely arbitrary properties.
 
-The OpenAPI Specification's base vocabulary is comprised of the following keywords:
+The OpenAPI Specification's base vocabulary is safe for standard validators to ignore, as indicated by its inclusion in the OAS dialect's [`$vocabulary`](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#section-8.1.2) keyword with a value of `false`.
```
karenetheridge, line 2586:
```suggestion
Validators MAY choose to ignore keywords defined by the OpenAPI Specification's base vocabulary, due to its inclusion in the OAS dialect with a [`$vocabulary`](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#section-8.1.2) value of `false`.
```

handrews:
@karenetheridge I did some further wordsmithing of that section to align more with how the JSON Schema spec talks about false vocabularies and unknown keywords - please take a look and let me know if that works!

ralfhandl (approved):

mikekistler (approved):
Looks good. 👍

ralfhandl:
@karenetheridge Please confirm

karenetheridge:
Looks good, thanks!

karenetheridge (approved):
Looks good, thanks!

PR #3911, opened 2024-06-14
Fix stray mentions of parameters in Encoding / Header Objects (3.0.4)

handrews:
Encoding Object and Header Object descriptions had a few incorrect usages of "parameter" or related text left in.


mikekistler (approved):
Looks good. 👍

PR #3919, opened 2024-06-17
Consisent x- Param/Encoding/Header wording (3.0.4)

handrews:
It was a bit challenging to figure out where to put the wording about allowing extension parameters when there are multiple fixed fields tables, each in their own subsection.

For the Parameter Object (the only one with multiple tables in past releases), it had been after the last table, but that got further and further away from what felt like the main part of the Object description.

I thought I had put it consistently after the initial "Common Fixed Fields" table, but I put it even before that in one place (which we don't do anywhere), and I forgot to include it in the Header Object at all.

This change puts it after the Common Fixed Fields table for all three, which means that for all Objects it is immediately after the first Fixed Fields table.  Note that the Encoding Object is not in this diff as the line was in the right place already for that Object.


mikekistler (approved):
Looks good! 👍

PR #3921, opened 2024-06-19
Port remaining 3.0.4 fixes with adjustments for 3.1.1

handrews:
This PR really has two parts, and you can click each "series of commits" to see the diff of just that part:
1.  A [series of commits](https://github.com/OAI/OpenAPI-Specification/pull/3921/files/9739dfe5afa8cb6fce5f64bbde04eb5f300658bc) that ports, with as little changes as possible, all of the outstanding 3.0.4 changes not yet in 3.1.1
2. A [series of commits](https://github.com/OAI/OpenAPI-Specification/pull/3921/files/9739dfe5afa8cb6fce5f64bbde04eb5f300658bc..31d23da8dc55d13c4cd16e15b07194df376dbc7f) that adjusts those ports for 3.1.1 differences, primarily the change from `format` to `contentEncoding`

***IMPORTANT NOTE:*** The 3.1.0 spec is inconsistent about whether `type: string` _without_ `contentEncoding` and any other primitive type (`number`, `integer`, `boolean`) regardless of `contentEncoding` defaults to a Content Type of `text/plain` (per [§4.8.1.4](https://spec.openapis.org/oas/v3.1.0#special-considerations-for-multipart-content)) or `application/octet-stream` (per the [`contentType` fixed fields entry](https://spec.openapis.org/oas/v3.1.0#fixed-fields-12)).  This PR settles on `text/plain` as that is what the [3.0.3 `contentType` fixed fields entry said](https://spec.openapis.org/oas/v3.0.3#fixed-fields-12), and the change from `format` to `contentEncoding` should not have impacted that.  See also #3922 for a similar fix to 3.0.4.

This is a good opportunity to look for consistency although I will ask that you ***DO NOT*** hold up this PR with consistency/editorial changes that might need to also go into 3.0.4.  Please either open a PR on 3.0.4 with those, or start an issue where we can track them.

I squashed the ported PR commits somewhat, and just retained separate commits when review feedback was substantial enough that I thought it warranted a separate commit for crediting or clarity purposes, or (in the case of 3861) the original PR was a grab-bag of different tiny fixes in the first place.

Ports:
* #3867 (2 consecutive commits)
    * 1st commit includes #3904
* #3864 
* #3818
* #3911 (2 commits, one combined with another PR noted further down this list)
* #3840
* #3841 
* #3894 
* #3861 (11 consecutive commits)
* #3857 (4 consecutive commits)
* #3859 (3 consecutive commits)
    * 3rd commit includes half of #3911) 
* #3917
[...]

PR #3922, opened 2024-06-19
format: byte also defaults to octet-stream (3.0.4)

handrews:
The description of defaults in the `contentType` fixed fields table in 3.0.3 did not match the description in the Media Type Object's "Considerations for multipart" section.

The "considerations for multipart" section appears to be more complete and correct in both 3.0.3 and 3.1.1, so this adjusts that table that replaced the fixed field defaults description to match the "considerations for multipart" section.

***NOTE:** The 3.1.1 version of this is incorporated into PR #3921, which fixes a slightly different discrepancy in the same place._

### DIFF
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -1717,8 +1717,8 @@ The default values for `contentType` are as follows, where an _n/a_ in the `form
 
 Property `type` | Property `format` | Default `contentType`
 ------------- | --------------- | ---------------------
-`string` | `binary` | `application/octet-stream`
-`string` | _none, or any except `binary`_ | `text/plain`
+`string` | `binary` _or_ `byte` | `application/octet-stream`
+`string` | _none, or any except `binary` or `byte`_ | `text/plain`
 `number`, `integer`, or `boolean` | _n/a_ | `text/plain`
 `object` | _n/a_ | `application/json`
 `array` | _n/a_ | according to the `type` and `format` of the `items` schema


versions/3.0.4.md
```diff
@@ -1717,8 +1717,8 @@ The default values for `contentType` are as follows, where an _n/a_ in the `form
 
 Property `type` | Property `format` | Default `contentType`
 ------------- | --------------- | ---------------------
-`string` | `binary` | `application/octet-stream`
-`string` | _none, or any except `binary`_ | `text/plain`
+`string` | `binary` _or_ `byte` | `application/octet-stream`
+`string` | _none, or any except `binary` or `byte`_ | `text/plain`
```
ralfhandl, line 1721: So `binary` and `byte` are synonyms for serializing parts of a multipart request body? Any guidance on which is the preferred synonym?

Also: why does `encoding` and the Encoding Object not apply to multipart _response_ bodies?

mikekistler: I am uncomfortable with this change. In general, I think `format: byte` means "base64 encoded" and the purpose of base64 encoding is to make the value safe for a string field.

Also, the "considerations for multipart" section in 3.0.3 does not actually mention `format: byte` -- it mentions `format: base64`, which not only is absent from the "formats defined by the OAS" table in the "Data Types" section but is not even listed in the Formats registry.

My conclusion here is that the "considerations for multipart" section is actually the place to be corrected by removing `format: base64`.

ralfhandl: Note: the `format: base64` typo in 3.0.3 has been corrected in 3.0.4 via #3182.

Not mentioning `format: byte` here (by closing this PR without merging) would make me feel more comfortable.

handrews: @ralfhandl
> So binary and byte are synonyms for serializing parts of a multipart request body? Any guidance on which is the preferred synonym?

 `binary` and `byte` are not synonyms, and are explained in he ["Data Types" section](https://handrews.github.io/renderings/oas/deploy/oas/v3.0.4.html#data-types), particularly in the ["Working with Binary Data" subsection](https://handrews.github.io/renderings/oas/deploy/oas/v3.0.4.html#working-with-binary-data).

> Also: why does encoding and the Encoding Object not apply to multipart response bodies?

It applies to `multipart` and `application/x-www-form-urlencoded`, but what does that have to do with this PR?

@mikekistler 

> I am uncomfortable with this change. In general, I think `format: byte` means "base64 encoded" and the purpose of base64 encoding is to make the value safe for a string field.

It's not a change.  I just made a mistake and forgot to include it and no one caught it because there had been a  mistake in another part of the text which made it look like only `binary` is correct  But that was wrong.  

The reason is that, in `multipart`, `Content-Type` refers to the actual content type of the data part.  The encoding that makes it safe for a string is placed in `Content-Transfer-Type`.  This should be a bit more clear after PR #3923.  So yes, it is made safe for a string, but no, it does not change the `Content-Type` header, and therefore does not change the default for `contentType` compared to `format: binary`.

The spec has always stated that `format: byte` defaults `application/octet-stream`, and in 3.1 it states even more clearly that _any_ `contentEncoding` (including but not limited to `base64`) defaults to `application/octet-stream`.  So this is not a change.

> Also, the "considerations for multipart" section in 3.0.3 does not actually mention `format: byte` -- it mentions `format: base64`, which not only is absent from the "formats defined by the OAS" table in the "Data Types" section but is not even listed in the Formats registry.

As @ralfhandl mentioned, `base64` was always an error and had been fixed - if you look in PR #3923 you'll see it was already `format: byte` in 3.0.4.  I know we discussed whether to add `base64` to the registry but I guess we decided not to, and to emphasize that `byte` was correct.  I can't find the discussion right now.  Either way, whether `base64` is in the registry is irrelevant to this PR.

> My conclusion here is that the "considerations for multipart" section is actually the place to be corrected by removing `format: base64`.

_goes with:_

@ralfhandl 

> Not mentioning `format: byte` here (by closing this PR without merging) would make me feel more comfortable.

That would make the spec incorrect.  It's inconsistent with the text from the Media Type object (which is more complete and correct in both 3.0 and 3.1) _and_ is inconsistent with how the analogous `contentEncoding` is used in 3.1.

This PR is just a bug fix.

handrews: @mikekistler in response to your question in the TDC call, the `Content-Transfer-Encoding` requirement is in the last paragraph of [this section](https://handrews.github.io/renderings/oas/deploy/oas/v3.0.4.html#encoding-multipart-media-types) (or, more accurately, it will be after PR #3923 is merged - that rendering includes both that PR and this one).

And yes, I want to acknowledge that 3.0.3 was not at all clear about `Content-Transfer-Encoding`, and the inconsistent use of `byte` and `base64` made things worse.  (`byte` was _always_ what was defined in the "Data Types" section, but 3.0.3 and earlier mistakenly used `base64` in some examples - we debated it a lot b/c obviously `base64` is what would be used anywhere else, but we ultimately agreed `byte` was what had been intended and is what should be used.)

ralfhandl (approved):

PR #3923, opened 2024-06-20
Reorganize and streamline form media type guidance (3.0.4)

handrews:
It's probably best to look at this as individual commits- `git` is pretty confused about the overall diff so I went back and divided it up into five separate steps:

* [moving blocks of text or example with ***no*** changes](https://github.com/OAI/OpenAPI-Specification/pull/3923/commits/72f0aa6ed513d62887d927a63daea0aa605fedf5)
* [add section titles and fix heading levels](https://github.com/OAI/OpenAPI-Specification/pull/3923/commits/7636fbf75c1d467aa7193c12e09d5cb9a83a3095) (this makes the remaining diffs much more clear)
* [changes to the Media Type Object](https://github.com/OAI/OpenAPI-Specification/pull/3923/commits/fd71328b0c0ac87568a4c9e271b53811b8a0eaf0)
* [changes for `form-urlencoded`](https://github.com/OAI/OpenAPI-Specification/pull/3923/commits/8cdbf839147df3d0a6ba693727b85fb50692ca2d)
* [changes for `form-data`](https://github.com/OAI/OpenAPI-Specification/pull/3923/commits/d8f014031b9c486dab96247b3536c186e3012a75)
* [fix for base64 example](https://github.com/OAI/OpenAPI-Specification/pull/3923/commits/e8cc358bfe8b02a47f56572ac65c012b3616f9c5)
* [restore note about `contentMediaType`](https://github.com/OAI/OpenAPI-Specification/pull/3923/commits/cf3e5f3d90d47c8425107a56f172643e03ed2540)

There has been a lot of duplication around `form-urlencoded` and `form-data` between the Media Type Object and the Encoding Object.

Figuring out form details has required reading through both sections, because some things only appear in one place, some only appear in the other (and not necessarily the way you'd guess), and some things are in both places but contradict each other (#3921 includes a fix for one in 3.1.1, #3922 fixes a similar problem in 3.0.4).

I've opted to handle everything under the Encoding Object, particularly as we have clarified that an absent `encoding` field in the Media Type Object is the same as having it with empty Encoding Objects for every field.  I've kept the old Media Type Object section headings and linked to the new sections for folks who might be used to looking there.

I've given each example a section heading because they do distinct things and it's helpful to see that in the TOC.

Porting this to 3.1.1 will have to wait for PR #3921 to be merged, and it will also be similar but not quite identical because of the differences in handling binary data.

### DIFF
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -1731,15 +1731,7 @@ name=example&icon=iVBORw0KGgoAAAANSUhEUgAAAAIAAAACCAYAAABytg0kAAAAEklEQVQIW2P8z8
 
 ##### Encoding `multipart` Media Types
 
-It is common to use `multipart/form-data` as a `Content-Type` when transferring request bodies to operations.  In contrast to 2.0, a `schema` is REQUIRED to define the input parameters to the operation when using `multipart` content.  This supports complex structures as well as supporting mechanisms for multiple file uploads.
-
-When passing in `multipart` types, boundaries MAY be used to separate sections of the content being transferred — thus, the following default `Content-Type`s are defined for `multipart`:
-
-* If the property is a primitive, or an array of primitive values, the default Content-Type is `text/plain`
-* If the property is complex, or an array of complex values, the default Content-Type is `application/json`
-* If the property is a `type: string` with `format: binary` or `format: byte` (aka a file object), the default Content-Type is `application/octet-stream`
-
-Note that only `multipart/*` media types with named parts can be described as shown here.  Note also that while `multipart/form-data` originally defined a per-part `Content-Transfer-Encoding` header that could indicate base64 encoding (`format: byte`), it has been deprecated for use with HTTP as of [RFC7578](https://www.rfc-editor.org/rfc/rfc7578#section-4.7).
+It is common to use `multipart/form-data` as a `Content-Type` when transferring forms as request bodies.  In contrast to 2.0, a `schema` is REQUIRED to define the input parameters to the operation when using `multipart` content.  This supports complex structures as well as supporting mechanisms for multiple file uploads.
 
 The `form-data` disposition and its `name` parameter are mandatory for `multipart/form-data` ([RFC7578 §4.2](https://www.rfc-editor.org/rfc/rfc7578.html#section-4.2)).
 Array properties are handled by applying the same `name` to multiple parts, as is recommended by [RFC7578 §4.3](https://www.rfc-editor.org/rfc/rfc7578.html#section-4.3) for supplying multiple values per form field.
@@ -1758,7 +1750,7 @@ See [Appendix E](#percentEncodingAndFormMediaTypes) for a detailed examination o
 
 ###### Example: Basic Multipart Form
 
-Examples:
+When the `encoding` attribute is _not_ used, the encoding is determined by the Encoding Object's defaults:
 
 ```yaml
 requestBody:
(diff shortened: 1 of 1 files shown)


ralfhandl (dismissed):
No relevant content seems to have been removed 😄

versions/3.0.4.md
```diff
@@ -1765,9 +1746,11 @@ Note also that `Content-Transfer-Encoding` is deprecated for `multipart/form-dat
 Using `format: byte` for a multipart field is equivalent to setting `Content-Transfer-Encoding: base64`.
```
mikekistler, line 1746: I'm confused by this. If `Content-Transfer-Encoding` is deprecated and a multipart field with `format: byte` is equivalent to setting `Content-Transfer-Encoding: base64`, doesn't this imply that a multipart field with `format: byte` is essentially deprecated?

I guess what I'm missing here is why `Content-Transfer-Encoding` was deprecated. It seems like this effectively disallows any encoding of individual fields in a multipart body, which seems undesirable. I'm sure this is just confusion on my part but would appreciate if someone could clear this up for me.

handrews:
I've added another commit (and edited the commit list in the PR text above) that fixes an error in the base64 example.  (@mikekistler I doubt it makes things more clear, but it's more correct).

versions/3.0.4.md
```diff
@@ -1726,9 +1726,13 @@ Given a name of `example` and a solid red 2x2-pixel PNG for `icon`, this
 would produce a request body of:
 
 ```urlencoded
-name=example&icon=iVBORw0KGgoAAAANSUhEUgAAAAIAAAACCAYAAABytg0kAAAAEklEQVQIW2P8z8AARAwMjDAGACwBA/+8RVWvAAAAAElFTk
+name=example&icon=iVBORw0KGgoAAAANSUhEUgAAAAIAAAACCAIAAAD91JpzAAAABGdBTUEAALGPC%2FxhBQAAADhlWElmTU0AKgAAAAgAAYdpAAQAAAABAAAAGgAAAAAAAqACAAQAAAABAAAAAqADAAQAAAABAAAAAgAAAADO0J6QAAAAEElEQVQIHWP8zwACTGCSAQANHQEDqtPptQAAAABJRU5ErkJggg%3D%3D
 ```
 
+Note that this base64-encoded value had to be futher percent-encoded, replacing `/` with `%2F` and each of two final `=` padding characters with `%3D`.
```
ralfhandl, line 1732: > replacing `/` with `%2F`

That's why [base64url](https://datatracker.ietf.org/doc/html/rfc4648#section-5) was invented - I wonder why isn't used more often in places where it makes sense.

ralfhandl (approved):

versions/3.0.4.md
handrews: Yes, it's allowed and used in an example in 3.1.  Oddly, the use of `=` for padding is not changed, so you still end up needing to do percent-encoding :-/

versions/3.0.4.md
mikekistler: I agree that we are obligated to follow the RFCs, which is exactly why I think we need to clearly state that `format: byte` is deprecated for a property in a multipart/form-data body. 

@darrelmiller Would appreciate your thoughts on this.

versions/3.0.4.md
```diff
@@ -1765,9 +1750,13 @@ Note also that `Content-Transfer-Encoding` is deprecated for `multipart/form-dat
 Using `format: byte` for a multipart field is equivalent to setting `Content-Transfer-Encoding: base64`.
 If `format: byte` is used along with setting a different `Content-Transfer-Encoding` value with the `headers` field, the result is undefined.
```
mikekistler, line 1751: Can we say this instead?
```suggestion
If `format: byte` is specified for a `multipart/form-data` field, the result is undefined.
```

handrews: No, because it is completely valid to do this.  "Deprecated" does not mean "forbidden", and the behavior is completely well-defined.  In RFC7578 it is a SHOULD NOT not a MUST NOT, and that RFC was only published in July 2017 – the same month as OAS 3.0.  We support all HTTP APIs, including ones that use behavior that is now deprecated.

The text I wrote in this paragraph provides all of the necessary context for authors to decide whether it is suitable for their use or not.  We wouldn't want to _replace_ this sentence with a `form-data`-specific sentence anyway as it's at least theoretically possible to describe other `multipart` types.

versions/3.0.4.md
handrews: Deprecated in OAS is not the same as deprecated in RFC7578, and the text I wrote _already_ makes it clear that the _usage_ is deprecated by RFC7578 in this context.

miqui (approved):

handrews:
@miqui @mikekistler had asked for changes here, and the discussion was not resolved.  This should not have been merged and I consider it to still be open for changes.

@mikekistler perhaps to avoid future confusion it would be best to use the "request changes" button in the review interface even if it's not yet clear what (if any) changes would be requested.

PR #3924, opened 2024-06-20
3.0.4 dev: typo "paramter"

ralfhandl:
Typo: "paramter" should be "parameter" (twice)


handrews (dismissed):

handrews (approved):
Thanks again - looks like we have everything that needs including from PR #3921 :-)

PR #3929, opened 2024-06-24
Clarify interaction of format: byte and Content-Transfer-Encoding header definition

mikekistler:
This PR clarifies (IMO) the interaction of `format: byte` in the schema of a multipart/form-data body with the `Content-Transfer-Encoding` header definition in the corresponding field of an Encoding Object.

I found the original text confusing because it talked about "setting the `Content-Transfer-Encoding` header", which I could not tell if this meant in the OpenAPI description or in the HTTP flow.  I think the new language makes clear that this is in the Encoding Object in the OpenAPI description.

I also made this its own paragraph since I think this is independent of the note about `Content-Transfer-Encoding` now being deprecated.

### DIFF
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -1754,8 +1754,8 @@ Note that there are significant restrictions on what headers can be used with `m
 
 Note also that `Content-Transfer-Encoding` is deprecated for `multipart/form-data` ([RFC7578 §4.7](https://www.rfc-editor.org/rfc/rfc7578.html#section-4.7)) where binary data is supported, as it is in HTTP.
 
-Using `format: byte` for a multipart field is equivalent to specifying an Encoding Object with a `headers` field containing `Content-Transfer-Encoding` with a schema that requires the value `base64`.
-If `format: byte` is used for a multipart field that has an encoding object with a `headers` field containing `Content-Transfer-Encoding` with a schema that disallows "base64", the result is undefined for serialization and parsing.
+Using `format: byte` for a multipart field is equivalent to specifying an [Encoding Object](#encodingObject) with a `headers` field containing `Content-Transfer-Encoding` with a schema that requires the value `base64`.
+If `format: byte` is used for a multipart field that has an Encoding Object with a `headers` field containing `Content-Transfer-Encoding` with a schema that disallows `base64`, the result is undefined for serialization and parsing.
 
 Per the JSON Schema specification, `contentMediaType` without `contentEncoding` present is treated as if `contentEncoding: identity` were present.  While useful for embedding text documents such as `text/html` into JSON strings, it is not useful for a `multipart/form-data` part, as it just causes the document to be treated as `text/plain` instead of its actual media type.  Use the Encoding Object without `contentMediaType` if no `contentEncoding` is required.
 
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -1755,7 +1755,7 @@ Note that there are significant restrictions on what headers can be used with `m
 Note also that `Content-Transfer-Encoding` is deprecated for `multipart/form-data` ([RFC7578 §4.7](https://www.rfc-editor.org/rfc/rfc7578.html#section-4.7)) where binary data is supported, as it is in HTTP.
 
 Using `format: byte` for a multipart field is equivalent to specifying an `encoding` object with a `headers` field containing `Content-Transfer-Encoding: { schema: { enum: [base64] } }`.
-If `format: byte` is used for a multipart field that has an encoding object with a `headers` field containing `Content-Transfer-Encoding` with a schema that permits values other than "base64", the result is undefined.
+If `format: byte` is used for a multipart field that has an encoding object with a `headers` field containing `Content-Transfer-Encoding` with a schema that disallows "base64", the result is undefined for serialization and parsing.
 
 Per the JSON Schema specification, `contentMediaType` without `contentEncoding` present is treated as if `contentEncoding: identity` were present.  While useful for embedding text documents such as `text/html` into JSON strings, it is not useful for a `multipart/form-data` part, as it just causes the document to be treated as `text/plain` instead of its actual media type.  Use the Encoding Object without `contentMediaType` if no `contentEncoding` is required.
 
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -1753,8 +1753,9 @@ It is not currently possible to correlate schema properties with unnamed, ordere
 Note that there are significant restrictions on what headers can be used with `multipart` media types in general ([RFC2046 §5.1](https://www.rfc-editor.org/rfc/rfc2046.html#section-5.1)) and `multi-part/form-data` in particular ([RFC7578 §4.8](https://www.rfc-editor.org/rfc/rfc7578.html#section-4.8)).
 
 Note also that `Content-Transfer-Encoding` is deprecated for `multipart/form-data` ([RFC7578 §4.7](https://www.rfc-editor.org/rfc/rfc7578.html#section-4.7)) where binary data is supported, as it is in HTTP.
-Using `format: byte` for a multipart field is equivalent to setting `Content-Transfer-Encoding: base64`.
-If `format: byte` is used along with setting a different `Content-Transfer-Encoding` value with the `headers` field, the result is undefined.
+
+Using `format: byte` for a multipart field is equivalent to specifying an `encoding` object with a `headers` field containing `Content-Transfer-Encoding: enum: ["base64"]`.
+If `format: byte` is used for a multipart field that has an encoding object with a `headers` field containing `Content-Transfer-Encoding` with a schema that permits values other than "base64", the result is undefined.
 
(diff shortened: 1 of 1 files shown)


mikekistler:
It appears that my editor did an auto-format. I will fix.

mikekistler:
Okay ... fixed silly whitespace diffs. I had to use VIM to make the change as it seemed nothing could convince VSCode to leave the formatting alone.

handrews:
> I had to use VIM to make the change as it seemed nothing could convince VSCode to leave the formatting alone.

weird- I wouldn't think VIM would do anything that strange (I do use VIM on Mac OS for all of my editing).

versions/3.0.4.md
```diff
@@ -1753,8 +1753,9 @@ It is not currently possible to correlate schema properties with unnamed, ordere
 Note that there are significant restrictions on what headers can be used with `multipart` media types in general ([RFC2046 §5.1](https://www.rfc-editor.org/rfc/rfc2046.html#section-5.1)) and `multi-part/form-data` in particular ([RFC7578 §4.8](https://www.rfc-editor.org/rfc/rfc7578.html#section-4.8)).
 
 Note also that `Content-Transfer-Encoding` is deprecated for `multipart/form-data` ([RFC7578 §4.7](https://www.rfc-editor.org/rfc/rfc7578.html#section-4.7)) where binary data is supported, as it is in HTTP.
-Using `format: byte` for a multipart field is equivalent to setting `Content-Transfer-Encoding: base64`.
-If `format: byte` is used along with setting a different `Content-Transfer-Encoding` value with the `headers` field, the result is undefined.
+
+Using `format: byte` for a multipart field is equivalent to specifying an `encoding` object with a `headers` field containing `Content-Transfer-Encoding: enum: ["base64"]`.
```
handrews, line 1757: You need `schema` and some `{}` around the `Content-Transfer-Encoding: {schema: {enum: ["base64"]}}` part (my apologies for leaving them out when we were chatting on IM).  And I'd probably either quote all the strings (like JSON) or none of them (like YAML).

versions/3.0.4.md
```diff
@@ -1753,8 +1753,9 @@ It is not currently possible to correlate schema properties with unnamed, ordere
...
+If `format: byte` is used for a multipart field that has an encoding object with a `headers` field containing `Content-Transfer-Encoding` with a schema that permits values other than "base64", the result is undefined.
```
handrews, line 1758: I'm not sure this is quite right, as if you use `format: byte` with, say, `headers: {Content-Transfer-Encoding: {schema: {enum: [base64, quoted-printable]}}}` then it would be kind-of pointless because the only valid option is `base64`, but it technically wouldn't be a problem.  Just confusing.  But you can do _a lot_ of things like this in JSON Schema and elsewhere in OpenAPI so we don't want to get into calling out every possible nonsensical combination that the specification syntactically allows.

Here's another attempt, feel free to try something else:

> The result of using `format: byte` and using the Encoding Object's `headers` field to require a `Content-Transfer-Encoding` other than `base64` is undefined, as this specification does not define a precedence for resolving the conflict.

Another reason this is undefined is that you can't rely on `format: byte` to actually be checked during JSON Schema validation, so unlike a lot of JSON Schema combinations that don't make sense, you can't rely on an error happening anywhere.

handrews (changes requested):
Minor quibbles trying to find the right wording for this confusing area.

> I also made this its own paragraph since I think this is independent of the note about Content-Transfer-Encoding now being deprecated.

Honestly I think this is the most important part of the change.  Now that I see it separated, your earlier confusion makes a lot more sense.  The way I had it made it look like the deprecation and undefined behavior were related, and they're not related at all.

versions/3.0.4.md
mikekistler: Okay ... I think I fixed that.

versions/3.0.4.md
mikekistler: Shouldn't "kind-of pointless" be "undefined"? If it is not undefined, then what exactly does this mean? Does it mean that `Content-Transfer-Encoding: quoted-printable` is allowed on this field?  You seem to say "No", because you say the only valid option is "base64", but that seems to mean that `format: byte` _overrides_ whatever is specified in the `headers` of the encoding object. But if that's the case, then why wouldn't that be the case for any schema of a `Content-Transfer-Encoding` header?

handrews: @mikekistler if you set it through `headers`, you can set `Content-Transfer-Encoding` to any valid encoding allowed by that header.  `quoted-printable` was the first one that came to mind other than `base64`.

Since the Header Object describes the _valid_ values for the specific usage with a schema, you can set it to allow multiple values.  So a schema of `{enum: [base64, quoted-printable]}` means that you can do either of those.

If you _also_ set `format: byte` in the main representation schema, then that is the thing that is equivalent to `base64`.  So now you're putting constraints on the header value in two places.  One of those places only allows one value, so the second value in the other place is pointless _in this specific usage_. However, you might do that if you `$ref` the Header Object in `headers` because maybe you use `headers` a lot and typically allow either encoding, but one particular usage uses `format` to restrict it to just one.

So "kinda pointless" doesn't mean there's never a valid reason it might happen.  And the behavior is completely well-defined.  If I tell you "you can use odd numbers" in one place and "you can use even numbers between 0 and 9 inclusive" in another, you know you can use 0, 2, 4, 6, and 8.  The interaction made one of these constraints superfluous, but not meaningless.

The problem here is that you're expecting this to make sense.  I don't think anyone ever thought through the implications of allowing all of these different ways to do things.  The `format` keyword was not meant for content encoding (which is why it doesn't do that in 3.1).  Whether it's `format` or `contentEncoding`, JSON Schema was not designed with awareness of any sort of headings, MIME or HTTP.  I assume the `headers` field was added to the Encoding Object for some other reason, and no one thought to consider the overlap with `format: base64`.  Or maybe they did, and they wanted to allow other `Content-Transfer-Encoding` values and it was just too much troubel to forbid `fomat: base64` in this specific case.  I have no idea.

But this is not going to make sense.  It's two completely independent mechanisms that were developed separately that collide, which happens a lot in Parameter/Header/Encoding/Schema Objects.  We're just stuck with it in 3.x and this is the best we can do for now.

handrews: @mikekistler there are tons of things in JSON Schema and OpenAPI that can be in conflict.  None of them are undefined.  To call this undefined would introduce a major inconsistency in how we treat descriptions that have conflicting or overlapping elements.

versions/3.0.4.md
```diff
@@ -1753,8 +1753,9 @@ It is not currently possible to correlate schema properties with unnamed, ordere
 Note that there are significant restrictions on what headers can be used with `multipart` media types in general ([RFC2046 §5.1](https://www.rfc-editor.org/rfc/rfc2046.html#section-5.1)) and `multi-part/form-data` in particular ([RFC7578 §4.8](https://www.rfc-editor.org/rfc/rfc7578.html#section-4.8)).
 
 Note also that `Content-Transfer-Encoding` is deprecated for `multipart/form-data` ([RFC7578 §4.7](https://www.rfc-editor.org/rfc/rfc7578.html#section-4.7)) where binary data is supported, as it is in HTTP.
-Using `format: byte` for a multipart field is equivalent to setting `Content-Transfer-Encoding: base64`.
-If `format: byte` is used along with setting a different `Content-Transfer-Encoding` value with the `headers` field, the result is undefined.
+
+Using `format: byte` for a multipart field is equivalent to specifying an `encoding` object with a `headers` field containing `Content-Transfer-Encoding: { schema: { enum: [base64] } }`.
```
ralfhandl, line 1757:
```suggestion
Using `format: byte` for a multipart field is equivalent to specifying an Encoding Object with a `headers` field containing `Content-Transfer-Encoding` with a schema that requires the value `base64`.
```

Avoid "partially condensed YAML"

versions/3.0.4.md
```diff
@@ -1753,8 +1753,9 @@ It is not currently possible to correlate schema properties with unnamed, ordere
...
+If `format: byte` is used for a multipart field that has an encoding object with a `headers` field containing `Content-Transfer-Encoding` with a schema that permits values other than "base64", the result is undefined.
```
ralfhandl, line 1758:
```suggestion
If `format: byte` is used for a multipart field that has an encoding object with a `headers` field containing `Content-Transfer-Encoding` with a schema that permits values other than `base64`, the result is undefined.
```

handrews (dismissed):
Thanks!

miqui (dismissed):

mikekistler:
Damn VSCode reformatted again! I will fix.

mikekistler:
Sorry to trouble you to re-review but GitHub dismissed the approvals when I moved the PR to draft temporarily while I fixed the reformatting issue.

handrews (approved):
Looks even better now with the latest tweaks!

ralfhandl (approved):

ralfhandl:
@OAI/tsc One more review please

miqui (approved):

PR #3937, opened 2024-07-01
Reorganize and streamline form media type guidance (3.1.1 modified port of 3923, 3929)

handrews:
Ports (with substantial changes due to `format` vs `contentEncoding` and `contentMediaType`):

* #3923 
* #3929

### DIFF
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -1550,6 +1550,8 @@ Field Name | Type | Description
 
 This object MAY be extended with [Specification Extensions](#specificationExtensions).
 
+An `encoding` attribute is introduced to give you control over the serialization of parts of `multipart` request bodies.  This attribute is _only_ applicable to request bodies, and _only_ for `multipart` and `application/x-www-form-urlencoded` media types.
+
 ##### Media Type Examples
 
 ```json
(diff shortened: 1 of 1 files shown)


versions/3.1.1.md
```diff
@@ -1655,26 +1655,77 @@ requestBody:
...
+<a name="encodingContentType"></a>contentType | `string` | The `Content-Type` for encoding a specific property. The value is a comma-separated list, each element of which is either a specific media type (e.g. `image/png`) or a wildcard media type (e.g. `image/*`). Default value depends on the property type as shown in the table below.
+<a name="encodingHeaders"></a>headers | Map[`string`, [Header Object](#headerObject) \| [Reference Object](#referenceObject)] | A map allowing additional information to be provided as headers. `Content-Type` is described separately and SHALL be ignored in this section. This property SHALL be ignored if the request body media type is not a `multipart`.
+
+This object MAY be extended with [Specification Extensions](#specificationExtensions).
+
+The default values for `contentType` are as follows, where an _n/a_ in the `contentEncoding` column means that the presence or value of `contentEncoding` is irrelevant:
+
+Property `type` | Property `contentEncoding` | Default `contentType`
+--------------- | -------------------------- | ---------------------
+_absent_ | _n/a_ | `application/octet-stream`
```
ralfhandl, line 1695: This line is not present in 3.0.4: why don't we need it there?

versions/3.1.1.md
```diff
@@ -1882,27 +1899,24 @@ requestBody:
...
-  encoding:
-    icon:
-      contentType: image/png, image/jpeg
+          # The property name 'file' will be used for all files.
+          file:
+            type: array
+            items: {}
 ```
 
+As seen in the section on `multipart/form-data` below, the empty schema for `items` indicates a media type of `application/octet-stream`.
```
ralfhandl, line 1918: Which section below? Appendix C or E?

Better add a link to the relevant section, and adjust "below" to "above" if needed.

ralfhandl (commented):
Minor nits

versions/3.1.1.md
handrews: Because `type` being absent has a specific meanint (see "Working With Binary Data") in draft 2020-12 / 3.1, which it does not in draft-04 / 3.0.  It is used for instances (such as binary data) that do not fit the JSON Schema data model.  That guidance was added to 2020-12 specifically to enable using the `content*` keywords with data that does not conform to the JSON data model at all.

versions/3.1.1.md
handrews: That text was left over from when things were in a different order - it's not about an appendix at all.  I'll sort it out and update - good catch!

versions/3.1.1.md
handrews: @ralfhandl honestly, I wanted to completely rearrange all of the data model things to put all of the rules in one place, but it would be a huge amount of work and would require a specification with a very different organization, so I decided against it.

handrews:
@ralfhandl I've updated both places with links to clarification, and (in the second place) made sure it mentions the correct section.

ralfhandl (approved):

ralfhandl:
@OAI/tsc Second review, please!

whitlockjc (approved):

PR #3940, opened 2024-07-04
Remove need for separate ReSpec clone for Arazzo

ralfhandl:
Currently OpenAPI and Arazzo specs use differently modified forks of ReSpec, just to set the headline of the "abstract" section.

This can be done within the HTML document, no need to tweak ReSpec, and no need for two different clones.


lornajane:
Abstract headline should read "What is the OpenAPI Specification?" still.

lornajane (approved):

PR #3944, opened 2024-07-04
Use latest respec

### LINKED ISSUE #3845: We should not keep a minified version of respec.js
## Problem
There are 2 minified versions of `respec.js` for 3.x and for 4.x versions. These files exists in the `gh-pages`:
1. `/js/respec-oai.js`
2. `/js/respec-arrazzo.js`

These files are basically unmaintainable and should instead try to store a human-readable version of it.

## Proposed solution
There is a `respec-oai.build.js.map` which might help us decouple the minified version `respec-oai.js`. 
After we succeed in decoupling the file and keep only the source code in our repo, we need to add a build step in our deploy process so we can push the minified version again.

## Alternative Proposal

Do not use a modified fork of ReSpec, instead use the unmodified original, see #3944.

### LINKED ISSUE #3852: Simplify HTML rendering build process
Currently, to build the HTML versions of the spec, you need to simulate `.github/workflows/respec.yam` by doing (in the root of a checked-out `main` branch of the repository):

1. make sure you have node version 20 installed
1. `npm install`
1. `./scripts/md2html/build.sh`
1. `git checkout gh-pages`
1. `cp -R js deploy/js`

Then you can look at the rendered HTML.  After the first time you only have to do the `./scripts/md2html/build.sh` (from `main`).

If you need to render a new version of the spec, you also need to add a new `MAINTAINERS_vX.Y.Z.md` file in the `scripts/md2html/history` directory (which the error message you get will show as `./history`).

We should fix this as we fix #3717, #3845, and #3715.

ralfhandl:
Fixes #3845
Fixes #3852

As discussed in TDC call on 2024-07-04
- pinned version of respec as dependency
- adjust frontmatter

As discussed in TDC call on 2024-07-11
- generate static HTML that renders faster and does not need `respec-w3c.js`

Differences due to new ReSpec defaults
- spacing around `<h1>` and `<h2>` - keep
- collapsible "More details about this document" was added - keep
- "Latest published version:" was added - keep
- section marker symbols (§) are now left of heading instead of right - keep

Reverted differences 
- code in text had text color instead of red color
- code in code blocks had different syntax coloring

Please compare
* with new respec: https://ralfhandl.github.io/OpenAPI-Specification/oas/latest.html
* current state: https://spec.openapis.org/oas/latest.html

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "editor.formatOnSave": false
-}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "editor.formatOnSave": false
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,6 +12,7 @@
         "cheerio": "^1.0.0-rc.5",
         "highlight.js": "^11.10.0",
         "markdown-it": "^14.1.0",
+        "respec": "35.1.1",
         "yargs": "^17.7.2"
       },
       "devDependencies": {
@@ -32,6 +33,43 @@
       },
       "engines": {
         "node": ">=6.0.0"
+      }
+    },
+    "node_modules/@babel/code-frame": {
+      "version": "7.24.7",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.24.7.tgz",
+      "integrity": "sha512-BcYH1CVJBO9tvyIZ2jVeXgSIMvGZ2FDRvDdOIVQyuklNKSsx+eppDEBq/g47Ayw+RqNFE+URvOShmf+f/qwAlA==",
+      "license": "MIT",
+      "dependencies": {
+        "@babel/highlight": "^7.24.7",
+        "picocolors": "^1.0.0"
+      },
+      "engines": {
+        "node": ">=6.9.0"
+      }
+    },
+    "node_modules/@babel/helper-validator-identifier": {
+      "version": "7.24.7",
+      "resolved": "https://registry.npmjs.org/@babel/helper-validator-identifier/-/helper-validator-identifier-7.24.7.tgz",
+      "integrity": "sha512-rR+PBcQ1SMQDDyF6X0wxtG8QyLCgUB0eRAGguqRLfkCA87l7yAP7ehq8SNj96OOGTO8OBV70KhuFYcIkHXOg0w==",
+      "license": "MIT",
+      "engines": {
+        "node": ">=6.9.0"
+      }
+    },
+    "node_modules/@babel/highlight": {
+      "version": "7.24.7",
+      "resolved": "https://registry.npmjs.org/@babel/highlight/-/highlight-7.24.7.tgz",
+      "integrity": "sha512-EStJpq4OuY8xYfhGVXngigBJRWxftKX9ksiGDnmlY3o7B/V7KIAc9X4oiK87uPJSc/vs5L869bem5fhZa8caZw==",
+      "license": "MIT",
+      "dependencies": {
+        "@babel/helper-validator-identifier": "^7.24.7",
+        "chalk": "^2.4.2",
+        "js-tokens": "^4.0.0",
+        "picocolors": "^1.0.0"
+      },
+      "engines": {
+        "node": ">=6.9.0"
       }
     },
     "node_modules/@esbuild/aix-ppc64": {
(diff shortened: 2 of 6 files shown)


handrews:
> code in text now has text color instead of red color

I'm not sure I like this change.  I'm trying to give it a chance, but the red text made it easier to scan for certain things.

> code in code blocks has different syntax coloring

What are the differences?  Is it a different package, a different configuration...?

handrews:
> > code in text now has text color instead of red color
> 
> I'm not sure I like this change. I'm trying to give it a chance, but the red text made it easier to scan for certain things.

If the red is too dramatic for folks, we could choose another color?  Even a gray that is distinctive enough to clearly stand out?  We should run A11Y (accessabilty) metrics on our rendering (no matter what we do with this color).

ralfhandl:
> We should run A11Y (accessabilty) metrics on our rendering (no matter what we do with this color).

Good idea, which tool(s) would you recommend?

ralfhandl:
> What are the differences?

- Delimiters are now black instead of red - seems the new code color shines through
- Numbers are now bronze instead of turquoise

New: https://ralfhandl.github.io/OpenAPI-Specification/oas/latest.html#format
Old: https://spec.openapis.org/oas/latest.html#format

stringtheory:
> > We should run A11Y (accessabilty) metrics on our rendering (no matter what we do with this color).
> 
> Good idea, which tool(s) would you recommend?

Accessibility Human here :) - here is the tldr; on a11y // and an overly informative answer with more tools and info. I also did a bunch of tests for you.
 
tldr;

If you are just using 1 tool - use [Google Lighthouse](https://developer.chrome.com/docs/lighthouse/overview/) in the Chrome Developer Tools

Run a Desktop and a Mobile test | Mode: Navigation | Categories: Just check them all

It will give you some great automatic feedback and is good to check before rolling new versions. 

### Background
**Background on Accessibility & Testing:**

Regarding accessibility standards, the primary specifications you should focus on are [WCAG](https://www.w3.org/TR/WCAG/) (Web Content Accessibility Guidelines) and [ARIA](https://www.w3.org/TR/wai-aria/) (Accessible Rich Internet Applications), both maintained by the W3C (World Wide Web Consortium).
It's important to note that there isn't a single, comprehensive automated test that covers the entire spectrum of accessibility compliance. Instead, achieving compliance typically involves a combination of:

- Manual testing
- Semi-automated browser extensions
- Automated tools (not necessary for your current needs)

### Your Results

1) Google Lighthouse
image

Here are your Desktop Results (mobile were similar)

_Issues:_ 
- color contrast
- Link spacing must allow for 48px x 48px touch target
- Links must be distinguishable without relying on color
-  Links must have discernible text

2) I did additional testing using the [Axe DevTools](https://chromewebstore.google.com/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd?pli=1) - which is a browser extension and works in a similar way to Lighthouse but the results can be easier to parse. 

image

Issues (with better stats): 
- 1175 Contrast Issues - 
- 31 - Links must be distinguishable without relying on color
- 2 - Links must have discernible text
- 7 - Scrollable region must have keyboard access

3) Manual Testing - you passed all of my initial checks. 

I did the top manual tests for the updated site - it scored really well: 

Keyboard - There are several to test here. I just did them & you are in good shape. 

- [x] Is tab order correct 
- [x] Are elements focusable when the user tabs through
- [x] Do you have a skip to the main content link which is only visible when tabbed through with a keyboard? 

ZOOM/Magnification 
- [x] Verify if the elements are displayed nicely and are still interactable.

Screen Readers
- [x] Tested for basic navigation. 
This is super advanced and takes practice to learn. If you are on Mac use the built-in VoiceOver, and on a PC use [NVDA](https://www.nvaccess.org/). I just did a brief test with Voice Over and nothing horrible stood out.

stringtheory:
I am glad to dig into the accessibility issues further and propose some recommendations to get you up to WCAG 2.1 AA Compliance (this is the current best-practice target to aim for). That seems far, far outside the scope of this issue - but I can start a thread if someone can recommend the best place for it.

lornajane:
Thank you @stringtheory! We would definitely like to improve our accessibility game, and you have already helped us with these pointers :) Could you create a new issue for anything we are missing (in this same repo), please? Even if we don't get all of it sorted out before this round of patches, we will keep iterating.

ralfhandl:
Created issue https://github.com/OAI/OpenAPI-Specification/issues/3946 to tackle this.

@stringtheory Please feel free to comment and add more recommendations.

handrews (dismissed):

ralfhandl:
Feedback:
* keep new default spacing
* keep new frontmatter
* use (only) static versions

miqui (approved):

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "editor.formatOnSave": false
```
romanblanco, line 2: @ralfhandl, editor settings probably weren't intended to be introduced to the codebase

ralfhandl: @OAI/tsc Opinions?

Add `.vscode` to `.gitignore`, or rather set "format on save" to true and encourage/enforce Prettier for code and markdown?

handrews: I would generally prefer not to have editor-specific directives in .gitignore.  People use a variety of editors, and treating anything like this as if it enforces some sort of policy is at best misleading.  If you want to enforce something on commit, use a commit hook.

ralfhandl: No traces of VS Code now

romanblanco (approved):
Editor configuration was removed out of the changes :+1:

darrelmiller (approved):

Issue #708, opened 2016-06-02
Discriminator field (unnecessarily) required


dmejia:
Is it possible to remove the restriction of making the field that acts as discriminator required (and assume that when not returned, the object is just of the 'base' type defined in the schema)?

i.e (using the Pet example from the docs) 

**Schema:**

```
(...)
properties: 
    pets: 
       type: array
       items:
           type: Pet
```

**Payload:**

```
{
   pets: [
   {
      name: "Some generic pet" 
   },
   {
      @type: Dog
      name: "Some dog"
      packSize: ...
   },
   {
      @type: Cat
      name: "Some cat"
      huntingSkill: ...
   },
   {
       name: "Some other generic pet"
   }
}
```

The first and third items are just of type _"Pet"_ which would be explicit from the schema of the "pets" field. _"Pet"_ would therefore define the @type field but it won't be required. 
We use schema.org and rely on polymorphic schemas extensively but we currently only return a type discriminator when is not obvious from the schema (i.e when returning a subtype). Unless I misunderstand it, adhering to the spec would require us to return @type everywhere which would significantly increase payload size (and thus would prob be a non-starter). 

Let me know if you have any thoughts or reasons to not go this route.

ePaul:
Related: #707.

dmejia:
Any comments on this?

ralfhandl:
I like this proposal, OData also omits the `@odata.type` field for instances of the base type.

webron:
Tackling PR: #741

Relequestual:
@webron #741 was closed in favour of merged #894.
Does that mean that this is now resolved? Appears not as of current branch:v3.0.2-dev

Is this being tracked on a new PR?


As an aside, having the ability to shortcut any/all/one Of sounds like a nice thing! Maybe we should look at putting this into JSON Schema itself!?

handrews:
@Relequestual `discriminator` does not fit well with other aspects of JSON Schema- I looked into it while we considered `if`/`then`/`else` and agains during the whole `unevaluatedProperties` saga.  I'd have to do some digging to explain exactly why.

Relequestual:
No, that's fine. The same functionality can be provided using if/then/else.

tmtron:
another use-case for an optional discriminator is API-evolution.

i.e. we may have an API that only supports one type (e.g. Dogs) and thus has no discriminator
But later we need to add support for Cats

Then we can change the API in a backwards compatible way
[...]

PR #3953, opened 2024-07-12
Align 3.1.1 with 3.0.4

ralfhandl:
Editorial changes taken over from 3.0.4.

Additional changes here are to be synced back to 3.0.4 via #3966

### DIFF
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -365,7 +365,6 @@ Field Name | Type | Description
 <a name="oasTags"></a>tags | [[Tag Object](#tagObject)] | A list of tags used by the document with additional metadata. The order of the tags can be used to reflect on their order by the parsing tools. Not all tags that are used by the [Operation Object](#operationObject) must be declared. The tags that are not declared MAY be organized randomly or based on the tools' logic. Each tag name in the list MUST be unique.
 <a name="oasExternalDocs"></a>externalDocs | [External Documentation Object](#externalDocumentationObject) | Additional external documentation.
 
-
 This object MAY be extended with [Specification Extensions](#specificationExtensions).
 
 #### <a name="infoObject"></a>Info Object
@@ -433,7 +432,7 @@ Contact information for the exposed API.
 Field Name | Type | Description
 ---|:---:|---
 <a name="contactName"></a>name | `string` | The identifying name of the contact person/organization.
-<a name="contactUrl"></a>url | `string` | The URI for to the contact information. This MUST be in the form of a URI.
+<a name="contactUrl"></a>url | `string` | The URI for the contact information. This MUST be in the form of a URI.
 <a name="contactEmail"></a>email | `string` | The email address of the contact person/organization. This MUST be in the form of an email address.
 
 This object MAY be extended with [Specification Extensions](#specificationExtensions).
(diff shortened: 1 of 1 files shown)


versions/3.1.1.md
```diff
@@ -2790,8 +2788,15 @@ The OpenAPI Specification allows combining and extending model definitions using
...
-To support polymorphism, the OpenAPI Specification adds the `discriminator` keyword.
+To support polymorphism, the OpenAPI Specification adds the `discriminator` field.
 When used, the `discriminator` indicates the name of the property that hints which schema definition is expected to validate the structure of the model.
+<!-- 3.0.4 contains the following text: does it also apply to 3.1.1?
+As such, the `discriminator` field MUST be a required field.
+There are two ways to define the value of a discriminator for an inheriting instance.
+- Use the schema name.
+- Override the schema name by overriding the property with a new value. If a new value exists, this takes precedence over the schema name.
+As such, inline schema definitions, which do not have a given id, *cannot* be used in polymorphism.
+-->
```
ralfhandl, line 2799: Does this 3.0.4 text also apply to 3.1.1?

versions/3.1.1.md
```diff
@@ -2790,8 +2788,15 @@ The OpenAPI Specification allows combining and extending model definitions using
 `allOf` takes an array of object definitions that are validated *independently* but together compose a single object.
 
 While composition offers model extensibility, it does not imply a hierarchy between the models.
-To support polymorphism, the OpenAPI Specification adds the `discriminator` keyword.
+To support polymorphism, the OpenAPI Specification adds the `discriminator` field.
```
ralfhandl, line 2791: I cross-checked our predominant use of "keyword" and "field":
- "keyword" if it is defined by JSON Schema
- "field" if it is defined by OAS (in one of the "Fixed Fields" sections)

miqui:
@handrews to check discriminator concern.

versions/3.1.1.md
handrews: @ralfhandl this is separate from the changes I made - I don't think I ever touched this section (which is the section that #3951 is about).  I can't recall the details of this MUST, but in #708 we apparently decided that the field would continue to be required.

Ideally, we would change this part in both versions, replacing "given id" with "component name" or "name within the Components Object":

> As such, inline schema definitions, which do not have a given id, *cannot* be used in polymorphism.

This is confusing, because the "id" here is the name under the Components Object, not the `$id` keyword (which is supported in 3.1).  I think I also avoided the phrase "inline schemas" when I did the other edits but can't remember right now- I'm less concerned about that.

lornajane (commented):
This looks good but we need to resolve the question about discriminator before we can approve

versions/3.1.1.md
ralfhandl: This [text is present in 3.1.0](https://spec.openapis.org/oas/v3.1.0.html#composition-and-inheritance-polymorphism) and apparently got lost during the refactoring, adding it back now.

The last sentence is not only confusing, it is also wrong as one can use `$ref` pointing to an "inline schema" anywhere in the OpenAPI document.

Adjusted the text accordingly and will "backport" this to 3.0.4 once we agree.

ralfhandl:
> This looks good but we need to resolve the question about discriminator before we can approve

@lornajane Question resolved, please review

versions/3.1.1.md
```diff
@@ -2790,8 +2788,12 @@ The OpenAPI Specification allows combining and extending model definitions using
 `allOf` takes an array of object definitions that are validated *independently* but together compose a single object.
 
 While composition offers model extensibility, it does not imply a hierarchy between the models.
-To support polymorphism, the OpenAPI Specification adds the `discriminator` keyword.
+To support polymorphism, the OpenAPI Specification adds the [`discriminator`](#schemaDiscriminator) field.
 When used, the `discriminator` indicates the name of the property that hints which schema definition is expected to validate the structure of the model.
+As such, the `discriminator` field MUST be a required field.
+There are two ways to define the value of a discriminator for an inheriting instance.
+- Use the schema name.
+- [Override the schema name](#discriminatorMapping) by overriding the property with a new value. If a new value exists, this takes precedence over the schema name.
```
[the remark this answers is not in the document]
handrews: Looks good- I checked that "schema name" is the term we're using elsewhere and it is, so this works for me!

handrews (approved):

PR #3973, opened 2024-07-24
3.1.1: improved RFC section links

ralfhandl:
* Link to `#section-x` instead of `#autoid-y` or `#page-z`
* All RFCs are added to appendix Normative References
* Informative reference to OpenAPI-Learn


mikekistler (dismissed):
Looks good. 👍

lornajane:
I think this looks good, but I'm not sure how someone reading the markdown source can use these links. Are they just expected to know where to look for the replacements in our build scripts, or is there a better way to tackle it?

ralfhandl:
> I'm not sure how someone reading the markdown source can use these links

If you mean [ReSpec references](https://respec.org/docs/#references-0) like

~~~md
is unrelated to [[JSON-Schema-2020-12|JSON Schema]].
~~~

then the answer unfortunately is
* Go to https://www.specref.org/ and search for `JSON-Schema-2020-12`

In this example the answer is https://www.specref.org/?q=JSON-Schema-2020-12.

Current assumption is
* Markdown is an "editor-friendly" source format, and ReSpec references guarantee that the referenced document is automatically listed as a normative reference, or an informative if preceded by `?`, and the build script will complain about mistyped SPEC-IDs
* The "reader-friendly" end product is the respec-ified static HTML

We can keep the markdown "reader-friendly" and educate editors to (1) stick to reference patterns recognized by the build script, and (2) adjust the build script for any new external reference that is not an RFC.

Which way do we want to go?

miqui (dismissed):

lornajane:
Discussion in TDC this week, we are overall not concerned about the destination of the hyperlinks being behind another level of abstraction

ralfhandl:
Closed in favor of #3996

PR #3987, opened 2024-08-01
Port 3.1.1 PRs to 3.2.0

handrews:
This should include almost all PRs that went into 3.1.1 except for the most recent ones synchronizing wording between 3.0.4 and 3.1.1 (I tried adding those but git couldn't figure out how to apply the commits so the wording changes probably aren't quite the same for 3.2.0 and will need to be done separately).

All of the referencing-related PRs are ported at the end (I was going to make 2 PRs at one point, but decided to consolidate it all).  You might have to click "Load more commits" to see them in the commit list below.

As with the larger 3.0.4->3.1.1 ports, all of the original commits are preserved and note the original 3.0.4 or 3.1.1 PR for reference.

The only substantive PR that is not ported is the change in deprecation status for the Path Item `$ref`, which is not a straight port and is being tracked separately in issue #3734.


ralfhandl (approved):
Looks familiar 😄

PR #3990, opened 2024-08-01
Link all specs and all versions

handrews:
Change the site to "Publications" as it is both specs and registries.


handrews:
This has some issues and I need to post a rendered version- I'll open a new PR shortly.

Issue #3938, opened 2024-07-02
Clarification about the meaning of an empty security array


MathieuVeber:
Hi, while I was trying to implement the specification, I ran into some existential questions about security requirements!

It's unclear to me if an empty security array (`[]`) is equivalent to a security array with an empty object (`[{}]`)... The spec doesn't even mention if it would be valid to have an empty security array at the OpenApi Object level and what would that mean...

- [OpenAPI Object](https://spec.openapis.org/oas/v3.1.0#openapi-object)
> A declaration of which security mechanisms can be used across the API. The list of values includes alternative security requirement objects that can be used. Only one of the security requirement objects need to be satisfied to authorize a request. Individual operations can override this definition. To make security optional, an empty security requirement ({}) can be included in the array.

- [Operation Object](https://spec.openapis.org/oas/v3.1.0#operation-object)
> A declaration of which security mechanisms can be used for this operation. The list of values includes alternative security requirement objects that can be used. Only one of the security requirement objects need to be satisfied to authorize a request. To make security optional, an empty security requirement ({}) can be included in the array. This definition overrides any declared top-level security. To remove a top-level security declaration, an empty array can be used.

handrews:
The empty object here is just a normal Security Requirement Object that doesn't have any entries.  And therefore does not require any security requirements to be fulfilled.  It does not have any special behavior - it is just the natural outcome of how the Security Requirement Object works that an empty object allows everything.

Since the OpenAPI Object and the Operation Object take arrays of Security Requirement Objects which are logically OR'd, an empty requirement list also allows everything.

You can stuff as many empty Security Requirement Objects into the array as you want, they still all get evaluated the same way, which means that `[]`, `[{}]`, and `[{}, {}, {}, {}, {}]` all behave the same.  It's like adding multiple `true`s to an OR.  `true OR true OR true` is still true.

MathieuVeber:
Thanks for your explanation 💡 

We thought for a moment that an empty security array `[ ]` meant that nothing is allowed whereas an empty security requirement object `[{}]` allowed everything 😅

@rvedotrc here is your answer

rvedotrc:
> Since the OpenAPI Object and the Operation Object take arrays of Security Requirement Objects which are logically OR'd, an empty requirement list also allows everything.

So you're saying logical-OR is defined for zero operands, and the result of that is "true", right?

[...]

PR #3994, opened 2024-08-05
Explicitly state what `security: []` means

rvedotrc:
It's currently not clear from the spec what `security: []` means. It's stated that it's _allowed_, but it doesn't say what it _does_. Obvious guesses are that it either authorizes everything, or that it authorizes nothing; but it's not stated which.

As per https://github.com/OAI/OpenAPI-Specification/issues/3938, I'm told that the answer is "authorize everything"; so this PR seeks to make that explicit.

On the other hand, if that answer is incorrect, please let me know what [] in fact means, and I'll see if I can update this PR accordingly.

See also: #3995.


ralfhandl (changes requested):
This seems to add a new feature, which we can't do in a patch release.

Please consider re-targeting this PR to the `v3.2.0-dev` branch.

rvedotrc:
I'm not sure what you mean by "adding a new feature" here, so I'll ask you to be a bit more black and white, if I may: are you saying that this rewording (or some variation of it) will not be accepted into 3.1, but could be OK in 3.2?

PR #3999, opened 2024-08-06
Specify that tag names are case-sensitive

rvedotrc:
It's not currently stated whether tags are case-sensitive or -insensitive (principally, for matching a tag on an Operation to a Tag object). This PR seeks to rectify that.

Or, maybe we should define them as case insensitive.


ralfhandl (approved):
Checked with a popular OAD display tool, and it treats tags case-sensitive.

ralfhandl:
@TSC: how are typical tools dealing with tags that differ only in case?

handrews:
@ralfhandl We already require them to be case-sensitive.  This is a bit more clear in 3.1.1 and 3.0.4 where there is now an "HTTP and Case-Sensitivity" section noting that everything is case-sensitive unless it directly maps to an HTTP concept that is defined as case-insensitive.  Tags clearly are not in the HTTP category, so they are case-sensitive.

ralfhandl:
@handrews How about renaming the new section to just "Case Sensitivity"? That would draw more attention, and we still can cover HTTP in the unchanged section text?

https://github.com/OAI/OpenAPI-Specification/blob/b7c20ee6f384b599d428c9f54c8a9a6d21ac4767/versions/3.0.4.md?plain=1#L62-L65

ralfhandl:
@rvedotrc If this new section also covers your concern, then please close this PR.

handrews:
@ralfhandl I'd support that rename!

PR #4001, opened 2024-08-07
3.0.4: minor nits

ralfhandl:
Mostly harmonization of terminology and formatting:
1. first mention of a normative reference or an OAS-defined Object in a (sub)*section is a link, additional mentions are not
2. OAS-defined Foo Bar Objects are written in this style, and are not `monospaced`
3. "example" instead of "sample" - this spec is not about statistics
4. "OpenAPI Object" instead of "root"
5. fixed fields are `monospaced`
6. field values are monospaced in JSON notation: `true`, `false`, `null`, `"header"`, ...
7. combination of fixed field name with example value uses JS notation: `in: "header"`, combining rules 5 and 6
8. exception to 5-7 is colloquial use, for example "values of type `array` or `object`" - "type" is not monospaced, so the monospaced values aren't enclosed in double quotes, I find this easier to read
9. "attribute" is only used in the XML context and means "XML attribute"
10. use Oxford commas, avoid Shatner commas


handrews (dismissed):
All of these changes are helpful!

handrews:
(oh wait it's a draft... well, I'll review it again when you're done :-)

versions/3.0.4.md
```diff
@@ -2409,22 +2411,22 @@ This mechanism is used by [Link Objects](#link-object) and [Callback Objects](#c
 The runtime expression is defined by the following [ABNF](https://tools.ietf.org/html/rfc5234) syntax
 
 ```abnf
-    expression = ( "$url" / "$method" / "$statusCode" / "$request." source / "$response." source )
-    source = ( header-reference / query-reference / path-reference / body-reference )
+    expression = "$url" / "$method" / "$statusCode" / "$request." source / "$response." source
```
ralfhandl, line 2414: Are we fine with the fact that ABNF literal text strings enclosed in double-quotes are case-insensitive?

For example `"$url"` matches values `$uRl`, `$url`, `$URL` and any other case variation.

Do we assume this to be general knowledge, or should we call it out?

handrews:
@ralfhandl could you include this clarification?

* https://github.com/OAI/OpenAPI-Specification/pull/3818#issuecomment-2282862601

versions/3.0.4.md
```diff
@@ -2682,7 +2684,7 @@ Other than the JSON Schema subset fields, the following fields MAY be used for f
...
 | ----------------------------------------------- | :-----------------------------------------------------------: | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| <a name="schema-nullable"></a>nullable           |                           `boolean`                           | A `true` value adds `"null"` to the allowed type specified by the `type` keyword, only if `type` is explicitly defined within the same Schema Object. Other Schema Object constraints retain their defined behavior, and therefore may disallow the use of `null` as a value. A `false` value leaves the specified or default `type` unmodified. The default value is `false`.                                                                    |
```
ralfhandl, line 2685: The old wording seems to mean

> Mentally put the string value of `type` into an array and add the string `"null"` to this array, then interpret the array in the JSON Schema sense because OpenAPI explicitly forbids array values for the `type` keyword in the preceding section.

The new text hopefully avoids this mental gymnastics and the possible confusion of `null` as an allowed value and `"null"` as an additional string item in the array value of the JSON Schema `type` keyword.

mikekistler (dismissed):
Looks good. 👍

versions/3.0.4.md
mikekistler: Not only are array values for type forbidden,  but `null` is also forbidden.
>  null is not supported as a type

But I'm not sure the new wording is better. In particular, the "allowed values" reference could be confused with `enum`. Oh, but actually we should add `null` to the enum if it exists. [JSON Schema](https://json-schema.org/draft-05/draft-wright-json-schema-validation-00.pdf) says:
> Elements in the array MAY be of any type, including null.
> An instance validates successfully against this keyword if its value is equal to one of the elements in this keyword's array value.

Oh my.

versions/3.0.4.md
```diff
@@ -1072,20 +1072,20 @@ These fields MAY be used with either `content` or `schema`.
...
+When `example` or `examples` are provided in conjunction with the `schema` field, the example SHOULD match the specified schema and follow the prescribed serialization strategy for the parameter.
 The `example` and `examples` fields are mutually exclusive, and if either is present it SHALL _override_ any `example` in the schema.
 
-Serializing with `schema` is NOT RECOMMENDED for `in: cookie` parameters, `in: header` parameters that use HTTP header parameters (name=value pairs following a `;`) in their values, or `in: header` parameters where values might have non-URL-safe characters; see [Appendix D](#appendix-d-serializing-headers-and-cookies) for details.
+Serializing with `schema` is NOT RECOMMENDED for `in: "cookie"` parameters, `in: "header"` parameters that use HTTP header parameters (name=value pairs following a `;`) in their values, or `in: "header"` parameters where values might have non-URL-safe characters; see [Appendix D](#appendix-d-serializing-headers-and-cookies) for details.
```
mikekistler, line 1083: I'm curious about the convention we are using for string values here and elsewhere. The original text has `in: cookie", which in YAML is valid and I think is preferred over the new text `in: "cookie"`. The new text is not valid JSON. So what is our convention and why did we choose this?

ralfhandl: @mikekistler Please see the PR description:

> 5. fixed fields are monospaced

This is the predominant style for fixed fields in the existing text.

> 6. field values are monospaced in JSON notation: `true`, `false`, `null`, `"header"`, ...

This is done for example in the first table in section https://spec.openapis.org/oas/v3.0.3.html#fixed-fields-9, and I find it helpful that field `names` and field `"values"` are clearly distinguished, so I applied this pattern wherever I wondered whether something is a field name or a field value.

The second table in the same section does not use that convention and is rather hard to read. For example

> When `style` is `form`, the default value is `true`.

Here `form` is the content of a JSON string: is `true` also the content of a JSON string, or is it a JSON boolean? How much context do I have to read to figure that out?

The pattern `in: query` was introduced by Henry in 3.0.4, and I first tried to rephrase that into a style already present in 3.0.3:
> if `in` is `"query"`

That felt clumsy in some places, so I opted for `in: "query"` etc., combining conventions 5 and 6.

The PR description is my stab at an "editor's style guide", and of course open for discussion, and destined for documentation, for example in `CONTRIBUTING.md`

versions/3.0.4.md
handrews: I agree with @mikekistler - I find the new wording more confusing.  I would go with something like:

> This keyword only takes effect if `type` is explicitly defined within the same Schema Object.  A `true` value indicates that both `null` values and values of the type specified by `type` are allowed.  Other Schema Object constraints retain their defined behavior, and therefore may disallow the use of `null` as a value. A `false` value leaves the specified or default `type` unmodified. The default value is `false`.

ralfhandl: Took over Henry's proposal

versions/3.0.4.md
mikekistler: Sorry I glossed over the PR description, and thank you for writing this! 

On the particular style choice, I think I would slightly prefer YAML to JS, just to avoid adding another syntax style to the doc, but I can live with either as long as we are consistent.

mikekistler (approved):
Looks good. 👍

PR #4007, opened 2024-08-08
Clarify `security`; state what `[]` means

rvedotrc:
A partial fix for #3938 (partial because this PR only addresses 3.1.1-dev). Note that I have also created **draft** PRs to port this to the other two `-dev` branches. I have no opinion on in which order they should be merged.

----

Thoughts as far as I can tell so far, as we still try to find some consensus.

Mostly we're talking about two cases: (1) security is not specified, and (2) security is specified but empty (`[]`); and for each of those two cases, obvious options include (a) define it as "allow all" aka "no security"; (b) define it as "deny all" aka "there is no way to authorize"; (c) state that the behaviour is undefined; (d) state that the behaviour is implementation-defined.

## Case 1: Leaving `security` unspecified

* (a) "allow all" aka "no security"
* (b) "deny all" aka "no way in". I don't think anyone is suggesting this option.
* (c) undefined (which would be a sign that `security` should actually be a required field). I don't think anyone is suggesting this option either.
* (d) implementation-defined

## Case 2: `security` is `[]`

* (a) "allow all" aka "no security". A perfectly reasonable possibility; albeit one which makes the empty list a special case, not following the usual rules of list logic.
* (b) "deny all" aka "no way in". Also a perfectly reasonable possibility, and one where the list logic follows the usual rules. Under this option, `security: []` is how you express "deny all", and `security: [{}]` is one way of expressing "allow all".
* (c) undefined
* (d) implementation-defined

Whichever one we opt for, could we also state that `security: []` is deprecated? "Implementers of services SHOULD NOT introduce new instances of `security: []`, and SHOULD seek an alternative formulation.", for example?


versions/3.1.1.md
```diff
@@ -314,12 +314,22 @@ This is the root object of the [OpenAPI document](#openapi-description).
...
 | <a name="oas-external-docs"></a>externalDocs            | [External Documentation Object](#external-documentation-object) | Additional external documentation.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      |
 
 This object MAY be extended with [Specification Extensions](#specification-extensions).
 
+###### <a name="oas-security-field"></a>The `security` field
+
+The `security` field describes how requests are authorized. If omitted, then requests are always authorized; otherwise, it is an array of security requirement objects, only one of which needs to be satisfied for the request to be authorized.
```
ralfhandl, line 325: What does
> then requests are always authorized
mean?

Top-of-mind interpretations:
1. Requests need authorization, and I don't tell you how, or
2. Requests don't need authorization

I'd argue that it means the first, because the second can be made explicit via `security: [{}]`.

versions/3.1.1.md
```diff
@@ -314,12 +314,22 @@ This is the root object of the [OpenAPI document](#openapi-description).
...
+
+Because the empty [security requirement object](#security-requirement-object) `{}` will always authorize the request, it follows that `security: [{}]` means the same as omitting `security` entirely: namely, all requests are authorized.
```
ralfhandl, line 327: I think "no info" means "no info", see above.

versions/3.1.1.md
rvedotrc: If you mean "what was the author's intention behind those words", then I can answer authoritatively: it means #2.

However, I can see how it could be interpreted the other way. Do you have a suggested alternative wording?

versions/3.1.1.md
rvedotrc: Could we perhaps be a little more precise with the language here, please? When you say that X means Y, should that be interpreted as:

1. "Personally, I have always taken X as implying Y" (i.e. a reference to an existing spec)
2. or: "I believe that the majority of people using OpenAPI have always taken X as implying Y" (existing spec)
3. or: "I would like X to imply Y" (i.e. with reference to a dev/future spec)

(1) doesn't need any rationale; if that's your interpretation, then that's your interpretation. Nobody can deny that.

(2) _may_ benefit from supporting evidence (depending on how controversial it seems, for example)

(3) may well benefit from a supporting argument, especially if "X implies Y" is not clearly true in the existing spec.

So which is it?

rvedotrc: For a long comment (by me), that sure was a waste of space. Sorry, ignore me. I'm not sure where I was going with that.

versions/3.1.1.md
```diff
@@ -314,12 +314,24 @@ This is the root object of the [OpenAPI document](#openapi-description).
...
-| <a name="oas-security"></a>security                    |  [[Security Requirement Object](#security-requirement-object)]  | A declaration of which security mechanisms can be used across the API. The list of values includes alternative security requirement objects that can be used. Only one of the security requirement objects need to be satisfied to authorize a request. Individual operations can override this definition. To make security optional, an empty security requirement (`{}`) can be included in the array.                                                                                                                                               |
+| <a name="oas-security"></a>security                    |  [[Security Requirement Object](#security-requirement-object)]  | See [the `security` field](#oas-security-field).
```
ralfhandl, line 317:
```suggestion
| <a name="oas-security"></a>security                    |  [[Security Requirement Object](#security-requirement-object)]  | A declaration of which security mechanisms can be used across the API. See [The `security` Field](#the-security-field).
```

versions/3.1.1.md
```diff
@@ -314,12 +314,24 @@ This is the root object of the [OpenAPI document](#openapi-description).
...
 | <a name="oas-external-docs"></a>externalDocs            | [External Documentation Object](#external-documentation-object) | Additional external documentation.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      |
 
 This object MAY be extended with [Specification Extensions](#specification-extensions).
 
+###### <a name="oas-security-field"></a>The `security` field
```
ralfhandl, line 323:
```suggestion
###### The `security` Field
```
We now rely on automatic anchors for headings

versions/3.1.1.md
```diff
@@ -314,12 +314,24 @@ This is the root object of the [OpenAPI document](#openapi-description).
...
+
+The `security` field describes how requests are authorized:
+
+- If omitted, then nothing can be inferred about the authorization requirements; the behaviour is implementation-defined.
+- If present but empty (`security: []`), then the behaviour is undefined.
+- Otherwise, it is an array of Security Requirement Objects, only one of which needs to be satisfied for the request to be authorized.
```
ralfhandl, line 329:
```suggestion
- Otherwise, it is an array of [Security Requirement Objects](#security-requirement-object), only one of which needs to be satisfied for the request to be authorized.
```

versions/3.1.1.md
```diff
@@ -314,12 +314,24 @@ This is the root object of the [OpenAPI document](#openapi-description).
...
+
+Because the empty [Security Requirement Object](#security-requirement-object) `{}` will always be satisfied, any `security` list that includes `{}` will allow all requests. In particular, `security: [{}]` means that no security schemes are in use (also known as "no security").
```
ralfhandl, line 331:
```suggestion
Because the empty Security Requirement Object `{}` will always be satisfied, any `security` list that includes `{}` will allow all requests. In particular, `security: [{}]` means that no security schemes are in use (also known as "no security").
```

versions/3.1.1.md
```diff
@@ -967,7 +979,7 @@ Describes a single API operation on a path.
...
-| <a name="operation-security"></a>security         |                [[Security Requirement Object](#security-requirement-object)]                | A declaration of which security mechanisms can be used for this operation. The list of values includes alternative security requirement objects that can be used. Only one of the security requirement objects need to be satisfied to authorize a request. To make security optional, an empty security requirement (`{}`) can be included in the array. This definition overrides any declared top-level [`security`](#oas-security). To remove a top-level security declaration, an empty array can be used.                                                                                            |
+| <a name="operation-security"></a>security         |                [[Security Requirement Object](#security-requirement-object)]                | If present, then this overrides any [top-level `security` field](#oas-security) for this operation. See the definition of that field for details.
```
ralfhandl, line 982:
```suggestion
| <a name="operation-security"></a>security         |                [[Security Requirement Object](#security-requirement-object)]                | A declaration of which security mechanisms can be used for this operation. If present, then this overrides any [top-level `security` field](#oas-security) for this operation. See the [definition of that field](#the-security-field) for details.
```

ralfhandl (commented):
+1, minor style suggestions

versions/3.1.1.md
```diff
@@ -967,7 +979,7 @@ Describes a single API operation on a path.
...
-| <a name="operation-security"></a>security         |                [[Security Requirement Object](#security-requirement-object)]                | A declaration of which security mechanisms can be used for this operation. The list of values includes alternative security requirement objects that can be used. Only one of the security requirement objects need to be satisfied to authorize a request. To make security optional, an empty security requirement (`{}`) can be included in the array. This definition overrides any declared top-level [`security`](#oas-security). To remove a top-level security declaration, an empty array can be used.                                                                                            |
+| <a name="operation-security"></a>security         |                [[Security Requirement Object](#security-requirement-object)]                | A declaration of which security mechanisms can be used for this operation. If present, then this overrides any [top-level `security` field](#oas-security) for this operation. See the [definition of that field](#the-security-field) for details.
```
rvedotrc, line 982:
```suggestion
| <a name="operation-security"></a>security         |                [[Security Requirement Object](#security-requirement-object)]                | A declaration of which security mechanisms can be used for this operation. If present, then this overrides any [top-level `security` field](#oas-security). See the [definition of that field](#the-security-field) for details.
```

@ralfhandl WDYT? The pre-suggestion text could be read as "there's a top-level security field, just for this operation", which is confusing and doesn't make sense.

ralfhandl: Yes, that is clearer. An alternative would be
```suggestion
| <a name="operation-security"></a>security         |                [[Security Requirement Object](#security-requirement-object)]                | A declaration of which security mechanisms can be used for this operation. If present, this overrides a `security` field specified at the [OpenAPI Object](#oas-security) level. See the [definition of that field](#the-security-field) for details.
```

rvedotrc: Yeah, except that reads strangely in the case where there _is_ no top-level `security`.

ralfhandl (dismissed):
Seems to cover the current state of discussion, see also comments on #3995 (which needs to be adjusted once this is approved & merged)

ralfhandl (approved):

versions/3.1.1.md
```diff
@@ -314,12 +314,24 @@ This is the root object of the [OpenAPI document](#openapi-description).
...
 | <a name="oas-external-docs"></a>externalDocs            | [External Documentation Object](#external-documentation-object) | Additional external documentation.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      |
 
 This object MAY be extended with [Specification Extensions](#specification-extensions).
 
+###### The `security` Field
+
+The `security` field describes how requests are authorized:
+
+- If omitted, then nothing can be inferred about the authorization requirements; the behaviour is implementation-defined.
```
handrews, line 327: While the text could have been more clear, I think this is actually pretty well-defined. I don't recall anyone arguing that any tools treat lack of security as anything but "allow all". The only question was whether [] is the same as that default or not.

versions/3.1.1.md
```diff
@@ -314,12 +314,24 @@ This is the root object of the [OpenAPI document](#openapi-description).
...
+- If present but empty (`security: []`), then the behaviour is undefined.
```
handrews, line 328: I think this is implementation-defined rather than undefined. There are only two viable options (which should be stated explicitly): allow-all or deny-all. It is implementation-defined as to which one of those two will happen. Undefined means that arbitrary (and arbitrarily wrong) behavior can happen, and I don't think that's a concern here. (Implementation-defined and undefined are terms that are defined within the OAS for 3.0.4 and 3.1.1, for anyone reading who was not aware of that)

versions/3.1.1.md
rvedotrc: Defined, no. Understood / assumed, yeah, I'll buy that.

Happy to change this to "allow all", but I'll wait for more opinions first.

versions/3.1.1.md
rvedotrc: So would we change this to (paraphrasing) "it's implementation-defined, but you probably want it to be either allow all or deny all"; OR, "the behaviour MUST be either allow-all or deny-all, but which of the two, is implementation-defined" ?

versions/3.1.1.md
miqui: I agree with @handrews  and I am not comfortable with omission being considered imp-defined. IMHO, this is counter productive.   "Allow all" seems (at least for now) a more reasonable approach.

versions/3.1.1.md
miqui: security[] == "allow all"     (+1)
I am not sure what would be the effect of making / interpreting security[] == "deny all".  Probably break some tools.

Again, I am leaning towards @handrews 's comment.  However (at least in my case) it is going to take a while to get used to security[] being in some cases deny-all if we were to go this route.

rvedotrc: https://github.com/OAI/OpenAPI-Specification/pull/4007#discussion_r1711137553

Sounds like there are still some pretty directly opposed viewpoints around.

versions/3.1.1.md
rvedotrc: https://github.com/OAI/OpenAPI-Specification/pull/4007#discussion_r1711137553

Sounds like there are still some pretty directly opposed viewpoints around.

versions/3.1.1.md
handrews: @rvedotrc 

> the behaviour MUST be either allow-all or deny-all, but which of the two, is implementation-defined

I think this would be preferable.  Pretty much everyone thinks it should be one or the other- no other options make sense.  So let's say that and leave it up to implementations which one to pick.  If we want, we can make one RECOMMENDED, but we should only do that if we are certain we will emphasize that direction in future releases.  So I'd probably avoid that for now.

versions/3.1.1.md
handrews: For omitting `security`, it should be straightforward to check what a few major implementations do.  I'm 99.999% sure it is allow-all.  No one seemed to suggest otherwise on the call, and it's pretty inconceivable that leaving out a field that not everyone even wants to use would make a description un-callable.  Some people just describe the HTTP basics.

We could also check what these implementations do with `security: []` although even if they all do the same thing I'd hesitate to use that to define the behavior.

versions/3.1.1.md
rvedotrc: Worksforme.

Because it's not yet clear to me that there's a sufficient consensus behind any specific path forwards, before I update the PR again, let's try this:

Please +1 this comment here, if you think you would approve the PR if it said this:

 * `security` not specified == allow all
 * `security: []` MUST be either allow-all or deny-all, implementation-defined

(or -1 this comment if you think you'd request changes).

ralfhandl: Can't agree to the first bullet point:
* The spec currently does not define a default for this field. Adding a default is a breaking change and needs to be deferred to v4.0

rvedotrc: > Adding a default is a breaking change

Is it? How can you tell, given that the spec did not say what the previous behaviour was?

ralfhandl: > How can you tell, given that the spec did not say what the previous behaviour was?

I can tell because 
* I am omitting `security` to express "I can't tell you which authentication mechanisms will work" and 
* @miqui omits it to express "No authentication required".

Adding a default will break at least one of these interpretations.

handrews:
My thoughts after today's TDC call:

* The `security` field is really a list of _authentication_ mechanisms, so using "allow all" / "deny all" doesn't quite fit because these are _authorization_-related terms (credit for this to @mikekistler and @rvedotrc, I'm just summarizing)
* A lack of `security` means that you don't know anything about authentication, which means that you assume none is needed (but might find out otherwise when you try something)
* An empty array (`security: []`) means that you are being told there are no authentication mechanisms.  It is ambiguous as to whether this is "there are no authentication mechanisms because you don't need any" (which we had been calling "allow all") or "there are no authentication mechanisms because you're not even authorized to know about them, which means you can't use this" (which we had been calling "deny all")

Note that we have a [whole section on "filtering" an OAD due to security constraints](https://spec.openapis.org/oas/v3.1.0#security-filtering), and also a [note that allows an empty Paths Object](https://spec.openapis.org/oas/v3.1.0#paths-object) because a filtered OAD might filter out all of the Path Item Objects if the target audience is not allowed to see them.

So I think the right thing to do here is to acknowledge the ambiguities:

* lack of `security` is basically a "^%$ around and find out" invitation
* logically, `security: []` ought to mean "you need to authenticate but you aren't allowed to know how, so you're out of luck"
* practically, `security: []` has been assumed to mean the same as `security` not being present, so we can acknowledge that ambiguity; this is partially because setting `security: []` at the Operation Object level is said to "remove" the OpenAPI Object-level `security`, _explicitly equating `security: []` with the `security` field being missing_
* `security: [{}]` is an unambiguous way to say "this API (or operation) does not require authentication", so it is RECOMMENDED that OAD authors use that construct if that is what they want to say

I no longer think there is an "undefined" or "implementation-defined" thing here.  Lack of `security` is "we're not telling you, just try it", and the "removes" wording in the Operation Object explicitly equates `security: []` with that state, so we can't make it mean anything else.

rvedotrc:
> explicitly equating `security: []` with the security field being missing

Explicit in the sense that it's hard to argue that there's a different, better interpretation, yes :-) 

I'll have a go at creating a wording to match what you just wrote @handrews

ralfhandl (approved):

mikekistler (changes requested):
After some thought and investigation, I do not think any changes are needed to clarify the behavior of the `security` field in the current specifications. I've given an explanation for this in a [comment on Issue 3938](https://github.com/OAI/OpenAPI-Specification/issues/3938#issuecomment-2308874884).

My recommendation is that we close this PR without merge.

miqui:
> After some thought and investigation, I do not think any changes are needed to clarify the behavior of the `security` field in the current specifications. I've given an explanation for this in a [comment on Issue 3938](https://github.com/OAI/OpenAPI-Specification/issues/3938#issuecomment-2308874884).
> 
> My recommendation is that we close this PR without merge.

I am 100% in agreement with @mikekistler

lornajane:
TDC agrees that this is too much additional content, we only need a small clarification on what the empty array can mean, we will try to work on wording to answer #3938

PR #4009, opened 2024-08-11
Build script: source-of-truth

### LINKED ISSUE #3576: What is our authoritative spec URL and how do people find it?
Currently, some official citation sites point to the rendered HTML rather than the authoritative markdown files.  There is also confusion over what the official, authoritative URL for the authoritative markdown actually is.  Is it the location on `main`?  Or is it the URL that uses a `tag` value?  And that brings up the question of whether a URL that is tied to GitHub internals and structure (which can be tricky to update when, for example, we have broken external links when other people move their documents) and domain is appropriate for a "stable" URL.

So, what is the authoritative URL, and how to we ensure that people find that, know what it is, and cite it appropriately?

ralfhandl:
Builds on
* #4008 

Fixes #3576

Adds
- [x] added thisVersion and canonicalURL pointing to `https://spec.openapis.org/oas/vM.m[.p].html`
- [x] section "Status of this Document" now refers to HTML version as the source of truth
- [x] Measure test coverage with c8 because vitest's built-in coverage doesn't consider child processes
- [x] Add test cases to increase coverage


miqui (approved):

rvedotrc:
Could you perhaps change the base of this PR, to the head of #4008?

handrews (approved):

mikekistler (approved):
Looks good! 👍

PR #4011, opened 2024-08-12
Clarify `security`; state that `[]` is undefined behaviour

rvedotrc:
Port of #4007 to v3.2.0-dev


handrews:
In today's TDC we agreed to re-try this based on @mikekistler 's "open world" view, so closing these three attempts.

@rvedotrc thank you for your contributions here- it is far from the first topic where we've had to go through multiple approaches, attempts, and PRs.  Regardless of who writes the final PR, this has helped us immensely in figuring the problem out.

PR #4012, opened 2024-08-12
Clarify `security`; state that `[]` is undefined behaviour

rvedotrc:
Port of #4007 to v3.0.4-dev


versions/3.0.4.md
```diff
@@ -249,12 +249,24 @@ This is the root object of the [OpenAPI document](#openapi-description).
...
 | <a name="oas-external-docs"></a>externalDocs | [External Documentation Object](#external-documentation-object) | Additional external documentation.                                                                                                                                                                                                                                                                                                                                                                        |
 
 This object MAY be extended with [Specification Extensions](#specification-extensions).
 
+###### The `security` Field
+
+The `security` field describes how requests are authorized:
+
+- If omitted, then nothing can be inferred about the authorization requirements; the behaviour is implementation-defined.
```
handrews, line 262: While the text could have been more clear, I think this is actually pretty well-defined.  I don't recall anyone arguing that any tools treat lack of `security` as anything but "allow all".  The only question was whether `[]` is the same as that default or not.

versions/3.0.4.md
```diff
@@ -249,12 +249,24 @@ This is the root object of the [OpenAPI document](#openapi-description).
...
+- If present but empty (`security: []`), then the behaviour is undefined.
```
handrews, line 263: I think this is implementation-defined rather than undefined.  There are only two viable options (which should be stated explicitly): allow-all or deny-all.  It is implementation-defined as to which one of those two will happen.  Undefined means that arbitrary (and arbitrarily wrong) behavior can happen, and I don't think that's a concern here.  (Implementation-defined and undefined are terms that are defined within the OAS for 3.0.4 and 3.1.1, for anyone reading who was not aware of that)

handrews:
Oh I didn't see that there is a 3.1.1 that we're doing before this 3.0.4?  

This is why we don't usually post the same fix to multiple branches at once- it's too confusing when there are a lot of notifications.  We usually do the oldest branch, and then *after* that is accepted and merged we post any forward-port PRs.

rvedotrc:
That's why this is a draft.

PR #4014, opened 2024-08-13
3.1.1: minor nits

ralfhandl:
This is a port of
* #4001

Mostly harmonization of terminology and formatting:
1. first mention of a normative reference or an OAS-defined Object in a (sub)*section is a link, additional mentions are not
2. OAS-defined Foo Bar Objects are written in this style, and are not `monospaced`
3. "example" instead of "sample" - this spec is not about statistics
4. "OpenAPI Object" instead of "root"
5. fixed fields are `monospaced`
6. field values are monospaced in JSON notation: `true`, `false`, `null`, `"header"`, ...
7. combination of fixed field name with example value uses JS notation: `in: "header"`, combining rules 5 and 6
8. exception to 5-7 is colloquial use, for example "values of type `array` or `object`" - "type" is not monospaced, so the monospaced values aren't enclosed in double quotes, I find this easier to read
9. "attribute" is only used in the XML context and means "XML attribute"
10. use Oxford commas, avoid Shatner commas

### DIFF
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -61,7 +61,7 @@ Some examples of possible media type definitions:
 The HTTP Status Codes are used to indicate the status of the executed operation.
 Status codes SHOULD be selected from the available status codes registered in the [IANA Status Code Registry](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml).
 
-### HTTP and Case Sensitivity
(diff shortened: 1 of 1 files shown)


mikekistler (approved):
Looks good! 👍

miqui (approved):

PR #4015, opened 2024-08-13
removes the use of integer type from registries format

baywet:
context https://github.com/microsoft/kiota/issues/5147
https://json-schema.org/draft/2020-12/json-schema-validation#name-type
https://spec.openapis.org/oas/latest.html#data-types

This adds the missing integer alternative type for integer formats.

### DIFF
--- a/_includes/format-entry.md
+++ b/_includes/format-entry.md
@@ -2,7 +2,7 @@
 
 ## {{ page.slug }} - {{ page.description }}
 
-Instance Type: `{{ page.base_type | join:', ' }}`.
+JSON data type: `{{ page.base_type | join:', ' }}`.
 
 {{ include.summary }}
 
--- a/_includes/format-entry.md
+++ b/_includes/format-entry.md
@@ -2,7 +2,7 @@
 
 ## {{ page.slug }} - {{ page.description }}
 
-JSON Data Type: `{{ page.base_type | join:', ' }}`.
+Instance Type: `{{ page.base_type | join:', ' }}`.
 
 {{ include.summary }}
 
--- a/_includes/format-entry.md
+++ b/_includes/format-entry.md
@@ -2,7 +2,7 @@
 
 ## {{ page.slug }} - {{ page.description }}
 
-JSON data type: `{{ page.base_type | join:', ' }}`.
+JSON Data Type: `{{ page.base_type | join:', ' }}`.
 
 {{ include.summary }}
 
--- a/_includes/format-entry.md
+++ b/_includes/format-entry.md
@@ -2,7 +2,7 @@
 
 ## {{ page.slug }} - {{ page.description }}
 
-Base type: `{{ page.base_type | join:', ' }}`.
+JSON Data Type: `{{ page.base_type | join:', ' }}`.
 
 {{ include.summary }}
 
--- a/registry/format.md
+++ b/registry/format.md
@@ -21,7 +21,9 @@ Please raise a [Pull-Request](https://github.com/OAI/OpenAPI-Specification/pulls
 
 ## Values
 
-|Value|Description|JSON Data Type|Source|Deprecated|
+For the purpose of [JSON Schema validation](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#section-7.1), each format should specify the set of instance types (primitive JSON types) for which it applies. In this registry, these types are shown in the "Instance Type" column.
+
+|Value|Description|Instance Type|Source|Deprecated|
 |---|---|----|---|---|----|
 {% for value in site.format %}| <a href="./{{ value.slug }}.html">{{ value.slug }}</a> | {{ value.description }} | {{ value.base_type | join: ', ' }} | {% if value.source %}<a href="{{ value.source }}">{% if value.source_label %}{{value.source_label}}{% else %}Open{% endif %}</a>{% endif %} | {% if value.deprecated_note %}Yes{% else %}No{% endif %} |
 {% endfor %}
--- a/registry/format.md
+++ b/registry/format.md
@@ -15,14 +15,13 @@ If tools choose to implement any format present in this registry, they SHOULD im
 
 The registry SHOULD NOT contain two entries that have the same meaning, unless all but one have been deprecated.
 
-
 ## Contributing
 
 Please raise a [Pull-Request](https://github.com/OAI/OpenAPI-Specification/pulls) or [Issue](https://github.com/OAI/OpenAPI-Specification/issues) to contribute or discuss a registry value.
 
 ## Values
 
-|Value|Description|Type|Source|Deprecated|
+|Value|Description|JSON Data Type|Source|Deprecated|
 |---|---|----|---|---|----|
 {% for value in site.format %}| <a href="./{{ value.slug }}.html">{{ value.slug }}</a> | {{ value.description }} | {{ value.base_type | join: ', ' }} | {% if value.source %}<a href="{{ value.source }}">{% if value.source_label %}{{value.source_label}}{% else %}Open{% endif %}</a>{% endif %} | {% if value.deprecated_note %}Yes{% else %}No{% endif %} |
 {% endfor %}
--- a/registry/format.md
+++ b/registry/format.md
@@ -21,9 +21,9 @@ Please raise a [Pull-Request](https://github.com/OAI/OpenAPI-Specification/pulls
 
 ## Values
 
-For the purpose of [JSON Schema validation](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#section-7.1), each format should specify the set of instance types (primitive JSON types) for which it applies. In this registry, these types are shown in the "Instance Type" column.
+For the purpose of [JSON Schema validation](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#section-7.1), each format should specify the set of JSON data types for which it applies. In this registry, these types are shown in the "JSON data type" column.
 
-|Value|Description|Instance Type|Source|Deprecated|
+|Value|Description|JSON data Type|Source|Deprecated|
 |---|---|----|---|---|----|
 {% for value in site.format %}| <a href="./{{ value.slug }}.html">{{ value.slug }}</a> | {{ value.description }} | {{ value.base_type | join: ', ' }} | {% if value.source %}<a href="{{ value.source }}">{% if value.source_label %}{{value.source_label}}{% else %}Open{% endif %}</a>{% endif %} | {% if value.deprecated_note %}Yes{% else %}No{% endif %} |
 {% endfor %}
--- a/registry/format.md
+++ b/registry/format.md
@@ -23,7 +23,7 @@ Please raise a [Pull-Request](https://github.com/OAI/OpenAPI-Specification/pulls
 
 For the purpose of [JSON Schema validation](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#section-7.1), each format should specify the set of JSON data types for which it applies. In this registry, these types are shown in the "JSON data type" column.
 
-|Value|Description|JSON data Type|Source|Deprecated|
+|Value|Description|JSON Data Type|Source|Deprecated|
 |---|---|----|---|---|----|
 {% for value in site.format %}| <a href="./{{ value.slug }}.html">{{ value.slug }}</a> | {{ value.description }} | {{ value.base_type | join: ', ' }} | {% if value.source %}<a href="{{ value.source }}">{% if value.source_label %}{{value.source_label}}{% else %}Open{% endif %}</a>{% endif %} | {% if value.deprecated_note %}Yes{% else %}No{% endif %} |
 {% endfor %}
--- a/registries/_format/double-int.md
+++ b/registries/_format/double-int.md
@@ -2,7 +2,7 @@
 owner: mikekistler
 issue: 
 description: an integer that can be stored in an IEEE 754 double-precision number without loss of precision
-base_type: integer
+base_type: number
 layout: default
 ---
 
--- a/registries/_format/int16.md
+++ b/registries/_format/int16.md
@@ -2,7 +2,7 @@
 owner: DarrelMiller
 issue: 
 description: signed 16-bit integer
-base_type: number
+base_type: [number, integer]
 layout: default
 ---
 
--- a/registries/_format/int16.md
+++ b/registries/_format/int16.md
@@ -2,7 +2,7 @@
 owner: DarrelMiller
 issue: 
 description: signed 16-bit integer
-base_type: [number, integer]
+base_type: number
 layout: default
 ---
 
--- a/registries/_format/int32.md
+++ b/registries/_format/int32.md
@@ -2,7 +2,7 @@
 owner: DarrelMiller
 issue: 
 description: signed 32-bit integer
-base_type: number
+base_type: [number, integer]
 layout: default
 source: https://spec.openapis.org/oas/latest.html#data-types
 source_label: OAS
--- a/registries/_format/int32.md
+++ b/registries/_format/int32.md
@@ -2,7 +2,7 @@
 owner: DarrelMiller
 issue: 
 description: signed 32-bit integer
-base_type: [number, integer]
+base_type: number
 layout: default
 source: https://spec.openapis.org/oas/latest.html#data-types
 source_label: OAS
--- a/registries/_format/int64.md
+++ b/registries/_format/int64.md
@@ -2,7 +2,7 @@
 owner: DarrelMiller
 issue: 
 description: signed 64-bit integer
-base_type: [number, string]
+base_type: [number, integer, string]
 layout: default
 source: https://spec.openapis.org/oas/latest.html#data-types
 source_label: OAS
--- a/registries/_format/int64.md
+++ b/registries/_format/int64.md
@@ -2,7 +2,7 @@
 owner: DarrelMiller
 issue: 
 description: signed 64-bit integer
-base_type: [number, integer, string]
+base_type: [number, string]
 layout: default
 source: https://spec.openapis.org/oas/latest.html#data-types
 source_label: OAS
--- a/registries/_format/int8.md
+++ b/registries/_format/int8.md
@@ -2,7 +2,7 @@
 owner: MikeRalphson
 issue: 845
 description: signed 8-bit integer
-base_type: [number, integer]
+base_type: number
 layout: default
 source: https://spec.openapis.org/oas/latest.html#data-types
 source_label: OAS
--- a/registries/_format/int8.md
+++ b/registries/_format/int8.md
@@ -2,7 +2,7 @@
 owner: MikeRalphson
 issue: 845
 description: signed 8-bit integer
-base_type: number
+base_type: [number, integer]
 layout: default
 source: https://spec.openapis.org/oas/latest.html#data-types
 source_label: OAS
--- a/registries/_format/sf-integer.md
+++ b/registries/_format/sf-integer.md
@@ -4,7 +4,7 @@ issue:
 description: structured fields integer as defined in [RFC8941]
 source: https://www.rfc-editor.org/rfc/rfc8941#name-integers
 source_label: RFC 8941
-base_type: [integer, number]
+base_type: number
 layout: default
 ---
 
--- a/registries/_format/uint8.md
+++ b/registries/_format/uint8.md
@@ -2,7 +2,7 @@
 owner: MikeRalphson
 issue: 845
 description: unsigned 8-bit integer
-base_type: [number, integer]
+base_type: number
 layout: default
 source: https://spec.openapis.org/oas/latest.html#data-types
 source_label: OAS
--- a/registries/_format/uint8.md
+++ b/registries/_format/uint8.md
@@ -2,7 +2,7 @@
 owner: MikeRalphson
 issue: 845
 description: unsigned 8-bit integer
-base_type: number
+base_type: [number, integer]
 layout: default
 source: https://spec.openapis.org/oas/latest.html#data-types
 source_label: OAS


ralfhandl (dismissed):

handrews (changes requested):
`integer` is not a type in the JSON data model.  This is [explicitly addressed in JSON Schema](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#section-4.2.1), and is why `integer` is never supposed to be present in `base_type`.

Both keyword and `format` applicability work with JSON types, not the `type` keyword values (`type: integer` is more like a format, except that it's actually reliable).

baywet:
@handrews I did come across this section of the specification, this other seems to contradict it though. (Link in my original post)

> String values MUST be one of the six primitive types ("null", "boolean", "object", "array", "number", or "string"), or "integer" which matches any number with a zero fractional part.

Not sure which one is correct, but it seems at least some descriptions in the wild are using this type. 

Also sf-integer in the registry here mentions this type.

handrews:
@baywet `sf-integer`shouldn't do that, I must have missed it when reviewing.

The sentence you quoted does not contradict anything- it expliclty states six primitive types, then lists the six, and _then_ says "or `integer` which matches any number with a zero fractional part", meaning that both `1` and `1.0` are integers.

The JSON RFC does not define integers as distinct from numbers, so JSON parsers do not necessarily make any distinction (in part because JavaScript does not either).

This was a significant point of discussion at one point in JSON Schema's history and I am very certain of what it means.

ralfhandl:
Before approving this PR I wondered whether `integer` is an allowed value for `type`, then I found in OAS 3.1.0 [Section 4.4 Data Types](https://spec.openapis.org/oas/v3.1.0.html#data-types) that

> Note that `integer` as a type is also supported and is defined as a JSON number without a fraction or exponent part.

and

> The formats defined by the OAS are:
> type | format | Comments
> -- | -- | --
> integer | int32 | signed 32 bits
> integer | int64 | signed 64 bits (a.k.a long)

So @baywet's changes to `int32` and `int64` are explicitly sanctioned, and the other three are consistent with the current specification text.

baywet:
> So @baywet's changes to int32 and int64 are explicitly sanctioned, and the other three are consistent with the current specification text.

I could revert the changes to uint etc to only include int32 and int64 to stop the bleeding here. But I guess if we did so, I'd have to leave a note. Plus what should we do with sf-integer then?

ralfhandl:
> I could revert the changes to uint etc

No, `int16`, `int8`, and `uint8` are consistent extrapolations of `int32` and `int64` and should follow the same rules: can be used with both types `number` and `integer`.

We could deprecate `integer` with 3.2.0 and remove it with Moonwalk 😄

baywet:
> We could deprecate integer with 3.2.0 and remove it with Moonwalk 

In that scenario, should we add a note in the registry for all the formats that mention the integer type?

mikekistler (dismissed):
Looks good. 👍 

I think it is pretty clear that `integer` is allowed/supported in OpenAPI, so these changes are appropriate. Further, I think we'd need to have compelling reasons to remove it in any future version.

handrews:
@ralfhandl 

> So @baywet's changes to int32 and int64 are explicitly sanctioned, and the other three are consistent with the current specification text.

This is not about OAS, this is about JSON Schema.  JSON Schema's `format` keyword _does not distinguish between integers and non-integer numbers_ when determining whether a given format applies to a given instance.  Putting `integer` in the format registry like this isn't going to change that, and misleads people to expect functionality that is not present.

handrews:
To try and make this a bit more clear, JSON Schema keywords and `format` values have a notion of applicability: to which JSON Schema _data model_ types does a given keyword or format apply?  If the keyword or format does _not_ apply, then it always produces a `true` (passing) validation result.

As I noted in a previous comment, the JSON Schema data model does _not_ include integers.  The `type` keyword is _not_ what defines the data model type.  An instance value has a data model type regardless of what keywords are present or absent.

If we were to treat integer as a data model type (which it is not), and we were to say that `format: uint8` only applies to instances of type integer, then the schema `format: uint8` would consider the instance `1.1` to be _valid_, because it would not be an integer and therefore the format would _not_ apply.  That is not what we want.  Which is fine because it's not possible anyway.

Consider `1.0`.  JSON Schema defines it as an integer, because it defines integers mathematically.  Some people expect `1.0` to be a non-integer because its textual representation contains a decimal point.  But this is not how JSON Schema works.  `1.0` is a place where trying to make a distinction between integers and non-integers is problematic because people expect their programming language's conventions for numeric literals to apply, which is, again, not how JSON Schema works.

baywet:
Thank you for the super detailed description of how things are supposed to work. I had no idea about the data model aspects and its subtilities.

I find it odd that JSON schema went with this opinionated way of dealing with numbers, which is different from JSON as far as I understand and loses information. When you have `1.0` as opposed to 1, it carries additional information that is the level of precision which is to the first decimal. 

In that context, it would make sense to me that an API producer would want to say this API only produces information with this or that level of precision. And consequently that somebody would want to validate the data it's receiving when compared with the spec.

But I understand this is not how JSON schema was designed to work and trying to make it work this way would be difficult because of that.

Now, to Ralph's earlier comment. It would make sense that we clean this up moving forward. 
I think this is a good candidate to have the table removed from 3.2.0. and I can update this pull request to remove any reference to the integer type for the time being and maybe we can even add a note or link to this issue/pull request for people who want to know why it has changed. What does everybody think about this solution?

ralfhandl:
Adjusted table in section "Data Types":
* #4031 
* #4032 
* #4035 

These PRs are part of the "property-field-keyword" cleanup

mikekistler:
> I can update this pull request to remove any reference to the integer type for the time being

I don't think that's a good idea. Integer is a type in OAS, and that might have some peculiarities, but there's lots of stuff in the spec that is peculiar.

ralfhandl:
I'm not quite sure what we are arguing about here.

My take on the current specification and format registry entries:
* [Spec](https://spec.openapis.org/oas/latest.html#data-types)
  > `integer` as a type is also supported and is defined as a JSON number [...]
* [Format `int32`](https://spec.openapis.org/registry/format/int32.html)
  > int32 - signed 32-bit integer
  > Base type: `number`.
  > The `int32` format represents a signed 32-bit integer, with the range −2,147,483,648 through 2,147,483,647.

Since `integer` implies `number`, I can combine `format: "int32"` with both `type: "number"` and `type: "integer"`.

An OpenAPI-aware payload validator that supports `format: "int32"` will produce the same validation result in both cases.

Do we need to explicitly list `integer` in the format's base type?
* No, because `number` implies that the format can also be used with `integer`.

So the argument seems to be:
* Does it harm to list both `number` and `integer` as a base type in the registry?

baywet:
> Does it harm to list both number and integer as a base type in the registry?

Which was my original thought process as well. To which I'm thinking "no it doesn't harm". But I'm ok if we say we need to clean it up instead. I just want us to be consistent.

handrews:
@ralfhandl @baywet It's objectively wrong according to the JSON Schema specification to list "integer" as a data model type for the purpose of keyword or format applicability.  Applicability of keywords and formats is _only_ determined by the six data model types:  number, string, boolean, null, object, and array.  Anything else is not in compliance with JSON Schema.

ralfhandl:
@handrews You are absolutely right from a pure JSON Schema perspective newer than [Draft 04](https://datatracker.ietf.org/doc/html/draft-zyp-json-schema-04#section-3.5).

OAS 3.0 and 3.1 (probably due to the removal of `integer` in JSON Schema Draft 05) explicitly add `integer` back as an allowed type in [Section 4.4 Data Types](https://spec.openapis.org/oas/v3.1.0.html#data-types), adding to the difference between the OAS [Schema Object](https://spec.openapis.org/oas/v3.1.0.html#schema-object) and pure JSON Schema.

I've opened #4038 to discuss and fix this.

mikekistler:
@handrews I am unclear what the issue is here. Is it:

- that "integer" is recognized as a valid value for "type" in OAS?
- that the OAS format registry defines values for "format", their meanings, and which "type" values they may be used with?
- that the OAS format registry associates some "format" values with the "integer" type value?

Or is it something else entirely?

handrews:
@ralfhandl @mikekistler I have tried to explain again in #4038.

I don't know how to be more clear about this.  The set of types that are _data model types_ and the set of values for the `type` keyword are not the same.  The OAS ***cannot*** change the set of data model types, and that is the set that JSON Schema uses to determine keyword and `format` value applicability.

We cannot change that.  It is not part of our spec.  It is part of JSON Schema.  It doesn't matter what we say, it's not going to change the meaning of JSON Schema.

baywet:
> The set of types that are data model types and the set of values for the type keyword are not the same

Why are they different? Is integer the only difference here? In which case, why was it introduced?

handrews:
@baywet it's literally in the text you quoted elsewhere:

> As an example, "integer" is a reasonable type for a vocabulary to define as a value for a keyword, but the data model makes no distinction between integers and other numbers.

This explicitly states that it is not part of the data model, it's just a convenience for a specific keyword.

mikekistler:
@handrews  I am not familiar with the concept of "data model types" in JSON Schema. Can you point to the part of the JSON Schema spec that describes this, so I can educate myself?

baywet:
> @baywet it's literally in the text you quoted elsewhere:
> 
> > As an example, "integer" is a reasonable type for a vocabulary to define as a value for a keyword, but the data model makes no distinction between integers and other numbers.
> 
> This explicitly states that it is not part of the data model, it's just a convenience for a specific keyword.

Allow me to rephrase my question. You probably have the most historical context among us here over why JSON schema has this integer type. 
Assuming this answers my question of "what's the Delta between the two sets", I'd like to understand why integer is a thing in JSON schema to begin with? Since it seems to be adding more confusion than value. It'd probably have been easier to match the two sets for everyone involved.

baywet:
> @handrews  I am not familiar with the concept of "data model types" in JSON Schema. Can you point to the part of the JSON Schema spec that describes this, so I can educate myself?

@mikekistler this is what is being referred to here I believe. https://json-schema.org/draft/2020-12/json-schema-core#name-instance-data-model

handrews:
@mikekistler @baywet is correct about the section on the data model.  See also [the actual JSON data model](https://www.rfc-editor.org/rfc/rfc8259.html#section-3).

@baywet 

> Assuming this answers my question of "what's the Delta between the two sets", I'd like to understand why integer is a thing in JSON schema to begin with?

Because `{"type": "number", "multipleOf": 1}` is really annoying to write when integers are so incredibly common as a data _constraint_.  Also, `multipleOf` in general has to be implemented with floating point, and no one wants to introduce that where it doesn't need to exist (yes, people could special case integer `multipleOf`, but do you want to rely on everyone doing that correctly?).

I can't tell you the full history because it goes way back to the original JSON Schema effort.  JSON Schema is on its (I think) 4th editorial team- I was part of the 3rd, which has some overlap with the 4th, but there was an almost complete turnover with a multi-year gap between the 2nd and 3rd.

But fundamentally, JSON Schema works with the JSON data model which does not have a distinct integer type.  That's part of JSON, and JSON Schema can't change JSON any more than OAS can change JSON Schema.

handrews:
@baywet 

> Since it seems to be adding more confusion than value. It'd probably have been easier to match the two sets for everyone involved.

You find it confusing because you are thinking about JSON Schema as a data definition system, which it is not.  It was not designed to be one.  `format` in particular was not even intended to be a mandatory feature.  The fact that people have built elaborate data definition and code generation systems on it is impressive, but very much a house of cards.  We should put effort into better solutions for Moonwalk, IMNSHO, instead of continuing to try

handrews:
@mikekistler also, apologies (to you and @baywet ) if I came across as expecting you to know things you did not- it got confusing with this conversation spread across several issues, and I found myself thinking I'd said things before and then couldn't find them.

This is a really non-intuitive corner of JSON Schema and the confusion you are both experiencing is entirely reasonable.

ralfhandl:
What we are going to do with this PR?

Seems that
* "integer" implies "number"
* "number" determines format applicability in JSON Schema's data model
* if OAS Schema Objects are transpiled into pure JSON Schema, a JSON Schema validator will do the right thing

Some formats in the OAS Format Registry already explicitly list "integer", and this PR adds this to the five that don't, and where "integer" makes sense.

So this PR increases the internal consistency of the OAS Format Registry's content, which I think is a good thing.

It could further increase consistency by
* adding "number" to format `double-int`
* use order "number, integer" also for format `sf-integer`

What do you think?

mikekistler:
I think we are getting close. I'm starting to understand @handrews concerns. This came from studying the section on 'Semantic validation with "format"' from the Draft 5 JSON Schema spec

https://json-schema.org/draft-04/draft-fge-json-schema-validation-00#rfc.section.7

And the key part is:
> A format attribute can generally only validate a given set of instance types. If the type of the instance to validate is not in this set, validation for this format attribute and instance SHOULD succeed.

where I think "instance type" means "the (JSON) type of the data in the instance".

Based on this, I see that an essential part of defining a new format is defining the set of "instance types" (i.e. JSON types) for which the format is applicable. 

So I think the crux of the issue is clearly defining what we mean by the "base type" column of the format registry.

If `base_type` means the "instance type" (as I think @handrews has taken it), then "integer" does not really belong.

I have always (prior to now) thought of `base_type` as meaning "the values of the 'type' keyword that may be used along with this format value".  This may _indirectly_ define the instance types to which the format is applicable, and I think this is the solution that @ralfhandl is proposing.

So I think we need to decide -- and then clearly document -- what `base_type` means in the format registry. That will determine whether or not we should include "integer" there.

ralfhandl:
@mikekistler Thanks for the clarification!

Yes, I want to define `base_type` as being the OAS value of the `type` keyword this format can be used with in OpenAPI descriptions.

> Explicitly state in the [Format Registry](https://ralfhandl.github.io/OpenAPI-Specification/registry/format/) that column "Type" uses the extended OAS definition.

mikekistler:
And then we'll need to explain how that implies the instance type applicability.

ralfhandl:
How about doing that in section [Data Types](https://spec.openapis.org/oas/latest.html#data-types) where we define `int32` and `int64` and claim they can be used with `type: "integer"`.

We could extend
* #4042

baywet:
@mikekistler thank you so much for your explanation here! Yes I went through this whole conversation assuming that in the registry "base_type" === "type" keyword.

Ok, so I think we've collectively agreed that type keyword !== instance type. What I'm still unclear about is how do you define the schema instance type? I've always assumed it was through the type keyword.

mikekistler:
> What I'm still unclear about is how do you define the schema instance type? 

I think "schema instance type" is mixing two ideas we need to keep separate. The "schema type" is the `type` defined in the schema. The "instance type" is the data type of data in a JSON "instance" -- object, array, or a primitive type. I think this type is always well-defined: `true` or `false` are boolean, numbers not surrounded by quotes are number, anything enclosed in quotes is string, etc.

baywet:
Would it be fair to say that "you can't _define_ the instance type"? i.e. it is "automatically" defined by the data on the wire?

handrews:
@ralfhandl 

> Yes, I want to define base_type as being the OAS value of the type keyword this format can be used with in OpenAPI descriptions.

That's not how JSON Schema works.  Feel free to go request a change from JSON Schema.  What you're trying to do here is equivalent to someone else redefining the OAS for their own purposes.

@baywet 

> Would it be fair to say that "you can't define the instance type"? i.e. it is "automatically" defined by the data on the wire?

Yes.

@mikekistler 

> I think "schema instance type" is mixing two ideas we need to keep separate. The "schema type" is the type defined in the schema. The "instance type" is the data type of data in a JSON "instance" -- object, array, or a primitive type. I think this type is always well-defined: true or false are boolean, numbers not surrounded by quotes are number, anything enclosed in quotes is string, etc.

There is no such thing as a "schema type".  The `type` and `format` keywords are constraints on the set of accepted instances.  JSON Schema is [just a constraint system](https://modern-json-schema.com/json-schema-is-a-constraint-system).  You don't need to have `type` or `format`.  You can have `format` without `type` or vice versa.  JSON Schema is not a data definition system.

darrelmiller:
I'm going to risk adding my 2c here.  The primary reason for doing what we did in OpenAPI 3.1 with JSON Schema was to enable that JSON Schemas in the OpenAPI document be processed by JSON Schema tooling.  If JSON Schema tooling will not validate JSON instances correctly if we couple formats to "integer" then we are defeating the purpose.  

We can't just say of OpenAPI changed this thing.  The whole point was to be compatible with existing JSON Schema tooling

baywet:
Thanks. I think I get it now.
So we could:
1. clean up integer from all the format registry, since it shouldn't work in theory (some tooling might have implemented it though)
2. Remove that table from the spec with OAS 4
3. Add a note in OAS 3.0/3.1 patch versions next to the table that formats should not be used in combination with type integer. It was a mistake.

Would that work for everyone?

handrews:
> Add a note in OAS 3.0/3.1 patch versions next to the table that formats should not be used in combination with type integer. It was a mistake.

This doesn't prevent formats from being used with `type: integer`.  It's about whether or not a given format is evaluated as a constraint based on the data model type, or whether it's treated as irrelevant (always passes).  Which has nothing to do with the `type` keyword at all.  It's entirely dependent on the actual instance JSON data type.

baywet:
I don't think we can prevent the use of integer type + format in OAS 3.0/3.1 at this point since this would be a "breaking change" and would mandate a major bump? Or can we?

handrews:
@baywet I don't understand where this "prevent the use of integer type + format" thing is coming from.

There are no constraints on whether or how you use the `type` and `format` keyword together.  That is the sort of thing OAS 3.1 got away from.  For example, in older OAS versions, you were not allowed to use the `items` keyword without `type: array`, which was not a restriction present in JSON Schema.  So in OAS 3.1, you can use `items` whether `type` is `array`, or something else, or not present at all.

What we are talking about is whether a particular format is _evaluated_ based on the _instance type_.  The `type` keyword is completely irrelevant.

handrews:
The current Data Types section is just a misleading error, which I somehow never previously noticed (or maybe I did and just filed it under "there are bigger problems to deal with right now". I honestly don't recall).

Let's walk through an example.  Currently (3.1.0) it says that `format: int32` is used with `integer`.  It does not include `number`.  So let's _pretend_ that `integer` is a data model type (which it is not).  Let's look at the behavior of `{"format": "int32"} (note the lack of `type`) with different instances, assuming it was restricted to the (fictional) integer data model type:

instance | valid? | notes
-------- | ------ | ------
1  | yes | integer, fits within 32 bits
1.0 | yes | mathematical integer, fits within 32 bits
5,000,000,000 | no | integer but too big for 32 bits
3.14 | yes | non-integer, so the `format` is not evaluated and instead automatically passes
"foo" | yes | non-integer, so the `format` is not evaluated and instead automatically passes

That second-to-last entry is where the problem is shown.  Of course, if you use `{"type": "integer", "format": "int32"}` then those last two entries flip from valid to invalid, and most people would do that.  But JSON Schema does not require `type`, and there are various not-entirely-intuitive-but-valid use cases where this happens.  Is it likely to happen the way most people use OpenAPI?  No.  But it is allowed by JSON Schema, and _some_ people might rely on it.  And as @darrelmiller notes, we don't get to change how JSON Schema works.

mikekistler:
In retrospect, I think we should have defined "format: integer" to be "a JSON number without a fraction or exponent part" and then defined "type: integer" to be equivalent to "type: number, format: integer".

baywet:
@handrews 

>  "prevent the use of integer type + format" thing is coming from.

From your previous reply

> This doesn't prevent formats from being used with type: integer

I guess my intent now that I understand why formats can't/shouldn't be used with type integer is two folds:
- prevent people from doing this it all together in OAS4. And I'm confident this is something you'll carry to that group given how much we've talked about it here.
- suggest people to avoid doing it in OAS3.0/3.1 since it has serious edge cases you've outlined with your examples they might not have anticipated.

Does that make sense? Is it a recommendation that should be part of the patch releases?

handrews:
I still don't understand this "can't/shouldn't be used with integer types."

The point here is not to change how or when any particular `format` can or should be used.  It is to correctly convey how standard JSON Schema tooling determines whether to pay attention to a `format`.

There is no usage of `format` that I can think of that was previously allowed that would no longer be allowed.  The behavior of JSON Schema never changed.  We just had the wrong text.  That's all this is.  We had the wrong text, and at least from OAS 3.1 onwards it simply had no effect because we stated definitively that we adopted JSON Schema draft 2020-12 in total.

For 3.1.1, at least, we're just fixing descriptive text.

For 3.0.4, I'd argue for aligning with 3.1.1 to minimize confusion (as much as possible).  It's not clear to me that that table in the Data Types section actually imposes any requirements.  There is no wording to the effect of "a given format MUST NOT be used if `type` is absent or does not match this table."

baywet:
> can't/shouldn't be used with integer types

Didn't we come to the conclusion that people should NOT use the format keyword in combination with the type integer as it WILL NOT work as expected at validation time? i.e. some things like decimal numbers will validate when most people would not expect it to.

handrews:
@baywet in an effort to make this more clear, I have posted PR #4045.

ralfhandl:
> Currently (3.1.0) it says that `format: int32` is used with `integer`. It does not include `number`.

Well, I thought that applicability to `type: number` was implied because
> `integer` as a type is also supported and is defined as a JSON _number_

As a consequence much of the discussion above seemed beside the point and was rather confusing.

@handrews Thanks for making this explicit in #4045

ralfhandl:
Once #4045 is approved and merged, we REALLY SHOULD adjust the header of the table in the [format registry](https://spec.openapis.org/registry/format/) and the wording in the [format-entry include](https://github.com/OAI/OpenAPI-Specification/blob/gh-pages/_includes/format-entry.md) to exactly match the improved OAS text.

baywet:
Alright @handrews @ralfhandl @mikekistler @darrelmiller thank you so much for the discussion here! I learnt a lot. I have rebased this PR and removed the use of integer across the board. Did we also want to add a not of why integer is not in use to clarify any potential future confusion?

ralfhandl:
> Did we also want to add a not of why integer is not in use to clarify any potential future confusion?

Definitely, see [my comment](https://github.com/OAI/OpenAPI-Specification/pull/4015#issuecomment-2301375695):
* Rename column "Type" in [format.md](https://github.com/OAI/OpenAPI-Specification/blob/fb7ab6b48c9e335e8ce6fd0991fcf5bd49c2064c/registry/format.md?plain=1#L25) to whatever #4045 ends up with, "Data Model Type", "JSON Data Type", ...
* Use the same terminology in [format-entry.md](https://github.com/OAI/OpenAPI-Specification/blob/fb7ab6b48c9e335e8ce6fd0991fcf5bd49c2064c/_includes/format-entry.md?plain=1#L5)

handrews (approved):
Thanks!

ralfhandl (approved):

ralfhandl:
Adjusted the other markdown file.

registry/format.md
```diff
@@ -15,14 +15,13 @@ If tools choose to implement any format present in this registry, they SHOULD im
 
 The registry SHOULD NOT contain two entries that have the same meaning, unless all but one have been deprecated.
 
-
 ## Contributing
 
 Please raise a [Pull-Request](https://github.com/OAI/OpenAPI-Specification/pulls) or [Issue](https://github.com/OAI/OpenAPI-Specification/issues) to contribute or discuss a registry value.
 
 ## Values
```
mikekistler, line 23:
```suggestion

For the purpose of [JSON Schema validation](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#section-7.1), each format should specify the set of instance types (primitive JSON types) for which it applies. In this registry, these types are shown in the "Instance Type" column.

```

registry/format.md
```diff
@@ -15,14 +15,13 @@ If tools choose to implement any format present in this registry, they SHOULD im
...
 
-|Value|Description|Type|Source|Deprecated|
+|Value|Description|JSON Data Type|Source|Deprecated|
```
mikekistler, line 24:
```suggestion
|Value|Description|Instance Type|Source|Deprecated|
```

baywet:
LGTM! (ralf's latest changes)
Just as a reminder, I do not have merge permissions, so I'll need somebody else to merge for me :)

mikekistler (changes requested):
I think we need a short explanation of the purpose of the Type column, and I think "Instance Type" is a slightly better term.

_includes/format-entry.md
```diff
@@ -2,7 +2,7 @@
 
 ## {{ page.slug }} - {{ page.description }}
 
-Base type: `{{ page.base_type | join:', ' }}`.
+JSON Data Type: `{{ page.base_type | join:', ' }}`.
```
baywet, line 5:
```suggestion
Instance Type: `{{ page.base_type | join:', ' }}`.
```

handrews:
As noted in #4045, I don't think most people know what an "instance" is, so "JSON Data Type" would be better.

baywet:
alright, I reverted back to JSON data type everyone. Final review and merge?

registry/format.md
```diff
@@ -21,9 +21,9 @@ Please raise a [Pull-Request](https://github.com/OAI/OpenAPI-Specification/pulls
 
 ## Values
 
-For the purpose of [JSON Schema validation](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#section-7.1), each format should specify the set of instance types (primitive JSON types) for which it applies. In this registry, these types are shown in the "Instance Type" column.
+For the purpose of [JSON Schema validation](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#section-7.1), each format should specify the set of JSON data types for which it applies. In this registry, these types are shown in the "JSON data type" column.
 
-|Value|Description|Instance Type|Source|Deprecated|
+|Value|Description|JSON data Type|Source|Deprecated|
```
ralfhandl, line 26:
```suggestion
|Value|Description|JSON Data Type|Source|Deprecated|
```
Sorry for being nitpicky

_includes/format-entry.md
```diff
@@ -2,7 +2,7 @@
 
 ## {{ page.slug }} - {{ page.description }}
 
-Instance Type: `{{ page.base_type | join:', ' }}`.
+JSON data type: `{{ page.base_type | join:', ' }}`.
```
ralfhandl, line 5:
```suggestion
JSON Data Type: `{{ page.base_type | join:', ' }}`.
```

handrews (approved):

registry/format.md
baywet: Why do we need a capital D in your view?

ralfhandl: Because its a table header, and we usually Uppercase Everything in a header cell.

Which @handrews did in #4045.

ralfhandl (approved):

ralfhandl:
@mikekistler Please check, and if you approve, please merge. Thanks!

baywet:
Thanks everyone! now we just need @mikekistler to approve.

handrews (approved):

ralfhandl:
@mikekistler Could you please re-review?

The wording is now in line with #4045

mikekistler (approved):
Looks good! 👍

baywet:
Thanks! We're ready to merge. Who has the permissions to do so?

ralfhandl:
> We're ready to merge. Who has the permissions to do so?

Me, already merged.

Thanks for all the work you put into this!

Issue #2660, opened 2021-07-29
Usage of `field` vs `property`


char0n:
I was wondering if authors of the spec would be open to make usage of `field` and `property` somehow consistent. When we have specification objects we refer to it's properties as fields, either fixed one, patterned one or mapped one. Then in field descriptions we refer to those fields as properties. 

Is there a semantic difference between field and property wording or some additional distinction between `field` vs `property`?

darrelmiller:
From the TDC discussion, we should make efforts to remove unnecessary ambiguity when using words such as keyword, property, attribute and field. 
Can we clarify when each of those terms should be used?
Can we have a generic meta model for the OpenAPI document that connects terms to externally referenced terms?

Can we limit terms to just object, property, map and array?  /cc @webron Can we get your insight on why `field` was used?

hudlow:
I think there's a case to be made that we should reserve "property" for a subschema an object schema: a subschema that defines valid members for a conforming object (member as defined by RFC 8259). 

To illustrate this distinction, we could say the `Pet` schema contains a `id` property even though we know the literal `Pet` schema object in an OpenAPI definition file won't have a member named `id`. Rather it will have a member named `properties` which has a member named `id`.

"Member" seems a little esoteric to me, but if we defined "field" to also refer to a literal key-value pair in a JSON object, then we could say:

* "The `Pet` schema contains an `id` property."
* "The `Pet` schema object contains a `properties` field."
* "The pet object contains an `id` field in conformance with the `Pet` schema."
* "The `id` field within the `properties` field of the `Pet` schema object defines an `id` property for the `Pet` schema."
* "Because the `Pet` schema defines a required `id` property, objects conforming to the `Pet` schema must contain an `id` field."

But there are some confusing subtleties to this for sure. One of them would be that the `required` array would really refer to required _fields_. (In other words, I'm not sure "required properties" makes sense if properties are schemas for fields.)

swaldron58:
[...]

PR #4020, opened 2024-08-15
3.0.4: replace "property" with "field" or "keyword" where appropriate

ralfhandl:
Part of
* #2660

Rules:
* JSON Schema keywords -> "keyword"
* OpenAPI fixed fields -> "field"
* property of a "plain" JSON object that is not an OpenAPI-defined Foo Object -> "property"


versions/3.0.4.md
```diff
@@ -2711,7 +2711,7 @@ There are two ways to define the value of a discriminator for an inheriting inst
 
 ###### XML Modeling
 
-The [xml](#schema-xml) property allows extra definitions when translating the JSON definition to XML.
+The [xml](#schema-xml) field allows extra definitions when translating the JSON definition to XML.
```
mikekistler, line 2714: I think there is some ambiguity in our rules here. In later versions of the spec (not this one since it is based on Draft 5 which does not have vocabularies), the `xml` field/keyword/thingy is defined in a JSON Schema vocabulary. When that is the case, do we consider it part of JSON Schema and call it a "keyword" or an OpenAPI fixed field and call it a "field"? "discriminator" is probably another instance of this.

ralfhandl: Checked:
* https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00: only refers to "xml-names"
* https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00 does not mention xml at all

So it is an OpenAPI field.

mikekistler: My question is more a philosophical one. If I write a JSON Schema vocabulary that adds keywords/fields/thingys, then are these called "keywords", following the JSON Schema terminology, or can I choose to call them "fields" or anything else I might choose to call them? Even if I _could_ call them _anything_, is that really a good thing to do?

ralfhandl: I personally define "terms" in a vocabulary, not "fields" or "keywords" 😎

ralfhandl: JSON Schema vocabularies define "extension keywords" according to @handrews

handrews (approved):

mikekistler (approved):
Looks good! 👍

PR #4022, opened 2024-08-15
EDITORS.md: list of current and former specification editors

ralfhandl:
- [x] Create EDITORS.md
- [x] Use it in BUILD.sh for 3.0.4 and 3.1.1

### DIFF
--- a/EDITORS.md
+++ b/EDITORS.md
@@ -1,16 +1,14 @@
-# OpenAPI Initiative Technical Steering Committee Members
-
+# OpenAPI Specification Editors
 
 ## Active
 * Darrel Miller [@darrelmiller](https://github.com/darrelmiller)
+* Henry Andrews [@handrews](https://github.com/handrews)
 * Jeremy Whitlock [@whitlockjc](https://github.com/whitlockjc)
-* Marsh Gardiner [@earth2marsh](https://github.com/earth2marsh)
-* Ron Ratovsky [@webron](https://github.com/webron)
-
-## Provisional
 * Lorna Mitchell [@lornajane](https://github.com/lornajane)
-* Mike Kistler [@mikekistler](https://github.com/mikekistler)
+* Marsh Gardiner [@earth2marsh](https://github.com/earth2marsh)
 * Miguel Quintero [@miqui](https://github.com/miqui)
+* Mike Kistler [@mikekistler](https://github.com/mikekistler)
+* Ron Ratovsky [@webron](https://github.com/webron)
 * Ralf Handl [@ralfhandl](https://github.com/ralfhandl)
 
 ## Emeritus
--- a/EDITORS.md
+++ b/EDITORS.md
@@ -8,8 +8,8 @@
 * Marsh Gardiner [@earth2marsh](https://github.com/earth2marsh)
 * Miguel Quintero [@miqui](https://github.com/miqui)
 * Mike Kistler [@mikekistler](https://github.com/mikekistler)
-* Ron Ratovsky [@webron](https://github.com/webron)
 * Ralf Handl [@ralfhandl](https://github.com/ralfhandl)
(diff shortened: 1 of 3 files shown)


MAINTAINERS.md
```diff
@@ -1,10 +1,13 @@
+# OpenAPI Specification Editors
```
earth2marsh, line 1: This list has served as the record of TSC membership, though I think that might be made explicit? By adding this title, it also changes the framing. For example, while TSC members are tasked with evolving the spec, that includes managing issues, encouraging community, etc.

MAINTAINERS.md
```diff
@@ -1,10 +1,13 @@
...
+
 ## Active
 * Darrel Miller [@darrelmiller](https://github.com/darrelmiller)
 * Jeremy Whitlock [@whitlockjc](https://github.com/whitlockjc)
 * Marsh Gardiner [@earth2marsh](https://github.com/earth2marsh)
 * Ron Ratovsky [@webron](https://github.com/webron)
 
-## Provisional
```
earth2marsh, line 7: Provisional does refer back to the governance processes. What do we lose by dropping it here?

MAINTAINERS.md
```diff
@@ -1,10 +1,13 @@
...
+## Trainee
```
earth2marsh, line 9: What makes Trainee better here? I'm not against it, just trying to understand why this word in particular.

earth2marsh (commented):
Historically this has been the TSC current and former membership list. Perhaps we should make that explicit? Consider whether "contributors.md" is a more appropriate place to recognize the impact that people like Henry make and/or carve out a different section to explain non-TSC maintainer roles?

ralfhandl:
We currently have
* MAINTAINERS.md: both TSC membership and OAS authors
* CONTRIBUTORS.md: people with more than 25 pull requests

Now TSC members and OAS authors are becoming non-identical groups of people and we may need three documents.

How about adding an EDITORS.md and use that (from now on) in the build script?

MAINTAINERS.md
ralfhandl: Added title to express original purpose of TSC membership

MAINTAINERS.md
ralfhandl: Reverted, added separate EDITORS.md

EDITORS.md
```diff
@@ -0,0 +1,18 @@
...
+## Active
+* Darrel Miller [@darrelmiller](https://github.com/darrelmiller)
+* Henry Andrews [@handrews](https://github.com/handrews)
+* Jeremy Whitlock [@whitlockjc](https://github.com/whitlockjc)
+* Lorna Mitchell [@lornajane](https://github.com/lornajane)
+* Marsh Gardiner [@earth2marsh](https://github.com/earth2marsh)
+* Miguel Quintero [@miqui](https://github.com/miqui)
+* Mike Kistler [@mikekistler](https://github.com/mikekistler)
+* Ron Ratovsky [@webron](https://github.com/webron)
+* Ralf Handl [@ralfhandl](https://github.com/ralfhandl)
```
earth2marsh, line 12: Ralf should come before Ron to preserve the alphabetical order! :)

earth2marsh (dismissed):
Thanks for the changes--this looks good to me (other than the minor suggested fix). Ideally we can get another +1, but if not, let's keep moving!

miqui (approved):

lornajane (approved):

earth2marsh:
Regarding the discussion in today's call, @ralfhandl I believe we can rename EDITORS.md > TSC.md without issue, as the openapis.org governance menu links to the MAINTAINERS.md (and fun fact, to the old branch name, not `main`!). So once we do the rename, we can work to update the website asynchronously. 

If that makes sense to you, LMK, and I'll make an issue and submit that change!

PR #3991, opened 2024-08-01
Markdown updates for 3.1.1

lornajane:
Replay of the steps from #3932 to update the markdown formatting and links for 3.1.1

(mysterious python script needed a few additions, included in a comment)

Issue #3596, opened 2024-02-20
Add whitespace/formatting automation to the build process


lornajane:
Over time the formatting on our key documents has become inconsistent in terms of spacing and quotes, etc.

Add a tool to our build process for pull requests and the main branch to that that the official spec documents conform to our standards (as established by our current conventions).

Bellangelo:
I have run some tests here and I have succeeded to format most of the cases using Prettier for formatitng and Markdownlint for linting. One thing that Prettier cannot catch is the optional quotes in the YAML properties and strings. There is an open [issue](https://github.com/prettier/prettier/issues/5350) for this case but couldn't find any solution to this.

ralfhandl:
@lornajane Can this issue be closed?

The [validate-markdown](https://github.com/OAI/OpenAPI-Specification/blob/dev/.github/workflows/validate-markdown.yaml) workflow now lints the published and work-in-progress spec documemts.

lornajane:
Yes, I think so

PR #4025, opened 2024-08-15
Markdown updates for 3.2.0

ralfhandl:
* 3.2.0 port of
  * #3991
* Run markdownlint after mdv on 3.1.1.md only because the older files won't pass
  * Part of #3596


mikekistler (approved):
Looks good! 👍

handrews (changes requested):
I'm confused as to what happens in which PR, so I'm just going to say the same thing on this one:


@ralfhandl can you re-do this with the tables not expanded since we agreed on that for the other branches? That will keep the change history so much less polluted on this branch.

Ideally I'd prefer not collapsing the spaces after periods, but I'm not going to my heels in on it (but it does create more noise and makes it harder to find sentence changes in a fixed-width editor).

ralfhandl:
The "table condenser" is currently only in #4040. @mikekistler has approved it half a dozen times already, and I'd happily merge it and port it to 3.1.1 and 3.2.0 with a second approval.

ralfhandl:
This is the first part of several PRs to sync 3.2.0 with 3.1.1
* #4025 - remove TOC, prettier, markdownlint
* #4026 - stylistic harmonization
* #4035 - property, field, keyword cleanup

We can merge them one by one, thus minimizing the differences, or just merge the last one with the cumulative changes and close the others.

We'll need an additional diff to 3.1.1 to add the changes we did since the last of these PRs was created.

versions/3.2.0.md
```diff
@@ -1183,76 +1124,76 @@ There are four possible parameter locations specified by the `in` field:
...
 
-See [Appendix E](#percentEncodingAndFormMediaTypes) for a discussion of percent-encoding, including when delimiters need to be percent-encoded and options for handling collisions with percent-encoded data.
+See [Appendix E](#appendix-e-percent-encoding-and-form-media-types) for a discussion of percent-encoding, including when delimiters need to be percent-encoded and options for handling collisions with percent-encoded data.
 
 ##### Style Examples
 
 Assume a parameter named `color` has one of the following values:
 
-```
+```js
```
handrews, line 1194: This isn't exactly accurate but looks nicer than plain text so OK :-)

handrews (approved):

ralfhandl:
Merging as @mikekistler has approved before table condensing.

PR #4030, opened 2024-08-16
gh-pages: fix schema links

ralfhandl:
Make links also work if site is deployed with non-empty base URL.

Has no visible effect now as it currently is deployed to the domain root.


rvedotrc (approved):
Looks believable, though I'm unfamiliar with the markup being used here (for example, to check whether or not this introduces any encoding issues).

handrews (approved):

PR #4033, opened 2024-08-16
build script: example links point to HTML wrapper on learn site

ralfhandl:
- example links point to HTML wrapper on learn site
- `npm test` (and thus the `schema-test` workflow) will complain if test coverage drops
- added comment to `md2html.js` explaining that the anchor tweak is needed for older spec versions

Depends on
* https://github.com/OAI/learn.openapis.org/pull/105


lornajane (approved):
Nice addition!

PR #4027, opened 2024-08-15
Add markdown style guide to CONTRIBUTING

mikekistler:
This PR adds a Markdown style guide to the CONTRIBUTING.md. I've lifted the bullets from @ralfhandl 's PRs on style fixup and added the .markdownlint.yaml file from @lornajane 's PRs for markdown updates.

There may be other things to add but I think this is a pretty good start and since we are already acting on these I thought we should make sure we have consensus and then capture that in the docs.

PR #4040, opened 2024-08-19
3.0.4: tables cleanup & markdownlint

ralfhandl:
- [x] Github Workflow: markdownlint after mdv on 3.0.4.md only (because the older files won't pass)
  - this ensures new pushes comply with our linter rules in the `main` branch
- [x] add npm script `format-markdown` to conveniently reformat markdown after changes
- [x] add .markdownlint.yaml because the npm script needs it
- [x] "compact" format for tables
- [x] extract formatting into bash script to avoid portability issues - tested on Windows and MacOS

Part of
* #3596


versions/.markdownlint.yaml
```diff
@@ -0,0 +1,16 @@
...
+        indent: 2
+
+      MD012: false # allow blank lines
+
+      MD013:
+        line_length: 800
+        tables: false
+
+      MD024: false # duplicate headings
+      MD033: false # inline HTML
```
mikekistler, line 16: I think we can remove, reduce leading whitespace here.
```suggestion
# Unordered list symbol
MD004:
    style: asterisk

# Unordered list indentation
MD007:
    indent: 2

MD012: false # allow blank lines

MD013:
    line_length: 800
    tables: false

MD024: false # duplicate headings
MD033: false # inline HTML
```

mikekistler (approved):
Looks good. 👍 

Left a comment to remove some unnecessary leading white space.

mikekistler (dismissed):
Looks good! 👍

ralfhandl:
Update after #4027 is merged, check out .markdownlint.yaml from `main` and use it for validation.

mikekistler (dismissed):
Looks good! 👍

lornajane:
I expected to see the action running in this pull request, should it?

ralfhandl:
Probably the merge conflict prevents the script from running.

ralfhandl:
Now it runs and is green 😄

lornajane (dismissed):

miqui (approved):

ralfhandl:
@handrews Added `sed` to make tables more compact:
- at most one space after/before a pipe character
- at most four hyphens in a row

package.json
```diff
@@ -27,6 +27,6 @@
     "API"
   ],
   "scripts": {
-    "format-markdown": "npx prettier --write --single-quote versions/3.0.4.md && npx --yes markdownlint-cli --fix --config .markdownlint.yaml versions/3.0.4.md"
+    "format-markdown": "npx prettier --write --single-quote versions/3.0.4.md && sed -i '' -E -e 's/  +\\|/ |/g' -e 's/\\|  +/| /g' -e 's/-----+/----/g' versions/3.0.4.md && npx --yes markdownlint-cli --fix --config .markdownlint.yaml versions/3.0.4.md"
```
mikekistler, line 30: I looked for a prettier configuration option to turn off formatting of tables but could not find one. I guess this is the next best option.

mikekistler (approved):
Looks good. 👍

package.json
handrews: @mikekistler thanks for checking- that's really disappointing :-(

I was going to ask what was wrong with how we were doing tables before, because this is still annoying in that it collapses the tables that _do_ fit within 80 columns and _were_ easier to read with spaces.

This overall does not seem like an improvement for markdown.  I guess we're using it to get it to handle the embedded JSON/YAML?  Meh.

handrews:
I can't say I'm at all happy with Prettier's markdown behavior, but this is better than what it was doing to tables before.  This double-change of every single table line is going to be such a pain for searching change history :-(

ralfhandl:
> I can't say I'm at all happy with Prettier's markdown behavior

Me neither, but it does format the code blocks correctly.

> this is better than what it was doing to tables before

Thanks.

mikekistler (dismissed):
Looks good. 👍 

I was concerned that the prettier extension in VSCode (my preferred editing environment) might reflow the tables if I made any change to the file, but I tested this and found that not to be the case.

ralfhandl:
> I was concerned that the prettier extension in VSCode (my preferred editing environment) might reflow the tables if I made any change to the file

My VS Code does if I explicitly format the file with the prettier extension enabled, or if I have "format on save" turned on and save the file.

Which is why I have both the extension disabled for this workspace and "format on save" turned off in the workspace settings 😎

handrews:
This is why I use vi 😛 

(ok technically vim, but that's my only concession to modernity!)

ralfhandl:
vi was my preferred tool for editing the makefile for Emacs 😎

lornajane (approved):

ralfhandl:
Merging as @mikekistler has already approved half a dozen times 😎

PR #4041, opened 2024-08-19
markdownlint for 3.1.1.md

ralfhandl:
- [x] Github Workflow: markdownlint after mdv on 3.1.1.md only (because the older files won't pass)
  - this ensures new pushes comply with our linter rules in the `main` branch
- [x] add npm script `format-markdown` to conveniently reformat markdown after changes
- [x] add .markdownlint.yaml because the npm script needs it
- [x] did _not_ run the script on 3.1.1.md to not mess up other open PRs

Part of
* #3596


mikekistler (dismissed):
Looks good! 👍

ralfhandl:
Update after #4027 is merged, check out .markdownlint.yaml from `main` and use it for validation.

.markdownlint.yaml
```diff
@@ -0,0 +1,16 @@
+# Unordered list symbol
```
mikekistler, line 1: Do you need this file if you are going to pull the .markdownlint.yaml from main? Or we could just commit the same version of .markdownlint.yaml in the root directory in all the branches -- that might be simpler.

ralfhandl: `validate-markdown.yaml` pulls `package.json` and `package-lock.json` from `main`, most likely to make sure that the same version of mdv is used across all branches, so it makes sense to also use the same linter ruleset across all branches.

The "branch-local" copy of .markdownlint.yaml allows formatting the current spec markdown locally.

mikekistler (approved):
Looks good. 👍

miqui (approved):

Issue #4038, opened 2024-08-19
Explicitly mention `type: "integer"` in definition of Schema Object


ralfhandl:
OAS 3.0 and 3.1 explicitly define `"integer"` as an allowed value of the `type` keyword in section [Data Types](https://spec.openapis.org/oas/v3.1.0.html#data-types).

This difference between OAS and pure JSON Schema is not mentioned in section [Schema Object](https://spec.openapis.org/oas/v3.1.0.html#schema-object).

### Proposal

Explicitly mention this significant difference and define OAS `type: "integer"` as a shortcut for JSON Schema `type: "number", multipleOf: 1`.

Explicitly state in the [Format Registry](https://ralfhandl.github.io/OpenAPI-Specification/registry/format/) that column "Type" uses the extended OAS definition.

handrews:
This is not a difference.  `integer` is an allowed value for the `type` keyword in all relevant versions of JSON Schema.

However, integers are ***not*** part of the JSON data model, and therefore it is not something that can control the applicability of JSON Schema keywords of `format` values.

These are two separate things, as the JSON Schema specification explicitly states.

We ***cannot*** change the applicability criteria for the format registry, because that is defined by JSON Schema and not the OAS.

baywet:
> String values MUST be one of the six primitive types ("null", "boolean", "object", "array", "number", or "string"), or "integer" which matches any number with a zero fractional part

[source](https://json-schema.org/draft/2020-12/json-schema-validation#section-6.1.1-2)

> Note that JSON Schema vocabularies are free to define their own extended type system. This should not be confused with the core data model types defined here. As an example, "integer" is a reasonable type for a vocabulary to define as a value for a keyword, but the data model makes no distinction between integers and other numbers.

[source](https://json-schema.org/draft/2020-12/json-schema-core#section-4.2.1-6)

> Note that the "type" keyword in this specification defines an "integer" type which is not part of the data model. Therefore a format attribute can be limited to numbers, but not specifically to integers. However, a numeric format can be used alongside the "type" keyword with a value of "integer", or could be explicitly defined to always pass if the number is not an integer, which produces essentially the same behavior as only applying to integers

[source](https://json-schema.org/draft/2020-12/json-schema-validation#section-7.1-2)

@handrews I know you've already tried to explain this multiple times. Thank you for you patience here.
Based on those, it seems that using format alongside with integer is explicitly permitted/defined in the specification.
[...]

PR #4042, opened 2024-08-19
Define `type: "integer"`

ralfhandl:
Part of 
* #4038

### DIFF
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -173,7 +173,7 @@ Note that no aspect of implicit connection resolution changes how [URLs are reso
 ### Data Types
 
 Data types in the OAS are based on the types supported by the [JSON Schema Specification Wright Draft 00](https://tools.ietf.org/html/draft-wright-json-schema-00#section-4.2).
-Note that `integer` as a type is also supported and is defined as a JSON number without a fraction or exponent part.
+Note that `integer` as a type is also supported and is defined as a JSON number that is a multiple of 1.
 `null` is not supported as a type (see [`nullable`](#schema-nullable) for an alternative solution).
 Models are defined using the [Schema Object](#schema-object), which is an extended subset of JSON Schema Specification Wright Draft 00.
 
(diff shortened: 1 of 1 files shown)


handrews (changes requested):
Why are we defining integer instead of leaving it to JSON Schema, which is very precise on this matter?

ralfhandl:
> Why are we defining integer instead of leaving it to JSON Schema, which is very precise on this matter?

This is a good question, although it comes a little too late. OAS 3.0.0 to 3.1.0 all explicitly define "integer" as an allowed value of the type keyword in section [Data Types](https://spec.openapis.org/oas/v3.1.0.html#data-types):

> Note that integer as a type is also supported and is defined as a JSON number without a fraction or exponent part. Models are defined using the [Schema Object](https://spec.openapis.org/oas/v3.1.0.html#schema-object), which is a superset of JSON Schema Specification Draft 2020-12.

This PR only tries to make the existing definition more precise and mention this extension in the Schema Objects section in addition to the Data Types section.

We cannot remove this feature in a minor version, let alone a patch version.

handrews:
What OpenAPI does with "integer" and formats can't change JSON Schema.  It doesn't matter if it's been that way in the spec and not been noticed or not.

handrews:
I think the correct thing to do here is to note that JSON Schema only scopes formats to its data model types (number, string, boolean, null, object, array), and therefore the text of the OAS has been corrected to use "number" rather than "integer".  We can add wording to the effect that this has nothing to do with the `type` keyword if we want.

Since this is a very tricky area, I would be happy to write up a PR that aligns with JSON Schema.

versions/3.0.4.md
```diff
@@ -173,7 +173,7 @@ Note that no aspect of implicit connection resolution changes how [URLs are reso
 ### Data Types
 
 Data types in the OAS are based on the types supported by the [JSON Schema Specification Wright Draft 00](https://tools.ietf.org/html/draft-wright-json-schema-00#section-4.2).
-Note that `integer` as a type is also supported and is defined as a JSON number that is a multiple of 1.
+Note that `integer` as a type is also supported and is defined as a JSON number without a fraction or exponent part.
```
handrews, line 176: I would prefer to backport the language about mathematical integers from #4045

handrews: Scratch the previous comment- I did not realize this was in the Schema Object section (there are way too many PRs and issues on this at once right now).  I am strongly against this change.  The presence of "integer" as a value for `type` is standard JSON Schema and should not be repeated.  That just adds to confusion by including it in a list of exceptions _when it is not an exception._

PR #4045, opened 2024-08-20
Clarify data model types and formats

handrews:
This is what I think is actually correct.  It aligns the table with how JSON Schema behaves.  While I dislike re-explaining other specs in the OAS, it seems that a certain amount of explanation is required (because this is truly confusing).

I started with 3.1.1 because it uses unmodified JSON Schema.  I think we should do the same with 3.0.4 as I don't think 3.0 actually adds any normative requirements that would change this, but we can sort that out if and when this PR is accepted.

Paging @baywet @mkistler @ralfhandl @darrelmiller


versions/3.1.1.md
```diff
@@ -211,13 +213,15 @@ Types that are not accompanied by a `format` property follow the type definition
...
 The formats defined by the OAS are:
 
-| [`type`](#data-types) | [`format`](#data-type-format) | Comments                     |
-| -------------------- | --------------------------- | ---------------------------- |
-| `integer`            | `int32`                     | signed 32 bits               |
-| `integer`            | `int64`                     | signed 64 bits (a.k.a long)  |
-| `number`             | `float`                     |                              |
-| `number`             | `double`                    |                              |
-| `string`             | `password`                  | A hint to obscure the value. |
+| [Data Model Type](#data-types) | [`format`](#data-type-format) | Comments                     |
```
mikekistler, line 216: I know you like the term "Data Model Type", but for me the light bulb went on when I read in the JSON Schema spec "Instance Type" -- the type of the instance being processed. Could we use that term instead?

ralfhandl (commented):
+1, minor nits.

How about mentioning `type: "integer"` in the Schema Object section, similar to #4042 (which I could then close)?

baywet (approved):
Thanks! I does make things a bit clearer. Was the plan to retro fit 3.0.4 as well?

ralfhandl (changes requested):
Avoid referencing multiple JSON RFCs

versions/3.1.1.md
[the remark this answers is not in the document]
handrews: I didn't add that, I just kept it as it was.

handrews: @mikekistler do most people understand the term "instance"?  Would perhaps "JSON Data Type" work better and be less jargon-y?

baywet (approved):

versions/3.1.1.md
ralfhandl: Yes, please, make it "JSON Data Type".

versions/3.1.1.md
```diff
@@ -199,10 +199,12 @@ Note that no aspect of implicit connection resolution changes how [URIs are reso
 
 ### Data Types
 
-Data types in the OAS are based on the types supported by the [JSON Schema Specification Draft 2020-12](https://tools.ietf.org/html/draft-bhutton-json-schema-00#section-4.2.1).
-Note that `integer` as a type is also supported and is defined as a JSON number without a fraction or exponent part.
 Models are defined using the [Schema Object](#schema-object), which is a superset of JSON Schema Specification Draft 2020-12.
 
+Data types in the OAS are based on the six types supported by the [JSON Schema Specification Draft 2020-12 data model](https://tools.ietf.org/html/draft-bhutton-json-schema-00#section-4.2.1): array, boolean, null, number, object, or string.  JSON Schema keywords and `format` values operate on these six types, with certain keywords and formats only applying to a specific type.  For example, the `pattern` keyword and the `date-time` format only apply to strings, and treat any instance of the other five types as _automatically valid._  This means JSON Schema keywords and formats do **NOT** implicitly require the expected type.  Use the `type` keyword to explicitly constrain the type.
```
mikekistler, line 204: This is an excellent addition! A common problem I see is schemas that contain "properties" and _assume_ that this implies "type: object" -- and I too made this error early on, until a wiser colleague educated me.

mikekistler, line 204: The JSON Schema spec uses the terminology "primitive types" in section 4.2.1, and I think we should mirror that here, to distinguish from the values allowed for the "type" keyword.
```suggestion
Data types in the OAS are based on the six primitive types supported by the [JSON Schema Specification Draft 2020-12 data model](https://tools.ietf.org/html/draft-bhutton-json-schema-00#section-4.2.1): array, boolean, null, number, object, or string.  JSON Schema keywords and `format` values operate on these six primitive types, with certain keywords and formats only applying to a specific type.  For example, the `pattern` keyword and the `date-time` format only apply to strings, and treat any instance of the other five types as _automatically valid._  This means JSON Schema keywords and formats do **NOT** implicitly require the expected type.  Use the `type` keyword to explicitly constrain the type.
```

versions/3.1.1.md
```diff
@@ -199,10 +199,12 @@ Note that no aspect of implicit connection resolution changes how [URIs are reso
...
+
+Note that the `type` keyword allows `"integer"` as a value for convenience, but keyword and format applicability does not recognize integers as being distinct from other numbers because [[RFC7159|JSON]] itself does not make that distinction.  JSON Schema defines integers [mathematically](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#section-6.3), meaning that both `1` and `1.0` are considered to be integers for the purpose of the `type` keyword.
```
mikekistler, line 206:
```suggestion
Note that the `type` keyword allows `"integer"` as a value for convenience, but JSON Schema keyword and format applicability does not recognize integers as being distinct from other numbers because [[RFC7159|JSON]] itself does not make that distinction.  JSON Schema defines integers [mathematically](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#section-6.1.1) as "any number with a zero fractional part", meaning that both `1` and `1.0` are considered to be integers for the purpose of the `type` keyword.
```

versions/3.1.1.md
```diff
@@ -211,13 +213,15 @@ Types that are not accompanied by a `format` property follow the type definition
...
-| `string`             | `password`                  | A hint to obscure the value. |
+| [Data Model Type](#data-types) | [`format`](#data-type-format) | Comments                     |
+| ------------------------------ | ----------------------------- | ---------------------------- |
+| number                         | `int32`                       | signed 32 bits               |
+| number                         | `int64`                       | signed 64 bits (a.k.a long)  |
+| number                         | `float`                       |                              |
+| number                         | `double`                      |                              |
+| string                         | `password`                    | A hint to obscure the value. |
+
+As noted above, both `type: number` and `type: integer` are considered to be numbers in the data model.
```
mikekistler, line 224: I think this table should be "flipped around" to better align with the format registry. Also changed to "primitive type" to align with the change above.
```suggestion
| [`format`](#data-type-format) | [primitive type](#data-types) | Comments                     |
| ------------------------------ | ----------------------------- | ---------------------------- |
| `int32`                      | number                         | signed 32 bits               |
| `int64`                      | number                          | signed 64 bits (a.k.a long)  |
| `float`                       | number                         |                              |
| `double`                      | number                        |                              |
| `password`                    | string                         | A hint to obscure the value. |

As noted above, both `type: number` and `type: integer` are considered to be numbers in the data model.
```

mikekistler (changes requested):
I've made a few edits that I hope are improvements. I'm not wed to any of these but requested changes just to ensure we get to discuss these before a merge.

versions/3.1.1.md
ralfhandl: @mikekistler #4015 uses "Instance Type": why not use the same terminology here?

mikekistler: Only because I slept in between making comments there and here ;-). I agree that we should make the terminology consistent and I think "Instance Type", with a brief explanation of the term at it first appearance, is now my preference. But open to other opinions.

versions/3.1.1.md
handrews: Section 6.3 "Mathematical Integers" is correct here.  There is no 6.1.1 in that draft of JSON Schema.

I am against "any number with a zero fractional part" as a description as it is ambiguous as to whether it refers to the value or the textual representation.

Again, this is a JSON Schema thing.  JSON Schema's wording is what matters, and OpenAPI ***CANNOT*** change the requirement.  The only thing we can or should do here is reference the appropriate section of JSON Schema.

versions/3.1.1.md
handrews: I thought I had updated this yesterday to say "JSON Type" but I guess maybe I just made a comment about it on one of the other 9000 issues or PRs open on this topic ;-)  I do not think "instance" is intuitive for most people.

versions/3.1.1.md
handrews: I'm glad this makes sense - I am generally opposed to repeating information from JSON Schema, except where it has clearly been problematic and there isn't an easy single spot to reference in the JSON Schema text.  That's why I'm OK with this but do not want to do anything but link to the section on what constitutes an integer.

handrews:
I have added a new commit based on the feedback, doing the following:

* Change from "Data Model Type" to "JSON Data Type" (which I could have sworn was being used elsewhere but now I can't find it)
* Flip the columns as @mikekistler suggested
* Replace the anchor on the paragraph before the table with a section heading that matches the former manually-added anchor (we now have enough text before that point to justify a subsection)
* Remove the self-reference links in the column headers as @ralfhandl suggested (the `format` link is still a self-reference after the subsection is added, and the link to data types heading has been moved to the comment under the table about the nature of integers and `type` as it seemed to me to fit better there.

ralfhandl (approved):

ralfhandl (dismissed):

ralfhandl:
@mikekistler If you are fine with the current state, please approve and merge. Thanks!

versions/3.1.1.md
mikekistler: Quite right that there is no 6.1.1 in that draft. Let me see if I can find what I was looking at. In any event, I don't believe section 6.3 "defines integers -- this is the problem I was trying to fix, so let's find a good solution to that.

mikekistler: Here's the correct link:

https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#autoid-11

ralfhandl: JSON Schema _Validation_ defines `"integer"` in [section 6.1.1](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#section-6.1.1)

mikekistler: I've updated my suggested change. @handrews please review.

versions/3.1.1.md
```diff
@@ -199,25 +199,31 @@ Note that no aspect of implicit connection resolution changes how [URIs are reso
...
+
+#### Data Type Format
+
+As defined by the [JSON Schema Validation specification](https://tools.ietf.org/html/draft-bhutton-json-schema-validation-00#section-7.3), data types can have an optional modifier property: `format`. As described in that specification, `format` is treated as a non-validating annotation by default; the ability to validate `format` varies across implementations.
 
 The OpenAPI Initiative also hosts a [Format Registry](https://spec.openapis.org/registry/format/) for formats defined by OAS users and other specifications. Support for any registered format is strictly OPTIONAL, and support for one registered format does not imply support for any others.
 
 Types that are not accompanied by a `format` property follow the type definition in the JSON Schema. Tools that do not recognize a specific `format` MAY default back to the `type` alone, as if the `format` is not specified.
```
mikekistler: BTW ... I think there is a minor conflict with JSON Schema here, since it has the Format Assertion Vocabulary and:
> When the Format-Assertion vocabulary is declared with a value of true, implementations MUST provide full validation support for all of the formats defined by this specification.

versions/3.1.1.md
```diff
@@ -199,25 +199,31 @@ Note that no aspect of implicit connection resolution changes how [URIs are reso
...
```
mikekistler, line 215: Let's include the same text from the format registry the explains the term "JSON Data Type":
```suggestion
For the purpose of [JSON Schema validation](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#section-7.1), each format should specify the set of JSON data types for which it applies. In this registry, these types are shown in the "JSON Data Type" column.

```

versions/3.1.1.md
handrews: @mikekistler by default what you get is the Format Annotation Vocabulary, to which implementations MAY add whatever opt-in validation they want to provide.  This is because some major JSON Schema implementations refused to fully validate `format`, but did a certain amount of best-effort validation.

I *really* don't want to get into the Format Assertion Vocabulary here.  I'm not sure many implementations support it, and you need to know how to change the meta-schema to include it.

versions/3.1.1.md
handrews: We use "primitive types" to mean "not array or object" in the OAS.  Let's leave JSON Schema's wording in JSON Schema.

versions/3.1.1.md
handrews: @mikekistler the point of this link is to explain why `1.0` is an integer.  The core spec §6.3 does that.  That is literally the entire purpose of that section.

This text here already explains that the `type` keyword allows `"integer"`.  If you _also_ want to link to the `type` keyword's section, we can do that on the word `type` (although I'd rather not- we don't do that for any other mention of any other JSON Schema keyword AFAICT so it would be a confusing inconsistency).

If you want to look up the details of `type`, it's obvious that you should look up its section in the validation spec.  On the other hand, the mathematical nature of integers in JSON Schema is a recurring point of confusion, and it's less obvious where to resolve it.  Hence linking to the section that explains it.

versions/3.1.1.md
mikekistler: Okay ... I think we agreed to use "JSON data types" elsewhere, so I've switched to that. I believe its important to qualify "types" here -- not let it stand alone -- to minimize the potential confusion with values of the "type" keyword.

versions/3.1.1.md
mikekistler: @handrews When I follow the link to section 6.3 I see this text:
> Some programming languages and parsers use different internal representations for floating point numbers than they do for integers.
> For consistency, integer JSON numbers SHOULD NOT be encoded with a fractional part.

I don't think there is anything here that "defines integers", and I don't know why I would conclude from this text that "1.0" is an integer.

handrews: @mikekistler OK let me think on this more.  §6.1.1 of the validation spec still isn't what should go here.

ralfhandl (approved):

handrews:
@mikekistler here's another pass linking to where the equivalence of things like `1` vs `1.0` are discussed.  Let me know what you think.  I am avoiding linking to the `type` section because:

1.  I want to keep the point focused on the data model (with `type: integer` only needing explanation in the context of that data model)
2. Everyone knows where to find the spec for `type` anyway
3.  I find showing these examples (`1` and `1.0`) makes it more clear to people than wording like "zero fractional part"
4. I think the section on equivalence helps flesh out the rationale- if `1` is an integer (no one disputes this), and `1` and `1.0` are considered to be equivalent by JSON Schema, then `1.0` must also be an integer because "equivalence" means it has to be treated the same

I tweaked some wording to hopefully focus things a bit better.

versions/3.1.1.md
```diff
@@ -199,25 +199,34 @@ Note that no aspect of implicit connection resolution changes how [URIs are reso
 
 ### Data Types
 
-Data types in the OAS are based on the types supported by the [JSON Schema Specification Draft 2020-12](https://tools.ietf.org/html/draft-bhutton-json-schema-00#section-4.2.1).
-Note that `integer` as a type is also supported and is defined as a JSON number without a fraction or exponent part.
-Models are defined using the [Schema Object](#schema-object), which is a superset of JSON Schema Specification Draft 2020-12.
+Models are defined using the [Schema Object](#schema-object), which is a superset of the JSON Schema Specification Draft 2020-12.
 
-<a name="data-type-format"></a>As defined by the [JSON Schema Validation specification](https://tools.ietf.org/html/draft-bhutton-json-schema-validation-00#section-7.3), data types can have an optional modifier keyword: `format`. As described in that specification, `format` is treated as a non-validating annotation by default; the ability to validate `format` varies across implementations.
+Data types in the OAS are based on the six JSON data types supported by the [JSON Schema's data model](https://tools.ietf.org/html/draft-bhutton-json-schema-00#section-4.2.1): array, boolean, null, number, object, or string.
```
mikekistler, line 204: Now that I understand that JSON Schema actually defines "integer" as a type, why shouldn't we include it here?

handrews: "integer" is a value for the `type` keyword.  It's not a JSON data type in the data model.

handrews:
Update:  @mikekistler and I have been having a good discussion about whether this section now over-emphasizes JSON Schema's mechanisms compared to OAS behavior, so there is likely to be some rewriting or perhaps another alternative PR.

versions/3.1.1.md
miqui: +1 @mikekistler

miqui:
> Update: @mikekistler and I have been having a good discussion about whether this section now over-emphasizes JSON Schema's mechanisms compared to OAS behavior, so there is likely to be some rewriting or perhaps another alternative PR.

@handrews, I am okay with this PR. However, I'll wait for your resolution on the above.

versions/3.1.1.md
```diff
@@ -199,25 +199,34 @@ Note that no aspect of implicit connection resolution changes how [URIs are reso
 
 ### Data Types
 
-Data types in the OAS are based on the types supported by the [JSON Schema Specification Draft 2020-12](https://tools.ietf.org/html/draft-bhutton-json-schema-00#section-4.2.1).
-Note that `integer` as a type is also supported and is defined as a JSON number without a fraction or exponent part.
-Models are defined using the [Schema Object](#schema-object), which is a superset of JSON Schema Specification Draft 2020-12.
+Models are defined using the [Schema Object](#schema-object), which is a superset of the JSON Schema Specification Draft 2020-12.
```
mikekistler, line 202: I think we should restore a slight variant of the first sentence in this first paragraph.
```suggestion
Data types in the OAS are based on the types defined by the [JSON Schema Validation Specification Draft 2020-12](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#section-6.1.1):
"null", "boolean", "object", "array", "number", "string", or "integer".
Models are defined using the [Schema Object](#schema-object), which is a superset of the JSON Schema Specification Draft 2020-12.
```

versions/3.1.1.md
mikekistler, line 204: And now remove this sentence
```suggestion
```

versions/3.1.1.md
```diff
@@ -199,25 +199,34 @@ Note that no aspect of implicit connection resolution changes how [URIs are reso
...
 
-<a name="data-type-format"></a>As defined by the [JSON Schema Validation specification](https://tools.ietf.org/html/draft-bhutton-json-schema-validation-00#section-7.3), data types can have an optional modifier keyword: `format`. As described in that specification, `format` is treated as a non-validating annotation by default; the ability to validate `format` varies across implementations.
+Data types in the OAS are based on the six JSON data types supported by the [JSON Schema's data model](https://tools.ietf.org/html/draft-bhutton-json-schema-00#section-4.2.1): array, boolean, null, number, object, or string.
+
+JSON Schema keywords and `format` values operate on these six JSON data types, with certain keywords and formats only applying to a specific type.  For example, the `pattern` keyword and the `date-time` format only apply to strings, and treat any instance of the other five types as _automatically valid._  This means JSON Schema keywords and formats do **NOT** implicitly require the expected type.  Use the `type` keyword to explicitly constrain the type.
```
mikekistler, line 206: And finally a minor modification here:
```suggestion
JSON Schema keywords and `format` values operate on JSON "instances" which may be one of the six JSON data types, "null", "boolean", "object", "array", "number", or "string", with certain keywords and formats only applying to a specific type.  For example, the `pattern` keyword and the `date-time` format only apply to strings, and treat any instance of the other five types as _automatically valid._  This means JSON Schema keywords and formats do **NOT** implicitly require the expected type.  Use the `type` keyword to explicitly constrain the type.
```

I know some folks are uncomfortable with the term "instance" but this is JSON Schema terminology so I think we should use it, in particular to emphasize the that data type is determined by the data value (instance) and _not_ from the `type` keyword. We could link to the section in the JSON Schema spec that defines "instance" if folks believe it is necessary.

https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#section-4.2

mikekistler (changes requested):
Here's my proposal for describing the OAS types but still clarifying that JSON Schema keywords and format values are applied conditionally base on JSON data type.

handrews (commented):
Great work, @mikekistler !  This is definitely the best yet, and with this comment I'm giving it my approval for merging (GitHub won't let me submit an approval since I originally submitted the PR).

I'm still pondering whether we ought to say anything about `type: number` vs `type: integer`, but that's a subset of the larger "how should programming languages handle type system mismatches" question and need not hold this up, I think.

I've taken the PR back out of draft.

mikekistler (dismissed):
Looks good! 👍

versions/3.1.1.md
```diff
@@ -199,25 +199,35 @@ Note that no aspect of implicit connection resolution changes how [URIs are reso
...
+
+Note that the `type` keyword allows `"integer"` as a value for convenience, but keyword and format applicability does not recognize integers as being of a distinct JSON type from other numbers because [[RFC7159|JSON]] itself does not make that distinction.  Since there is no distinct JSON integer type, JSON Schema defines integers mathematically.  This means that both `1` and `1.0` are [equivalent](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#section-4.2.2), and are both considered to be integers.
+
+#### Data Type Format
+
+As defined by the [JSON Schema Validation specification](https://tools.ietf.org/html/draft-bhutton-json-schema-validation-00#section-7.3), data types can have an optional modifier property: `format`. As described in that specification, `format` is treated as a non-validating annotation by default; the ability to validate `format` varies across implementations.
```
ralfhandl, line 213:
```suggestion
As defined by the [JSON Schema Validation specification](https://tools.ietf.org/html/draft-bhutton-json-schema-validation-00#section-7.3), data types can have an optional modifier keyword: `format`. As described in that specification, `format` is treated as a non-validating annotation by default; the ability to validate `format` varies across implementations.
```

We call it a "keyword" in line 217, [JSON Schema Validation 2020-12](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#section-7.1) calls it "annotation keyword", and [JSON Schema Validation 2020-12](https://datatracker.ietf.org/doc/html/draft-wright-json-schema-validation-00#section-7.1) just "keyword".

ralfhandl (approved):

versions/3.1.1.md
```diff
@@ -199,25 +199,35 @@ Note that no aspect of implicit connection resolution changes how [URIs are reso
 
 ### Data Types
 
-Data types in the OAS are based on the types supported by the [JSON Schema Specification Draft 2020-12](https://tools.ietf.org/html/draft-bhutton-json-schema-00#section-4.2.1).
-Note that `integer` as a type is also supported and is defined as a JSON number without a fraction or exponent part.
-Models are defined using the [Schema Object](#schema-object), which is a superset of JSON Schema Specification Draft 2020-12.
+Data types in the OAS are based on the types defined by the [JSON Schema Validation Specification Draft 2020-12](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#section-6.1.1):
```
mikekistler, line 202: Here we are using a markdown link -- should this be a ReSpec link instead?

versions/3.1.1.md
```diff
@@ -199,25 +199,35 @@ Note that no aspect of implicit connection resolution changes how [URIs are reso
...
+
+Note that the `type` keyword allows `"integer"` as a value for convenience, but keyword and format applicability does not recognize integers as being of a distinct JSON type from other numbers because [[RFC7159|JSON]] itself does not make that distinction.  Since there is no distinct JSON integer type, JSON Schema defines integers mathematically.  This means that both `1` and `1.0` are [equivalent](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#section-4.2.2), and are both considered to be integers.
+
+#### Data Type Format
+
+As defined by the [JSON Schema Validation specification](https://tools.ietf.org/html/draft-bhutton-json-schema-validation-00#section-7.3), data types can have an optional modifier keyword: `format`. As described in that specification, `format` is treated as a non-validating annotation by default; the ability to validate `format` varies across implementations.
```
mikekistler, line 213: Another markdown link.

versions/3.1.1.md
```diff
@@ -199,25 +199,35 @@ Note that no aspect of implicit connection resolution changes how [URIs are reso
...
 
 The OpenAPI Initiative also hosts a [Format Registry](https://spec.openapis.org/registry/format/) for formats defined by OAS users and other specifications. Support for any registered format is strictly OPTIONAL, and support for one registered format does not imply support for any others.
 
 Types that are not accompanied by a `format` keyword follow the type definition in the JSON Schema. Tools that do not recognize a specific `format` MAY default back to the `type` alone, as if the `format` is not specified.
+For the purpose of [JSON Schema validation](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#section-7.1), each format should specify the set of JSON data types for which it applies. In this registry, these types are shown in the "JSON Data Type" column.
```
mikekistler, line 218: Another markdown link.

mikekistler (approved):
Looks good. 👍 

I flagged a few links that may?? need to be converted to Respec links. I haven't grokked the Respec link stuff yet, so this might be just misunderstanding on my part.

versions/3.1.1.md
ralfhandl: This is fine.

Direct section links don't seem possible with ReSpec, so we have special handling for RFC links with section fragments, converting them into a ReSpec document reference _plus_ a direct link to the RFC section.

Links to JSON Schema specs without a section fragment are turned into ReSpec document references, and one of them is sufficient for the spec to show up in Normative References.

Direct section links (with a fragment) are left untouched.

PR #4050, opened 2024-08-26
"Object" definition, "with" in titles, fix level of "Working with Examples"

handrews:
Define capital-O "Object" to make it clear why it is only occasionally capitalized.  I complained about some of these being lower-cased in 3.1.1 and 3.2.0 and will fix those when forward-porting.  For whatever reason, they were not changed in 3.0.4, so there's nothing to fix here other than adding the definition.

Use lowercase "with" in titles consistently (it was more common than capitalized "With".  This is one of those rules that is different depending on whose style guide you use- about half the major guides say always lowercase, the other say always captial.  We were using lowercase more often.

Also, fix "Working with Examples" to be a subsection of the Example Object as otherwise it breaks the pattern of all of the headings of its level being Object definitions.  I remember messing with this a lot when I first posted that PR and either I just missed this or I had some reason I have now forgotten to do it this way.


mikekistler (approved):
Looks good. 👍

ralfhandl (approved):

PR #4052, opened 2024-08-27
"Object" definition, etc. (3.1.1 port of #3050)

handrews:
Includes two capitalizations (object -> Object) as a second commit on top of the ported commit.


mikekistler (approved):
Looks good. 👍

ralfhandl (approved):

PR #4053, opened 2024-08-27
Port format / integer changes from 3.1.1

handrews:
This is a slightly different change due to the different JSON Schema draft being referenced.  Please check it carefully as it would have been easy for me to miss something!

Most notably, the older draft has a section on type applicability, so there is an extra link here that is not present in 3.1.1.

### DIFF
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -176,32 +176,37 @@ Note that no aspect of implicit connection resolution changes how [URLs are reso
 
 ### Data Types
 
-Data types in the OAS are based on the types supported by the [JSON Schema Specification Wright Draft 00](https://tools.ietf.org/html/draft-wright-json-schema-00#section-4.2).
-Note that `integer` as a type is also supported and is defined as a JSON number without a fraction or exponent part.
-`null` is not supported as a type (see [`nullable`](#schema-nullable) for an alternative solution).
-Models are defined using the [Schema Object](#schema-object), which is an extended subset of JSON Schema Specification Wright Draft 00.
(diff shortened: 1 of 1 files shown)


versions/3.0.4.md
```diff
@@ -176,32 +176,37 @@ Note that no aspect of implicit connection resolution changes how [URLs are reso
 
 ### Data Types
 
-Data types in the OAS are based on the types supported by the [JSON Schema Specification Wright Draft 00](https://tools.ietf.org/html/draft-wright-json-schema-00#section-4.2).
-Note that `integer` as a type is also supported and is defined as a JSON number without a fraction or exponent part.
-`null` is not supported as a type (see [`nullable`](#schema-nullable) for an alternative solution).
-Models are defined using the [Schema Object](#schema-object), which is an extended subset of JSON Schema Specification Wright Draft 00.
+Data types in the OAS are based on the non-`null` types supported by the [JSON Schema Validation Specification Draft Wright-00](https://datatracker.ietf.org/doc/html/draft-wright-json-schema-validation-00#autoid-32):
+"boolean", "object", "array", "number", "string", or "integer".
+See [`nullable`](#schema-nullable) for an alternative solution to "null" as a type.
```
mikekistler, line 181: Oh wow! I totally forgot about "null" when we were working on the 3.0.4 version. We need to exclude null there too, right?

mikekistler (approved):
Looks good! 👍 

Do we need to fix the 3.0.4 spec to exclude "null" as an OAS type as is done here?

versions/3.0.4.md
handrews: @mikekistler this is the 3.0.4 version- I did this change backwards (3.1.1 first) because 3.1.1 was easier as it uses JSON Schema as-is.  Hopefully I didn't miss anything in this PR where `null` needed to be removed.

PR #4054, opened 2024-08-28
3.0.4: use same reference style for draft wright as #4053

ralfhandl:
Follow-up to
* #4053

- [x] use same style for referencing "JSON Schema Specification Draft Wright-00" everywhere


mikekistler (approved):
Looks good! 👍

PR #4055, opened 2024-08-28
3.1.1: example for "raw JSON" as message body

ralfhandl:
- [x] add example for "raw JSON" as a message body - external feedback
- [x] prevent line-break in monospaced table cell content (feedback from @handrews)

### DIFF
--- a/versions/3.1.1.md
+++ b/versions/3.1.1.md
@@ -267,8 +267,8 @@ The following table shows how to migrate from OAS 3.0 binary data descriptions, 
 
 | OAS < 3.1                            | OAS 3.1                                                                          | Comments                                                                                |
 | ------------------------------------ | -------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------- |
-| `type: string`<br /><code style="white-space:nowrap">format: binary</code> | <code style="white-space:nowrap">contentMediaType: image/png</code> | if redundant, can be omitted, often resulting in an empty [Schema Object](#schema-object) |
-| `type: string`<br />`format: byte` | `type: string`<br />`contentMediaType: image/png`<br />`contentEncoding: base64` | note that `base64url` can be used to avoid re-encoding the base64 string to be URL-safe |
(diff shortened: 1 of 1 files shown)


versions/3.1.1.md
```diff
@@ -1591,6 +1591,12 @@ content:
   application/octet-stream: {}
 ```
 
+```yaml
+# arbitrary JSON without constraints beyond being well-formed:
```
handrews, line 1595: I vaguely wonder if "being well-formed" will throw anyone.  Perhaps "being syntactically valid JSON"?  I am fine with it as it is, though.

versions/3.1.1.md
```diff
@@ -267,8 +267,8 @@ The following table shows how to migrate from OAS 3.0 binary data descriptions,
...
 | ------------------------------------ | -------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------- |
-| `type: string`<br />`format: binary` | `contentMediaType: image/png`                                                    | if redundant, can be omitted, often resulting in an empty [Schema Object](#schema-object) |
-| `type: string`<br />`format: byte`   | `type: string`<br />`contentMediaType: image/png`<br />`contentEncoding: base64` | note that `base64url` can be used to avoid re-encoding the base64 string to be URL-safe |
+| `type: string`<br /><code style="white-space:nowrap">format: binary</code> | <code style="white-space:nowrap">contentMediaType: image/png</code> | if redundant, can be omitted, often resulting in an empty [Schema Object](#schema-object) |
+| `type: string`<br />`format: byte` | `type: string`<br />`contentMediaType: image/png`<br />`contentEncoding: base64` | note that `base64url` can be used to avoid re-encoding the base64 string to be URL-safe |
```
handrews, line 271: I think it might be best to do all of the bits with `<code style...` because either someone will have a device on which the other parts wrap unexpectedly (even if I don't see how that would happen offhand) or someone will slightly tweak this and then one of the non-`<code...` bits will end up longer and wrapping.

But I'm not going to hold this up if you'd rather leave it as-is.

pk-work (approved):

handrews (commented):
Minor nits- if others are fine with this as-is, feel free to merge.

handrews (approved):

lornajane (approved):

PR #4056, opened 2024-08-29
3.1.1: tables cleanup

ralfhandl:
3.1.1 port of
* #4040

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,6 @@
     "API"
   ],
   "scripts": {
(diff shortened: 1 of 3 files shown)


mikekistler (approved):
Looks good. 👍 

Tip to other reviewers: You can "hide whitespace" when reviewing on GitHub and that makes the spec file diff much smaller -- only table header changes remain.

miqui (approved):

lornajane (approved):

PR #4058, opened 2024-08-29
gh-pages: schema list on landing page: call out "view" and "download" links

ralfhandl:
Attempted UX improvement:
- call out which links are "view", which are "download", and what can be viewed or downloaded
- also list "meta" and "dialect" for v3.1

image

Live preview:
* https://ralfhandl.github.io/OpenAPI-Specification/


miqui (approved):
@ralfhandl looks good to me.

handrews (approved):

PR #4059, opened 2024-08-30
3.0.4: absent, empty, or incomplete security list

ralfhandl:
Trying to sum up our [lively and highly educating discussion](https://github.com/OAI/OpenAPI-Specification/issues/3938#issuecomment-2308874884) in one sentence - improvement suggestions are welcome.

Part of
* #3938


mikekistler (approved):
Looks good. 👍 

In future specs we may want something in the intro that explains the whole "open world" concept, but for now I think this addresses the specific concern with an empty security array.

miqui (approved):

PR #4060, opened 2024-08-30
3.0.4: fix json schema reference

ralfhandl:
Changing the link text broke our ReSpec magic in the build script.

Now use ReSpec link in source and don't rely on magic.


mikekistler (approved):
Looks good! 👍

handrews (approved):

PR #4061, opened 2024-08-30
3.1.1: File Upload - difference also to 3.0.x

ralfhandl:
- The contrast here is due to the different JSON Schema version and also affects 3.0
- Stylistic harmonization: we use "OAS 3" everywhere else instead of "OpenAPI 3"
- We use "OpenAPI 2.0" eight times


pk-work (approved):

mikekistler (approved):
Looks good. 👍

handrews (approved):

PR #4062, opened 2024-08-30
3.0.4: port the undisputed parts of #2140

ralfhandl:
This is an attempt to revive/port
* #2140 

It contains all undisputed changes:
- rework confusing description in example
- remove confusing reference to the Schema Object's treatment of default values

Part of
* #1675

### DIFF
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -488,7 +488,7 @@ An object representing a Server Variable for server URL template substitution.
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
 | <a name="server-variable-enum"></a>enum | [`string`] | An enumeration of string values to be used if the substitution options are from a limited set. The array SHOULD NOT be empty. |
-| <a name="server-variable-default"></a>default | `string` | **REQUIRED**. The default value to use for substitution, which SHALL be sent if an alternate value is _not_ supplied. If the [`enum`](#server-variable-enum) is defined, the value SHOULD exist in the enum's values. |
+| <a name="server-variable-default"></a>default | `string` | **REQUIRED**. The default value to use for substitution, which SHALL be sent if an alternate value is _not_ supplied. If the [`enum`](#server-variable-enum) is defined, the value SHOULD exist in the enum's values. Note that this behavior is different from the [Schema Object](#schema-object)'s `default` keyword, which documents the receiver's behavior rather than inserting the value into the data. |
 | <a name="server-variable-description"></a>description | `string` | An optional description for the server variable. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation. |
 
 This object MAY be extended with [Specification Extensions](#specification-extensions).
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -445,7 +445,7 @@ The following shows how variables can be used for a server configuration:
       "variables": {
         "username": {
           "default": "demo",
-          "description": "this value is assigned by the service provider, in this example `gigantic-server.com`"
+          "description": "A user-specific subdomain. Use `demo` for a free sandbox environment."
         },
         "port": {
           "enum": ["8443", "443"],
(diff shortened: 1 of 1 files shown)


versions/3.0.4.md
```diff
@@ -488,7 +488,7 @@ An object representing a Server Variable for server URL template substitution.
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
 | <a name="server-variable-enum"></a>enum | [`string`] | An enumeration of string values to be used if the substitution options are from a limited set. The array SHOULD NOT be empty. |
-| <a name="server-variable-default"></a>default | `string` | **REQUIRED**. The default value to use for substitution, which SHALL be sent if an alternate value is _not_ supplied. Note this behavior is different than the [Schema Object's](#schema-object) treatment of default values, because in those cases parameter values are optional. If the [`enum`](#server-variable-enum) is defined, the value SHOULD exist in the enum's values. |
+| <a name="server-variable-default"></a>default | `string` | **REQUIRED**. The default value to use for substitution, which SHALL be sent if an alternate value is _not_ supplied. If the [`enum`](#server-variable-enum) is defined, the value SHOULD exist in the enum's values. |
```
mikekistler, line 491: While the sentence that was delete here is a bit confusing, I think there is something lost if we don't call out the difference in the meaning of "default" here. Would something like this be okay?
> Note this differs from a default value in the [Schema Object's](#schema-object), which specify the value applied by the server when none is provided by the client.

ralfhandl: The preceding sentence says
> SHALL be sent

That is a MUST, not sending a value is not an allowed option.

The Schema Object's `default` talks about what 
> would be assumed [...] if one is not provided

Referencing something that does not apply is confusing.

mikekistler: I think we are referencing something to explain that, while it looks very similar -- same property name `default` in a very similar context -- it is in fact different and does not apply. I think that will be helpful to users of the spec, no?

ralfhandl: @darrelmiller suggested to remove this sentence in https://github.com/OAI/OpenAPI-Specification/pull/2140#discussion_r380796394. 

I agree because I find it confusing here.

handrews: I think it would be better to change the sentence to something like:

Note that this behavior is different from the Schema Object's `default` keyword, which documents the default behavior rather than inserting the value into the data.

miqui: >Note that this behavior is different from the Schema Object's default keyword, which documents the default behavior rather than inserting the value into the data.

That sounds good to me. However I don't really have a problem with Ralf's PR as is.

miqui (approved):

versions/3.0.4.md
ralfhandl: Note that keeping this sentence here is the option that was least favored by both @darrelmiller and the PR author @tedepstein, see [his comment](https://github.com/OAI/OpenAPI-Specification/pull/2140#discussion_r379936894).

handrews: @ralfhandl I'm certainly not proposing keeping the current sentence, which I also find confusing.  But people find `default` _endlessly_ confusing.  They are particularly likely to think that JSON Schema's `default` means that the value of `default` should be written into the instance, ***which is not how it works at all***.

So the fact that we have a _different_ `default` that _does_ work that way is a very, very likely point of confusion.  The text that was there before did not really help.  However, I think the text that I proposed explains the real difference succinctly, and will keep people who read this `default` from thinking that JSON Schema's `default` works similarly.  Which was always one of the endless recurring questions/complaints in the JSON Schema project, so it's a very practical and real concern, with years of experience answering this exact question to back it up.

I think my experience here is a bit deeper than was available during the previous conversation.

ralfhandl: @handrews Good point, have added your proposed sentence to the end of the table cell's text, and used "receiver's behavior" instead of "default behavior" to stress the difference.

handrews (approved):

ralfhandl:
@mikekistler does the added explanation meet your expectations?

miqui (approved):

PR #4065, opened 2024-09-02
3.0.4: update from main

ralfhandl:
This merges current `main` into `v3.0.4-dev`, resolving the two merge conflicts in 
- .gitignore: taken from main
- validate-markdown.yaml: taken from v3.0.4-dev


mikekistler (approved):
Looks good. 👍 

I reviewed this by checking out the branch locally and then diffing with main.  Only 5 files are changed:
```text
>git diff main --name-only               
.github/workflows/validate-markdown.yaml
package-lock.json
package.json
scripts/format-markdown.sh
versions/3.0.4.md
```

Changes to versions/3.04.md are what we want. I checked the diffs of the other files and all look right.

handrews:
Same question here- why are we doing this when we're about to merge in the other direction?

handrews:
From a DM conversation, apparently this is how it's preferred to resolve conflicts, which seems messy to me but I won't object to it.

handrews (approved):
`3.0.4.md` was not changed so this looks good.

PR #4035, opened 2024-08-16
3.2.0: replace "property" with "field" or "keyword" where appropriate

ralfhandl:
This is the 3.2.0 port of
* #4032
* #4020

Builds on and contains
* #4025 
* #4026

PR #4066, opened 2024-09-02
3.2.0: sync with 3.1.1

### LINKED ISSUE #3938: Clarification about the meaning of an empty security array

ralfhandl:
Remaining changes to 3.1.1
* #4091 
* #4088 
* #4085 
* #4079 
* #4071 
* #4064
* #4056
* #4055 
* #4052 
* #4045 
* #4044 
* #4041

Depends on and includes
* #4035

Fixes #3938

### DIFF
--- a/versions/3.2.0.md
+++ b/versions/3.2.0.md
@@ -28,6 +28,10 @@ A "schema" is a formal description of syntax and structure.
 This document serves as the [schema](#schema) for the OpenAPI Specification format; a non-authoritative JSON Schema based on this document is also provided on [spec.openapis.org](https://spec.openapis.org) for informational purposes.
 This specification also _uses_ schemas in the form of the [Schema Object](#schema-object).
 
+### Object
+
+When capitalized, the word "Object" refers to any of the Objects that are named by section headings in this document.
+
 ### Path Templating
 
 Path templating refers to the usage of template expressions, delimited by curly braces (`{}`), to mark a section of a URL path as replaceable using path parameters.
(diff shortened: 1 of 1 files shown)


handrews (dismissed):

ralfhandl:
Added latest updates to 3.1.1

handrews (approved):

handrews:
@ralfhandl if we follow the proposal in https://github.com/OAI/OpenAPI-Specification/issues/3677#issuecomment-2427757769, we will not need this PR as we will just abandon `v3.2.0-dev` in favor of a reconstructed `v3.2-dev` (there have only been four 3.2-specific changes so far, and porting those makes for a much nicer 3.2 version history!)

ralfhandl:
@handrews Thanks for taking over!

PR #4074, opened 2024-09-06
3.1.1: update from main

ralfhandl:
This merges current `main` into `v3.1.1-dev`, resolving the three merge conflicts in 
- .gitignore: taken from main
- validate-markdown.yaml: taken from v3.1.1-dev
- 3.1.0.md: taken from main


handrews:
Why are we merging main into 3.1.1 when we're just about to merge 3.1.1 into main?  Apologies if I missed somewhere that documents this.

handrews:
Same as 3.0.4: From a DM conversation, apparently this is how it's preferred to resolve conflicts, which seems messy to me but I won't object to it.

handrews (approved):
I think all I really care about here is that `3.1.1.md` was _not_ changed, and that looks correct here.  `3.1.0.md` being taken from `main` is also correct and fixes an annoying problem with branch divergence.

miqui (commented):
almost done

miqui (approved):

PR #4077, opened 2024-09-07
Clarify complete vs self-contained documents

handrews:
With apologies for the last-minute change, I encountered some confusion around the term "complete document", where it was mistaken for a single, self-contained OpenAPI Description.  This is an attempt to solve that- we don't use "complete document" outside of this bit, so the awkward "syntactically complete" (with "self-contained" defined separately for clarity) probably does what we need.  It is a bit awkward, though, so if anyone else can think of better words here please do speak up.

### DIFF
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -121,7 +121,9 @@ In order to preserve the ability to round-trip between YAML and JSON formats, YA
 
 An OpenAPI Description (OAD) MAY be made up of a single document or be divided into multiple, connected parts at the discretion of the author. In the latter case, [Reference Object](#reference-object) and [Path Item Object](#path-item-object) `$ref` keywords, as well as the [Link Object](#link-object) `operationRef` keyword, are used.
 
-Any document consisting entirely of an [OpenAPI Object](#openapi-object) is known as a **complete OpenAPI document**.
+Any document consisting entirely of an [OpenAPI Object](#openapi-object) is known as a **syntactically complete OpenAPI document**.
+An OpenAPI document that does ***not*** reference any other documents is known as a **self-contained OpenAPI document.***
+A single-document description is therefore _both_ syntactically complete _and_ self-contained.
 In a multi-document description, the document containing the OpenAPI Object where parsing begins for a specific API's description is known as that API's **entry OpenAPI document**, or simply **entry document**.
 
 It is RECOMMENDED that the entry OpenAPI document be named: `openapi.json` or `openapi.yaml`.
@@ -130,7 +132,7 @@ It is RECOMMENDED that the entry OpenAPI document be named: `openapi.json` or `o
 
 When parsing an OAD, JSON or YAML objects are parsed into specific Objects (such as [Operation Objects](#operation-object), [Response Objects](#response-object), [Reference Objects](#reference-object), etc.) based on the parsing context. Depending on how references are arranged, a given JSON or YAML object can be parsed in multiple different contexts:
 
-* As a complete OpenAPI Description document
+* As a syntactically complete OpenAPI Description document
 * As the Object type implied by its parent Object within the document
 * As a reference target, with the Object type matching the reference source's context
 
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -122,7 +122,7 @@ In order to preserve the ability to round-trip between YAML and JSON formats, YA
 An OpenAPI Description (OAD) MAY be made up of a single document or be divided into multiple, connected parts at the discretion of the author. In the latter case, [Reference Object](#reference-object) and [Path Item Object](#path-item-object) `$ref` keywords, as well as the [Link Object](#link-object) `operationRef` keyword, are used.
 
 Any document consisting entirely of an [OpenAPI Object](#openapi-object) is known as a **syntactically complete OpenAPI document**.
-An OpenAPI document that does ***not*** reference any other documents is known as a **self-contained OpenAPI document.***
+An OpenAPI document that does _not_ reference any other documents is known as a **self-contained OpenAPI document**.
 A single-document description is therefore _both_ syntactically complete _and_ self-contained.
 In a multi-document description, the document containing the OpenAPI Object where parsing begins for a specific API's description is known as that API's **entry OpenAPI document**, or simply **entry document**.
 


versions/3.0.4.md
```diff
@@ -121,7 +121,9 @@ In order to preserve the ability to round-trip between YAML and JSON formats, YA
 
 An OpenAPI Description (OAD) MAY be made up of a single document or be divided into multiple, connected parts at the discretion of the author. In the latter case, [Reference Object](#reference-object) and [Path Item Object](#path-item-object) `$ref` keywords, as well as the [Link Object](#link-object) `operationRef` keyword, are used.
 
-Any document consisting entirely of an [OpenAPI Object](#openapi-object) is known as a **complete OpenAPI document**.
+Any document consisting entirely of an [OpenAPI Object](#openapi-object) is known as a **syntactically complete OpenAPI document**.
+An OpenAPI document that does ***not*** reference any other documents is known as a **self-contained OpenAPI document.***
```
ralfhandl, line 125:
```suggestion
An OpenAPI document that does _not_ reference any other documents is known as a **self-contained OpenAPI document**.
```
Two changes:
- Three stars after "document" didn't pass markdown validation.
- Making "not" bold and italics would be a first in this spec, reduced it to italics.

ralfhandl (changes requested):

ralfhandl (approved):

versions/3.0.4.md
handrews: Thanks, @ralfhandl !

versions/3.0.4.md
```diff
@@ -130,7 +132,7 @@ It is RECOMMENDED that the entry OpenAPI document be named: `openapi.json` or `o
 
 When parsing an OAD, JSON or YAML objects are parsed into specific Objects (such as [Operation Objects](#operation-object), [Response Objects](#response-object), [Reference Objects](#reference-object), etc.) based on the parsing context. Depending on how references are arranged, a given JSON or YAML object can be parsed in multiple different contexts:
 
-* As a complete OpenAPI Description document
+* As a syntactically complete OpenAPI Description document
```
mikekistler, line 135: When do we ever parse a document in the context of "a (syntactically) complete OAD" rather than one of the other options?

mikekistler, line 135:
```suggestion
* As a syntactically complete OpenAPI Description
```

mikekistler (changes requested):
I have always found this section confusing.

I would like to suggest:
- We eliminate any appearance of "OpenAPI document" - I think all these should be "OpenAPI description". "OpenAPI description document" should likewise be simply "OpenAPI description".
- "document" should refer to a JSON or Yaml structure.

I left some suggestions to illustrate how this might look, but I think more work might still be needed.

versions/3.0.4.md
```diff
@@ -121,7 +121,9 @@ In order to preserve the ability to round-trip between YAML and JSON formats, YA
 
 An OpenAPI Description (OAD) MAY be made up of a single document or be divided into multiple, connected parts at the discretion of the author. In the latter case, [Reference Object](#reference-object) and [Path Item Object](#path-item-object) `$ref` keywords, as well as the [Link Object](#link-object) `operationRef` keyword, are used.
```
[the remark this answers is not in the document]
handrews: @mikekistler I like the first part of this, but I would prefer not to introduce "document identifier" given that "document identifier" is not a term I recall ever hearing.  The various reference keywords and the sections on URIs and URLs explain all of that just fine.  The reason that last sentence is there is because too many people aren't aware that the Path Item `$ref` is distinct, and the Link Object `operationRef` is similar.  Actually, the Discriminator Object's `mapping` can work with URIs as well although it's more complicated 🤔

versions/3.0.4.md
handrews: It is the objects that are parsed "in the context", not documents.  If you have a syntactically complete document, then you know that the root object is an OpenAPI Object.

You can't parse the root object of the entry document by either parent context or reference context because neither of those contexts exist.  But you can only parse a syntactically complete document without those contexts.

Basically, if you just get some random document without any other context, you _MUST_ treat it as an entry document, and error if the root object is not an OpenAPI Object.  It might be better to state that more directly and reserve the "context" wording for the other two cases?

handrews: This change would be incorrect- "syntactically complete" applies to _individual documents_, not descriptions.

versions/3.0.4.md
```diff
@@ -130,7 +132,7 @@ It is RECOMMENDED that the entry OpenAPI document be named: `openapi.json` or `o
 
 When parsing an OAD, JSON or YAML objects are parsed into specific Objects (such as [Operation Objects](#operation-object), [Response Objects](#response-object), [Reference Objects](#reference-object), etc.) based on the parsing context. Depending on how references are arranged, a given JSON or YAML object can be parsed in multiple different contexts:
 
-* As a complete OpenAPI Description document
+* As a syntactically complete OpenAPI Description document
 * As the Object type implied by its parent Object within the document
```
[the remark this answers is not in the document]
handrews: This is also not correct.  Part of the point of this section is to highlight that you can end up with a different implied Object type depending on whether you are looking at the parent _within the document structure_ or if you are looking at the _reference context_ (which could be considered the "parent" in the OAD depending on how you look at it).

In JSON Schema, the second concept is referred to as the "evaluation path", but we don't have that concept in the OAS and it seemed overly complex to introduce it as it's not quite the same usage.

handrews:
@mikekistler there are some very important distinctions between "document" and "description" that need to be maintained here.  All of this is really about precisely separating those concepts and highlighting the ambiguities that we've created by _not_ having made that separation clear in the past.

versions/3.0.4.md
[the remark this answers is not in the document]
mikekistler: I didn't know the right terminology so I just took a stab. What I was trying to get across with the document identifier is that refs may be internal and these should not be considered to impose a "multi-document" OAD. Happy to rephrase to get this across.

versions/3.0.4.md
mikekistler: Now I understand. I think you are saying that if there is no parent object, then the context is an OpenAPI description. Let me think about how this could be more clear.

versions/3.0.4.md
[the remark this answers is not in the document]
mikekistler: Okay ... let me try again.

versions/3.0.4.md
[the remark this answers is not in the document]
handrews: >  What I was trying to get across with the document identifier is that refs may be internal and these should not be considered to impose a "multi-document" OAD. Happy to rephrase to get this across.

I think this "multiple, connected documents" wording (with your added word "documents") gets this across sufficiently.  The listing of reference keywords is explicitly dependent on that ("In the latter case").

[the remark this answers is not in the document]
handrews: This sentence was an error in the previous comment:

> The listing of reference keywords is explicitly dependent on that ("In the latter case").

What I mean is that the wording "multipe, connected documents" is _not_ dependent on the next sentence about reference objecs and keywords.

It might be good to just separate out that sentence into a new paragraph noting that any of the listed objects/keywords are able to connect additional documents.  Or something like that?

versions/3.0.4.md
```diff
@@ -121,16 +121,18 @@ In order to preserve the ability to round-trip between YAML and JSON formats, YA
...
 
-Any document consisting entirely of an [OpenAPI Object](#openapi-object) is known as a **complete OpenAPI document**.
-In a multi-document description, the document containing the OpenAPI Object where parsing begins for a specific API's description is known as that API's **entry OpenAPI document**, or simply **entry document**.
+Any document consisting entirely of an [OpenAPI Object](#openapi-object) is known as a **syntactically complete OpenAPI document**.
+An OpenAPI document that does _not_ reference any other documents is known as a **self-contained OpenAPI document**.
+A single-document description is therefore _both_ syntactically complete _and_ self-contained.
+In a multi-document description, the document containing the OpenAPI Object where parsing begins for a specific API's description is known as that API's **entry document**.
 
 It is RECOMMENDED that the entry OpenAPI document be named: `openapi.json` or `openapi.yaml`.
```
ralfhandl, line 129:
```suggestion
It is RECOMMENDED that the entry document be named: `openapi.json` or `openapi.yaml`.
```
"entry OpenAPI document" is no longer defined

versions/3.0.4.md
```diff
@@ -121,16 +121,18 @@ In order to preserve the ability to round-trip between YAML and JSON formats, YA
 
 An OpenAPI Description (OAD) MAY be made up of a single document or be divided into multiple, connected parts at the discretion of the author. In the latter case, [Reference Object](#reference-object) and [Path Item Object](#path-item-object) `$ref` keywords, as well as the [Link Object](#link-object) `operationRef` keyword, are used.
 
-Any document consisting entirely of an [OpenAPI Object](#openapi-object) is known as a **complete OpenAPI document**.
-In a multi-document description, the document containing the OpenAPI Object where parsing begins for a specific API's description is known as that API's **entry OpenAPI document**, or simply **entry document**.
+Any document consisting entirely of an [OpenAPI Object](#openapi-object) is known as a **syntactically complete OpenAPI document**.
+An OpenAPI document that does _not_ reference any other documents is known as a **self-contained OpenAPI document**.
+A single-document description is therefore _both_ syntactically complete _and_ self-contained.
+In a multi-document description, the document containing the OpenAPI Object where parsing begins for a specific API's description is known as that API's **entry document**.
```
mikekistler, line 127: Here's a new suggestion for this part of the text. Let me know what you think of this (GitHub won't let me do this as a suggestion so I'll just quote it here:

> An OpenAPI Description (OAD) MAY describe certain features of the API "by reference", using the [Reference Object](#reference-object) or [Path Item Object](#path-item-object) `$ref` keywords, as well as the [Link Object](#link-object) `operationRef` keyword. When any of these references are used, the meaning of the OAD is equivalent to replacing the reference with description element at the referenced location.
> 
> Any of these references may specify content in a document other than the document containing the reference. Structuring an OAD as a single JSON or YAML document or composed from multiple, connected documents is at the discretion of the author. 
> 
> In a multi-document OAD, the document containing the OpenAPI Object where parsing begins for a specific API's description is known as that OAD's **entry document**.

ralfhandl: I'm stumbling over the sentence

> When any of these references are used, the meaning of the OAD is equivalent to replacing the reference with description element at the referenced location.

The "replacing" is only true for the Reference Object.

For Path Item Objects `$ref` means "merge the referenced object into this object and deal with conflicts as you like".

Link Objects are instructions on how to parameterize the referenced operation, the referenced operation does not replace the Link Object.

If this sentence is removed, I'd be ok with the suggestion.

mikekistler: Okay ... I didn't realize these were different. I'll remove that sentence.

handrews: @mikekistler I'm really reluctant to get into this much rewriting.  I was just trying to fix a tiny little thing on what I thought was already approved text.  We're on the verge of releasing this, is it truly necessary to revisit this in depth?

I don't agree with these further proposed changes and really do not want to delve more into referencing, or what "by reference" means, or whether references are replacements or not, etc.  in this section at all.

handrews:
@mikekistler @ralfhandl given how late we are in this cycle I'd really like to pull back and determine the minimum thing that is _absolutely necessary_ to change.

I encountered a problem where someone from  the workflows group read "complete document" and told other people that that meant it had no external references.  That is an understandable error to me, so I added a word ("syntactically") and a term ("self-contained") to clarify.

How much of the rest of this is similarly needed to avoid _egregiously wrong interpretations?_  If we had folks treating "complete document" as "no references", that would be a big problem as it's not what the text was intended to mean at all.  I feel like the rest of this is just trying to optimize language, which I would prefer to do in 3.2.

I'd really like to just revert this to the minimal necessary change, get 3.0.4 and 3.1.1 out the door, and delve more deeply into how we present this in 3.2.

ralfhandl (approved):

versions/3.0.4.md
mikekistler: Then perhaps we can just drop this PR altogether? Because honestly I think it only makes a muddy topic even muddier. 

But maybe it's just me ... and I suppose I should have spoken up when the earlier language was added. So let me see if I can just clear my review and let others decide.

handrews: @mikekistler as I explained in another reply further down, this was a tiny fix to a very specific problem:  the misunderstanding that someone who is very familiar with the spec had when they read "complete document" to mean "self-contained with no external references."

I don't understand how my insertion of the word "syntactically" and defintion of a "self-contained" document makes anythig more confusing.  The things you brought up are pretty much all outside of the scope of that tiny fix, so I'm not sure why you are insisting on either a lot more change or no change at all.

If you want to put up a PR with more changes, I have no objection to that, I just object to expanding the scope of my very specific fix for a concrete, observed problem that could really cause a lot of confusion, with a more general debate on the wording.

handrews:
After an DM chat with @mikekistler I am backing out the one suggestion from him that I had accepted (as it did not change that wording in every occurrence) and he will consider a separate PR for more of a rewrite.

This PR is again just about fixing the "complete document" misunderstanding.

ralfhandl (approved):

lornajane (approved):

PR #4078, opened 2024-09-09
Further clarify link operation ambiguity (3.0.4)

handrews:
There are even more ways this can go wrong!

### DIFF
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -2310,9 +2310,10 @@ For computing links and providing instructions to execute them, a [runtime expre
 This object MAY be extended with [Specification Extensions](#specification-extensions).
 
 A linked operation MUST be identified using either an `operationRef` or `operationId`.
-In the case of an `operationId`, it MUST be unique and resolved in the scope of the OpenAPI description.
-Because of the potential for name clashes, the `operationRef` syntax is preferred
-for multi-document OpenAPI descriptions.
+The identified or reference operation MUST be unique, and in the case of an `operationId`, it MUST be resolved within the scope of the OpenAPI Description (OAD).
+Because of the potential for name clashes, the `operationRef` syntax is preferred for multi-document OADs.
+However, because use of an operation depends on its URL path template in the [Paths Object](#paths-object), operations from any [Path Item Object](#path-item-object) that is referenced multiple times within the OAD cannot be resolved unambiguously.
+In such ambiguous cases, the resulting behavior is implementation-defined and MAY result in an error.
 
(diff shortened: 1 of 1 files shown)


versions/3.0.4.md
```diff
@@ -2310,9 +2310,10 @@ For computing links and providing instructions to execute them, a [runtime expre
 This object MAY be extended with [Specification Extensions](#specification-extensions).
 
 A linked operation MUST be identified using either an `operationRef` or `operationId`.
-In the case of an `operationId`, it MUST be unique and resolved in the scope of the OpenAPI description.
-Because of the potential for name clashes, the `operationRef` syntax is preferred
-for multi-document OpenAPI descriptions.
+The identified or reference operation MUST be unique, and in the case of an `operationId`, it MUST be resolved within the scope of the OpenAPI Description (OAD).
+Because of the potential for name clashes, the `operationRef` syntax is preferred for multi-document OADs.
+However, because use of an operation depends on its URL path template in the [Paths Object](#paths-object), operations from any [Path Item Object](#path-item-object) that is referenced multiple times within the OAD cannot be resolved unambiguously.
```
ralfhandl, line 2315: Sounds like: reusable Path Item Objects were a bad idea, use them with extreme care 😎

ralfhandl (approved):

versions/3.0.4.md
mikekistler: I'm not entirely sure I follow the problem here. I think the original language in the spec was intending to mean that the referenced operation is "well-defined" -- that it could not resolve to potentially multiple operations. Clearly this could happen if the reference is by operationId, particularly when the operations are in different documents, though I think this case could be considered as invalid since the spec requires that operationIds be unique.

The change here suggests that an operation ref could potential resolve to multiple operations, and we want to disallow this as well. But I'm struggling to understand, even with the last sentence that was added, how this can actually occur. How could a reference possibly resolve to multiple operations?

mikekistler: Okay ... I get it now. The problem is essentially that an "operation" is ambiguous / incomplete without the path template, and the operationRef may point to an operation in a Path Item that does not have a parent path template. Could we fix this by requiring the operationRef to include a path template segment?

ralfhandl: That would work.

handrews: @mikekistler we can't fix it in the patch release, hence the warning here.  We should probably open an issue for it in 3.2, though.

ralfhandl: Created 
* #4084

miqui (approved):

PR #4081, opened 2024-09-11
Minor param serialization and wording fixes

handrews:
A few last (_really_!  I hope...) changes from feedback on Slack today and a few things spotted while compiling release notes.  **If anything here is controversial beyond a slight back-and-forth, I'd rather drop it than bog down the patch releases at this stage.**

* Directly address `style` delimiters in parameter values by explaining that we don't define how to handle them. _(NOTE:  I had language about this in here at some point and it got dropped somewhere along the line- this uses an example from @zlondrej who brought it up on Slack today)_
* Explicitly set `explode: false` in an example as the default with `style: form` is `explode: true`; the `explode: true` example was also left explicit to reduce confusion. _(also caught by @zlondrej)_
* Tidy up overly conversational (e.g. "our document") language that I'd meant to revisit but forgot about. _(@jeremyfiel commented on this but I didn't want to wordsmith at the time, and then I forgot where it was!)_
* Include the Header Object as one of the places where the `style` keyword is used (even if it is the simplest case)
* Fix a missing space before an RFC reference.
* Minor grammar fix.

### DIFF
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -3789,7 +3789,7 @@ security:
 
 See [Resolving Implicit Connections](#resolving-implicit-connections) for more information.
 
-First, our [entry document](#openapi-description-structure) is where parsing begins. It defines the `MySecurity` security scheme to be JWT-based, and it defines a Path Item as a reference to a component in another document:
+First, the [entry document](#openapi-description-structure) is where parsing begins. It defines the `MySecurity` security scheme to be JWT-based, and it defines a Path Item as a reference to a component in another document:
 
 ```HTTP
 GET /api/description/openapi HTTP/1.1
@@ -3832,7 +3832,7 @@ paths:
     $ref: 'other#/components/pathItems/Foo'
 ```
 
-Next, we have our referenced document, `other`. The fact that we don't use file extensions gives the client the flexibility to choose an acceptable format on a resource-by-resource basis, assuming both representations are available:
+This entry document references another document, `other`, without using a file extension.  This gives the client the flexibility to choose an acceptable format on a resource-by-resource basis, assuming both representations are available:
 
 ```HTTP
 GET /api/description/other HTTP/1.1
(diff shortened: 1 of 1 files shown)


jeremyfiel (approved):

versions/3.0.4.md
```diff
@@ -1127,6 +1127,10 @@ In order to support common ways of serializing simple parameters, a set of `styl
 
 See [Appendix E](#appendix-e-percent-encoding-and-form-media-types) for a discussion of percent-encoding, including when delimiters need to be percent-encoded and options for handling collisions with percent-encoded data.
 
+Note that parameter data can contain characters used as delimiters by certain `style` and `explode` configurations.
+For example, when using `style: "simple"`, the arrays `["monospace", "serif", "Arial,Helvetica,sans-serif"]` and `["monospace", "serif", "Arial", "Helvetica", "sans-serif"]` would both be encoded as `"monospace,serif,Arial,Helvetica,sans-serif"`.
```
ralfhandl, line 1131: Style "simple" explicitly references [RFC6570, section 3.2.2](https://datatracker.ietf.org/doc/html/rfc6570#section-3.2.2) which has an example for encoding commas in parameter values as `%2C`:

       {keys}             semi,%3B,dot,.,comma,%2C
       {keys*}            semi=%3B,dot=.,comma=%2C

This example and the explicitly referenced [section 3.2.1](https://datatracker.ietf.org/doc/html/rfc6570#section-3.2.1) imply that this example is encoded as `monospace,serif,Arial%2CHelvetica%2Csans-serif`.

There is no need and actually no lee-way for API providers to document something different.

I think this paragraph and the next should be removed from this PR.

We could (and probably should) extend the [style examples](https://github.com/OAI/OpenAPI-Specification/blob/v3.1.1-dev/versions/3.1.1.md#style-examples) and change one of the color values to contain a comma, for example `"black,shiny"`.

ralfhandl (changes requested):

versions/3.0.4.md
handrews: @ralfhandl oh good catch!  I'm going to take this part ouot of this PR and go read up and see what makes sense to do.  I'll submit another PR assuming we should still say something.

handrews:
@ralfhandl ready for re-review

ralfhandl (approved):

ralfhandl:
@jeremyfiel Are you still fine with this PR?

It is now a subset of what you approved, and @handrews will follow up on the removed part.

jeremyfiel:
Good catch, Ralf. I didn't recognize that either. 

I'm good with the rest of the pr.

ralfhandl (approved):

lornajane (approved):

PR #4091, opened 2024-09-12
3.1.1 port of Move complex Sec Req example to appendix F

handrews:
The extended example with a multi-document OAD and a Security Requirement in the referenced document
did to fit well where it was, and there wasn't room in the Resolving Implicit Connections area.  So this moves it to an Appendix linked from both locations.


ralfhandl (approved):

mikekistler (approved):
Looks good! 👍

PR #4094, opened 2024-09-13
3.2.0: back out changes to Path Item Object and bring its `$ref` field back

### LINKED ISSUE #3734: Did we break 3.2 compatibility by removing the Path Item $ref special case?

ralfhandl:
Fixes #3734 by
- adding `$ref` field back to Path Item Object
- removing Reference Object as an alternative to Path Item Object in four places

Builds on and includes
* #4066

### DIFF
--- a/versions/3.2.0.md
+++ b/versions/3.2.0.md
@@ -28,6 +28,10 @@ A "schema" is a formal description of syntax and structure.
 This document serves as the [schema](#schema) for the OpenAPI Specification format; a non-authoritative JSON Schema based on this document is also provided on [spec.openapis.org](https://spec.openapis.org) for informational purposes.
 This specification also _uses_ schemas in the form of the [Schema Object](#schema-object).
 
+### Object
+
+When capitalized, the word "Object" refers to any of the Objects that are named by section headings in this document.
+
 ### Path Templating
 
 Path templating refers to the usage of template expressions, delimited by curly braces (`{}`), to mark a section of a URL path as replaceable using path parameters.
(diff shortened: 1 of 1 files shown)


mikekistler (approved):
Looks good. 👍

handrews:
@ralfhandl I'll look at this once the "built on" stack is resolved :-)

ralfhandl:
> I'll look at this once the "built on" stack is resolved

In case you are curious, these are the changes: https://github.com/ralfhandl/OpenAPI-Specification/compare/3.2.0-sync-with-3.1.1...3.2.0-back-out-changes-to-path-item-object

- bring back `$ref` field in Path Item Object
- remove Reference Object as an alternative to Path Item Object in four places

handrews:
@ralfhandl if we follow the proposal in https://github.com/OAI/OpenAPI-Specification/issues/3677#issuecomment-2427757769, we will not need this PR as we will just abandon `v3.2.0-dev` in favor of a reconstructed `v3.2-dev`, where the reconstruction just omits the commits that made this change that we want to back out.

ralfhandl:
@handrews Thanks for taking over!

PR #4097, opened 2024-09-15
Update release process notes

lornajane:
The notes about a Liason and releasing patches in a certain week of the month are outdated, so they are removed from the DEVELOPMENT file.


ralfhandl (approved):

ralfhandl:
Do we need the Liason's approval to this change?

handrews (approved):

handrews:
@ralfhandl not if we can't even remember who the Liason was... and definitely not if it's the Linux Foundation liason/pm as we don't have one at the moment.

PR #4107, opened 2024-09-23
ReSpec build script: no automatic syntax highlighting for unspecified or "unknown" languages

ralfhandl:
We intentionally format code blocks with `highlight.js` if the explicitly specified language of the code block is known to this package.

Formerly we left highlighting of unknown languages or code blocks with unspecified language to ReSpec, which now leads to erratic results, see #4103.

This PR suppresses automatic syntax highlighting by ReSpec for code blocks whose language is not specified or unknown to `highlight.js`.

It also adds two custom languages `uri` and `uritemplate` that are used in
* #4110 
* #4109 

Preview:
* https://ralfhandl.github.io/OpenAPI-Specification/oas/v3.1.1.html#rfc6570-equivalent-expansion

image

### DIFF
--- a/scripts/md2html/md2html.js
+++ b/scripts/md2html/md2html.js
@@ -36,7 +36,7 @@ const md = require('markdown-it')({
       '</code></pre>';
     }
 
-    return '<pre class="highlight '+lang+'" tabindex="0"><code>' + md.utils.escapeHtml(str) + '</code></pre>';
+    return '<pre class="nohighlight" tabindex="0"><code>' + md.utils.escapeHtml(str) + '</code></pre>';
   }
 });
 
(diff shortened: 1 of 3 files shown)


handrews:
Can we update highlight.js to get more languages?  I really wish urlencoded highlighting were supported (I believe it's possible).

ralfhandl:
> Can we update highlight.js to get more languages?

We are already using the latest version of highlight.js, and we can either try to find a [supported language](https://github.com/highlightjs/highlight.js/blob/main/SUPPORTED_LANGUAGES.md) that looks nice for URL templates and query parts, or register a custom language for it:

image

Both approaches would require to differentiate between `uritemplate` and `urlencoded`, otherwise URI templates will look funny:

image

handrews:
At least some packages have plugins for urlencoded. Not sure I’ve seen uritemplates but yes def different format!

PR #4113, opened 2024-09-25
Add in new full-TSC members

earth2marsh:
Following the provisional period, welcome to the new maintainers! Thank you for all the work you've put in until now and for the progress yet to come. 🙏

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -5,8 +5,6 @@
 * Jeremy Whitlock [@whitlockjc](https://github.com/whitlockjc)
 * Marsh Gardiner [@earth2marsh](https://github.com/earth2marsh)
 * Ron Ratovsky [@webron](https://github.com/webron)
-
-## Provisional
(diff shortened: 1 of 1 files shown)


ralfhandl (commented):
Do we want to sort by first name?

earth2marsh:
> Do we want to sort by first name?

Hmmm, fair question! I think there are two general implicit principles that could be made explicit. 

1. The emeritus section has had names added (prepended) at the top of the list as members move out of active status (a sort of trickling down from above).
2. The active section has had new members appended. When a cohort contains more than one member, it is sorted by first name.

Of course that's just the organic way it has evolved! Given that the active list is going to be a fairly small number, the value of sorting it alphabetically by first name might not be that critical? But if it makes more sense to people viewing it, I'm not opposed!

ralfhandl (approved):

mikekistler (approved):
Looks good! 👍

PR #4100, opened 2024-09-17
Fix description vs document terminology

mikekistler:
This PR attempts to fully distinguish "description" and "document" by using "description" consistently for the concept of an API description in the OpenAPI format and "document" to only refer to structural features.

I also revised the "OpenAPI Description Structure" section a bit in part to address the above and also in an attempt to simplify and clarify that section.

Some specifics:
- Every instance of "OpenAPI document" and "OpenAPI description document" has been eliminated. Most are changed to "OpenAPI Description".
- All occurrences of "OpenAPI Description" capitalize the "D" in "Description".
- I changed some instances of "OpenAPI Description" to "OAD", but there are still 33 occurrences of "OpenAPI Description" and at least some of these might be good candidates to change to "OAD".
- the previous text defined "self-contained OpenAPI document" -- I changed that "single-document OpenAPI Description" which I think is clearer and simpler.
- the previous text defined "syntactically complete OpenAPI document" -- I removed that as I believe that other wording changes I made made it unnecessary.

I think there are some good changes here but I'm definitely open to feedback / suggestions on how make it even better.

PR #4130, opened 2024-10-10
Note relative URL-reference resolution ambiguity

handrews:
This explicitly lists out the cases and notes which ones were (as far as we can recall) intended to behave like `$ref`, but could be reasonably read to behave like API URLs.

See:
* https://github.com/OAI/OpenAPI-Specification/pull/4100#discussion_r1794569584

### DIFF
--- a/versions/3.0.4.md
+++ b/versions/3.0.4.md
@@ -232,9 +232,14 @@ OpenAPI Description authors SHOULD consider how text using such extensions will 
 ### Relative References in URLs
 
 Unless specified otherwise, all fields that are URLs MAY be relative references as defined by [RFC3986](https://tools.ietf.org/html/rfc3986#section-4.2).
+
 Relative references are resolved using the URLs defined in the [Server Object](#server-object) as a Base URI.
 
 Relative references used in `$ref` are processed as per [JSON Reference](https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03), using the URL of the current document as the base URI. See also the [Reference Object](#reference-object).
+
+The wording above is intended to distinguish between API URLs (which use the Server Object to determine the base URI or the prefix for the URL path template), and OAD URLs (which use the current document's base URI per RFC3986).
+However, only `$ref` was mentioned explicitly in versions 3.0.0 through 3.0.3 of this specification, leaving the behavior of the following fields ambiguous: the `operationRef` field of the [Link Object](#link-object), the URI form of the `mapping` field of the [Discriminator Object](#discriminator-object), the `externalValue` field of the [Example Object](#example-object), and the `url` fields of the [External Documentation](#external-documentation-object), [Contact](#contact-object), and [License](#license-object) Objects.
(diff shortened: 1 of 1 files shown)


ralfhandl (dismissed):

versions/3.0.4.md
```diff
@@ -232,10 +232,15 @@ OpenAPI Description authors SHOULD consider how text using such extensions will
...
 
+The wording above is intended to distinguish between API URLs (which use the Server Object to determine the base URI or the prefix for the URL path template), and OAD URLs (which use the current document's base URI per RFC3986).
+However, only `$ref` was mentioned explicitly in versions 3.0.0 through 3.0.3 of this specification, leaving the behavior of the following fields ambiguous: the `operationRef` field of the [Link Object](#link-object), the URI form of the `mapping` field of the [Discriminator Object](#discriminator-object), the `externalValue` field of the [Example Object](#example-object), and the `url` fields of the [External Documentation](#external-documentation-object), [Contact](#contact-object), and [License](#license-object) Objects.
+Therefore, which of the above two rules is used for each of these fields is _implementation-defined_, although it is RECOMMENDED to use the `$ref` process for all of them for compatibility with future versions of this specification.
+
```
mikekistler, line 243: How about this slightly more concise wording:
```suggestion
Resolution of relative references in the `operationRef` field of the [Link Object](#link-object), the URI form of the `mapping` field of the [Discriminator Object](#discriminator-object), the `externalValue` field of the [Example Object](#example-object), and the `url` fields of the [External Documentation](#external-documentation-object), [Contact](#contact-object), and [License](#license-object) Objects is _implementation-defined_. For compatibility with future versions of this specification, It is RECOMMENDED to use the same process as $ref, but MAY instead use the URLs defined in the [Server Object](#server-object) as a Base URI, as either interpretation was possible in previous versions of this specification.
```

mikekistler (approved):
Looks good! 👍 

I think the wording as is conveys the key information, but I left an alternate wording that I think might read better in the spec. I'll leave it up to you if you prefer the original text.

versions/3.0.4.md
handrews: @mikekistler I like most of the wording change but want to tweak it a bit further- will update shortly.

handrews:
@mikekistler see how you like this latest change.  I kept most of your deletions, but decided to entirely drop mention of previous drafts, and kept wording that highlights two important points:

* It is only implementation-defined _which of the two alternatives_ is used; it is not implementation-defined without constraint
* The each / all wording indicates that tools don't have to choose the same process for every keyword in the list.  For example, @darrelmiller (in our conversation after the TDC call) drew the line in a different place, which would put some in the list using Server Objects and some using the `$ref` process, and we don't want to invalidate that either

ralfhandl (approved):

mikekistler (approved):
Looks good! 👍

handrews:
Two TSC approvals, so I'm hitting the merge button.

Issue #4125, opened 2024-10-03
schema updates for 3.0.4


karenetheridge:
The 3.0.4 schemas need a review pass for accuracy.

- [ ] update `$comment`, `description` and `$id` URIs for v3.0.3->v3.0.4 (where can an html preview of the specification be viewed?)
  - will be available at https://spec.openapis.org/oas/v3.0.4.html
- [ ] ensure links in specification (referencing schema files) are correct
  - 3.0.4 and 3.1.1 will link to https://spec.openapis.org/ which now has a subsection "Non-Normative JSON Schemas" with "view" and "download" links
- [ ] compare specification and schema diffs from 3.0.3 and 3.0.4, thinking about any spec changes that haven't yet been reflected in the schemas, and what changes have been made to the schemas already and confirm they are accurate (this SHOULD be a no-op, but mistakes have occasionally happened before ;) )
- [x] ensure the .json and .yaml files are consistent
- [x] verify validity against the draft4 json metaschema  
  (I am converting locally to draft7 so as to allow: `apt-get libjson-schema-modern-perl; json-schema-eval --strict --validate-schema schemas/v3.0/schema.json`)

handrews:
There are definitely improvements to be made to the 3.0 schemas, but there aren't any schema-impacting differences between 3.0.3 and 3.0.4.  The only version-ish thing that should change is the date in the `$id` of the schemas.

AFAIK, this is true across 3.0.0 - 3.0.4, and any schema applying to one applies to all and should only mention 3.0 in comments or other text (no patch release number).  Some might make an argument regarding `nullable` in 3.0.0 – 3.0.2 va 3.0.3 – 3.0.4.  However, the official position is that the clarifications to `nullable` in 3.0.3 were only clarifications (it made the relationship between `nullable` and `type` behave the same way as the relationship between draft-04 `exclusiveMinimum` and `minimum` (and `exclusiveMaximum` and `maximum`).

[...]

Issue #3715, opened 2024-04-18
Define and document our schema publishing process


handrews:
We need to figure this out, so I'm extracting the recent/relevant comments from #151 as it is so old and long that potential contributors are discouraged from figuring out where they might help.

In addition to what's discussed below, we might want to think about how people might find different schema versions readily visible in GitHub on `main` that differ from the `gh-pages`-published ones.  This will be less of a concern if schema work moves to release branches as is being considered in #3677 and #3716.  Separately, see also #2400 for discussion of publishing the schemas through NPM; this issue is just for publishing them ourselves.

-----

From @handrews (https://github.com/OAI/OpenAPI-Specification/issues/151#issuecomment-1556268486)

I think we agreed this should be the issue tracking how/when we deploy schema updates.  Including deploying new revisions of the OAS 3.1 JSON Schema vocabulary and dialect  meta-schemas which don't currently have dates in their `$id`s.  ~~Which I think happened because I wrote them as tentative proposals and then by the time OAS 3.1 shipped I wasn't around to catch that sort of thing.~~

-----

From @jdesrosiers (https://github.com/OAI/OpenAPI-Specification/issues/151#issuecomment-1561690640)

I've been handling schema deployment. I don't usually prepare a deployment for every change, but tend to wait a little while in case more changes come in (they usually come in waves). There are currently two changes that have been merged but not deployed yet and one open PR.

Here's the process I've been following for deployment
1. Update release dates in
    * schema.yaml
    * schema-base.yaml (Don't forget the one in the `$ref`)
    * tests/v3.1/test.js
2. Generate the JSON versions from the YAML versions
    * The README says to only modify the YAML version, but contributors usually include changes to both. So, this step is usually just a sanity check.
3. Run the tests
    * The tests are very minimal and haven't been added to since they were first added, but I make sure that we at least don't break what we have.
3. Deploy the schemas
    * Add the schemas to the `gh-pages` branch
    * Update the `latest` symlink to point to the newest version
4. Other stuff I'm forgetting about ???

Here's an example of a deploy I've done in the past
* https://github.com/OAI/OpenAPI-Specification/pull/3042
* https://github.com/OAI/OpenAPI-Specification/pull/3043

[...]

PR #4132, opened 2024-10-10
v3.0.4 schema updates

karenetheridge:
Brings the schemas up to date for the v3.0.4 release (see #4125), and adds some support scripts which are used to check for validation errors and inconsistencies between the yaml and json versions of the files.

```
$ json-schema-eval --validate-schema schemas/v3.0/schema.json
{
  "valid": true
}

$ scripts/compare schemas/v3.0/schema
schemas/v3.0/schema.yaml and schemas/v3.0/schema.json are identical.
```

### DIFF
--- a/schemas/v3.0/schema.yaml
+++ b/schemas/v3.0/schema.yaml
@@ -43,6 +43,7 @@ definitions:
       '^\$ref$':
         type: string
         format: uri-reference
+
   Info:
     type: object
     required:
@@ -65,7 +66,6 @@ definitions:
     patternProperties:
       '^x-': {}
     additionalProperties: false
-
 
   Contact:
     type: object
@@ -592,7 +592,6 @@ definitions:
       '^x-': {}
     minProperties: 1
     additionalProperties: false
-
 
   SecurityRequirement:
     type: object
--- a/schemas/v3.0/schema.yaml
+++ b/schemas/v3.0/schema.yaml
@@ -1,4 +1,4 @@
-id: https://spec.openapis.org/oas/3.0/schema/2021-09-28
+id: https://spec.openapis.org/oas/3.0/schema/2024-10-10
 $schema: http://json-schema.org/draft-04/schema#
 description: The description of OpenAPI v3.0.x documents, as defined by https://spec.openapis.org/oas/v3.0.4
 type: object
--- a/schemas/v3.0/schema.yaml
+++ b/schemas/v3.0/schema.yaml
@@ -1,6 +1,6 @@
 id: https://spec.openapis.org/oas/3.0/schema/2021-09-28
 $schema: http://json-schema.org/draft-04/schema#
-description: The description of OpenAPI v3.0.x documents, as defined by https://spec.openapis.org/oas/v3.0.3
+description: The description of OpenAPI v3.0.x documents, as defined by https://spec.openapis.org/oas/v3.0.4
 type: object
 required:
   - openapi
--- a/scripts/compare
+++ b/scripts/compare
@@ -1,0 +1,40 @@
+#!/usr/bin/env perl
+# vim: set ts=8 sts=2 sw=2 tw=100 et :
+
+# given a filename base, compare the .yaml and .json versions of the file for differences.
+# to run:
+#   apt-get libjson-schema-modern-perl # (or equivalent on your distribution)
+#   scripts/compare schemas/v3.0/schema
+#   scripts/compare schemas/v3.1/schema
+#   scripts/compare schemas/v3.1/schema-base
+
+# Author: Karen Etheridge - karen@etheridge.ca, ether@cpan.org
+# see also https://github.com/karenetheridge/JSON-Schema-Modern
+
+use 5.020;
+use warnings;
+use Path::Tiny;
+use YAML::PP;
+use JSON::Schema::Modern ();
+use JSON::Schema::Modern::Utilities 0.591 'is_equal';
+
+my $base = shift; # the filename, without .yaml or .json extension
+
+my $yaml_decoder = YAML::PP->new(boolean => 'JSON::PP');
+my $json_decoder = JSON::Schema::Modern::_JSON_BACKEND()->new->utf8(1);
+
+my $yaml_path = path($base.'.yaml');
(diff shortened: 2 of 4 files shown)


scripts/yaml2json/yaml2json.pl
ralfhandl, line 1: Why do we need this? We already have https://github.com/OAI/OpenAPI-Specification/blob/main/scripts/yaml2json/yaml2json.js

scripts/compare
ralfhandl, line 1: Why do we need this?

If we are to accept
* #3814

or build something similar, the JSON variant becomes a downstream artifact that is always regenerated from the YAML variant.

schemas/v3.0/schema.yaml
```diff
@@ -1,6 +1,6 @@
-id: https://spec.openapis.org/oas/3.0/schema/2021-09-28
+id: https://spec.openapis.org/oas/3.0/schema/2024-10-10
 $schema: http://json-schema.org/draft-04/schema#
-description: The description of OpenAPI v3.0.x documents, as defined by https://spec.openapis.org/oas/v3.0.3
```
handrews, line 3: @karenetheridge I realize you are not the person who put the "as defined by ...v3.0.N" here, but I'd really like to drop it. This is for 3.0. All of 3.0.

schemas/v3.0/schema.yaml
```diff
@@ -1,6 +1,6 @@
-id: https://spec.openapis.org/oas/3.0/schema/2021-09-28
```
handrews, line 1: I think part of the problem with out publishing process is has to do with this date.  I'm fine with manually setting it when we push it out, but I think there was some expectation to set it automatically.  Perhaps documented in #3715

handrews (changes requested):
I agree with @ralfhandl 's questions about why we should re-do the format synchronization scripts.  Also, as much as I dislike JavaScript, we should have one tooling language if at all possible.

I'd be happy if we changed it, but let's either plan to change it all or plan to keep it, or at least have a clear division of labor (e.g. GitHub Pages requires Ruby to build so we can't get rid of it, but we don't use it otherwise).

On a related note, why do we even check in the JSON version here on `main`?  It's confusing.  Why don't we just generate it for publication in the `gh-pages` branch?

scripts/compare
karenetheridge: I edited the json files by hand in order to preserve the same property order as was in the yaml files. Autogenerating the json from the yaml would lose that, and cause a much larger diff.

Since the file was almost identical already (and just lacking a few of the recent changes to the yaml file) it was easier to add those by hand and the diff was smaller.

scripts/yaml2json/yaml2json.pl
karenetheridge: We don't "need" it. I wrote this script for my own use as I am not a javascript person. I simply thought it would be helpful to commit it.

karenetheridge:
> Why don't we just generate it [the JSON file] for publication in the gh-pages branch?

I don't object to that if it wouldn't inconvenience contributors or tool vendors. It's not difficult to convert the yaml file in main (or a topic branch) to json on the fly, but the barrier is not zero.

schemas/v3.0/schema.yaml
karenetheridge: The file in git should still have an `id`/`$id` of some kind though.. perhaps something with "in_progress" or "latest" or something else non-dateish to indicate that it's not reflecting an officially-published version. Otherwise, it's yet one more barrier to tool vendors (on top of converting yaml to json) to make use of the interim file for development and testing.

handrews: @karenetheridge I'd lean towards something like "in_progress" that discourages casual use.

There really shouldn't be interim files if our publication process is running smoothly.  It might be different for pre-publication releases, but we should handle that separately as that's not a process we've defined at all yet.

In theory, every time we change a published schema, we just re-publish it immediately.  If we ever have a reason _not_ to do that, then we probably also don't want people to use the schemas out of `main`.  We really only want the published ones to be usable, I think (again, excluding work-in-progress versions where we just haven't defined a process).

I do agree that if we _do_ want people to use things directly from `main`, then they need `$id`/`id` values that are somehow reasonable, and not identical to any published version.

handrews:
@karenetheridge we already have a YAML-to-JSON script that was run every time we published, as changes were (and are) _only_committed directly to the YAML version.  So this isn't a new thing.  The only baffling thing to me is why we ever checked the JSON in to `main`.  We have, AFAICT, never used it for anything as it was regenerated on each publication.  All it does is confuse people into thinking they need to keep it in sync on `main`.

scripts/compare
handrews: There are definitely order-preserving translators.  I believe current versions of JavaScript maintain object property order.  If the JSON is in a different order from the YAML, we should probably fix that.

schemas/v3.0/schema.yaml
ralfhandl: The widely used [@apidevtools/openapi-schemas](https://www.npmjs.com/package/@apidevtools/openapi-schemas) package (1.7 million weekly downloads) seems to pull the OpenAPI schemas directly from `main`:

> This package contains **[the official JSON Schemas](https://github.com/OAI/OpenAPI-Specification/tree/master/schemas)** for every version of Swagger/OpenAPI Specification

And there may be other uses out there we don't know of.

Better have up-to-date YAML and JSON files with the most current identifier in the `main` branch.

ralfhandl: Well, then let's replace the date with "in-progress" and see what gets published to the NPM package registry 😎

handrews: @ralfhandl I wouldn't object to trying to coordinate with them a bit.  This is part of my (myriad) frustration with the schemas- we don't exactly make it easy for people who want to do something like package and distribute to do the right thing.  It was only a couple of months ago that I even realized that "published" meant "put on the GitHub pages site" since we didn't even link to them on the index page of that site.   It's far easier to find the "wrong" files than the "right" ones.

ralfhandl:
Converted to draft as it needs to be adjusted after merging
* #4146

PR #4133, opened 2024-10-11
gh-pages: no download link for "latest" schema

ralfhandl:
[This comment](https://github.com/OAI/OpenAPI-Specification/issues/4125#issuecomment-2392563767) suggests that

> we should not have a "latest" schema symlink because refining the constraints to be more accurate could cause automated systems to suddenly invalidate previously-valid OADs which is not a nice thing to do

The wrapper markdown page for the "view" links now has the same name as the schema file it wraps and the `latest` symlinks have been removed.

> [!NOTE]  
> The new publishing process MUST _rename_ the wrapper markdown pages for date-named schema files to match the most recent schema file.

Now:
image

With this PR:
image

Preview: https://ralfhandl.github.io/OpenAPI-Specification/


handrews:
@ralfhandl I'm a bit confused as the preview still has lots of "latest" links that go to `latest.html` files?

ralfhandl:
Updated the PR description:
* The "view latest" link remains, it leads to an HTML page with a syntax-colored code block
* The download link named "latest" is removed, and only the download links for the date-named files remain

handrews (dismissed):
@ralfhandl thanks- I'm approving this because it does what it says it does, and that's definitely an improvement! :-)

I do ask why we're using a "latest" link even for the HTML pages rather than just using the latest date as that link.  I feel like having a "latest" HTML link will encourage people to look for a "latest" schema, while a design like:

* **view latest schema: [2022-10-07](https://spec.openapis.org/oas/3.1/schema/latest.html)**

would set the users' expectations more accurately through the use of a date as the link name plus a filename of `schema/2022-10-07.html` (or whatever is appropriate for including the date).  But if there's interest in discussing this further, we can do that in another issue if you'd like.

ralfhandl:
@handrews next step: removed word "latest" from the view link

mikekistler (approved):
Looks good. 👍

handrews (approved):
Thanks @ralfhandl !

PR #4137, opened 2024-10-15
Run tests with YAML metaschemas

ralfhandl:
The schema-test workflow was using the JSON metaschemas, whereas the README.md files claim that the YAML files are the source of truth.

- [x] Use YAML metaschemas in schema tests
- [x] Use YAML metaschemas in `validate.mjs`
- [x] Remove out-of-sync JSON files
- [x] Update README.md files


.github/workflows/schema-tests.yaml
```diff
@@ -4,14 +4,15 @@ name: schema-test
...
 #
-# This workflow runs the npm test script to validate passing and failing
-# testcases for the metaschema.
+# This workflow
+# - converts the YAML metaschemas to JSON
+# - runs the npm test script to validate passing and failing testcases for the metaschemas
+# - checks in changed JSON files
 #
 
 # run this on push to any branch and creation of pull-requests
```
mikekistler, line 13: Should this comment change too?
```suggestion
# run this on push to any branch
```

mikekistler (approved):
Looks good. 👍 

One minor comment.

handrews (changes requested):
I'd still prefer to _not_ have the JSON on `main` at all.  I think the problem of "these schemas look published but they are not" is the bigger problem here, and I don't think we should invest in the current arrangement.

Per #3677, I think we should focus on setting up schema development correctly on the new `dev` branch (and release-line-specific branches off of it), which is where all work-in-progress will live.

If there's an argument that this is a necessary step, I'm open to it, but this feels like going backwards to me by trying to bless `main` as an alternate publication location.

ralfhandl:
> I'd still prefer to _not_ have the JSON on `main` at all

The JSON is used by the schema tests on `main`. We can
- Generate the JSON on the fly the YAML before running the tests, or
- Change the tests to read the YAML instead

In addition we can remove the JSON from `main`.

handrews (approved):

PR #4146, opened 2024-10-18
schemas: add WORK-IN-PROGRESS placeholders, don't mention patch numbers, update terminology

handrews:
I'd like to merge this and address it in the build scripts rather than resuming publication of schemas with absent dates and misleading patch release guidance.

* Replace all dates with WORK-IN-PROGRESS
* Append a date segement (as WORK-IN-PROGRESS) where missing
* Adjust refs and other fields to WORK-IN-PROGRESS
* Remove mention of specific patch releases
* Use capital-D Document and Description according to new rules


schemas/v3.1/schema.yaml
```diff
@@ -13,7 +13,7 @@ properties:
   jsonSchemaDialect:
     type: string
     format: uri
-    default: 'https://spec.openapis.org/oas/3.1/dialect/base'
+    default: 'https://spec.openapis.org/oas/3.1/dialect/base/WORK-IN-PROGRESS'
```
ralfhandl, line 16: Do we really need this change?

It will complicate the build script, so what is the return on investment?

schemas/v3.1/schema.yaml
```diff
@@ -1,7 +1,7 @@
-$id: 'https://spec.openapis.org/oas/3.1/schema/2022-10-07'
+$id: 'https://spec.openapis.org/oas/3.1/schema/WORK-IN-PROGRESS'
 $schema: 'https://json-schema.org/draft/2020-12/schema'
 
-description: The description of OpenAPI v3.1.x documents without schema validation, as defined by https://spec.openapis.org/oas/v3.1.0
+description: The description of OpenAPI v3.1.x Documents without Schema Object validation
```
ralfhandl, line 4: "without Schema Object Validation": should we rather use `schema-base.yaml` in our schema tests?

schemas/v3.1/dialect/base.schema.yaml
```diff
@@ -1,8 +1,8 @@
-$id: https://spec.openapis.org/oas/3.1/dialect/base
+$id: https://spec.openapis.org/oas/3.1/dialect/base/WORK-IN-PROGRESS
```
ralfhandl, line 1: We currently don't publish or reference date-stamped iterations for this file. 

Do we want to change this?

schemas/v3.1/meta/base.schema.yaml
```diff
@@ -1,7 +1,7 @@
-$id: https://spec.openapis.org/oas/3.1/meta/base
+$id: https://spec.openapis.org/oas/3.1/meta/base/WORK-IN-PROGRESS
```
ralfhandl, line 1: We currently don't publish or reference date-stamped iterations for this file. 

Do we want to change this?

schemas/v3.1/schema-base.yaml
```diff
@@ -1,20 +1,20 @@
...
-$ref: 'https://spec.openapis.org/oas/3.1/schema/2022-10-07'
+$ref: 'https://spec.openapis.org/oas/3.1/schema/WORK-IN-PROGRESS'
 properties:
   jsonSchemaDialect:
     $ref: '#/$defs/dialect'
 
 $defs:
   dialect:
-    const: 'https://spec.openapis.org/oas/3.1/dialect/base'
+    const: 'https://spec.openapis.org/oas/3.1/dialect/base/WORK-IN-PROGRESS'
```
ralfhandl, line 13: Is this necessary? See comment on the referenced file.

schemas/v3.1/schema-base.yaml
```diff
@@ -1,20 +1,20 @@
...
 
   schema:
     $dynamicAnchor: meta
-    $ref: 'https://spec.openapis.org/oas/3.1/dialect/base'
+    $ref: 'https://spec.openapis.org/oas/3.1/dialect/base/WORK-IN-PROGRESS'
```
ralfhandl, line 17: Is this necessary? See comment on the referenced file.

ralfhandl (dismissed):

schemas/v3.1/schema.yaml
handrews: @ralfhandl I'm not sure I understand the question.  Each document can be iterated independently.  These needed to have dates from the beginning.

schemas/v3.1/dialect/base.schema.yaml
handrews: @ralfhandl How else would we handle changes to this document?  Changing the valid set of a schema without changing it's `$id` is a problem.

schemas/v3.1/meta/base.schema.yaml
handrews: @ralfhandl yes, because it never should have been this way in the first place.

schemas/v3.1/schema-base.yaml
handrews: Yes.

schemas/v3.1/schema-base.yaml
handrews: Yes.

handrews:
@ralfhandl As I noted in my comments on #4139, I am aware that this is more complex.  I also noted there (and apparently forgot to note here, for which I apologize- if you read this first it was no doubt confusing) that it's probably fine to just change all of the dates at once for now.  Certainly the first time we publish things we'll have changes to `schema` and need to set initial dates on `dialect/base` and `meta/base`.  We'll also need to update `base-schema` to point to the new `schema` every time _any_ of the other three are iterated.

So for now, a global search-and-replace is fine.  We can then refine it.  The rules are fairly straightforward:

* Changes to anything require a change in `base-schema`
* Changes to `meta-base` require a change to `dialect-base`

I think this will be easier to manage once we fix the branching.  We need a `schema-dev` branch analogous to the `dev` branch so that we can manage schemas in parallel.  They can't just be on the `dev` branch because they don't change at the same rate.  But particularly with 3.1+, it's best that we be able to make a series of changes and decide when to publish to `main`, just like we do with the spec.  It's just that the threshold for publishing should be much lower.

But doing this will make it easier to manage which of the various schema dates need updating.

Keep in mind also that you can search/replace `schema/WORK-IN-PROGRESS`, `schema-base/WORK-IN-PROGRESS` `dialect/base/WORK-IN-PROGRESS`, and `meta/base/WORK-IN-PROGRESS` separately, as none of them are substrings of each other.

schemas/v3.1/schema.yaml
handrews: Definitely, but see also #4147 - `schema-base` would work for us but it's not quite the default OAS behavior, and we should take advantage of that to provide a more intuitively-named schema that validates that behavior.  Note that `schema-base` is correct for a slightly different, more restrictive use case.

schemas/v3.1/schema.yaml
ralfhandl: I assumed it was a conscious decision to _not_ date-stamp the `dialect` and `meta` files, and not to have a `vocab` file:

~~~mermaid
flowchart LR
    schema_base
    schema
    dialect
    meta
    vocab
    schema --> dialect
    schema_base --> schema
    schema_base --> dialect
    dialect --> meta
    dialect --> vocab
    meta --> vocab
~~~

ralfhandl:
If we remove the ready-to-use `id` or `$id` from the YAML files, we should add a link to https://spec.openapis.org/ to each of the README.md files.

schemas/v3.1/schema.yaml
handrews: @ralfhandl I wasn't here when that decision was made as I would never have approved it.

If these aren't dated, **_how do you plan to handle it if the dialect or vocabulary schemas need to be changed for any reason?_**

Yes, they are simpler and less likely to change.  That doesn't mean they'll never need to change- if not for 3.1 then in the future, in which case we need to have the pattern set correctly and not be stuck.  I don't know why they were done this way but as far as I'm concerned it was an error.

Again, in practice we can set them once and they probably won't change after that, so the build script needn't get much more complicated right now.  We could:
* Manually publish the dialect and vocabulary schemas with an initial dated URI
* Hardcode a dependency on that initial dated URI for now:
    * Always use those hardcoded initial dated URIs when `schema-base` is updated
    * Error out if the dialect or vocabulary schemas change, so we don't have to code support for that unlikely scenario now (or deal with triggering on multi-document interdependent changes)
    * Changes to `schema` aren't impacted by this at all

ralfhandl: Let's do the needful, even if it hurts 😎

handrews:
I realized that we can't both host a schema ending in `/base` _and_ have a directory of schemas like `/base/2024-10-18`.  Absent any better idea, I just replaced `base` with `WORK-IN-PROGRESS` instead of appending it.

Note that the _vocabulary_ URI (not to be confused with the vocabulary _metaschema_ URI) does not get a date as it is about the semantics, which aren't going to change within a minor release line even if we correct or add commentary to the corresponding metaschema.

It's not ideal to have this `base` alongside `2024-10-18`, and I'm open to other date-including patterns.

ralfhandl:
Resolved conflicts and aligned the two README.md files.

schemas/v3.0/README.md
```diff
@@ -1,27 +1,40 @@
...
+## Schema `id` dates
+
+The published schemas on the spec site have an _iteration date_ in their `id`s.
+This allows the schemas for a release line (in this case 3.0) to be updated independent of the spec patch release cycle.
 
 The iteration version of the JSON Schema can be found in the `id` field.
 For example, the value of `id: https://spec.openapis.org/oas/3.0/schema/2019-04-02` means this iteration was created on April 2nd, 2019.
 
-## Contributing
+The special URL with `latest` in place of the date is intended to provide an HTTP redirect to the most recent dated schema.
```
ralfhandl, line 17:
```suggestion
<!-- The special URL with `latest` in place of the date is intended to provide an HTTP redirect to the most recent dated schema. -->
```
Has someone figured out how to do this with Jekyll?

https://stackoverflow.com/questions/39657226/jekyll-redirect-link may be the answer, if it also works with extension-less non-MD/HTML files.

schemas/v3.1/README.md
```diff
@@ -1,41 +1,54 @@
...
+This allows the schemas for a release line (in this case 3.0) to be updated independent of the spec patch release cycle.
 
 The iteration version of the JSON Schema can be found in the `$id` field.
 For example, the value of `$id: https://spec.openapis.org/oas/3.1/schema/2021-03-02` means this iteration was created on March 2nd, 2021.
 
-The `schema.yaml` schema doesn't validate the JSON Schemas in your OpenAPI
-document because 3.1 allows you to use any JSON Schema dialect you choose. We
-have also included `schema-base.yaml` that extends the main schema to validate
-that all schemas use the default OAS base vocabulary.
+The special URL with `latest` in place of the date is intended to provide an HTTP redirect to the most recent dated schema.
```
ralfhandl, line 30:
```suggestion
<!-- The special URL with `latest` in place of the date is intended to provide an HTTP redirect to the most recent dated schema. -->
```
Has someone figured out how to do this with Jekyll?

schemas/v3.1/schema.yaml
ralfhandl, line 16:
```suggestion
    default: 'https://spec.openapis.org/oas/3.1/dialect/WORK-IN-PROGRESS'
```

ralfhandl (changes requested):

schemas/v3.0/README.md
handrews: @ralfhandl it does not.  We cannot do a non-HTML-based redirect with GitHub pages.  I have been down this path before, and I tried it again yesterday.  GitHub pages is just inadequate for our needs and they have made clear statements in the past that they will never fix this because it's out-of-scope (as is proper setting of media types).

schemas/v3.1/README.md
handrews: It can't be done with Jekyll.

schemas/v3.0/README.md
ralfhandl: Then we should remove this sentence unless/until we decide whether to have a `latest` or `latest.json` symlink.

ralfhandl (approved):

lornajane (approved):

PR #4164, opened 2024-10-28
gh-pages: skip dialect base & meta base

ralfhandl:
Follow-up to
* #4162 

- [x] Only list date-stamped schema iterations for `dialect` and `meta`

Preview:
* https://ralfhandl.github.io/OpenAPI-Specification/

### DIFF
--- a/index.md
+++ b/index.md
@@ -37,6 +37,9 @@ This site contains the OpenAPI Initiative Registry and content for the HTML vers
 * **v{{ last_version }}**
 {%- endif -%}
 {%- if segments[3] != last_kind -%}
+{%- if segments[4] == "base" -%}
+{%- continue -%}
+{%- endif -%}
 {%- assign last_kind = segments[3] %}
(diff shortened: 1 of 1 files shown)


handrews:
Why do we need to treat these differently?  We want all of the schema files to be treated the same.  If we need to group them to show the dependencies, then we should figure out how to do that (probably by generating some YAML file to iterate over rather than trying to do anything complex in Jekyll/Liquid).  But for now, just showing all revisions for all four is correct (and ideally also showing the old `*/base` ones since we should `schema/$date` and `schema-base/$date` that depend on them.

ralfhandl:
> We want all of the schema files to be treated the same.

That would mean that `base` is before the newer date-stamped iterations because "the same" means "in descending lexicographical order". Moving `base` after the newer date-stamped iterations exceeds my Liquid skills - I've tried and failed.

handrews (approved):

handrews:
I approve this, but I can't tell if it needs to go in before or after #4177

PR #4154, opened 2024-10-23
Describe new spec branch and file layout

handrews:
With related process tweaking.

View the file as rendered markdown to see a diagram of the branching strategy!

PR #4166, opened 2024-10-29
main: move YAML examples to test fixtures, remove examples folder

ralfhandl:
- Move `examples/v3.1/*.yaml` files to `tests/v3.1/pass/`
  - These files are automatically processed by `tests/v3.0/schema.test.mjs`
  - Remove `tests/v3.0/examples.test.js` as it does nothing now
- Move `examples/v3.0/*.yaml` files to `tests/v3.0/pass`
  - Adjust `tests/v3.0/examples.test.js` to read from this folder
  - Rename `tests/v3.0/examples.test.js` to `schema.test.mjs`
- Remove `examples` folder
- Remove workflow `convert-examples-to-json.yaml`


handrews:
@ralfhandl are these the same examples we moved to the Learn site?  If so, would it be better to check them out from there and only maintain them in one place?

ralfhandl:
I like to have tests right next to the code, meaning in the same branch of the same repo, and that's where they are now.

That they happen to be identical to something that is also living somewhere else doesn't matter, the two locations serve different purposes and I would not be surprised or concerned if they evolve into different directions.

handrews:
@ralfhandl fair enough.  I think this ties into my questions in #4154 about the purpose of each branch, and clarity on what is staying in this repository and what needs to move.  Both that change and this one, and several others we have discussed, will be so much easier to agree on and approve if we have a coherent framework.

I'm struggling with these things popping up separately because I can't visualize a coherent end state.  I don't really want to block this as I agree it's an improvement, but I'd like to plot the simplest path from where we are to where we want to be.  That includes figuring out where our infrastructure lives.

With three specs that _should_ use the same build, publish, and test infrastructure, all of the code / infrastructure should live in a separate neutral space as much as possible.  But the test _data_ that is specific to a spec should live with the spec.

So I might actually be ending up at a different perspective here:  I think test infrastructure and test data probably live in different places as they change for different reasons at different times.  But I don't actually quite know what constitutes "test infrastructure" and what pieces can be spec-neutral and what pieces cannot.  And I'd like to understand that before rearranging things.

ralfhandl:
I understand your concerns, and I think this PR does not interfere with whatever we will come up with in the future. It merely finishes what we began with https://github.com/OAI/learn.openapis.org/pull/102 almost half a year ago: move (not just copy) the examples to the learn repo.

It also preserves and simplifies the schema tests, which we will need anyway in the future.

Wherever we go from here, this cleanup is useful and can be done now.

lornajane (approved):

mikekistler (approved):
Looks good. 👍

PR #4182, opened 2024-11-08
Use GitHub project for contributor lists instead of a file

### LINKED ISSUE #4148: Remove/replace/migrate the CONTRIBTORS file
The CONTRIBUTORS.md file seems to be a manually maintained list of people who have opened more than 25 PRs (to a released spec? in general?  not sure).  @ralfhandl confirms it is not used by the build system.

In OAI/Overlay-Specification#80, @lornajane suggested removing it because this PR metric does not capture community contribution very well.  I would support that (and we could come back to figuring out a better metric and place to credit later on. one that works for all of our projects).

Alternatively, if we are not comfortable just immediately deleting it, let's migrate the info to a page on spec.openapis.org with some context.

Bonus:  Removing CONTRIBUTORS.md will make it far less confusing to find CONTRIBUTING.md, and will make auto-complete easier.

lornajane:
Fixes #4148 by removing the file.


handrews (approved):

ralfhandl:
Will `dev`, `v3.1-dev`, and `v3.2-dev` be updated from `main` after merging this?

handrews:
@ralfhandl up to the TSC.  I definitely do not want to be the person in charge of approving every branch/merge step.

lornajane:
Yes, we'll merge `main` into those dev branches on an informal schedule - I suggest we work through more of these docs updates and then do some dev branch updates to keep everyone aligned.

PR #4184, opened 2024-11-10
Add criteria for minor and patch releases

### LINKED ISSUE #3528: 3.x.y patch release approach

### LINKED ISSUE #3529: 3.x minor release approach

lornajane:
We had some great discussions in the TDC meetings when we were finalising what should go into the recent patch releases and what must be a minor release. This pull request takes those discussions into our CONTRIBUTING file.

Fixes #3528 
Fixes #3529


CONTRIBUTING.md
```diff
@@ -153,10 +153,34 @@ The following additional rules should be followed but currently are not enforced
...
+Changes in minor releases (such as 3.2, 3.3) meet the following criteria:
+
+* Are **backwards-compatible** and be reasonably easy to implement in tooling that already supports the previous minor version.
+  For example, new optional fields can be added.
+* Drive quality-of-life improvements to support how OpenAPI is used by practitioners, so that OpenAPI evolves to continue to meet user needs.
+  For example, adding fields to support changes in other standards, or adopting common `x-*` extension fields into the specification.
+* Bring the future closer by making changes that are in line with future 3.x releases and the planned OpenAPI 4.x (Moonwalk) specification as the details of that become available.
+* Make the specification document clearer or easier to understand.
+
+A minor release is due when there are some meaningful features (including one or a small nummber of headline features).
```
ralfhandl, line 171:
```suggestion
A minor release is due when there are some meaningful features (including one or a small number of headline features).
```

CONTRIBUTING.md
```diff
@@ -153,10 +153,34 @@ The following additional rules should be followed but currently are not enforced
...
+
+### Patch Releases
+
+Patch releases reflect a constant quest for improving the active minor versions of OpenAPI.
+Since we do not edit specification documents after publication, even the smallest change has to be in a new release.
+
+Changes in patch releases meet the following criteria:
+
+* Minor updates such as spelling or formatting fixes, including link updates.
```
ralfhandl, line 180:
```suggestion
* Editorial changes such as spelling or formatting fixes, including link updates.
```

ralfhandl (commented):
+1, minor nits

CONTRIBUTING.md
```diff
@@ -33,7 +33,7 @@ As of October 2024 (post-OAS 3.0.4 and 3.1.1), the OAS is developed in the `src/
...
 For example, if a change in November 2024 apples to both 3.1 and 3.2, the PR would go to the `v3.1-dev` branch, which will be merged up to `v3.2-dev` before the next 3.2 release.
 
-Releases are published to the [spec site](https://spec.opanapis.org) by creating an `X.Y.Z-rel` branch where `src/oas.md` is renamed to the appropriate `versions/X.Y.Z.md` file and them merged to `main`.  The HTML versions of the OAS are automatically generated from the `versions` directory on `main`.  This renaming on the `X.Y.Z-rel` branch preserves the commit history for the published file on `main` when using `git log --follow` (as is the case for all older published files).
+Releases are published to the [spec site](https://spec.openapis.org) by creating an `X.Y.Z-rel` branch where `src/oas.md` is renamed to the appropriate `versions/X.Y.Z.md` file and them merged to `main`.  The HTML versions of the OAS are automatically generated from the `versions` directory on `main`.  This renaming on the `X.Y.Z-rel` branch preserves the commit history for the published file on `main` when using `git log --follow` (as is the case for all older published files).
```
ralfhandl, line 36:
```suggestion
Releases are published to the [spec site](https://spec.openapis.org) by creating an `X.Y.Z-rel` branch where `src/oas.md` is renamed to the appropriate `versions/X.Y.Z.md` file and then merged to `main`.  The HTML versions of the OAS are automatically generated from the `versions` directory on `main`.  This renaming on the `X.Y.Z-rel` branch preserves the commit history for the published file on `main` when using `git log --follow` (as is the case for all older published files).
```
Typo spotted by @jeremyfiel in #4186

mikekistler (dismissed):
Looks good. 👍

CONTRIBUTING.md
lornajane: If this has been added elsewhere then it should be accepted in that pull request. If we make the same change here, we may cause conflicts?

ralfhandl: I would have guessed no, and guessed that we'll get a conflict if we don't make the same change here.

The second guess has been confirmed, the first remains unproven 😎

ralfhandl (approved):

PR #4189, opened 2024-11-11
The schemas are (mostly) not metaschemas

handrews:
OpenAPI schemas are not metaschemas.  Well, the bits for the Schema Object are, but the overall OAS is more important and it's already driving me bonkers to see these described as "metaschemas."  An OAD is not a JSON Schema, so these schemas are just schemas. _(yes, I am fixating on something fundamentally unimportant, please let me have this small nitpick thx)_

### DIFF
--- a/.github/workflows/schema-publish.yaml
+++ b/.github/workflows/schema-publish.yaml
@@ -49,8 +49,8 @@ jobs:
         path: deploy
         labels: Housekeeping,Schema
         reviewers: darrelmiller,webron,earth2marsh,webron,lornajane,mikekistler,miqui,ralfhandl,handrews,karenetheridge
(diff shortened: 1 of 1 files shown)


ralfhandl (approved):

ralfhandl:
What about https://github.com/OAI/OpenAPI-Specification/blob/ac6ba3d7afb7140de33d1b4429dcb818e0d8d2f6/schemas/v3.1/README.md?plain=1#L14-L15

Is "metaschema" used correctly there?

handrews:
@ralfhandl yes, those are metaschemas.  But from the point of view of most users, they are implementation details, and either `schema` or `schema-base` is "the schema"

ralfhandl:
Merging as it only affects a github workflow

PR #4191, opened 2024-11-12
respec workflow: create symlinks for spec minor versions

ralfhandl:
- [x] Create symlink `vX.Y.html` for each minor version except 2.0
- [x] make `latest.html` also a symlink instead of a copy


lornajane:
We agreed to remove the `-latest` suffix

handrews (approved):

PR #4186, opened 2024-11-10
CONTRIBUTING.md: Schema development process, improved mermaid diagram, etc.

handrews:
More work towards #3715 and #3716.
Also paging @jeremyfiel as GitHub won't let me add you as a reviewer.

PR #4197, opened 2024-11-13
v3.1-dev: schema tests use src/schemas/validation

ralfhandl:
Run schema tests with current work-in-progress schema in `src/schemas/validation` folder

Open:
- [ ] Should the `pass` and `fail` folders with schema test instances also be copied into the `src/schemas/validation` folder?

### DIFF
--- a/tests/v3.1/schema.test.mjs
+++ b/tests/v3.1/schema.test.mjs
@@ -26,7 +26,7 @@ const parseYamlFromFile = (filePath) => {
 
 setMetaSchemaOutputFormat(BASIC);
 
-const validateOpenApi = await validate("./schemas/v3.1/schema.yaml");
+const validateOpenApi = await validate("./src/schemas/validation/schema.yaml");
 
 describe("v3.1", () => {
   describe("Pass", () => {


handrews:
@ralfhandl I'm now realizing that actually (per the update to CONTRIBUTING.md that I just made in #4186) that this (and, sadly, #4196 which I merged before thinking about it) should start on `dev` and then be merged out to the `vX.Y-dev` branches from there.

`dev` should be where everything fully works, so that each time we start a new `vX.Y-dev` branch **_from `dev`_**, it always gets the most up-to-date infrastructure.  We never want to have to think about which branch is active or most current or whatever when starting new branches in the future.  We just want to keep `dev` in good shape as the place where new branches _always_ start.

ralfhandl:
See
* #4198

ralfhandl:
I cherry-picked the only commit in this PR for #4198.

Do we want to merge this PR, or open a new one updating `v3.1-dev` from `dev`?

lornajane:
As discussed in TDC call, closing in favour of #4198

PR #4198, opened 2024-11-13
dev: schema tests use src/schemas/validation

ralfhandl:
Run schema tests with current work-in-progress schema in src/schemas/validation folder

### DIFF
--- a/tests/v3.1/schema.test.mjs
+++ b/tests/v3.1/schema.test.mjs
@@ -26,7 +26,7 @@ const parseYamlFromFile = (filePath) => {
 
 setMetaSchemaOutputFormat(BASIC);
 
-const validateOpenApi = await validate("./schemas/v3.1/schema.yaml");
+const validateOpenApi = await validate("./src/schemas/validation/schema.yaml");
 
 describe("v3.1", () => {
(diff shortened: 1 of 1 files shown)


handrews:
I would prefer to merge rather than cherry-pick.  The branches haven't diverged yet so a bit of merging back and forth while we learn our new process doesn't seem like a bad thing.  Sub-optimal, but we'll all likely do a few sub-optimal things while learning!

If you feel strongly that cherry-pick is better, I'm open to understanding that position.  I'd like for us to have a clear policy on when we use different git functions rather than doing it ad-hoc.

lornajane:
I feel strongly that cherry pick is better for a use case where we do want to make the same change to more than one `v*-dev` branch because otherwise we'll merge too many things between branches when they diverge more and it'll be chaotic. There's definitely an alternative where we make a second pull request with a different base but I vote for a change in one place, and a cherry pick to the other branch that needs it. If that makes sense then I totally agree that one of us should write that down!

ralfhandl:
Yes, please, let's document the preferred processes!

> The branches haven't diverged yet

Except for src/oas.md 🫤

handrews (approved):

PR #4200, opened 2024-11-13
Removing the CoC copy in favor of the org-level template

earth2marsh:
IIUC, the [org level CoC](https://github.com/OAI/.github/blob/main/.github/CODE_OF_CONDUCT.md) will be used if no file exists at the repo level. Unless there is some clear reason why we want to keep separate documents in sync, I think we should remove it. Then, like [Arazzo](https://github.com/OAI/Arazzo-Specification), it should pick up the CoC on the landing page as well as the various appropriate areas like issue filing.

... and if there are clear reasons why we don't want to remove this copy, then please close this PR without merging.


mikekistler:
My vague recollection is that we added the CoC to get a green check for this in the "Community Standards" page on GitHub:

image

Will will lose the green check if we remove this file? If so, do we care?

ralfhandl:
> Will will lose the green check if we remove this file? If so, do we care?

Let's find out:sunglasses:

ralfhandl (approved):

PR #4244, opened 2024-12-05
adds ABNF for path templating

baywet:
partial #3256

This pull request adds an ABNF for path templating.

It only partially addresses the issue, my goal is to submit additional pull requests for:

- additional links to annex C where missing
- ABNF for server templating
- relationship between server and path templating (if any)

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -44,6 +44,34 @@ Each template expression in the path MUST correspond to a path parameter that is
 
 The value for these path parameters MUST NOT contain any unescaped "generic syntax" characters described by [RFC3986](https://tools.ietf.org/html/rfc3986#section-3): forward slashes (`/`), question marks (`?`), or hashes (`#`).
 
+The path templating expression is defined by the following [ABNF](https://tools.ietf.org/html/rfc5234) syntax
+
+```abnf
+; OpenAPI Path Templating ABNF syntax
+path-template                  = path [ query-marker query ] [ fragment-marker fragment ]
+path                           = slash *( path-segment slash ) [ path-segment ]
+path-segment                   = 1*( path-literal / template-expression )
+query                          = *( query-literal )
+query-literal                  = 1*( unreserved / pct-encoded / sub-delims / ":" / "@" / "/" / "?" / "&" / "=" )
+query-marker                   = "?"
+fragment                       = *( fragment-literal )
+fragment-literal               = 1*( unreserved / pct-encoded / sub-delims / ":" / "@" / "/" / "?" )
+fragment-marker                = "#"
+slash                          = "/"
+path-literal                   = 1*( unreserved / pct-encoded / sub-delims / ":" / "@" )
+template-expression            = "{" template-expression-param-name "}"
+template-expression-param-name = 1*( unreserved / pct-encoded / sub-delims / ":" / "@" )
+
+; Characters definitions (from RFC 3986)
+unreserved          = ALPHA / DIGIT / "-" / "." / "_" / "~"
+pct-encoded         = "%" HEXDIG HEXDIG
+sub-delims          = "!" / "$" / "&" / "'" / "(" / ")"
+                    / "*" / "+" / "," / ";" / "="
+ALPHA               = %x41-5A / %x61-7A   ; A-Z / a-z
+DIGIT               = %x30-39            ; 0-9
+HEXDIG              = DIGIT / "A" / "B" / "C" / "D" / "E" / "F"
+```
+
 ### Media Types
 
 Media type definitions are spread across several resources.


src/oas.md
```diff
@@ -44,6 +44,31 @@ Each template expression in the path MUST correspond to a path parameter that is
 
 The value for these path parameters MUST NOT contain any unescaped "generic syntax" characters described by [RFC3986](https://tools.ietf.org/html/rfc3986#section-3): forward slashes (`/`), question marks (`?`), or hashes (`#`).
 
+The path templating expression is defined by the following [ABNF](https://tools.ietf.org/html/rfc5234) syntax
+
+```abnf
+; OpenAPI Path Templating ABNF syntax
+path-template                  = path [ query-marker query ]
```
baywet, line 51: I have removed the fragment part from the original work but left the query parameter part:
- fragments: never go to the service, no point in adding them to the spec anyway
- query parameters: probably shouldn't be here either, but wanted to triple confirm we want to remove it since some descriptions use it to describe "constant query parameters", and removing those would technically break them.

src/oas.md
```diff
@@ -44,6 +44,31 @@ Each template expression in the path MUST correspond to a path parameter that is
...
+path                           = slash *( path-segment slash ) [ path-segment ]
+path-segment                   = 1*( path-literal / template-expression )
+query                          = *( query-literal )
+query-literal                  = 1*( unreserved / pct-encoded / sub-delims / ":" / "@" / "/" / "?" / "&" / "=" )
+query-marker                   = "?"
+slash                          = "/"
+path-literal                   = 1*( unreserved / pct-encoded / sub-delims / ":" / "@" )
+template-expression            = "{" template-expression-param-name "}"
+template-expression-param-name = 1*( unreserved / pct-encoded / sub-delims / ":" / "@" )
```
ralfhandl, line 60: The only explicit restriction we state is that

> Each template expression in the path MUST correspond to a path parameter

and there are no restrictions on path parameter names.

The only implicit restriction is that the `}` cannot be used in a template expression because we don't define an escaping mechanism for it.

src/oas.md
```diff
@@ -44,6 +44,31 @@ Each template expression in the path MUST correspond to a path parameter that is
...
 
+The path templating expression is defined by the following [ABNF](https://tools.ietf.org/html/rfc5234) syntax
+
+```abnf
+; OpenAPI Path Templating ABNF syntax
+path-template                  = path [ query-marker query ]
+path                           = slash *( path-segment slash ) [ path-segment ]
+path-segment                   = 1*( path-literal / template-expression )
+query                          = *( query-literal )
+query-literal                  = 1*( unreserved / pct-encoded / sub-delims / ":" / "@" / "/" / "?" / "&" / "=" )
```
ralfhandl, line 55:
```suggestion
query-literal                  = 1*( unreserved / pct-encoded / sub-delims / ":" / "@" / "/" / "?" )
```

`&` and `=` are part of `sub-delims`

src/oas.md
```diff
@@ -44,6 +44,31 @@ Each template expression in the path MUST correspond to a path parameter that is
...
+template-expression            = "{" template-expression-param-name "}"
+template-expression-param-name = 1*( unreserved / pct-encoded / sub-delims / ":" / "@" )
+
+; Characters definitions (from RFC 3986)
+unreserved          = ALPHA / DIGIT / "-" / "." / "_" / "~"
+pct-encoded         = "%" HEXDIG HEXDIG
+sub-delims          = "!" / "$" / "&" / "'" / "(" / ")"
+                    / "*" / "+" / "," / ";" / "="
+ALPHA               = %x41-5A / %x61-7A   ; A-Z / a-z
+DIGIT               = %x30-39            ; 0-9
```
ralfhandl, line 68:
```suggestion
DIGIT               = %x30-39             ; 0-9
```

ralfhandl (changes requested):
Minor nits

src/oas.md
baywet: so you'd extend what's in there in essence? I double checked RFC5234 and there does not appear to be any way so simply say "anything but this character", and I think it's on purpose.

What do you think is missing?  `%[]^"'<>?` ?

src/oas.md
ralfhandl: I'd remove the query part because the specification text clearly talks about paths, and section [4.8.12.1 Parameter Locations](https://spec.openapis.org/oas/v3.1.1.html#parameter-locations) clearly distinguishes path parameters and query parameters.

Making this explicit in the ABNF may prevent future loose interpretation of the spec.

src/oas.md
```diff
@@ -44,6 +44,31 @@ Each template expression in the path MUST correspond to a path parameter that is
 
 The value for these path parameters MUST NOT contain any unescaped "generic syntax" characters described by [RFC3986](https://tools.ietf.org/html/rfc3986#section-3): forward slashes (`/`), question marks (`?`), or hashes (`#`).
 
+The path templating expression is defined by the following [ABNF](https://tools.ietf.org/html/rfc5234) syntax
+
+```abnf
+; OpenAPI Path Templating ABNF syntax
+path-template                  = path [ query-marker query ]
+path                           = slash *( path-segment slash ) [ path-segment ]
```
ralfhandl, line 52:
```suggestion
path-template                  = slash *( path-segment slash ) [ path-segment ]
```

src/oas.md
```diff
@@ -44,6 +44,31 @@ Each template expression in the path MUST correspond to a path parameter that is
...
+path-segment                   = 1*( path-literal / template-expression )
+query                          = *( query-literal )
+query-literal                  = 1*( unreserved / pct-encoded / sub-delims / ":" / "@" / "/" / "?" )
+query-marker                   = "?"
```
baywet, line 56:
```suggestion
```

src/oas.md
baywet: Thanks for confirming, removed

src/oas.md
ralfhandl: Trick question: which ABNF rule is equivalent to regexp `^[^}]+$` 😎

baywet: that's effectively what I was looking for, couldn't find an equivalent, but maybe I missed something?

ralfhandl: We'd probably have to list all characters except the `}`, which raises the question what "all characters" are:
* all octets `%x00-FF`, or
* just all characters allowed in URIs: `unreserved / reserved / pct-encoded`

One could argue that a path template is not restricted to URI characters and is just a string that becomes a URI path _after_ replacing all template expressions with their values.

baywet: could we start with a more restrictive set (unreserved and al) and maybe open it up based on the feedback we get in another patch release?

ralfhandl (approved):

src/oas.md
baywet: based on @handrews 's suggestion, should we go with that instead?

`%x00-79 / %x7C / %x7E-FF ; every unicode character except { and }`

I'm worried this will be too broad as it includes special characters like CR/LF/NUL space etc...

ralfhandl: We could omit the nonprinting control characters and start with `%x20`, or even exclude the space and start with `%x21`.

karenetheridge: There are unicode characters beyond %xFF.

baywet: Then we should probably also exclude 7F to A0 as well. Anything else?
https://symbl.cc/en/unicode-table/

baywet: > There are unicode characters beyond %xFF.

@karenetheridge do we think it's broadly needed? could we stick with basic latin and latin1 (minus what we're planning to exclude already) and see whether people ask us to extend it later in a patch release?

handrews: @baywet we should not forbid anything that is currently allowed.  So unless the spec says you cannot use those characters, then you an use them and they should be allowed.  Keep in mind that [Appendix C.4.4](https://spec.openapis.org/oas/v3.1.1#illegal-variable-names-as-parameter-names) shows an example of a parameter with the heart emoji as a name.

We can probably get away with excluding the non-printing control characters, as that's pretty pathological.  I'm not sure we should forbid the space character.

ralfhandl: The most generous rule would be
```suggestion
template-expression-param-name = 1*( %x00-7C / %x7E-10FFFF ) ; every unicode character except "}"
```
This relies on path templates being represented as JSON or YAML strings and thus using UTF-8.

ralfhandl (changes requested):

src/oas.md
baywet: @ralfhandl should we also at least exclude `{` ? it'd probably make parser's lives much easier.

ralfhandl: @handrews?

handrews:
@ralfhandl 

> This relies on path templates being represented as JSON or YAML strings and thus using UTF-8.

We primarily define thing in terms of JSON, and (as mentioned in one of the new appendixes), JSON strings are only broadly interoperable in UTF-8, per the JSON RFC.  It might be worth mentioning that here, but I think it's reasonable for us to write the spec based on what is interoperable in JSON.

src/oas.md
baywet: Thanks for all this additional information!

I think we're reaching a consensus to this

```anbf
%x00-79 / %x7C / %x7E-10FFFF ; every UTF8 character except { and }
```

Thoughts?

char0n:
Here is the latest version of ABNF I've come up in https://github.com/swaggerexpert/openapi-path-templating/blob/next/src/path-templating.bnf

```abnf
; OpenAPI Path Templating ABNF syntax
; Aligned with RFC 3986 (https://datatracker.ietf.org/doc/html/rfc3986#section-3.3)
path-template                  = slash [ path-template-nz ]
path-template-nz               = path-segment *( slash path-segment )
slash                          = "/"
path-segment                   = 1*( path-literal / template-expression )
path-literal                   = 1*pchar
template-expression            = "{" template-expression-param-name "}"
template-expression-param-name = 1*pchar

; Characters definitions (from RFC 3986)
pchar               = unreserved / pct-encoded / sub-delims / ":" / "@"
unreserved          = ALPHA / DIGIT / "-" / "." / "_" / "~"
pct-encoded         = "%" HEXDIG HEXDIG
sub-delims          = "!" / "$" / "&" / "'" / "(" / ")"
                    / "*" / "+" / "," / ";" / "="
ALPHA               = %x41-5A / %x61-7A   ; A-Z / a-z
DIGIT               = %x30-39            ; 0-9
HEXDIG              = DIGIT / "A" / "B" / "C" / "D" / "E" / "F"
```

This grammar is now as closely aligned aligned with https://datatracker.ietf.org/doc/html/rfc3986#section-3.3 as it can be.

char0n:
Just a small note related to licensing. I do support this effort fully, but seeing this PR, I got a bit worried how this ABNF grammar is being copied here from `openapi-path-templating` without any attribution or copyright notice. 

I've created the original ABNF grammar in https://github.com/swaggerexpert/openapi-path-templating and is licensed under `Apache-2.0` with following copyright notice:

```
openapi-path-templating
Copyright 2023, Vladimír Gorej
openapi-path-templating is licensed under Apache 2.0 license.
Copy of the Apache 2.0 license can be found in `LICENSE` file.
```

The [NOTICE file](https://github.com/swaggerexpert/openapi-path-templating/blob/main/NOTICE) is part of https://github.com/swaggerexpert/openapi-path-templating repo and npm package distribution. 

---

Can we as part of this repo (https://github.com/OAI/OpenAPI-Specification) create a NOTICE file ([standard practice with Apache-2.0 licensing](https://vladimirgorej.com/blog/how-to-apply-apache2-license-to-your-open-source-software-project/)) and mention the attribution of `openapi-path-templating` there?

src/oas.md
char0n: > Trick question: which ABNF rule is equivalent to regexp `^[^}]+$` 😎

[SABNF](https://github.com/ldthomas/apg-js2/blob/master/SABNF.md) does allow that, be we want to keep to standard ABNF which does not

ralfhandl (approved):

char0n:
@ralfhandl if there is a consensus on https://github.com/OAI/OpenAPI-Specification/pull/4244/commits/fc576083e3b9a40fa7d2066d66ce7f26c1ba8983, I'd strongly suggest to go for a following ABNF:

```abnf
; OpenAPI Path Templating ABNF syntax
; Aligned with RFC 3986 (https://datatracker.ietf.org/doc/html/rfc3986#section-3.3)
path-template                  = "/" [ path-template-nz ]
path-template-nz               = path-segment *( "/" path-segment )
path-segment                   = 1*( path-literal / template-expression )
path-literal                   = 1*pchar
template-expression            = "{" template-expression-param-name "}"
template-expression-param-name = 1*( %x00-79 / %x7C / %x7E-10FFFF ) ; every UTF8 character except { and }

; Characters definitions (from RFC 3986)
pchar               = unreserved / pct-encoded / sub-delims / ":" / "@"
unreserved          = ALPHA / DIGIT / "-" / "." / "_" / "~"
pct-encoded         = "%" HEXDIG HEXDIG
sub-delims          = "!" / "$" / "&" / "'" / "(" / ")"
                    / "*" / "+" / "," / ";" / "="
ALPHA               = %x41-5A / %x61-7A   ; A-Z / a-z
DIGIT               = %x30-39            ; 0-9
HEXDIG              = DIGIT / "A" / "B" / "C" / "D" / "E" / "F"
```

Reasoning 

1. this grammar now reflects https://datatracker.ietf.org/doc/html/rfc3986#section-3.3 as closely as possible, taking into account OpenAPI template expressions
2. This grammar is as simplest as it can be (there is need for defining `slash` as non-terminal; I only originally did it for my internal parsing purposes). Using "/" litera is more readable, less verbose and always preferred (as per RFC3986)

src/oas.md
```diff
@@ -44,6 +44,28 @@ Each template expression in the path MUST correspond to a path parameter that is
 
 The value for these path parameters MUST NOT contain any unescaped "generic syntax" characters described by [RFC3986](https://tools.ietf.org/html/rfc3986#section-3): forward slashes (`/`), question marks (`?`), or hashes (`#`).
 
+The path templating expression is defined by the following [ABNF](https://tools.ietf.org/html/rfc5234) syntax
+
+```abnf
+; OpenAPI Path Templating ABNF syntax
+path-template                  = slash *( path-segment slash ) [ path-segment ]
+path-segment                   = 1*( path-literal / template-expression )
+slash                          = "/"
```
baywet, line 53:
```suggestion
path-template                  = "/" [ path-template-nz ]
path-template-nz               = path-segment *( "/" path-segment )
path-segment                   = 1*( path-literal / template-expression )
```

(porting the suggestion from @char0n so we can have a focused discussion)

@char0n I think one of the challenges of the syntax you're proposing is that it doesn't allow a trailing slash `/foo/bar/` where the current one does.

I don't have a strong opinion on the slash being a terminal or not.

baywet:
@char0n I just pushed changes for the pchar. Thanks for the suggestion!

Also made this comment so we can have a focused discussion https://github.com/OAI/OpenAPI-Specification/pull/4244/files#r1886798408

Lastly, around the license aspect, isn't openapi-path-templating derivative work of this specification? (under Apache 2) Which means:
- the notice in your repo should mention the use of the specification.
- re-incorporating derivative work into the original work doesn't need attribution? (not sure about this one, but it'd create an attribution loop, which sounds problematic)

src/oas.md
char0n: > @char0n I think one of the challenges of the syntax you're proposing is that it doesn't allow a trailing slash /foo/bar/ where the current one does.

It should allow, as per RFC 3986. This ABNF MUST correctly reflect that.

I've updated to reflect that as well - https://github.com/OAI/OpenAPI-Specification/pull/4244#issuecomment-2545553855

baywet: just to confirm I'm not missing anything, the only difference is the slash not being a terminal, correct?

char0n:
> @char0n I just pushed changes for the pchar. Thanks for the suggestion!
> 
> Also made this comment so we can have a focused discussion https://github.com/OAI/OpenAPI-Specification/pull/4244/files#r1886798408
> 
> Lastly, around the license aspect, isn't openapi-path-templating derivative work of this specification? (under Apache 2) Which means:
> 
> * the notice in your repo should mention the use of the specification.
> * re-incorporating derivative work into the original work doesn't need attribution? (not sure about this one, but it'd create an attribution loop, which sounds problematic)

I don't want to get into black hole of interpretations or opinions.

> Lastly, around the license aspect, isn't openapi-path-templating derivative work of this specification?

IMHO no, not at all. While my original ABNF syntax is inspired by OpenAPI’s description of path templating, it implements ideas, not specific expressions or text, from the OpenAPI spec. It also builds upon public domain concepts from RFC 3986 for URL syntax and path structures. Thus, my ABNF does not qualify as a derivative work of OpenAPI Specification under copyright law.

---

Overall I consider the matter closed. I've raised a small concern here because I've seen ABNF work I did copied without any attribution. Let's focus of getting the best ABNF into the OpenAPI going forward.

src/oas.md
baywet: sorry, I miss-used the terminology here. What I meant was: can you confirm this is the only difference remaining please?

```suggestion
path-template                  = "/" *( path-segment "/" ) [ path-segment ]
path-segment                   = 1*( path-literal / template-expression )
```

char0n: Yes I confirm. These are the only last changes needed.

baywet:
@char0n thanks for the suggestions here!
@ralfhandl @handrews this should be ready for (hopefully final) review :)

src/oas.md
```diff
@@ -44,6 +44,27 @@ Each template expression in the path MUST correspond to a path parameter that is
 
 The value for these path parameters MUST NOT contain any unescaped "generic syntax" characters described by [RFC3986](https://tools.ietf.org/html/rfc3986#section-3): forward slashes (`/`), question marks (`?`), or hashes (`#`).
 
+The path templating expression is defined by the following [ABNF](https://tools.ietf.org/html/rfc5234) syntax
+
+```abnf
+; OpenAPI Path Templating ABNF syntax
```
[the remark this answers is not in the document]
char0n: @baywet there is this one. I'm not 100% sure - it's just my understanding of  IETF’s copyright policies under BCP 78

src/oas.md
```diff
@@ -44,6 +44,27 @@ Each template expression in the path MUST correspond to a path parameter that is
 
 The value for these path parameters MUST NOT contain any unescaped "generic syntax" characters described by [RFC3986](https://tools.ietf.org/html/rfc3986#section-3): forward slashes (`/`), question marks (`?`), or hashes (`#`).
 
+The path templating expression is defined by the following [ABNF](https://tools.ietf.org/html/rfc5234) syntax
```
char0n, line 47:
```suggestion
The path templating is defined by the following [ABNF](https://tools.ietf.org/html/rfc5234) syntax
```

Isn't it either `path templating` or `path template expression`?  Mixing it together doesn't sound right. This ABNF is defining whole `path templating`, not just the `path templating expression`.

My suggestion is aligned with https://github.com/OAI/OpenAPI-Specification/blob/1da9fbbbce16ea22da73951fdefdd1d32c447664/src/oas.md#runtime-expressions

src/oas.md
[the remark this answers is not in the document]
baywet: Why don't we instead put the link at line 57 instead? This way it's clear where we got that from.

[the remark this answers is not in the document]
char0n: Line `57` defines just the Characters definitions copied from various sections of RFC 3986.

Line `50` (`; OpenAPI Path Templating ABNF syntax`) and all non-terminals defined under it, are directly derived from specific section https://datatracker.ietf.org/doc/html/rfc3986#section-3.3, which defines what `Path` is. 

If you look at [runtime expressions](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#runtime-expressions) it doesn't use RFC attribution directly in ABNF grammar, but rather uses a normal sentence in which it attributes various RFCs. We should probably align with runtime expressions in this form to be consistent.

--- form start ---

```abnf
path-template                  = "/" *( path-segment "/" ) [ path-segment ]
path-segment                   = 1*( path-literal / template-expression )
path-literal                   = 1*pchar
template-expression            = "{" template-expression-param-name "}"
template-expression-param-name = 1*( %x00-79 / %x7C / %x7E-10FFFF ) ; every UTF8 character except { and }

pchar               = unreserved / pct-encoded / sub-delims / ":" / "@"
unreserved          = ALPHA / DIGIT / "-" / "." / "_" / "~"
pct-encoded         = "%" HEXDIG HEXDIG
sub-delims          = "!" / "$" / "&" / "'" / "(" / ")"
                    / "*" / "+" / "," / ";" / "="
```

Here, all characters definitions are taken from [RFC 3986](https://tools.ietf.org/html/rfc3986). The `path-template` is directly derived from [RFC 3986, section 3.3](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3).

--- form end ---

I've taken out `ALPHA`, `DIGIT`, `HEXDIG` which are part of https://datatracker.ietf.org/doc/html/rfc2234, so it's implied that they exists in the grammar by default.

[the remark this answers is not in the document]
baywet: Thank you for the additional information.

I'm still struggling to understand why this is needed/what are we trying to achieve here? (keep in mind this is the first time I'm building those sort of things)

Is the goal to say "we have a dependency on X, make sure you understand this" or to say "this is inspired on that other thing, read this if you want more context"? (or something else entirely?)
If the goal is the former, I'd go with the derived from suggestion.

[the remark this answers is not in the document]
baywet: Thank you for the additional information.

If this information is not normative then, should it be instead outside of the ANBF block, like runtime expression currently does?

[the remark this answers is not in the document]
baywet: Thanks! I just pushed exactly those changes!

src/oas.md
```diff
@@ -44,6 +44,27 @@ Each template expression in the path MUST correspond to a path parameter that is
...
+unreserved          = ALPHA / DIGIT / "-" / "." / "_" / "~"
+pct-encoded         = "%" HEXDIG HEXDIG
+sub-delims          = "!" / "$" / "&" / "'" / "(" / ")"
+                    / "*" / "+" / "," / ";" / "="
+ALPHA               = %x41-5A / %x61-7A   ; A-Z / a-z
+DIGIT               = %x30-39             ; 0-9
+HEXDIG              = DIGIT / "A" / "B" / "C" / "D" / "E" / "F"
+```
+
+Here, all characters definitions are taken from [RFC 3986](https://tools.ietf.org/html/rfc3986). The path-template is directly derived from [RFC 3986, section 3.3](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3).
```
char0n, line 66:
```suggestion
Here, all characters definitions are taken from [RFC 3986](https://tools.ietf.org/html/rfc3986). The `path-template` is directly derived from [RFC 3986, section 3.3](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3).
```

Nitpick (alignment with how runtime expression section references non-terminals in text).

src/oas.md
```diff
@@ -44,6 +44,27 @@ Each template expression in the path MUST correspond to a path parameter that is
...
+path-literal                   = 1*pchar
+template-expression            = "{" template-expression-param-name "}"
+template-expression-param-name = 1*( %x00-79 / %x7C / %x7E-10FFFF ) ; every UTF8 character except { and }
+
+pchar               = unreserved / pct-encoded / sub-delims / ":" / "@"
+unreserved          = ALPHA / DIGIT / "-" / "." / "_" / "~"
+pct-encoded         = "%" HEXDIG HEXDIG
+sub-delims          = "!" / "$" / "&" / "'" / "(" / ")"
+                    / "*" / "+" / "," / ";" / "="
+ALPHA               = %x41-5A / %x61-7A   ; A-Z / a-z
```
char0n, line 61: Nitpick (ALPHA, DIGIT, HEXDIG) are part of RFC 5234 (ABNF) syntax itself and are usually dropped from final ABNFs. Do we want to align with `runtime expression` ABNF and not explicitly define those?

baywet: Thanks for pointing this out! Removed since 3986 doesn't redefine them either.

char0n (approved):
Thank you for your patience! Looks great!

src/oas.md
```diff
@@ -44,6 +44,24 @@ Each template expression in the path MUST correspond to a path parameter that is
...
+template-expression-param-name = 1*( %x00-79 / %x7C / %x7E-10FFFF ) ; every UTF8 character except { and }
+
+pchar               = unreserved / pct-encoded / sub-delims / ":" / "@"
+unreserved          = ALPHA / DIGIT / "-" / "." / "_" / "~"
+pct-encoded         = "%" HEXDIG HEXDIG
+sub-delims          = "!" / "$" / "&" / "'" / "(" / ")"
+                    / "*" / "+" / "," / ";" / "="
+```
+
+Here, all characters definitions are taken from [RFC 3986](https://tools.ietf.org/html/rfc3986). The `path-template` is directly derived from [RFC 3986, section 3.3](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3).
```
ralfhandl, line 63: What is meant by "characters definitions"?

baywet: Would you prefer to call out the individual items?

baywet: I went ahead and implemented the change!

ralfhandl (approved):

mikekistler (approved):
Looks good! 👍

Issue #4223, opened 2024-11-22
Create dev build infrastructure


handrews:
I had a check fail on #4188 on `v3.2-dev` and realized that there's no way to build without copying things over to the publish location (and even then something went weird when it was trying to build the old versions b/c the dev branches aren't set up for that).

We should be able to build just `src/oas.md` for testing purposes on `dev` and all branches derived from `dev`.

This change should go into `dev` and be merged/cherry-picked out from there.

If Arazzo and Overlay are going to use the same system, this might also be a good time to start building a shared system instead of continuing to patch things in triplicate.

- [x] Build `src/oas.md`
- [x] Build schemas from `src`
  - #4361 
- [x] Test schemas in `src` upon push
- [x] Validate & lint `src/oas.md` upon push
- [x] Format `src/oas.md`
  - #4336

ralfhandl:
@handrews How luxurious do you want it to be?

The simplest solution would be to add a script `npm run build-src` that creates a static `oas.html` next to the `oas.md` (or anywhere you like) and add that file to `.gitignore`.

A nicer experience would be to start a tiny express server via `npm run serve-src` that dynamically produces an ephemeral `oas.html` on every change of `oas.md` and tells the browser to auto-refresh. That way you could have a browser window next to your editor window, type in one and live-check the other.

I have copy templates for both approaches lying around 😎

handrews:
@ralfhandl I'm perfectly happy with the simple version (and would be more likely to use it TBH because all of my dev tools tend to be the lowest-tech that works for me).  But if you want to do the more complicated thing I'm not going to stop you.

Although... express seems like a bit dependency to add, and I'd kind-of like to be able to just build and look at the HTML without running a server.  I guess just make sure I don't have to understand express _at all_ to use the build system.

ralfhandl:
@handrews regarding

> - [ ] Build schemas from src

What would you expect this to do?

The work-in-progress schema in `src/schemas/validation/schema.yaml` is already validated by the schema tests, and the build step just converts the YAML files into JSON, renames them, and adjusts the corresponding `$ref`erences.

handrews:
@ralfhandl I don't even remember writing that!  I think I did mean "build the JSON from the YAML and update the ids and references in both", with the intention being to load the built schemas and make sure that the references all line up.

[...]

PR #4254, opened 2024-12-16
dev: build HTML from src/oas.md

ralfhandl:
- [x] new npm script `npm run build-src` to build HTML from file file `src/oas.md`
- [x] produce `deploy-preview/oas.html`
- [x] git-ignore folder `deploy-preview` 
- [x] update`CONTRIBUTING.md` from `main`: #4271
- [x] describe build scripts in `CONTRIBUTING.md`

Part of
- #4223


handrews:
I tried this out and it works as advertised- thanks!

One question: Could we put `src-preview/` under `deploy/`?  Or maybe have a `deploy-dev/`?  I kind of like keeping the build artifacts together, or at least following the same sort of naming.  But this isn't something I'd insist on, just a preference.

ralfhandl:
We can put `oas.html` wherever you like, I just thought it is easier to find next to `src/`, and we could also build the JSON Schema files from `src/schemas/validation/` into it.

lornajane:
We agreed to name the output directory `deploy-preview` (the main build script outputs to `deploy/`. Also this pull request needs some usage documentation in the CONTRIBUTING file so we can remember what we built and how to use it.

scripts/md2html/build-src.sh
```diff
@@ -0,0 +1,26 @@
...
+mkdir -p deploy/oas
+mkdir -p deploy/js
+mkdir -p deploy-preview
+
+cp -p node_modules/respec/builds/respec-w3c.* deploy/js/
+
+cd scripts/md2html
+
+filename=src/oas.md
+tempfile=../../deploy/oas/oas.html
```
handrews, line 16: it seems odd to me that we're putting a tempfile under /deploy when our output is under /deploy-preview.  Should we not have the tempfile there and named something obvious (eg. with `tmp` or `temp` in the name?) so that if something breaks the mess is left in the same place we were expecting output?  And we can have an `npm run clean-src` that just deletes that directory so that it works whether the build was successful or not? (I don't care if you add clean-src in this PR or not, I'm just thinking it might be nice at some point).

Homedefense78 (changes requested):
We reserve the right as the sole owners to change laws and policies with in reason to the new terms and conditions with our new global responsibilities to continue our growth in quantum capabilities since I am the only way to do this process successfully first we will be changing things momentarily please advise all staff owner Dain Maurice Scruggs thanks we appreciate you cooperation

handrews (approved):

PR #4257, opened 2024-12-17
Editorial change: Link Object points to Operation Object

char0n:

PR #4260, opened 2024-12-17
Editorial change: Link Object points to Operation Object

char0n:
Unfortunately `>=3.0.4` is not going to be addressed by this, as discussed in https://github.com/OAI/OpenAPI-Specification/pull/4257


ralfhandl (approved):

mikekistler (approved):
Looks good! 👍 

Thanks for the fix.

PR #4264, opened 2024-12-18
adds server url template abnf

baywet:
partial #3256
related #4244

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -563,14 +563,13 @@ The server URL templating is defined by the following [ABNF](https://tools.ietf.
 server-url-template            = 1*( literals / template-expression )
 template-expression            = "{" template-expression-param-name "}"
 template-expression-param-name = 1*( %x00-79 / %x7C / %x7E-10FFFF ) ; every UTF8 character except { and }
+
 literals                       = 1*( %x21 / %x23-24 / %x26 / %x28-3B / %x3D / %x3F-5B
                                   / %x5D / %x5F / %x61-7A / %x7E / ucschar / iprivate
                                   / pct-encoded)
                                       ; any Unicode character except: CTL, SP,
                                       ;  DQUOTE, "'", "%" (aside from pct-encoded),
                                       ;  "<", ">", "\", "^", "`", "{", "|", "}"
-
-
 pct-encoded    =  "%" HEXDIG HEXDIG
 ucschar        =  %xA0-D7FF / %xF900-FDCF / %xFDF0-FFEF
                /  %x10000-1FFFD / %x20000-2FFFD / %x30000-3FFFD
--- a/src/oas.md
+++ b/src/oas.md
@@ -581,7 +581,7 @@ ucschar        =  %xA0-D7FF / %xF900-FDCF / %xFDF0-FFEF
 iprivate       =  %xE000-F8FF / %xF0000-FFFFD / %x100000-10FFFD
 ```
 
-Here, `pct-encoded`, `uschar` and `iprivate` definitions are taken from [RFC 6570](https://www.rfc-editor.org/rfc/rfc6570).
+Here, `literals`, `pct-encoded`, `uschar` and `iprivate` definitions are taken from [RFC 6570](https://www.rfc-editor.org/rfc/rfc6570).
 
 ##### Fixed Fields
 
--- a/src/oas.md
+++ b/src/oas.md
@@ -564,11 +564,11 @@ server-url-template            = 1*( literals / template-expression )
 template-expression            = "{" template-expression-param-name "}"
 template-expression-param-name = 1*( %x00-79 / %x7C / %x7E-10FFFF ) ; every UTF8 character except { and }
 
-literals       = 1*( %x21 / %x23-24 / %x26 / %x28-3B / %x3D / %x3F-5B
+literals       = 1*( %x21 / %x23-24 / %x26-3B / %x3D / %x3F-5B
                / %x5D / %x5F / %x61-7A / %x7E / ucschar / iprivate
                / pct-encoded)
                     ; any Unicode character except: CTL, SP,
-                    ;  DQUOTE, "'", "%" (aside from pct-encoded),
+                    ;  DQUOTE, "%" (aside from pct-encoded),
                     ;  "<", ">", "\", "^", "`", "{", "|", "}"
 pct-encoded    =  "%" HEXDIG HEXDIG
 ucschar        =  %xA0-D7FF / %xF900-FDCF / %xFDF0-FFEF
--- a/src/oas.md
+++ b/src/oas.md
@@ -580,7 +580,7 @@ ucschar        =  %xA0-D7FF / %xF900-FDCF / %xFDF0-FFEF
 iprivate       =  %xE000-F8FF / %xF0000-FFFFD / %x100000-10FFFD
 ```
 
-Here, `literals`, `pct-encoded`, `uschar` and `iprivate` definitions are taken from [RFC 6570](https://www.rfc-editor.org/rfc/rfc6570).
+Here, `literals`, `pct-encoded`, `ucschar` and `iprivate` definitions are taken from [RFC 6570](https://www.rfc-editor.org/rfc/rfc6570), incorporating the corrections specified in [Errata 6937](https://www.rfc-editor.org/errata/eid6937) for `literals`.
 
 ##### Fixed Fields
 
--- a/src/oas.md
+++ b/src/oas.md
@@ -560,19 +560,28 @@ An object representing a Server Variable for server URL template substitution.
 The server URL templating is defined by the following [ABNF](https://tools.ietf.org/html/rfc5234) syntax.
 
 ```abnf
-server-url-template            = 1*( server-literal / template-expression )
-server-literal                 = 1*pchar
+server-url-template            = 1*( literals / template-expression )
 template-expression            = "{" template-expression-param-name "}"
 template-expression-param-name = 1*( %x00-79 / %x7C / %x7E-10FFFF ) ; every UTF8 character except { and }
+literals                       = 1*( %x21 / %x23-24 / %x26 / %x28-3B / %x3D / %x3F-5B
+                                  / %x5D / %x5F / %x61-7A / %x7E / ucschar / iprivate
+                                  / pct-encoded)
+                                      ; any Unicode character except: CTL, SP,
+                                      ;  DQUOTE, "'", "%" (aside from pct-encoded),
+                                      ;  "<", ">", "\", "^", "`", "{", "|", "}"
 
-pchar          = unreserved / pct-encoded / sub-delims / ":" / "@"
-unreserved     =  ALPHA / DIGIT / "-" / "." / "_" / "~"
+
 pct-encoded    =  "%" HEXDIG HEXDIG
-sub-delims     =  "!" / "$" / "&" / "'" / "(" / ")"
-               /  "*" / "+" / "," / ";" / "="
+ucschar        =  %xA0-D7FF / %xF900-FDCF / %xFDF0-FFEF
+               /  %x10000-1FFFD / %x20000-2FFFD / %x30000-3FFFD
+               /  %x40000-4FFFD / %x50000-5FFFD / %x60000-6FFFD
+               /  %x70000-7FFFD / %x80000-8FFFD / %x90000-9FFFD
+               /  %xA0000-AFFFD / %xB0000-BFFFD / %xC0000-CFFFD
+               /  %xD0000-DFFFD / %xE1000-EFFFD
+iprivate       =  %xE000-F8FF / %xF0000-FFFFD / %x100000-10FFFD
 ```
 
-Here, `pchar`, `unreserved`, `pct-encoded` and `sub-delims` definitions are taken from [RFC 3986](https://tools.ietf.org/html/rfc3986).
+Here, `pct-encoded`, `uschar` and `iprivate` definitions are taken from [RFC 6570](https://www.rfc-editor.org/rfc/rfc6570).
 
 ##### Fixed Fields
 
--- a/src/oas.md
+++ b/src/oas.md
@@ -557,6 +557,23 @@ servers:
 
 An object representing a Server Variable for server URL template substitution.
 
+The server URL templating is defined by the following [ABNF](https://tools.ietf.org/html/rfc5234) syntax.
+
+```abnf
+server-url-template            = 1*( server-literal / template-expression )
+server-literal                 = 1*pchar
+template-expression            = "{" template-expression-param-name "}"
+template-expression-param-name = 1*( %x00-79 / %x7C / %x7E-10FFFF ) ; every UTF8 character except { and }
+
+pchar          = unreserved / pct-encoded / sub-delims / ":" / "@"
+unreserved     =  ALPHA / DIGIT / "-" / "." / "_" / "~"
+pct-encoded    =  "%" HEXDIG HEXDIG
+sub-delims     =  "!" / "$" / "&" / "'" / "(" / ")"
+               /  "*" / "+" / "," / ";" / "="
+```
+
+Here, `pchar`, `unreserved`, `pct-encoded` and `sub-delims` definitions are taken from [RFC 3986](https://tools.ietf.org/html/rfc3986).
+
 ##### Fixed Fields
 
 | Field Name | Type | Description |
--- a/src/oas.md
+++ b/src/oas.md
@@ -51,7 +51,7 @@ path-template                  = "/" *( path-segment "/" ) [ path-segment ]
 path-segment                   = 1*( path-literal / template-expression )
 path-literal                   = 1*pchar
 template-expression            = "{" template-expression-param-name "}"
-template-expression-param-name = 1*( %x00-79 / %x7C / %x7E-10FFFF ) ; every UTF8 character except { and }
+template-expression-param-name = 1*( %x00-7A / %x7C / %x7E-10FFFF ) ; every UTF8 character except { and }
 
 pchar               = unreserved / pct-encoded / sub-delims / ":" / "@"
 unreserved          = ALPHA / DIGIT / "-" / "." / "_" / "~"
--- a/src/oas.md
+++ b/src/oas.md
@@ -564,12 +564,12 @@ server-url-template            = 1*( literals / template-expression )
 template-expression            = "{" template-expression-param-name "}"
 template-expression-param-name = 1*( %x00-79 / %x7C / %x7E-10FFFF ) ; every UTF8 character except { and }
 
-literals                       = 1*( %x21 / %x23-24 / %x26 / %x28-3B / %x3D / %x3F-5B
-                                  / %x5D / %x5F / %x61-7A / %x7E / ucschar / iprivate
-                                  / pct-encoded)
-                                      ; any Unicode character except: CTL, SP,
-                                      ;  DQUOTE, "'", "%" (aside from pct-encoded),
-                                      ;  "<", ">", "\", "^", "`", "{", "|", "}"
+literals       = 1*( %x21 / %x23-24 / %x26 / %x28-3B / %x3D / %x3F-5B
+               / %x5D / %x5F / %x61-7A / %x7E / ucschar / iprivate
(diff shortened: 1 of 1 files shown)


src/oas.md
```diff
@@ -557,6 +557,23 @@ servers:
 
 An object representing a Server Variable for server URL template substitution.
```
char0n, line 559: Hi @baywet,

As we've exchanged in https://github.com/OAI/OpenAPI-Specification/issues/3256#issuecomment-2514431402, IMHO IServer URL templating is distinct thing from Path templating. 

I would suggest following ABNF (It's basically a reduced and modified version of [openapi-server-url-templating](https://github.com/swaggerexpert/openapi-server-url-templating); modifications come from concensus in https://github.com/OAI/OpenAPI-Specification/pull/4244):

```abnf
server-url-template    = 1*( literals / server-variable )
server-variable        = "{" server-variable-name "}"
server-variable-name   = 1*( %x00-79 / %x7C / %x7E-10FFFF ) ; every UTF8 character except { and }
literals               = 1*( %x21 / %x23-24 / %x26 / %x28-3B / %x3D / %x3F-5B
                       / %x5D / %x5F / %x61-7A / %x7E / ucschar / iprivate
                       / pct-encoded)
                           ; any Unicode character except: CTL, SP,
                           ;  DQUOTE, "'", "%" (aside from pct-encoded),
                           ;  "<", ">", "\", "^", "`", "{", "|", "}"

pct-encoded    =  "%" HEXDIG HEXDIG
ucschar        =  %xA0-D7FF / %xF900-FDCF / %xFDF0-FFEF
               /  %x10000-1FFFD / %x20000-2FFFD / %x30000-3FFFD
               /  %x40000-4FFFD / %x50000-5FFFD / %x60000-6FFFD
               /  %x70000-7FFFD / %x80000-8FFFD / %x90000-9FFFD
               /  %xA0000-AFFFD / %xB0000-BFFFD / %xC0000-CFFFD
               /  %xD0000-DFFFD / %xE1000-EFFFD
iprivate       =  %xE000-F8FF / %xF0000-FFFFD / %x100000-10FFFD
```

**Reasoning**:

1. Alignment with https://www.rfc-editor.org/rfc/rfc6570. IMHO Server URL templating has nothing to do with RFC 3986, as server variables can appear in any place within the server URL not just the [Path](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3) section as in Path Templating. This means that `literals` non-terminal needs to be used from RFC 6570 to take that into account.
2. Naming - Server URL templates don't don't mention explicitly concept of `template expression` at all. Server URL templating mentions concept of `server variable substitution`. So the ABNF naming is revolving around this. 

---

`literals`, `iprivate`, `ucschar` and `pct-encoded` are taken directly from https://www.rfc-editor.org/rfc/rfc6570

baywet: Thanks for the feedback!

1. I'm not sure why RFC6570 is relevant here? OAS does not depend on it, it only has a few interop considerations in the appendix.
2. Names: I tried to align with what we've built so far for path templates, but I'm happy to change the names if we want to clear confusion/think they are semantically different.

char0n: > I'm not sure why RFC6570 is relevant here? OAS does not depend on it, it only has a few interop considerations in the appendix.

I'll try to elaborate fully on this, so that we build shared understanding. I'll explain my reasoning which should help us move forward or dismiss my logic as nonsensical.

The current `server-url-template` non-terminal as defined in this PR, will not parse an absolute URI (https://datatracker.ietf.org/doc/html/rfc3986#section-4.3) like `https://example.com`. It will fail on a server URL template like this as well `{scheme}://example.com`. If you build ABNF grammar on top of RFC 3986 for `server URL templating`, which I've tried during the first implementation of [openapi-server-url-templating](https://github.com/swaggerexpert/openapi-server-url-templating), you'll end up with extremely large and unreliable grammar, which will have to alternate a lot (I've never finished it as it was so ambiguous that I was having problem keeping it together). And then you'll get into a situation when `{server-variable}` can span multiple non-terminals of absolute URI like `scheme` or  `hier-part` (fully or partially). 

`pchar` from `path templating` is used to represent group of literals allowed in [path](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3). It also represents characters allowed in [query](https://datatracker.ietf.org/doc/html/rfc3986#section-3.4) and [fragment](https://datatracker.ietf.org/doc/html/rfc3986#section-3.5) - but as established, query and fragment cannot be part of path template. We specifically use it to express `path-absolute` non-terminal (https://datatracker.ietf.org/doc/html/rfc3986#section-3.3) with support of template expressions. It doesn't encompass allowed characters from the entire URI structure. 

And that is where `path templating` differs from `server URL templating`. `Path templating` only deals with `path-absolute` non-terminal (https://datatracker.ietf.org/doc/html/rfc3986#section-3.3), apart from `server URL templating` which deals with the whole URI. This has signification implications on the ABNF grammar.

Now to avoid all the complexities of alteration and ambiguity described in the second paragraph of this elaboration, we can build on ideas of smart people that defined [RFC 6570](https://www.rfc-editor.org/rfc/rfc6570). They faced exactly the same problem - coming up with ABNF grammar that will allow substituting template expressions for values in any part of the URI. Their ABNF start rule is:

```abnf
URI-Template  = *( literals / expression )
```


So the entire URI si decomposed to either URI literals or template expressions. URI literals must be defined carefully to allow to encompass any part of the URI.

```abnf
 literals      =  %x21 / %x23-24 / %x26 / %x28-3B / %x3D / %x3F-5B
                   /  %x5D / %x5F / %x61-7A / %x7E / ucschar / iprivate
                   /  pct-encoded
                        ; any Unicode character except: CTL, SP,
                        ;  DQUOTE, "'", "%" (aside from pct-encoded),
                        ;  "<", ">", "\", "^", "`", "{", "|", "}"
```

literals in RFC 6570 is like an abstraction layer designed to simplify handling URI templates without having to explicitly parse or manage the individual parts of a URI as defined in RFC 3986.

IMHO We cannot really say that `server URL templating` is a subset of  URI-templating because of following definitions (but IMHO we can say it's derived from it):

```anbf
server-url-template    = 1*( literals / server-variable ) ; vs URI-Template  = *( literals / expression )
server-variable-name   = 1*( %x00-79 / %x7C / %x7E-10FFFF ) ; every UTF8 character except { and }; vs variable-list =  varspec *( "," varspec )
```

The first one is the implication of following OpenAPI excerpt:
> If the servers field is not provided, or is an empty array, the default value would be a [Server Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.4.md#server-object) with a [url](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.4.md#server-url) value of /

The second one is local consensus of how the server variable name can be defined.

ralfhandl: I am not sold yet on basing server URL templates on a different RFC than path templates. The main difference between the two is that server URL templates may also contain scheme and authority components in RFC3986 terminology, so I would actually prefer basing both flavors of URL templates on RFC3986.

baywet: Thank you for taking the time to provide a detailed explanation of your thinking process here.

That's a question I had for the group: do we want to ensure people can't build server URIs that end up being invalid based on the "template" structure?

I was thinking that no we don't want to because:
- it might be more restrictive than what can currently be done/is supported by the current tools. And the work on path templating clearly established we don't want to restrain the "existing situation" any further.
- ultimately, we don't control what's in the variable values, which could render the whole resulting server URI invalid anyway.

This is why I put together this initial draft with simpler grammar that's similar to the path templating, granted we can adjust it to account for missing characters etc...

handrews:
After discussion in TDC we agreed that `pchar` is not the right rule and we should use the general literals per RFC6570.

src/oas.md
baywet: Further discussions on the TSC call today with @ralfhandl and @handrews allowed me to understand things a bit better.
I've just pushed a change with pchar being replaced by literal (and the other definitions you've provided)

I have kept the naming for template-expression and template-expression-param-name, since they are identical to the ones in path templating. We could imagine "refactoring" them, but it'd probably make things harder. At least implementer know they are the same (should we add a comment in that sense?)

char0n: > I am not sold yet on basing server URL templates on a different RFC than path templates. The main difference between the two is that server URL templates may also contain scheme and authority components in RFC3986 terminology, so I would actually prefer basing both flavors of URL templates on RFC3986.

@ralfhandl, sure I understand the need to stay within the RFC 3986. My initial experimentation about six months about around creating ABNF for Server URL templating in [openapi-server-url-templating](https://github.com/swaggerexpert/openapi-server-url-templating) was actually trying to achieve this. After couple of unsuccessful attempts I ended up with RFC 6570.

> That's a question I had for the group: do we want to ensure people can't build server URIs that end up being invalid based on the "template" structure?

IMHO we MUST NOT. This would break a lot of stuff. We can only say that after successful server variable substitution the server URL must be valid URL or relative reference. So I totally agree with your two conclusions.

> Further discussions on the TSC call today with @ralfhandl and @handrews allowed me to understand things a bit better. I've just pushed a change with pchar being replaced by literal (and the other definitions you've provided).

ACK

src/oas.md
```diff
@@ -557,6 +557,32 @@ servers:
...
+ucschar        =  %xA0-D7FF / %xF900-FDCF / %xFDF0-FFEF
+               /  %x10000-1FFFD / %x20000-2FFFD / %x30000-3FFFD
+               /  %x40000-4FFFD / %x50000-5FFFD / %x60000-6FFFD
+               /  %x70000-7FFFD / %x80000-8FFFD / %x90000-9FFFD
+               /  %xA0000-AFFFD / %xB0000-BFFFD / %xC0000-CFFFD
+               /  %xD0000-DFFFD / %xE1000-EFFFD
+iprivate       =  %xE000-F8FF / %xF0000-FFFFD / %x100000-10FFFD
+```
+
+Here, `pct-encoded`, `uschar` and `iprivate` definitions are taken from [RFC 6570](https://www.rfc-editor.org/rfc/rfc6570).
```
char0n, line 584: `literals` comes from RFC 6570 as well. It's been omitted from the description sentence.

```suggestion
Here, `literals`, `pct-encoded`, `uschar` and `iprivate` definitions are taken from [RFC 6570](https://www.rfc-editor.org/rfc/rfc6570).
```

src/oas.md
```diff
@@ -557,6 +557,32 @@ servers:
...
+template-expression            = "{" template-expression-param-name "}"
+template-expression-param-name = 1*( %x00-79 / %x7C / %x7E-10FFFF ) ; every UTF8 character except { and }
+literals                       = 1*( %x21 / %x23-24 / %x26 / %x28-3B / %x3D / %x3F-5B
+                                  / %x5D / %x5F / %x61-7A / %x7E / ucschar / iprivate
+                                  / pct-encoded)
+                                      ; any Unicode character except: CTL, SP,
+                                      ;  DQUOTE, "'", "%" (aside from pct-encoded),
+                                      ;  "<", ">", "\", "^", "`", "{", "|", "}"
+
+
```
baywet, line 573:
```suggestion

literals                       = 1*( %x21 / %x23-24 / %x26 / %x28-3B / %x3D / %x3F-5B
                                  / %x5D / %x5F / %x61-7A / %x7E / ucschar / iprivate
                                  / pct-encoded)
                                      ; any Unicode character except: CTL, SP,
                                      ;  DQUOTE, "'", "%" (aside from pct-encoded),
                                      ;  "<", ">", "\", "^", "`", "{", "|", "}"
```

baywet:
@handrews @char0n @ralfhandl I think we're ready for final review here :)

src/oas.md
```diff
@@ -557,6 +557,32 @@ servers:
 
 An object representing a Server Variable for server URL template substitution.
 
+The server URL templating is defined by the following [ABNF](https://tools.ietf.org/html/rfc5234) syntax.
+
+```abnf
+server-url-template            = 1*( literals / template-expression )
+template-expression            = "{" template-expression-param-name "}"
+template-expression-param-name = 1*( %x00-79 / %x7C / %x7E-10FFFF ) ; every UTF8 character except { and }
```
char0n, line 565: > > I have kept the naming for template-expression and template-expression-param-name, since they are identical to the ones in path templating. We could imagine "refactoring" them, but it'd probably make things harder. At least implementer know they are the same (should we add a comment in that sense?)

@baywet is it possible to reconsider the naming here? This naming `template-expression` and `template-expression-param-name` comes directly from [openapi-path-templating](https://github.com/swaggerexpert/openapi-path-templating). My intention when creating this naming was to have `template-expression` and `template-expression-param-name` specific to `OpenAPI Path Templating` concept only. The intention is quite explicit from the `template-expression-param-name` non-terminal name specifically - it tries to tells us that the identifier (template variable) inside the brackets represents an OpenAPI `Parameter Object`.`name` fixed field. (which doesn't apply to Server URL templating).

By that logic, that's why I've suggested non-terminals in this ABNF to be server URL templating specific:

```abnf
server-variable        = "{" server-variable-name "}"
server-variable-name   = 1*( %x00-79 / %x7C / %x7E-10FFFF ) ; every UTF8 character except { and }
```

If we want to go for something generic that is identical in both ABNF grammars, I would probably suggest something like:

```abnf
template-expression                = "{" template-expression-identifier "}"
template-expression-identifier = 1*( %x00-79 / %x7C / %x7E-10FFFF ) ; every UTF8 character except { and }
``` 
or
```abnf
template-expression                = "{" template-variable "}"
template-variable = 1*( %x00-79 / %x7C / %x7E-10FFFF ) ; every UTF8 character except { and }
``` 

But IMHO there is benefit in keeping the naming distinct as it will allow to evolve those non-terminals independently in future. I can imagine, future OpenAPI versions might adopt further concepts (expadning) from RFC 6570, while it doesn't make sense for Server URL templating to adopt those.

src/oas.md
```diff
@@ -557,6 +557,31 @@ servers:
...
+
+literals       = 1*( %x21 / %x23-24 / %x26 / %x28-3B / %x3D / %x3F-5B
```
char0n, line 567: @baywet do we want to apply your own errata here https://www.rfc-editor.org/errata/eid6937 ?

src/oas.md
```diff
@@ -557,6 +557,31 @@ servers:
...
+               / %x5D / %x5F / %x61-7A / %x7E / ucschar / iprivate
+               / pct-encoded)
+                    ; any Unicode character except: CTL, SP,
+                    ;  DQUOTE, "'", "%" (aside from pct-encoded),
+                    ;  "<", ">", "\", "^", "`", "{", "|", "}"
```
ralfhandl, line 572:
```suggestion
literals       = 1*( %x21 / %x23-24 / %x26-3B / %x3D / %x3F-5B
               / %x5D / %x5F / %x61-7A / %x7E / ucschar / iprivate
               / pct-encoded)
                    ; any Unicode character except: CTL, SP,
                    ;  DQUOTE, "%" (aside from pct-encoded),
                    ;  "<", ">", "\", "^", "`", "{", "|", "}"
```

ralfhandl (changes requested):
Align with https://www.rfc-editor.org/errata/eid6937

src/oas.md
baywet: 🤦 I had forgotten about this one... Thanks for pointing it out. Since Ralf made a suggestion, I'll go merge it and resolve this one.

src/oas.md
baywet: I can buy the future changes that diverge, creating confusion. Updated to a distinct name, let me know if I missed anything.

char0n (approved):
LGTM, I've tested it as well, compiles and parses correctly.

src/oas.md
```diff
@@ -557,6 +557,31 @@ servers:
...
+ucschar        =  %xA0-D7FF / %xF900-FDCF / %xFDF0-FFEF
+               /  %x10000-1FFFD / %x20000-2FFFD / %x30000-3FFFD
+               /  %x40000-4FFFD / %x50000-5FFFD / %x60000-6FFFD
+               /  %x70000-7FFFD / %x80000-8FFFD / %x90000-9FFFD
+               /  %xA0000-AFFFD / %xB0000-BFFFD / %xC0000-CFFFD
+               /  %xD0000-DFFFD / %xE1000-EFFFD
+iprivate       =  %xE000-F8FF / %xF0000-FFFFD / %x100000-10FFFD
+```
+
+Here, `literals`, `pct-encoded`, `uschar` and `iprivate` definitions are taken from [RFC 6570](https://www.rfc-editor.org/rfc/rfc6570).
```
char0n, line 583: Final note: do we want to mention the application of [errata](https://www.rfc-editor.org/errata/eid6937) here? It might be confusing to parser authors to see that literals is different that the actual RFC 6570.

baywet: What do you think of that?
```suggestion
Here, `literals`, `pct-encoded`, `uschar` and `iprivate` definitions are taken from [RFC 6570](https://www.rfc-editor.org/rfc/rfc6570), the following [errata](https://www.rfc-editor.org/errata/eid6937) has also been taken into account.
```

char0n: Feel like a separate sentence:

```suggestion
Here, `literals`, `pct-encoded`, `ucschar` and `iprivate` definitions are taken from [RFC 6570](https://www.rfc-editor.org/rfc/rfc6570). The following [errata](https://www.rfc-editor.org/errata/eid6937) has also been taken into account for `literals`.
```

Alternative suggestion which allows to read the sentence more naturaly:

```suggestion
Here, `literals`, `pct-encoded`, `ucschar` and `iprivate` definitions are taken from [RFC 6570](https://www.rfc-editor.org/rfc/rfc6570), incorporating the corrections specified in [Errata ID 6937](https://www.rfc-editor.org/errata/eid6937) for `literals`.
```

---

**NOTE**: there is also a type in the sentence: `uschar` vs `ucschar`. Both suggestions address that.

handrews:
Random question- why does this PR start with "docs:"?  It's a spec PR, not a docs PR.  I keep overlooking it because of this.

baywet:
> Random question- why does this PR start with "docs:"?  It's a spec PR, not a docs PR.  I keep overlooking it because of this.

Muscle memory: on most repos I work on, we follow semantic commits. And usually when I touch Md files there, it's documentation. 

GitHub CLI: -f uses the first commit when creating a PR.

handrews:
@baywet thank you for satisfying my curiosity!  Yeah, this project is a little unusual in that the spec markdown is really the "source".  "Docs" to me are things like CONTRIBUTING.md and the README.

ralfhandl (approved):

src/oas.md
```diff
@@ -557,6 +557,31 @@ servers:
 
 An object representing a Server Variable for server URL template substitution.
 
+The server URL templating is defined by the following [ABNF](https://tools.ietf.org/html/rfc5234) syntax.
+
+```abnf
+server-url-template  = 1*( literals / server-variable )
+server-variable      = "{" server-variable-name "}"
+server-variable-name = 1*( %x00-79 / %x7C / %x7E-10FFFF ) ; every UTF8 character except { and }
```
char0n, line 565: This is incorrect as it excludes the `z` character. Whenever the server variable name contains `z` the ABNF fails.
Here is a quick suggestion to fix this temporarily including `z` explicitly. This also affects **ABNF** grammar for Path Templating in the same way.

```suggestion
server-variable-name = 1*( %x00-79 / %x7C / %x7E-10FFFF / "z" ) ; every UTF8 character except { and }
```

char0n:
```suggestion
server-variable-name = 1*( %x00-7A / %x7C / %x7E-10FFFF ) ; every UTF8 character except { and }
```

baywet: ooooh thanks for catching that! this is probably a case of me not seeing properly and skipping a line in the ASCII table. I'll also correct the path templating one in a minute.

char0n: np. FYI for path templating I've issued a separate fixing PR in https://github.com/OAI/OpenAPI-Specification/pull/4279

char0n:
I'm doing some final testing on extended test suite - will check back here when everything glows green.

char0n (approved):
LGTM, all tests on extended test-suite turned  green

miqui (approved):
looks good - discussed during TDC.

PR #4265, opened 2024-12-18
Editorial change: Include Header Object into 'Working With Examples' section

char0n:
When studying changes of OpenAPI 3.0.4 (from the OpenAPI 3.0.3), I've noticed a section [Working with examples](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.4.md#working-with-examples). Everything that is written there does apply to `Header Object`. In other section when something apply to `Header Object` it is enumerated explicitly. But in `Working with examples` section, the header is missing. IMHO it's an unintended omission, and this PR is trying to remedy it.

### Appendix B.

image

### Appendix C.

image

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -2211,7 +2211,7 @@ Tooling implementations MAY choose to validate compatibility automatically, and 
 
 ##### Working with Examples
 
-Example Objects can be used in [Parameter Objects](#parameter-object), [Header Objects](#header-object) and [Media Type Objects](#media-type-object).
+Example Objects can be used in [Parameter Objects](#parameter-object), [Header Objects](#header-object), and [Media Type Objects](#media-type-object).
 In both Objects, this is done through the `examples` (plural) field.
 However, there are several other ways to provide examples: The `example` (singular) field that is mutually exclusive with `examples` in all three Objects, and two keywords (the deprecated singular `example` and the current plural `examples`, which takes an array of examples) in the [Schema Object](#schema-object) that appears in the `schema` field of all three Objects.
 Each of these fields has slightly different considerations.
--- a/src/oas.md
+++ b/src/oas.md
@@ -2223,7 +2223,7 @@ The mutually exclusive fields in the Parameter, Header, or Media Type Objects ar
 The exact serialization and encoding is determined by various fields in the Parameter Object, Header Object, or in the Media Type Object's [Encoding Object](#encoding-object).
 Because examples using these fields represent the final serialized form of the data, they SHALL _override_ any `example` in the corresponding Schema Object.
 
-The singular `example` field in the Parameter, Header or Media Type Object is concise and convenient for simple examples, but does not offer any other advantages over using Example Objects under `examples`.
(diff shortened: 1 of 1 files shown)


src/oas.md
```diff
@@ -2211,19 +2211,19 @@ Tooling implementations MAY choose to validate compatibility automatically, and
...
-The exact serialization and encoding is determined by various fields in the Parameter Object, or in the Media Type Object's [Encoding Object](#encoding-object).
+The mutually exclusive fields in the Parameter, Header or Media Type Objects are used to show example values which SHOULD both match the schema and be formatted as they would appear as a serialized parameter or within a media type representation.
+The exact serialization and encoding is determined by various fields in the Parameter Object, Header Object, or in the Media Type Object's [Encoding Object](#encoding-object).
 Because examples using these fields represent the final serialized form of the data, they SHALL _override_ any `example` in the corresponding Schema Object.
 
-The singular `example` field in the Parameter or Media Type Object is concise and convenient for simple examples, but does not offer any other advantages over using Example Objects under `examples`.
+The singular `example` field in the Parameter, Header or Media Type Object is concise and convenient for simple examples, but does not offer any other advantages over using Example Objects under `examples`.
```
ralfhandl, line 2226:
```suggestion
The singular `example` field in the Parameter, Header, or Media Type Object is concise and convenient for simple examples, but does not offer any other advantages over using Example Objects under `examples`.
```

ralfhandl (changes requested):
Minor changes

char0n:
@ralfhandl thanks for catching all inconsistencies!

handrews (dismissed):

src/oas.md
```diff
@@ -2211,19 +2211,19 @@ Tooling implementations MAY choose to validate compatibility automatically, and
...
 Each of these fields has slightly different considerations.
 
 The Schema Object's fields are used to show example values without regard to how they might be formatted as parameters or within media type representations.
 The `examples` array is part of JSON Schema and is the preferred way to include examples in the Schema Object, while `example` is retained purely for compatibility with older versions of the OpenAPI Specification.
 
-The mutually exclusive fields in the Parameter or Media Type Objects are used to show example values which SHOULD both match the schema and be formatted as they would appear as a serialized parameter or within a media type representation.
-The exact serialization and encoding is determined by various fields in the Parameter Object, or in the Media Type Object's [Encoding Object](#encoding-object).
+The mutually exclusive fields in the Parameter, Header, or Media Type Objects are used to show example values which SHOULD both match the schema and be formatted as they would appear as a serialized parameter or within a media type representation.
```
mikekistler, line 2222: Do we also need this?
```suggestion
The mutually exclusive fields in the Parameter, Header, or Media Type Objects are used to show example values which SHOULD both match the schema and be formatted as they would appear as a serialized parameter, serialized header, or within a media type representation.
```

mikekistler (approved):
Looks good. 👍 

I left one question/suggestion that you may want to consider.

src/oas.md
char0n: Let's go for it, make sense to me. Thank you!

ralfhandl (approved):

handrews (approved):

Issue #3866, opened 2024-05-30
Add dark mode support to the HTML versions of the OpenAPI specs


mikekistler:
Update the rendering process for the HTML versions of the OpenAPI specs (using ReSpec) to generate pages that support dark mode.

ReSpec has added dark mode support: https://github.com/speced/respec/issues/2651

Related: #3576

handrews:
Paging @Bellangelo (we're working on making it so that we can assign issues to you, but haven't quite sorted that out yet).

Bellangelo:
Unfortunately, this seems not to be so easy. Our `js/respec-oai.js` file loads this css file `https://www.w3.org/StyleSheets/TR/2016/base.css` which overrides most styles. There is a `https://www.w3.org.org/StyleSheets/TR/2026/dark.css` but even if we add it, it still gets overridden by the `base.css`.
It looks like we need to give more effort in this issue https://github.com/OAI/OpenAPI-Specification/issues/3845 which will help us have more control over the `respec-oai.js`

ralfhandl:
@mikekistler Don't know what changed in the last two weeks, now dark mode seems to work:
* https://ralfhandl.github.io/OpenAPI-Specification/oas/latest.html

Quite some fine-tuning needed for dark mode to really work:
- different OAS logos for light/dark mode
- table and code-block background color depending on light/dark mode
- ...

I'll gladly leave this to someone fluent in CSS 😁

pavelkornev:
I have started on this, please assign to my name.

pavelkornev:
> Hi [@pavelkornev](https://github.com/pavelkornev), I used some of your CSS in this [PR](https://github.com/OAI/OpenAPI-Specification/pull/4268) to continue on this work. If you have already a PR that I missed then I can close mine and use yours instead.

The approach i tried won't work because then switcher between Dark/Light mode doesn't work properly. Try both scenarious:
1. System settings set to **Light** mode: try switch between Dark/Light using switcher
2. System settings set to **Dark** mode: try switch between Dark/Light using switcher

Bellangelo:
> > Hi [@pavelkornev](https://github.com/pavelkornev), I used some of your CSS in this [PR](https://github.com/OAI/OpenAPI-Specification/pull/4268) to continue on this work. If you have already a PR that I missed then I can close mine and use yours instead.
> 
> The approach i tried won't work because then switcher between Dark/Light mode doesn't work properly. Try both scenarious:
> 
[...]

PR #4268, opened 2024-12-19
Add dark mode support to the HTML versions of the OpenAPI specs

Bellangelo:
Closes https://github.com/OAI/OpenAPI-Specification/issues/3866

Except from adding the necessary CSS rules I have also moved it into its own file here: https://github.com/OAI/OpenAPI-Specification/commit/b3949bdf468dbdf565d0854a9abef8c6bbdcbd88

This declutters a little bit the `md2html.js`. Plus, we can also use the IDE for styling and formatting.

The CSS is heavily inspired by the work of @pavelkornev, see https://github.com/OAI/OpenAPI-Specification/issues/3866#issuecomment-2327406991

End results: https://jsfiddle.net/gu23x9kb/


ralfhandl (changes requested):
As @pavelkornev points out in https://github.com/OAI/OpenAPI-Specification/issues/3866#issuecomment-2553095801 this only works on my Mac if the system appearance is set to Dark. If it is set to Light, or set to Auto and is currently Light, then the switch doesn't work.

Bellangelo:
> As @pavelkornev points out in [#3866 (comment)](https://github.com/OAI/OpenAPI-Specification/issues/3866#issuecomment-2553095801) this only works on my Mac if the system appearance is set to Dark. If it is set to Light, or set to Auto and is currently Light, then the switch doesn't work.

@ralfhandl Oh yes... I had encapsulated the code under the `prefers-color-scheme` which was running only on dark mode. Fixed in https://github.com/OAI/OpenAPI-Specification/pull/4268/commits/973999a6a7c1e774b72ca44aa24cf53d4647f782

ralfhandl:
Still doesn't work if system is in Light mode:
image

and
image

Background stays white, only text colors change.

ralfhandl (changes requested):
On MacOS only works if system is in dark mode.

Bellangelo:
> On MacOS only works if system is in dark mode.

@ralfhandl That's weird.. I am not getting the same results as you. I am testing it on a MacOS and it perfectly works with Firefox and Safari. Could it be a browser setting? I will further look into it.

Bellangelo:
@ralfhandl I have managed to reproduce it with Firefox. The problem is that respec adds this link
```html
<link rel="stylesheet" media="(prefers-color-scheme: dark)" href="https://www.w3.org/StyleSheets/TR/2021/dark.css">
```
Once again, due to the `(prefers-color-scheme: dark)` it doesn't work correctly. For this reason, i have copied/pasted the changes from the file into our `main.css` here: https://github.com/OAI/OpenAPI-Specification/pull/4268/commits/e25d7530c2a6a5d7c5397b6da4cdb93758e98076

Although, this seems to be fixing the issue it still loads unnecessarily the `dark.css`. Would it be ok if I changed the `build.sh`to parse the output and remove it? Or are we ok with the extra request?

**Note:** I will fix the tests after we agree on the solution.

ralfhandl:
> i have copied/pasted the changes from the file into our `main.css`

Is that necessary? I'd prefer to keep our CSS as slim as possible.

ralfhandl:
> Would it be ok if I changed the `build.sh`to parse the output and remove it?

Why not, if that works better?

lornajane:
@swaldron58 can you check if our theme fits this?

Bellangelo:
> Is that necessary? I'd prefer to keep our CSS as slim as possible.
@ralfhandl I think so. The implementation depends on how Respec handles dark mode, which is by adding the `darkmode` class in the `body` tag. Since we apply the styling based on this class and not by checking the preferred mode the `media="(prefers-color-scheme: dark)"` becomes an obstacle.

> Why not, if that works better?
@ralfhandl I have changed the `build.sh` in this [commit](https://github.com/OAI/OpenAPI-Specification/pull/4268/commits/d5f765eaa96f2bd367e6d583cb66cbbdd42bcbfe) and its tests [here](https://github.com/OAI/OpenAPI-Specification/pull/4268/commits/ffa5176bc27e721624a43430fad3f0378799063e).

ralfhandl:
@Bellangelo Thanks for the detailed analysis!

> It seems that we cannot stop the loading of the `dark.css` without changing quite a few things.

I'd prefer not to go that path, given how fiddly it was to get back from a modified ReSpec v21 to an unmodified ReSpec v35.

How far can we get with just additional CSS?

Bellangelo:
> How far can we get with just additional CSS?

@ralfhandl What exactly do you mean? The dark mode should be working right now.

lornajane:
To clarify, we don't want any customisation on respec itself, we will use the published version of that library. @ralfhandl 's question is whether we can achieve dark mode purely with CSS styles?

ralfhandl (approved):
Works fine in Chrome and Firefox on Windows. Preview:
* https://ralfhandl.github.io/OpenAPI-Specification/oas/v3.1.1.html

mikekistler (approved):
I'm not qualified to review the actual code changes, but I tested this on my Mac in both Safari and Chrome and it works great.

PR #4272, opened 2024-12-19
Introduce constraints for Server Object url fixed field

char0n:
I went ahead with this suggestion, as it is something that is not obvious from reading the OpenAPI spec, but it becomes obvious for the tooling authors during implementations.

`Server Object`.`url` field is defined as `A URL to the target host`.

OpenAPI further stipulates: 

> Unless specified otherwise, all fields that are URLs MAY be relative references as defined by [RFC3986](https://tools.ietf.org/html/rfc3986#section-4.2)

Looking at RFC 3986, the `Server Object`.`url` can be one of the following:

```abnf
URI           = scheme ":" hier-part [ "?" query ] [ "#" fragment ]
relative-ref  = relative-part [ "?" query ] [ "#" fragment ]
```

Both of these non-terminals allow `query` and `fragment`. But effectively, we cannot allow them as OpenAPI further stipulates:

> The path is appended to the URL from the [Server Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.1.md#server-object) in order to construct the full URL 

Appending to URL or resolved relative reference containing query or fragment will end up corrupting the resulting URL.

---

There is another aspect of `Server Object`.`url`, and that's being [Scheme-Based Normalization](https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.3). When we define server URL as `https://exmaple.com/`, the tooling tends to remove the last character from the URL, if the character=`/` before the path from `Paths Object` is appended to it. This is possible because of equivalence as described by Scheme-Based Normalization.

```
// server URL
http://example.com/

// paths
/path1
/path2

// we want to end up here
http://example.com/path1
http://example.com/path2

// we might end up here
http://example.com//path1
http://example.com//path2
```

AFAICT we cannot apply [Scheme-Based Normalization](https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.3) to URL like `http://example.com/initial-path`. `http://example.com/initial-path` is not equivalent anymore to `http://example.com/initial-path/`.

I guess my question is, if we shouldn't restrict  `Server Object`.`url` not to end with explicit `/`?

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -455,7 +455,7 @@ An object representing a Server.
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
-| <a name="server-url"></a>url | `string` | **REQUIRED**. A URL to the target host. This URL supports Server Variables and MAY be relative, to indicate that the host location is relative to the location where the document containing the Server Object is being served. Query and fragment MUST not be part of this URL. Variable substitutions will be made when a variable is named in `{`braces`}`. |
+| <a name="server-url"></a>url | `string` | **REQUIRED**. A URL to the target host. This URL supports Server Variables and MAY be relative, to indicate that the host location is relative to the location where the document containing the Server Object is being served. Query and fragment MUST NOT be part of this URL. Variable substitutions will be made when a variable is named in `{`braces`}`. |
 | <a name="server-description"></a>description | `string` | An optional string describing the host designated by the URL. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation. |
 | <a name="server-variables"></a>variables | Map[`string`, [Server Variable Object](#server-variable-object)] | A map between a variable name and its value. The value is used for substitution in the server's URL template. |
 
--- a/src/oas.md
+++ b/src/oas.md
@@ -455,7 +455,7 @@ An object representing a Server.
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
-| <a name="server-url"></a>url | `string` | **REQUIRED**. A URL to the target host. This URL supports Server Variables and MAY be relative, to indicate that the host location is relative to the location where the document containing the Server Object is being served. Query and fragment MUST not be part of a URL. Variable substitutions will be made when a variable is named in `{`braces`}`. |
+| <a name="server-url"></a>url | `string` | **REQUIRED**. A URL to the target host. This URL supports Server Variables and MAY be relative, to indicate that the host location is relative to the location where the document containing the Server Object is being served. Query and fragment MUST not be part of this URL. Variable substitutions will be made when a variable is named in `{`braces`}`. |
 | <a name="server-description"></a>description | `string` | An optional string describing the host designated by the URL. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation. |
 | <a name="server-variables"></a>variables | Map[`string`, [Server Variable Object](#server-variable-object)] | A map between a variable name and its value. The value is used for substitution in the server's URL template. |
 
--- a/src/oas.md
+++ b/src/oas.md
@@ -455,7 +455,7 @@ An object representing a Server.
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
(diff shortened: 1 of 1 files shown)


atimark (approved):

char0n:
@ralfhandl thanks, what about the [Scheme-Based Normalization](https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.3) issue? Do you think it is warranted to to verbalize this into a `url` field description? Something like `This URL SHOULD not end with `/` character.`

src/oas.md
```diff
@@ -455,7 +455,7 @@ An object representing a Server.
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
-| <a name="server-url"></a>url | `string` | **REQUIRED**. A URL to the target host. This URL supports Server Variables and MAY be relative, to indicate that the host location is relative to the location where the document containing the Server Object is being served. Variable substitutions will be made when a variable is named in `{`braces`}`. |
+| <a name="server-url"></a>url | `string` | **REQUIRED**. A URL to the target host. This URL supports Server Variables and MAY be relative, to indicate that the host location is relative to the location where the document containing the Server Object is being served. Query and fragment MUST not be part of this URL. Variable substitutions will be made when a variable is named in `{`braces`}`. |
```
char0n, line 458: One additional change: `MUST not` vs `MUST NOT`

```suggestion
| <a name="server-url"></a>url | `string` | **REQUIRED**. A URL to the target host. This URL supports Server Variables and MAY be relative, to indicate that the host location is relative to the location where the document containing the Server Object is being served. Query and fragment MUST NOT be part of this URL. Variable substitutions will be made when a variable is named in `{`braces`}`. |
```

miqui (approved):
Perhaps we should wait until the next tdc to merge?

ralfhandl (approved):

miqui:
@baywet ping! check it out as well.

miqui:
I might have missed this, but this fixed should target 3.2 only? (cc: @ralfhandl , @lornajane )

baywet (approved):
Changes look good, but I think the wrong branch is being targeted at the moment.

baywet:
Some more context around the branches: I asked that same specific question on the TDC call a couple of weeks ago. The agreement we came to with @ralfhandl and @handrews was that, unless things are specific to a version of the spec, they should target 3.2-dev. Once 3.2-dev is released, all the changes that can be applied to 3.1 will be collected and applied in a single shot to avoid confusion. They _might_ also be applied to 3.0, even though we agreed last time that 3.0.4 was the last release.

karenetheridge:
> I guess my question is, if we shouldn't restrict Server Object.url not to end with explicit /?

How about something like "Server Object.url values that end with `/` MUST be treated as equivalent to values without a trailing `/` when prepended to path values" (or "path templates"?) ?

This relates to what we need to when matching request URIs against `/paths/<path template>` entries in the absence of any Server objects -- for a request URI of "https://example.com", the path portion of the URI is not "" but rather "/", when considering a potential match against Paths Objects.  (ref https://github.com/karenetheridge/OpenAPI-Modern/blob/v0.076/lib/OpenAPI/Modern.pm#L536)

ralfhandl:
> Changes look good, but I think the wrong branch is being targeted at the moment.

CONTRIBUTING.md states that pull requests should [Target the earliest active `*-dev` branch](https://github.com/OAI/OpenAPI-Specification/blob/main/CONTRIBUTING.md#target-the-earliest-active--dev-branch), which is the case here.

ralfhandl:
> How about something like "Server Object.url values that end with `/` MUST be treated as equivalent to values without a trailing `/` when prepended to path values" (or "path templates"?) ?

URLs with and without a trailing slash are different. Some servers redirect from one to the other, and some servers serve the same content without redirect, so in some cases they may seem equivalent, but I would hesitate to make that a rule or even a recommendation in OpenAPI.

@karenetheridge If you feel we should discuss this in more detail, please open a new issue for it.

baywet:
> > Changes look good, but I think the wrong branch is being targeted at the moment.
> 
> CONTRIBUTING.md states that pull requests should [Target the earliest active `*-dev` branch](https://github.com/OAI/OpenAPI-Specification/blob/main/CONTRIBUTING.md#target-the-earliest-active--dev-branch), which is the case here.

@ralfhandl I'm confused then, should I have targeted 3.1 as well with my recent pull requests then?

PR #4279, opened 2024-12-25
Fix Path Templating ABNF grammar

char0n:
template-expression-param-name was excluding "z" character.

---

The issue was originally uncovered in [Server URL templating PR](https://github.com/OAI/OpenAPI-Specification/pull/4264#discussion_r1897325305), but applies to Path Templating ABNF as well.


ralfhandl (approved):

miqui (approved):
looks good  - discussed during TDC

PR #4287, opened 2025-01-03
Add tag kinds registry to the spec site

lornajane:
This goes with #4288 , the additions to the Tag object. We agreed that we should have a registry of some common values for the `kind` field, so this pull request adds that registry with some starter values and a warning about the feature not being released yet

### DIFF
--- a/registries/_tag-kind/nav.md
+++ b/registries/_tag-kind/nav.md
@@ -7,7 +7,7 @@ layout: default
 
 {% capture summary %}
 Tags with `kind: nav` are used in documentation to group operations into sections.
-In most cases, an operation only has one tag of this tag so it belongs in one section.
+In most cases, an operation only has one tag of this type so it belongs in one section.
 {% endcapture %}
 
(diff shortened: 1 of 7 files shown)


registries/_tag-kind/nav.md
```diff
@@ -0,0 +1,13 @@
+---
+owner: lornajane
+issue:
+description: "Tags with `kind: nav` are used in documentation to group operations into sections"
+layout: default
+---
+
+{% capture summary %}
+Tags with `kind: nav` are used in documentation to group operations into sections.
+In most cases, an operation only has one tag of this tag so it belongs in one section.
```
mikekistler, line 10: Maybe this should be ?
```suggestion
In most cases, an operation only has one tag of this type so it belongs in one section.
```

karenetheridge: or "kind", if that's the word we're going with?

mikekistler (approved):
Looks good! 👍

PR #4288, opened 2025-01-03
Additional fields for the tag object

lornajane:
Implementation of the [proposal](https://github.com/OAI/OpenAPI-Specification/blob/main/proposals/2024-09-01-Tags-Improvement.md) to extend the functionality offered by tags. The main features:
 - tags get a `summary` field in addition to their description
 - tags can have a parent tag, enabling building a tree-ish hierarchy
 - tags can have a `kind` to say what sort of tag they are, so we can use tags for many; #4287 adds a registry for the most common types

Pull request still in draft because I would like to link to the registry but it doesn't exist yet so we should not merge - review feedback very welcome though!

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -2691,7 +2691,7 @@ It is not mandatory to have a Tag Object per tag defined in the Operation Object
 | <a name="tag-summary"></a>summary | `string` | A short summary of the tag, used for display purposes. |
 | <a name="tag-description"></a>description | `string` | A description for the tag. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation. |
 | <a name="tag-external-docs"></a>externalDocs | [External Documentation Object](#external-documentation-object) | Additional external documentation for this tag. |
-| <a name="tag-parent"></a>parent | `string` | The `name` of a tag that this tags is nested under. The named tag MUST exist in the API description, and circular references between parent and child tags MUST NOT be used. |
+| <a name="tag-parent"></a>parent | `string` | The `name` of a tag that this tag is nested under. The named tag MUST exist in the API description, and circular references between parent and child tags MUST NOT be used. |
 | <a name="tag-kind"></a>kind | `string` | A machine-readable string to categorize what sort of tag it is. Any string value can be used; common uses are `nav` for Navigation, `badge` for visible badges, `audience` for APIs used by different groups. A [registry of the most commonly used values](https://spec.openapis.org/registry/tag-kind/) is available. |
 
 This object MAY be extended with [Specification Extensions](#specification-extensions).
--- a/src/oas.md
+++ b/src/oas.md
@@ -2699,43 +2699,47 @@ This object MAY be extended with [Specification Extensions](#specification-exten
 ##### Tag Object Example
 
 ```json
-{
-  "name": "account-updates",
-  "summary": "Account Updates",
-  "description": "Account update operations",
-  "kind": "nav"
-},
-{
-  "name": "partner",
-  "summary": "Partner",
-  "description": "Operations available to the partners network",
-  "parent": "external",
-  "kind": "audience"
-},
-{
-  "name": "external",
-  "summary": "External",
-  "description": "Operations available to external consumers",
-  "kind": "audience"
-}
+"tags": [
+  {
+    "name": "account-updates",
+    "summary": "Account Updates",
+    "description": "Account update operations",
+    "kind": "nav"
+  },
+  {
+    "name": "partner",
+    "summary": "Partner",
+    "description": "Operations available to the partners network",
+    "parent": "external",
+    "kind": "audience"
+  },
+  {
+    "name": "external",
+    "summary": "External",
+    "description": "Operations available to external consumers",
+    "kind": "audience"
+  }
+]
 ```
 
 ```yaml
-- name: account-updates
-  summary: Account Updates
-  description: Account update operations
-  kind: nav
+tags:
 
-- name: partner
-  summary: Partner
-  description: Operations available to the partners network
-  parent: external
-  kind: audience
+  - name: account-updates
+    summary: Account Updates
+    description: Account update operations
+    kind: nav
 
-- name: external
-  summary: External
-  description: Operations available to external consumers
-  kind: audience
+  - name: partner
+    summary: Partner
+    description: Operations available to the partners network
+    parent: external
+    kind: audience
+
+  - name: external
+    summary: External
+    description: Operations available to external consumers
+    kind: audience
 ```
 
 #### Reference Object
--- a/src/oas.md
+++ b/src/oas.md
@@ -2687,9 +2687,12 @@ It is not mandatory to have a Tag Object per tag defined in the Operation Object
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
-| <a name="tag-name"></a>name | `string` | **REQUIRED**. The name of the tag. |
+| <a name="tag-name"></a>name | `string` | **REQUIRED**. The name of the tag. Use this value in the `tags` array of an Operation. |
+| <a name="tag-summary"></a>summary | `string` | A short summary of the tag, used for display purposes. |
 | <a name="tag-description"></a>description | `string` | A description for the tag. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation. |
 | <a name="tag-external-docs"></a>externalDocs | [External Documentation Object](#external-documentation-object) | Additional external documentation for this tag. |
+| <a name="tag-parent"></a>parent | `string` | The `name` of a tag that this tags is nested under. The named tag MUST exist in the API description, and circular references between parent and child tags MUST NOT be used. |
+| <a name="tag-kind"></a>kind | `string` | A machine-readable string to categorize what sort of tag it is. Common uses are `nav` for Navigation, `badge` for badges, `internal` for internal APIs, but any string value can be used. |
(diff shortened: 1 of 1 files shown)


baywet (approved):

src/oas.md
```diff
@@ -2687,24 +2687,59 @@ It is not mandatory to have a Tag Object per tag defined in the Operation Object
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
-| <a name="tag-name"></a>name | `string` | **REQUIRED**. The name of the tag. |
+| <a name="tag-name"></a>name | `string` | **REQUIRED**. The name of the tag. Use this value in the `tags` array of an Operation. |
+| <a name="tag-summary"></a>summary | `string` | A short summary of the tag, used for display purposes. |
 | <a name="tag-description"></a>description | `string` | A description for the tag. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation. |
 | <a name="tag-external-docs"></a>externalDocs | [External Documentation Object](#external-documentation-object) | Additional external documentation for this tag. |
+| <a name="tag-parent"></a>parent | `string` | The `name` of a tag that this tags is nested under. The named tag MUST exist in the API description, and circular references between parent and child tags MUST NOT be used. |
+| <a name="tag-kind"></a>kind | `string` | A machine-readable string to categorize what sort of tag it is. Common uses are `nav` for Navigation, `badge` for badges, `internal` for internal APIs, but any string value can be used. |
```
lornajane, line 2695: Should internal be audience in this context?

miqui (approved):

mikekistler (approved):
Looks good! 👍

src/oas.md
```diff
@@ -2687,24 +2687,59 @@ It is not mandatory to have a Tag Object per tag defined in the Operation Object
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
-| <a name="tag-name"></a>name | `string` | **REQUIRED**. The name of the tag. |
+| <a name="tag-name"></a>name | `string` | **REQUIRED**. The name of the tag. Use this value in the `tags` array of an Operation. |
+| <a name="tag-summary"></a>summary | `string` | A short summary of the tag, used for display purposes. |
 | <a name="tag-description"></a>description | `string` | A description for the tag. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation. |
```
LasneF, line 2692: What about updating the Jsonschema as it has been recently refreshed so that it got at munch in synch as it could

karenetheridge:
[ x ] Schema changes are required for this PR and are not done yet

src/oas.md
lornajane: Updated this part!

src/oas.md
```diff
@@ -2687,24 +2687,59 @@ It is not mandatory to have a Tag Object per tag defined in the Operation Object
...
 | <a name="tag-external-docs"></a>externalDocs | [External Documentation Object](#external-documentation-object) | Additional external documentation for this tag. |
+| <a name="tag-parent"></a>parent | `string` | The `name` of a tag that this tags is nested under. The named tag MUST exist in the API description, and circular references between parent and child tags MUST NOT be used. |
```
mikekistler, line 2694: Minor typo fix.
```suggestion
| <a name="tag-parent"></a>parent | `string` | The `name` of a tag that this tag is nested under. The named tag MUST exist in the API description, and circular references between parent and child tags MUST NOT be used. |
```

src/oas.md
```diff
@@ -2687,24 +2687,59 @@ It is not mandatory to have a Tag Object per tag defined in the Operation Object
...
+| <a name="tag-kind"></a>kind | `string` | A machine-readable string to categorize what sort of tag it is. Any string value can be used; common uses are `nav` for Navigation, `badge` for visible badges, `audience` for APIs used by different groups. A [registry of the most commonly used values](https://spec.openapis.org/registry/tag-kind/) is available. |
```
mikekistler, line 2695: Does "machine-readable" add anything here? Isn't the entire description machine readable? It's a minor point but it might raise doubts for some readers.

mikekistler (dismissed):
Looks good! 👍 

I left a couple minor comments but nothing serious enough to block merging.

src/oas.md
ralfhandl: I like it because it makes clear that this field is not intended for prose text.

ralfhandl (approved):

src/oas.md
```diff
@@ -2687,24 +2687,59 @@ It is not mandatory to have a Tag Object per tag defined in the Operation Object
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
-| <a name="tag-name"></a>name | `string` | **REQUIRED**. The name of the tag. |
+| <a name="tag-name"></a>name | `string` | **REQUIRED**. The name of the tag. Use this value in the `tags` array of an Operation. |
+| <a name="tag-summary"></a>summary | `string` | A short summary of the tag, used for display purposes. |
 | <a name="tag-description"></a>description | `string` | A description for the tag. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation. |
 | <a name="tag-external-docs"></a>externalDocs | [External Documentation Object](#external-documentation-object) | Additional external documentation for this tag. |
+| <a name="tag-parent"></a>parent | `string` | The `name` of a tag that this tag is nested under. The named tag MUST exist in the API description, and circular references between parent and child tags MUST NOT be used. |
```
handrews, line 2694: > The named tag MUST exist in the API description,

This interacts with the following text in 3.1.1:

> For resolving component and tag name connections from a referenced (non-entry) document, it is RECOMMENDED that tools resolve from the entry document, rather than the current document. This allows Security Scheme Objects and Tag Objects to be defined next to the API’s deployment information (the top-level array of Server Objects), and treated as an interface for referenced documents to access.

However, that was written to resolve Operation Object `tags` fields to Tag Objects, which means you need to locate a particular OpenAPI Object with Tag Objects under its `tags` field.  For this `parent` field, if we follow the same guidance, a Tag Object in a referenced document would search for its parent in the entry document.

Resolving to the entry document would be the most consistent thing, but not necessarily the most intuitive.

On the one hand, making `parent` only resolve within the current document feels more intuitive.  On the other hand, it's also limiting if for some reason you wanted to break tags across multiple OpenAPI Documents in a single OpenAPI Description, and it would make two fields that reference via a tag name behave differently, which is confusing as well.

A third option that ends up being more problematic than it seems at first would be to make `parent` take a URI-reference, but since the OpenAPI Object's `tags` field is an array, JSON Pointers end up being cryptic and fragile.

A fourth option, which would be annoyingly complex, would be to offer both `parent` (resolves to the entry document, since that could be different if the same other document is reference from multiple entry documents) and `parentRef` (like `operationRef`) which could be used with a URI-ref, e.g. `#/tags/4`.

A fifth option would be to define plain name fragments for tags, e.g "tag.{tagname}" so that URI-references would not have to have a numerical index.  But we've avoided doing that sort of thing (although JSON Schema's `$anchor` keyword does it).

handrews: I'm playing around with writing up a `tagRefs` field to allow tags to resolve by URI, and if that works out then maybe `parentRef` would be the consistent option.

lornajane: I think we're in agreement here, although perhaps the wording needs work. An OpenAPI description has the tags defined in its entry document, because there's nowhere else to define them. They don't need to be refs, because like the `info` section, `tags` are a document-level thing and I don't see how they can be ref'd in. Operations can use tags that are not defined, but if the tag is defined and has a parent, then I think that parent should be defined too.

miqui: > but if the tag is defined and has a parent, then I think that parent should be defined too.
+1

src/oas.md
```diff
@@ -2687,24 +2687,59 @@ It is not mandatory to have a Tag Object per tag defined in the Operation Object
...
+    "summary": "Account Updates",
+    "description": "Account update operations",
+    "kind": "nav"
+  },
+  {
+    "name": "partner",
+    "summary": "Partner",
+    "description": "Operations available to the partners network",
+    "parent": "external",
+    "kind": "audience"
```
miqui, line 2714: @lornajane Overall PR looks good. Thanks for the examples. However, wondering if a child tag can have a different kind from its parent tag ?  Honestly I don't want to delay this being merged, but just throwing this out here to get feedback.  Perhaps a followup PR if anything around this comes up.

miqui (approved):

src/oas.md
lornajane: It doesn't make sense to be too rigid, so I suggest we don't add restrictions here

src/oas.md
handrews: @lornajane I agree with the decision to merge this, and I'll pick up the subtleties elsewhere since I need to do so for `tags` in the Operation Object.

This gets tricky when you have multiple [OpenAPI Documents](https://spec.openapis.org/oas/v3.1.1#openapi-document) (capital D) in your OAD, each of which can have a list of Tag Objects.  Similar problems exist for the Security Requirement and Security Scheme Objects.

LoHertel:
@lornajane I'm looking forward to this addition. Is it intended for the v3.2.0 specification? I'm asking because it's not added to the [v3.2.0 milestone](https://github.com/OAI/OpenAPI-Specification/milestone/12) yet.

lornajane:
@LoHertel yes it's in 3.2, already merged in that branch. The milestones are mostly used for internal planning, they're not always a good representation of what's included

PR #4292, opened 2025-01-07
adds links to annex C in path and server templating sections

### LINKED ISSUE #3256: Formally define (ABNF) 3.x URL templating behavior

baywet:
fixes #3256
This continues the work started in #4244 and #4264 to add links to annex C in the path and server templating sections.
CC @char0n

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -602,7 +602,7 @@ iprivate       =  %xE000-F8FF / %xF0000-FFFFD / %x100000-10FFFD
 
 Here, `literals`, `pct-encoded`, `ucschar` and `iprivate` definitions are taken from [RFC 6570](https://www.rfc-editor.org/rfc/rfc6570), incorporating the corrections specified in [Errata 6937](https://www.rfc-editor.org/errata/eid6937) for `literals`.
 
-See also [Appendix C: Using RFC6570-Based Serialization](#appendix-c-using-rfc6570-based-serialization) for additional guidance.
+See the [Paths Object](#paths-object) for guidance on constructing full request URLs.
 
 ##### Fixed Fields
 


src/oas.md
```diff
@@ -600,6 +602,8 @@ iprivate       =  %xE000-F8FF / %xF0000-FFFFD / %x100000-10FFFD
 
 Here, `literals`, `pct-encoded`, `ucschar` and `iprivate` definitions are taken from [RFC 6570](https://www.rfc-editor.org/rfc/rfc6570), incorporating the corrections specified in [Errata 6937](https://www.rfc-editor.org/errata/eid6937) for `literals`.
 
+See also [Appendix C: Using RFC6570-Based Serialization](#appendix-c-using-rfc6570-based-serialization) for additional guidance.
```
handrews, line 605: I'm less sure this makes sense given that the Server Variable Object works very differently.  I've not actually dug into how it works, though, so maybe I'm missing something?

baywet: so would it make sense to instead add this link to the appendix ONLY for path templating, and then some kind of statement that says "string concat the result of the server template with your path template, and there you go to resolve that" ?

handrews: Yes, although we already say to do that concatenation in the main description of the [Paths Object](https://spec.openapis.org/oas/v3.0.4#paths-object):

> The path is appended to the URL from the [Server Object](https://spec.openapis.org/oas/v3.0.4#server-object) in order to construct the full URL.

baywet: do you have any objection to me duplicating this statement in the server object variable section? (and removing the annex C I just added to this section)

handrews: Maybe "see the Paths Object for guidance on constructing full request URLs" instead?  Duplicating the guidance is a maintenance problem waiting to happen if we ever change it. (witness all of the errors made trying to keep the Header Object and Parameters Object text in sync).

baywet: Thanks! just made the update. Let me know what you think!

mikekistler (approved):
Looks good! 👍

miqui (approved):

lornajane (approved):

PR #4294, opened 2025-01-09
Fix copy-paste "format: binary" error

### LINKED ISSUE #4293: Discrepancy between encoding multipart/ examples and working with binary data sections
In the Open API 3.1.1 spec, there seems to be a discrepancy between how to represent binary data, and the examples around encoding multipart parts.

[Working with binary data](https://spec.openapis.org/oas/latest.html#working-with-binary-data) and [migrating binary descriptions from OAS 3.0](https://spec.openapis.org/oas/latest.html#migrating-binary-descriptions-from-oas-3-0) indicate that `type` and `format` can be removed for raw binary data, and `type` with `contentEncoding` can be used (instead of `format`) for encoded binary data.

The [multipart form with multiple files](https://spec.openapis.org/oas/latest.html#example-multipart-form-with-multiple-files) example follows this guidance, but the [multipart form with encoding objects](https://spec.openapis.org/oas/latest.html#example-multipart-form-with-encoding-objects) still specifies `type` and `format` for binary data.

The roughly same example in the Open API 3.1.0 spec uses an empty schema for raw binary data instead of `format` and `type` which is what I was expecting: https://spec.openapis.org/oas/v3.1.0.html#encoding-object-example

So I'm wondering if the change between 3.1.0 and 3.1.1 was intentional. Are there any scenarios where we should still be using `format` of `byte` or `binary` with OAS 3.1?

handrews:
Fixes #4293 

These examples got copied from 3.0.4 and apparently I forgot to adjust them for 3.1.1 and no one else noticed.

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -1805,7 +1805,7 @@ requestBody:
             type: string
             format: uuid
 
-          # default for a schema withhout `type` is `application/octet-stream`
+          # default for a schema without `type` is `application/octet-stream`
           profileImage: {}
 
           # default for arrays is based on the type in the `items`
--- a/src/oas.md
+++ b/src/oas.md
@@ -1800,17 +1800,17 @@ requestBody:
       schema:
         type: object
         properties:
+          # default for a string without `contentEncoding` is `text/plain`
           id:
-            # default for primitives without a special format is text/plain
             type: string
             format: uuid
-          profileImage:
-            # default for string with binary format is `application/octet-stream`
-            type: string
-            format: binary
+
+          # default for a schema withhout `type` is `application/octet-stream`
+          profileImage: {}
+
+          # default for arrays is based on the type in the `items`
+          # subschema, which is an object, so `application/json`
           addresses:
-            # default for arrays is based on the type in the `items`
-            # subschema, which is an object, so `application/json`
             type: array
             items:
               $ref: '#/components/schemas/Address'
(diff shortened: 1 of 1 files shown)


ralfhandl:
Lgtm

ralfhandl (approved):

src/oas.md
```diff
@@ -1800,17 +1800,17 @@ requestBody:
       schema:
         type: object
         properties:
+          # default for a string without `contentEncoding` is `text/plain`
```
mikekistler, line 1803: Could we instead say:
```suggestion
          # default `contentEncoding` for a string is `text/plain`
```

src/oas.md
```diff
@@ -1800,17 +1800,17 @@ requestBody:
...
             format: uuid
-          profileImage:
-            # default for string with binary format is `application/octet-stream`
-            type: string
-            format: binary
+
+          # default for a schema without `type` is `application/octet-stream`
+          profileImage: {}
+
+          # default for arrays is based on the type in the `items`
```
mikekistler, line 1811: and
```suggestion
          # default `contentEncoding` for arrays is based on the type in the `items`
```

src/oas.md
```diff
@@ -1828,31 +1828,27 @@ requestBody:
       schema:
         type: object
         properties:
+          # No Encoding Object, so use default `text/plain`
```
mikekistler, line 1831: and 
```suggestion
          # No Encoding Object, so use default `contentEncoding` `text/plain`
```

src/oas.md
```diff
@@ -1828,31 +1828,27 @@ requestBody:
...
           id:
-            # default is `text/plain`
             type: string
             format: uuid
+
+          # Encoding Object overrides the default `application/json`
```
mikekistler, line 1836: and maybe
```suggestion
          # Encoding Object overrides the default `contentEncoding` `application/json`
```

mikekistler (approved):
Looks good. 👍 

I left some comments to add `contentEncoding` in places where I thought it added clarity. When I first read this line:
> default for a string without `contentEncoding` is `text/plain`

my thought was "default what"? So my changes are intended to fix that but I'll defer to your judgement to decide if this is necessary.

src/oas.md
handrews: @mikekistler no, because `text/plain` is a content type (Encoding Object `contentType`, or alternatively Schema Object `contentMediaType`) and not a content encoding (Schema Object `contentEncoding`, or alternatively Encoding Object `headers` field, header `Content-TransferEncoding`) which is something like `"base64"`

But we could say:

```YAML
# default content type for a string without `contentEncoding` is `text/plain`
```

so I'll do that.  I totally see how you read it the way you did and I had to think a moment before realizing what was wrong with it.

PR #4301, opened 2025-01-15
More contentEncoding example improvements (from feedback)

handrews:
This incorporates ideas from @mikekistler that did not make it into #4294.  Then I'll cherry-pick both commits over to v3.2-dev

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -1747,7 +1747,7 @@ requestBody:
             type: string
           icon:
             # The default content type with "contentEncoding" present
-            # is application/octet-stream, # so we need to set the correct
+            # is application/octet-stream, so we need to set the correct
             # image media type(s) in the Encoding Object.
             type: string
             contentEncoding: base64url
--- a/src/oas.md
+++ b/src/oas.md
@@ -1746,8 +1746,9 @@ requestBody:
           name:
             type: string
           icon:
-            # The default with "contentEncoding" is application/octet-stream,
-            # so we need to set image media type(s) in the Encoding Object.
+            # The default content type with "contentEncoding" present
+            # is application/octet-stream, # so we need to set the correct
+            # image media type(s) in the Encoding Object.
             type: string
             contentEncoding: base64url
   encoding:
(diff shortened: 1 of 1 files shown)


src/oas.md
```diff
@@ -1746,8 +1746,9 @@ requestBody:
           name:
             type: string
           icon:
-            # The default with "contentEncoding" is application/octet-stream,
-            # so we need to set image media type(s) in the Encoding Object.
+            # The default content type with "contentEncoding" present
+            # is application/octet-stream, # so we need to set the correct
```
mikekistler, line 1750:
```suggestion
            # is application/octet-stream, so we need to set the correct
```

mikekistler (dismissed):
Looks good! 👍 

Thanks @handrews! I left one very minor suggestion.

src/oas.md
```diff
@@ -1800,16 +1801,19 @@ requestBody:
       schema:
         type: object
         properties:
-          # default for a string without `contentEncoding` is `text/plain`
+          # default content type for a string without `contentEncoding`
```
ralfhandl, line 1804:
```suggestion
          # default content type for a string without "contentEncoding"
```
Align with line 1749

src/oas.md
```diff
@@ -1800,16 +1801,19 @@ requestBody:
...
+          # is `text/plain`
           id:
             type: string
             format: uuid
 
-          # default for a schema without `type` is `application/octet-stream`
+          # default content type for a schema without `type`
+          # is `application/octet-stream`
```
ralfhandl, line 1811:
```suggestion
          # default content type for a schema without "type"
          # is application/octet-stream
```
Align with lines 1749-1750

src/oas.md
```diff
@@ -1800,16 +1801,19 @@ requestBody:
...
           profileImage: {}
 
-          # default for arrays is based on the type in the `items`
-          # subschema, which is an object, so `application/json`
+          # default content type for arrays is based on the type
+          # in the `items` subschema, which is an object here,
+          # so the default content type for each item is `application/json`
```
ralfhandl, line 1816:
```suggestion
          # in the "items" subschema, which is an object here,
          # so the default content type for each item is application/json
```

src/oas.md
```diff
@@ -1833,7 +1837,7 @@ requestBody:
             type: string
             format: uuid
 
-          # Encoding Object overrides the default `application/json`
+          # Encoding Object overrides the default `application/json` content type
```
ralfhandl, line 1840:
```suggestion
          # Encoding Object overrides the default application/json content type
```

ralfhandl (commented):
We use an inconsistent mix of single-quotes, double-quotes, backticks, and no decoration for field names and field values within YAML comments.

The style of the very first example looks best, so propagating its style to the other examples:
- double-quotes around field names
- no decoration of media type values

handrews:
@ralfhandl we use backtics in YAML comments in many places.  I would like to leave these as-is and then fix them throughout the document in a separate PR.  There are some stylistic things to consider about the right way to handle this, including nested quotes of various sorts, so let's not introduce new variations without picking an overall style.

handrews:
@ralfhandl I have filed #4305 to track the further work.

mikekistler (approved):
Looks good. 👍 

I am fine with deferring quoting fixes to a later PR.

ralfhandl (approved):

PR #4318, opened 2025-01-22
Use branch-local package-lock.json

ralfhandl:
This change allows trusting the checks on dependabot PRs as they will now run with the updated dependencies.

### DIFF
--- a/.github/workflows/schema-tests.yaml
+++ b/.github/workflows/schema-tests.yaml
@@ -29,9 +29,7 @@ jobs:
         node-version: '20.x'
 
     - name: Install dependencies from main
-      run: |
-        git checkout remotes/origin/main -- package.json package-lock.json
-        npm ci
+      run: npm ci
 
     - name: Run tests
       run: npm run test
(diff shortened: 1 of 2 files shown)


handrews:
@ralfhandl I assume this means we have to keep the JS environment in sync across branches.  Or at least as in-sync as makes sense.  Are you adding this because we want some to be different?

Can you update CONTRIBUTING.md with whatever the new policy is for making this work?

ralfhandl:
With the current state - running tests and validation with the `package-lock.json` from `main`, breaking changes in the validation and test tools and their dependencies will _not_ be caught by the automated checks on the PR branch, which especially means that green checks on dependabot PRs have no meaning.

This PR improves that situation - green checks now mean that updating the dependencies will not break the validation and test tools.

I couldn't find documentation on why these workflows currently fetch `package-lock.json` from `main`.

lornajane:
Perhaps there should be some documentation covering the workflow of updating the other branches with dependencies. I think that's separate to this change, though, so it would be okay to capture what is needed as an issue.

lornajane (approved):
I agree it makes more sense to use the files in the current branch context.

PR #4328, opened 2025-02-04
Adjust uri to uri-reference for security schema fixed fields

### LINKED ISSUE #4327: `tokenUrl` in 3.1 schema does not support relative urls as specified
I am not super familiar with the OpenAPI 3.1 specification, so please excuse me if I misunderstood anything.

It appears that `tokenUrl`, a [fixed field](https://spec.openapis.org/oas/v3.1.1#fixed-fields-25), should have been modeled in the 3.1 JSONSchema as a `uri-reference` to [support relative urls](https://spec.openapis.org/oas/v3.1.1#relative-references-in-api-urls).

REF: https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.1/schema.yaml#L899-L901

Note that the [3.0 schema](https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.0/schema.yaml#L938-L940) has the format as `uri-reference`, as I would expect.

Happy to put in a pull request if this makes sense to someone with more experience in the repository.

REL: https://github.com/seriousme/openapi-schema-validator/issues/178

markrzen:
Resolves #4327 

### DIFF
--- a/schemas/v3.1/schema.yaml
+++ b/schemas/v3.1/schema.yaml
@@ -881,10 +881,10 @@ $defs:
         properties:
           tokenUrl:
             type: string
-            format: uri
+            format: uri-reference
           refreshUrl:
             type: string
-            format: uri
+            format: uri-reference
           scopes:
             $ref: '#/$defs/map-of-strings'
         required:
@@ -898,10 +898,10 @@ $defs:
         properties:
           tokenUrl:
             type: string
-            format: uri
+            format: uri-reference
           refreshUrl:
             type: string
-            format: uri
+            format: uri-reference
           scopes:
             $ref: '#/$defs/map-of-strings'
         required:
@@ -915,13 +915,13 @@ $defs:
         properties:
           authorizationUrl:
             type: string
-            format: uri
+            format: uri-reference
           tokenUrl:
             type: string
-            format: uri
+            format: uri-reference
           refreshUrl:
             type: string
-            format: uri
+            format: uri-reference
           scopes:
             $ref: '#/$defs/map-of-strings'
         required:
--- a/schemas/v3.1/schema.yaml
+++ b/schemas/v3.1/schema.yaml
@@ -12,7 +12,7 @@ properties:
     $ref: '#/$defs/info'
   jsonSchemaDialect:
     type: string
-    format: uri
+    format: uri-reference
     default: 'https://spec.openapis.org/oas/3.1/dialect/WORK-IN-PROGRESS'
   servers:
     type: array
@@ -64,7 +64,7 @@ $defs:
         type: string
       termsOfService:
         type: string
-        format: uri
+        format: uri-reference
       contact:
         $ref: '#/$defs/contact'
       license:
@@ -85,7 +85,7 @@ $defs:
         type: string
       url:
         type: string
-        format: uri
+        format: uri-reference
       email:
         type: string
         format: email
@@ -102,7 +102,7 @@ $defs:
         type: string
       url:
         type: string
-        format: uri
+        format: uri-reference
     required:
       - name
     dependentSchemas:
@@ -297,7 +297,7 @@ $defs:
         type: string
       url:
         type: string
-        format: uri
+        format: uri-reference
     required:
       - url
     $ref: '#/$defs/specification-extensions'
@@ -598,7 +598,7 @@ $defs:
       value: true
       externalValue:
         type: string
-        format: uri
+        format: uri-reference
     not:
       required:
       - value
@@ -830,7 +830,7 @@ $defs:
           properties:
             openIdConnectUrl:
               type: string
-              format: uri
+              format: uri-reference
           required:
             - openIdConnectUrl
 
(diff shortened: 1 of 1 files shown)


mikekistler:
Why not change all the `format: uri` to `format: uri-reference`. I could not find any fields defined as URIs and forbid relative references (but I am surprised that the OAuth ones allow them).

markrzen:
@handrews On it right now.

markrzen:
@handrews I could only find the xml object in the meta.yaml, and that document already looked correct.

schemas/v3.1/schema.yaml
```diff
@@ -12,7 +12,7 @@ properties:
     $ref: '#/$defs/info'
   jsonSchemaDialect:
     type: string
-    format: uri
+    format: uri-reference
```
karenetheridge, line 15: This one is the twin to JSON Schema's `$schema` keyword, and therefore cannot be a uri-reference.

karenetheridge (dismissed):
I found one change that should be reverted; the others look ok to me (not sure about the security ones though).

schemas/v3.1/schema.yaml
handrews: @karenetheridge philosophically I agree with you, but the way the OAS field is written unfortunately allows a relative URI-reference.  Of course you'd have to resolve that before you could use it as a dialect URI, and the only way that might work is with a custom dialect, but there's no reason it couldn't work that way.

I do agree that it should have been written as requiring a non-relative URI.

markrzen: Happy to make the change if needed. Waiting for maintainer to come to an agreement.

karenetheridge: @handrews what do you mean "the way the OAS field is written unfortunately allows a relative URI-reference"? What are you referencing here?

handrews: @karenetheridge It's the combination of two things:

§4.6.7 states:

> Unless specified otherwise, all fields that are URIs MAY be relative references as defined by [[RFC3986](https://spec.openapis.org/oas/v3.1.1#bib-rfc3986)] [Section 4.2](https://datatracker.ietf.org/doc/html/rfc3986#section-4.2).

while the fixed fields description for `jsonSchemaDialect` says:

> The default value for the `$schema` keyword within [Schema Objects](https://spec.openapis.org/oas/v3.1.1#schema-object) contained within this OAS document. This MUST be in the form of a URI.

Strictly speaking, this text allows a URI reference.  The _value_ used with `$schema` of course must be a full URI, but that can be produced by resolving the relative URI-reference normally, just as with any other field that the OAS says "MUST be in the form of a URI".  What is used with `$schema` is the resolved value, not the literal text.

At least that is my interpretation.  We can always kick it up to @OAI/tsc for a ruling if you disagree.  Again, I prefer your reading (and really dislike that we don't just explicitly say "URI-reference" and instead say "URI means URI-reference unless otherwise stated") but that's what we have to work with.

handrews: @karenetheridge take a look at the XML Object's `namespace` field for how we say a field is a fully-resolve URI (again, I agree we should use the correct terminology per RFC3986 and not what we're doing now, but we _do_ have ways to say we don't want a non-relative reference).  We are very explicit about how OAD URI-referencess and API URL-references are resolved.  Each of those topics gets a full section in the 3.1 specification.  In any event, this is drifting outside of the scope of this PR and should continue in a discussion or issue.

mikekistler (approved):
Looks good. 👍

handrews (approved):

handrews (approved):
@karenetheridge it sounded like you accepted my reading of the spec for `jsonSchemaDialect` in the last comment, so I'm going to go ahead and merge this.  If I'm wrong, lmk and we'll sort it out.

handrews:
@karenetheridge or not... can you remove your "changes requested"?

handrews:
Oh this is because there's only one approval- it's too early in the morning for me to be doing code reviews apparently.  @karenetheridge or @mikekistler if either of you is willing to give an approval that will do it.

mikekistler (approved):
Looks good! 👍

seriousme:
Apparently  it needs explicit approval from @karenetheridge as the two approvals from @mikekistler and @handrews are in but the flow is still blocked.

markrzen:
Someone with write access to the repo can likely dismiss the request for change.

handrews:
I'll check with @karenetheridge on this directly to resolve, she might just not have seen this in the last few days.

ralfhandl (approved):

ralfhandl:
@markrzen Thanks for your contribution, and extra thanks for your patience 😊

seriousme:
Thanks to all for making this work!
Sorry if the question is inappropriate here, but can you indicate when the `v3.1-dev` will be "officially released" as update to v3.1 in the gh-pages branch?

Reason for asking is that I maintain tooling that builds on the "officially released" schema's and I'd like to offer @markrzen the update merged in this PR in my tooling. (see Mark's initial post in #4327)

Kind regards,
Hans

PR #4333, opened 2025-02-06
dev: test schemas in ./src and ./schemas

ralfhandl:
Part of
* #4223

Changes:
- separate test fixtures for published schemas in `./schemas` and next schema in `./src`
- one test script for all schema versions
- schema test coverage also for next schema in `./src`


handrews:
@ralfhandl I think I might just suggest calling "vNext" "dev" to match the branch.  It's not clear to me that what's on the dev branch will always match the next version under development.  In fact I feel like it's most likely to just be the same as whatever branch was most recently released as that's when cnanges are merged down to dev.

PR #4336, opened 2025-02-06
dev: format src/oas.md

ralfhandl:
Part of
* #4223

New npm script `format-markdown` formats `src/oas.md`.


baywet:
@ralfhandl to re-target to dev instead of main

handrews:
@ralfhandl does this mean that the published markdown will be different from the markdown in github?  If so, wouldn't it be better to make format-markdown a precommit hook instead?

ralfhandl:
This script allows ad-hoc formatting of `src/oad.md`, that is the file that will eventually be released to `main`. 

It is the same script that we used to format `3.1.1.md` and `3.0.4.md` in their respective dev branches before releasing them to `main`.

We can later additionally automate this.

handrews (approved):

PR #4339, opened 2025-02-06
Allow discriminator to be optional

mikekistler:
This PR changes the `discriminator` to allow the discriminator property to be an optional field, so that when it is not present the `default` field can be used to specify which schema of the `anyOf` or `oneOf` is expected to validate the structure of the model.

I've also massaged the descriptions of polymorphism to emphasize that JSON Schema can and should be used to describe polymorphism and that discriminator is an optional extension of that support.

I also dropped some of the JSON examples when the same example was shown in yaml just to cut down on redundancy.

- [x] schema changes are needed for this pull request but not done yet

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -3484,7 +3484,7 @@ MyResponseType:
     - $ref: '#/components/schemas/Lizard'
 ```
 
-which means the payload _MUST_, by validation, match exactly one of the schemas described by `Cat`, `Dog`, or `Lizard`. Deserialization of a `oneOf` can be a costly operation, as it requires determining which schema matches the payload and thus should be used in deserialization. This problem also exists for `anyOf` schemas. A `discriminator` MAY be used as a "hint" to improve the efficiency of selection of the matching schema. The Discriminator Object cannot change the validation result of the `oneOf`, it can only help make the deserialization more efficient and provide better error messaging. We can specify the exact field that tells us which schema is expected to match the instance:
+which means the payload _MUST_, by validation, match exactly one of the schemas described by `Cat`, `Dog`, or `Lizard`. Deserialization of a `oneOf` can be a costly operation, as it requires determining which schema matches the payload and thus should be used in deserialization. This problem also exists for `anyOf` schemas. A `discriminator` MAY be used as a "hint" to improve the efficiency of selection of the matching schema. The [Discriminator Object](#discriminator-object) cannot change the validation result of the `oneOf`, it can only help make the deserialization more efficient and provide better error messaging. We can specify the exact field that tells us which schema is expected to match the instance:
 
 ```yaml
 MyResponseType:
--- a/src/oas.md
+++ b/src/oas.md
@@ -3245,6 +3245,57 @@ components:
         - packSize
 ```
 
+###### Models with Polymorphism Support using allOf and a Discriminator Object
+
+It is also possible to describe polymorphic models using `allOf`. The following example uses `allOf` with a [Discriminator Object](#discriminator-object) to describe a polymorphic `Pet` model.
+
+```yaml
+components:
+  schemas:
+    Pet:
+      type: object
+      discriminator:
+        propertyName: petType
+      properties:
+        name:
+          type: string
+        petType:
+          type: string
+      required:
+        - name
+        - petType
+    Cat: # "Cat" will be used as the discriminating value
+      description: A representation of a cat
+      allOf:
+        - $ref: '#/components/schemas/Pet'
+        - type: object
+          properties:
+            huntingSkill:
+              type: string
+              description: The measured skill for hunting
+              enum:
+                - clueless
+                - lazy
+                - adventurous
+                - aggressive
+          required:
+            - huntingSkill
+    Dog: # "Dog" will be used as the discriminating value
+      description: A representation of a dog
+      allOf:
+        - $ref: '#/components/schemas/Pet'
+        - type: object
+          properties:
+            packSize:
+              type: integer
+              format: int32
+              description: the size of the pack the dog is from
+              default: 0
+              minimum: 0
+          required:
+            - packSize
+```
+
 ###### Generic Data Structure Model
 
 ```JSON
--- a/src/oas.md
+++ b/src/oas.md
@@ -3406,7 +3406,7 @@ However, the exact nature of such conversions are implementation-defined.
 
 When the discriminating property is defined as optional, the `discriminator` field must include a `default` field that specifies a schema that is expected to validate the structure of the model when the discriminating property is not present in the payload. This allows the schema to still be validated correctly even if the discriminator property is missing.
 
-The primary use case for an optional discriminating property is to allow a schema to be extended with a discriminator without breaking existing clients that do not provide the discriminator property.
+The primary use case for an optional discriminating property is to allow a schema to be extended with a discriminator without breaking existing clients that do not provide the discriminating property.
 
 Typically the schema specified in the `default` field will specify that the discriminating property is not present, e.g.
 
--- a/src/oas.md
+++ b/src/oas.md
@@ -2877,7 +2877,7 @@ These keywords can be used to describe polymorphism, where a single field can ac
 
 The OpenAPI specification extends the JSON Schema support for polymorphism by adding the [`discriminator`](#schema-discriminator) field whose value is a [Discriminator Object](#discriminator-object).
 When used, the Discriminator Object indicates the name of the property that hints which schema of an `anyOf` or `oneOf` is expected to validate the structure of the model.
-The discriminating property may be defined as required or optional, but when defined as an optional property the Discriminator Object must include a `defaultMapping` field that specifies which schema of the `anyOf` or `oneOf` is expected to validate the structure of the model when the discriminating property is not present.
+The discriminating property MAY be defined as required or optional, but when defined as an optional property the Discriminator Object MUST include a `defaultMapping` field that specifies which schema of the `anyOf` or `oneOf`, or which schema that references the current schema in an `allOf`, is expected to validate the structure of the model when the discriminating property is not present.
 
 There are two ways to define the value of a discriminating property for an inheriting instance.
 
@@ -3362,7 +3362,7 @@ components:
 
 When request bodies or response payloads may be one of a number of different schemas, these should use the JSON Schema `anyOf` or `oneOf` keywords to describe the possible schemas (see [Composition and Inheritance](#composition-and-inheritance-polymorphism)).
 
-A polymorphic schema MAY include a Discriminator Object, which defines the name of the property that may be used as a hint for which schema of the `anyOf` or `oneOf` is expected to validate the structure of the model.
+A polymorphic schema MAY include a Discriminator Object, which defines the name of the property that may be used as a hint for which schema of the `anyOf` or `oneOf`, or which schema that references the current schema in an `allOf`, is expected to validate the structure of the model.
 This hint can be used to aid in serialization, deserialization, and validation.
 The Discriminator Object does this by implicitly or explicitly associating the possible values of a named property with alternative schemas.
 
@@ -3372,7 +3372,7 @@ Note that `discriminator` MUST NOT change the validation outcome of the schema.
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
-| <a name="property-name"></a>propertyName | `string` | **REQUIRED**. The name of the discriminating property in the payload that will hold the discriminating value. The discriminating property may be defined as required or optional, but when defined as optional the Discriminator Object must include a `defaultMapping` field that specifies which schema is expected to validate the structure of the model when no discriminating property is present. |
+| <a name="property-name"></a>propertyName | `string` | **REQUIRED**. The name of the discriminating property in the payload that will hold the discriminating value. The discriminating property MAY be defined as required or optional, but when defined as optional the Discriminator Object MUST include a `defaultMapping` field that specifies which schema is expected to validate the structure of the model when the discriminating property is not present. |
 | <a name="discriminator-mapping"></a> mapping | Map[`string`, `string`] | An object to hold mappings between payload values and schema names or URI references. |
 | <a name="default"></a> defaultMapping | `string` | The schema name or URI reference to a schema that is expected to validate the structure of the model when the discriminating property is not present in the payload. |
 
@@ -3396,7 +3396,7 @@ The behavior of any configuration of `oneOf`, `anyOf`, `allOf` and `discriminato
 The value of the property named in `propertyName` is used as the name of the associated schema under the [Components Object](#components-object), _unless_ a `mapping` is present for that value.
 The `mapping` entry maps a specific property value to either a different schema component name, or to a schema identified by a URI.
 When using implicit or explicit schema component names, inline `oneOf` or `anyOf` subschemas are not considered.
-The behavior of a `mapping` value or `default` value that is both a valid schema name and a valid relative URI reference is implementation-defined, but it is RECOMMENDED that it be treated as a schema name.
+The behavior of a `mapping` value or `defaultMapping` value that is both a valid schema name and a valid relative URI reference is implementation-defined, but it is RECOMMENDED that it be treated as a schema name.
 To ensure that an ambiguous value (e.g. `"foo"`) is treated as a relative URI reference by all implementations, authors MUST prefix it with the `"."` path segment (e.g. `"./foo"`).
 
 Mapping keys MUST be string values, but tooling MAY convert response values to strings for comparison.
@@ -3404,7 +3404,7 @@ However, the exact nature of such conversions are implementation-defined.
 
 ##### Optional discriminating property
 
-When the discriminating property is defined as optional, the [Discriminator Object](#discriminator-object) must include a `defaultMapping` field that specifies a schema that is expected to validate the structure of the model when the discriminating property is not present in the payload. This allows the schema to still be validated correctly even if the discriminating property is missing.
+When the discriminating property is defined as optional, the [Discriminator Object](#discriminator-object) MUST include a `defaultMapping` field that specifies a schema that is expected to validate the structure of the model when the discriminating property is not present in the payload. This allows the schema to still be validated correctly even if the discriminating property is missing.
 
 The primary use case for an optional discriminating property is to allow a schema to be extended with a discriminator without breaking existing clients that do not provide the discriminating property.
 
@@ -3476,7 +3476,7 @@ Here the discriminating value of `dog` will map to the schema `#/components/sche
 
 When used in conjunction with the `anyOf` construct, the use of the discriminator can avoid ambiguity for serializers/deserializers where multiple schemas may satisfy a single payload.
 
-When the discriminating property is defined as optional, the Discriminator Object must include a `defaultMapping` field that specifies a schema of the `anyOf` or `oneOf` is expected to validate the structure of the model when the discriminating property is not present in the payload. This allows the schema to still be validated correctly even if the discriminator property is missing.
+When the discriminating property is defined as optional, the Discriminator Object MUST include a `defaultMapping` field that specifies a schema of the `anyOf` or `oneOf` is expected to validate the structure of the model when the discriminating property is not present in the payload. This allows the schema to still be validated correctly even if the discriminator property is missing.
 
 For example:
 
--- a/src/oas.md
+++ b/src/oas.md
@@ -2877,7 +2877,7 @@ These keywords can be used to describe polymorphism, where a single field can ac
 
 The OpenAPI specification extends the JSON Schema support for polymorphism by adding the [`discriminator`](#schema-discriminator) field.
 When used, the `discriminator` indicates the name of the property that hints which schema of an `anyOf` or `oneOf` is expected to validate the structure of the model.
-The discriminating property may be defined as required or optional, but when defined as an optional property the `discriminator` field must include a `default` field that specifies which schema of the `anyOf` or `oneOf` is expected to validate the structure of the model.
+The discriminating property may be defined as required or optional, but when defined as an optional property the `discriminator` field must include a `defaultMapping` field that specifies which schema of the `anyOf` or `oneOf` is expected to validate the structure of the model when the discriminating property is not present.
 
 There are two ways to define the value of a discriminator for an inheriting instance.
 
@@ -3372,9 +3372,9 @@ Note that `discriminator` MUST NOT change the validation outcome of the schema.
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
-| <a name="property-name"></a>propertyName | `string` | **REQUIRED**. The name of the discriminating property in the payload that will hold the discriminating value. The discriminating property may be defined as required or optional, but when defined as optional the `discriminator` field must include a `default` field that specifies which schema is expected to validate the structure of the model. |
+| <a name="property-name"></a>propertyName | `string` | **REQUIRED**. The name of the discriminating property in the payload that will hold the discriminating value. The discriminating property may be defined as required or optional, but when defined as optional the `discriminator` field must include a `defaultMapping` field that specifies which schema is expected to validate the structure of the model when no discriminating property is present. |
 | <a name="discriminator-mapping"></a> mapping | Map[`string`, `string`] | An object to hold mappings between payload values and schema names or URI references. |
-| <a name="default"></a> default | `string` | The schema name or URI reference to a schema that is expected to validate the structure of the model when the discriminating property is not present in the payload. |
+| <a name="default"></a> defaultMapping | `string` | The schema name or URI reference to a schema that is expected to validate the structure of the model when the discriminating property is not present in the payload. |
 
 This object MAY be extended with [Specification Extensions](#specification-extensions).
 
@@ -3404,11 +3404,11 @@ However, the exact nature of such conversions are implementation-defined.
 
 ##### Optional discriminating property
 
-When the discriminating property is defined as optional, the `discriminator` field must include a `default` field that specifies a schema that is expected to validate the structure of the model when the discriminating property is not present in the payload. This allows the schema to still be validated correctly even if the discriminator property is missing.
+When the discriminating property is defined as optional, the `discriminator` field must include a `defaultMapping` field that specifies a schema that is expected to validate the structure of the model when the discriminating property is not present in the payload. This allows the schema to still be validated correctly even if the discriminator property is missing.
 
 The primary use case for an optional discriminating property is to allow a schema to be extended with a discriminator without breaking existing clients that do not provide the discriminating property.
 
-Typically the schema specified in the `default` field will specify that the discriminating property is not present, e.g.
+Typically the schema specified in the `defaultMapping` field will specify that the discriminating property is not present, e.g.
 
 ```yaml
 OtherPet:
@@ -3417,7 +3417,7 @@ OtherPet:
     required: ['petType']
 ```
 
-This will prevent the default schema from validating a payload that includes the discriminating property, which would cause a validation of the payload to fail when polymorphism is described using the `oneOf` JSON schema keyword.
+This will prevent the default mapping schema from validating a payload that includes the discriminating property, which would cause a validation of the payload to fail when polymorphism is described using the `oneOf` JSON schema keyword.
 
 ##### Examples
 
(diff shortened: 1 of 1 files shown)


handrews:
@jeremyfiel `default` could instead be `ifAbsent` or `whenAbsent` or something like that

src/oas.md
```diff
@@ -3458,6 +3476,28 @@ Here the discriminating value of `dog` will map to the schema `#/components/sche
 
 When used in conjunction with the `anyOf` construct, the use of the discriminator can avoid ambiguity for serializers/deserializers where multiple schemas may satisfy a single payload.
 
+When the `discriminator` property is defined as optional, the `discriminator` field must include a `default` field that specifies a schema of the `anyOf` or `oneOf` is expected to validate the structure of the model when the `discriminator` property is not present in the payload. This allows the schema to still be validated correctly even if the discriminator property is missing.
```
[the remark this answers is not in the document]
handrews: We use the term "discriminating value" when talking about the thing in the instance, so "discriminating property" would make more sense.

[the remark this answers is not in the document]
mikekistler: I like @handrews suggestion of  "discriminating property" -- I'll make that change.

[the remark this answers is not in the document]
mikekistler: @jeremyfiel If "default" was defined within "mapping", wouldn't that prevent "default" from being used as a valid discriminator value?

As for the keyword name "default", I am not wed to this and happy to use something else. Any suggestions?

src/oas.md
```diff
@@ -3372,8 +3372,9 @@ Note that `discriminator` MUST NOT change the validation outcome of the schema.
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
-| <a name="property-name"></a>propertyName | `string` | **REQUIRED**. The name of the property in the payload that will hold the discriminating value. This property SHOULD be required in the payload schema, as the behavior when the property is absent is undefined. |
+| <a name="property-name"></a>propertyName | `string` | **REQUIRED**. The name of the property in the payload that will hold the discriminating value. This property may be defined as required or optional, but when defined as an optional property the `discriminator` field must include a `default` field that specifies which schema is expected to validate the structure of the model. |
```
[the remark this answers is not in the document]
mikekistler: I took a stab at improving this.

src/oas.md
```diff
@@ -3402,13 +3402,13 @@ To ensure that an ambiguous value (e.g. `"foo"`) is treated as a relative URI re
 Mapping keys MUST be string values, but tooling MAY convert response values to strings for comparison.
 However, the exact nature of such conversions are implementation-defined.
 
-##### Optional `discriminator` property
+##### Optional discriminating property
 
-When the `discriminator` property is defined as optional, the `discriminator` field must include a `default` field that specifies a schema that is expected to validate the structure of the model when the `discriminator` property is not present in the payload. This allows the schema to still be validated correctly even if the discriminator property is missing.
+When the discriminating property is defined as optional, the `discriminator` field must include a `default` field that specifies a schema that is expected to validate the structure of the model when the discriminating property is not present in the payload. This allows the schema to still be validated correctly even if the discriminator property is missing.
```
jeremyfiel, line 3407:
```suggestion
When the discriminating property is defined as optional, the `discriminator` schema must include a `default` field that specifies a schema that is expected to validate the structure of the model when the discriminating property is not present in the payload. This allows the schema to still be validated correctly even if the discriminator property is missing.
```

src/oas.md
```diff
@@ -2877,7 +2877,7 @@ These keywords can be used to describe polymorphism, where a single field can ac
 
 The OpenAPI specification extends the JSON Schema support for polymorphism by adding the [`discriminator`](#schema-discriminator) field.
 When used, the `discriminator` indicates the name of the property that hints which schema of an `anyOf` or `oneOf` is expected to validate the structure of the model.
-The `discriminator` property may be defined as required or optional, but when defined as an optional property the `discriminator` field must include a `default` field that specifies which schema of the `anyOf` or `oneOf` is expected to validate the structure of the model.
+The discriminating property may be defined as required or optional, but when defined as an optional property the `discriminator` field must include a `default` field that specifies which schema of the `anyOf` or `oneOf` is expected to validate the structure of the model.
```
jeremyfiel, line 2880:
```suggestion
The discriminating property may be defined as required or optional, but when defined as an optional property the `discriminator` schema must include a `default` field that specifies which schema of the `anyOf` or `oneOf` is expected to validate the structure of the model.
```

src/oas.md
```diff
@@ -3402,13 +3402,13 @@ To ensure that an ambiguous value (e.g. `"foo"`) is treated as a relative URI re
...
+##### Optional discriminating property
 
-When the `discriminator` property is defined as optional, the `discriminator` field must include a `default` field that specifies a schema that is expected to validate the structure of the model when the `discriminator` property is not present in the payload. This allows the schema to still be validated correctly even if the discriminator property is missing.
+When the discriminating property is defined as optional, the `discriminator` field must include a `default` field that specifies a schema that is expected to validate the structure of the model when the discriminating property is not present in the payload. This allows the schema to still be validated correctly even if the discriminator property is missing.
 
-The primary use case for an optional `discriminator` property is to allow a schema to be extended with a discriminator without breaking existing clients that do not provide the discriminator property.
+The primary use case for an optional discriminating property is to allow a schema to be extended with a discriminator without breaking existing clients that do not provide the discriminator property.
```
jeremyfiel, line 3409:
```suggestion
The primary use case for an optional discriminating property is to allow a schema to be extended with a discriminator without breaking existing clients that do not provide the discriminating property.
```

src/oas.md
```diff
@@ -3476,7 +3476,7 @@ Here the discriminating value of `dog` will map to the schema `#/components/sche
 
 When used in conjunction with the `anyOf` construct, the use of the discriminator can avoid ambiguity for serializers/deserializers where multiple schemas may satisfy a single payload.
 
-When the `discriminator` property is defined as optional, the `discriminator` field must include a `default` field that specifies a schema of the `anyOf` or `oneOf` is expected to validate the structure of the model when the `discriminator` property is not present in the payload. This allows the schema to still be validated correctly even if the discriminator property is missing.
+When the discriminating property is defined as optional, the `discriminator` field must include a `default` field that specifies a schema of the `anyOf` or `oneOf` is expected to validate the structure of the model when the discriminating property is not present in the payload. This allows the schema to still be validated correctly even if the discriminator property is missing.
```
jeremyfiel, line 3479:
```suggestion
When the discriminating property is defined as optional, the `discriminator` schema must include a `default` field that specifies a schema of the `anyOf` or `oneOf` is expected to validate the structure of the model when the discriminating property is not present in the payload. This allows the schema to still be validated correctly even if the discriminator property is missing.
```

jeremyfiel (commented):
a few minor edits for clarification

src/oas.md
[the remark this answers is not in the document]
jeremyfiel: You're right, that would prevent the use of the `default` key. 

some suggestions for the keyword

`defaultSchema`
`defaultMapping` 
`fallbackSchema`

src/oas.md
mikekistler: I don't think changing "field" to "schema" is right here. Can you explain why you have suggesting this change?

jeremyfiel: the `discriminator` keyword defines a schema. It's not a standalone field. It requires the `propertyName` keyword per the spec.

```yaml
discriminator:
  propertyName: thing
  default: Dog # < your new keyword
```
vs

```yaml
discriminator: ??
```

src/oas.md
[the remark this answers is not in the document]
mikekistler: I have changed `default` to `defaultMapping`.

src/oas.md
jeremyfiel: there were a few locations where this wasn't updated.  They are in the `Outdated` conversations below. Is that an oversight or intentional?

jeremyfiel:
I feel like this line should be updated with the language of `discriminating value`

This was not part of your changes, but it's related and the perfect time to make additional clarifications

https://github.com/OAI/OpenAPI-Specification/blob/7ebde1428df16d1e007c913a8e838c3f3d340ace/src/oas.md?plain=1#L3459

src/oas.md
mikekistler: I can change `discriminator` field to `discriminator` keyword. Would that address your concerns?

handrews: I feel like OAS fields should be fields even in the Schema Object.  Sometimes when talking about the Schema Object as one of many Objects I just use "fields" because the context is clearly OAS.  I use "keyword" when the more important context it JSON Schema.  Although I have no stronger argument for this usage.  Perhaps @ralfhandl has an opinion here?

jeremyfiel: the reason I raised this is because the discriminator field is not actually a thing. it's a schema.  You can't use the field by itself, as shown in my example.  Referring to it as a field is not correct, imo.

mikekistler: @jeremyfiel I don't know what you mean when you say "the discriminator field is not actually a thing". `discriminator` is a ["fixed field" defined by OpenAPI](https://spec.openapis.org/oas/v3.0.4.html#fixed-fields-20), so by [our established terminology](https://github.com/OAI/OpenAPI-Specification/blob/v3.2-dev/CONTRIBUTING.md#use-of-keyword-field-property-and-attribute), we refer to it as a "field".

jeremyfiel: I understand the field is described by OAS, but the field itself is a schema. It's never a standalone field such as additionalProperties: false or additionalProperties: {} where it can be both. 

In this reference language, I believe the discriminator field should be referenced as a schema because you are referencing the behavior of the field. The use of _field_ is correct when referencing the `defaultMapping` requirement.

> When the discriminating property is defined as optional, the `discriminator` schema must include a `defaultMapping` field that specifies a schema

mikekistler: Okay ... I think I see a way to resolve this. What we are talking about here is the "Discriminator Object" -- the value of the `discriminator` field. So we could use "Discriminator Object" rather than "`discriminator` field`" in the places you have flagged. I think this would be consistent with language/terminology elsewhere in the spec. Would this work for you @jeremyfiel ?

jeremyfiel: Sounds good! 👍

jeremyfiel (approved):

src/oas.md
```diff
@@ -3395,12 +3396,29 @@ The behavior of any configuration of `oneOf`, `anyOf`, `allOf` and `discriminato
 The value of the property named in `propertyName` is used as the name of the associated schema under the [Components Object](#components-object), _unless_ a `mapping` is present for that value.
 The `mapping` entry maps a specific property value to either a different schema component name, or to a schema identified by a URI.
 When using implicit or explicit schema component names, inline `oneOf` or `anyOf` subschemas are not considered.
-The behavior of a `mapping` value that is both a valid schema name and a valid relative URI reference is implementation-defined, but it is RECOMMENDED that it be treated as a schema name.
+The behavior of a `mapping` value or `default` value that is both a valid schema name and a valid relative URI reference is implementation-defined, but it is RECOMMENDED that it be treated as a schema name.
```
jeremyfiel, line 3399:
```suggestion
The behavior of a `mapping` value or `defaultMapping` value that is both a valid schema name and a valid relative URI reference is implementation-defined, but it is RECOMMENDED that it be treated as a schema name.
```

src/oas.md
```diff
@@ -3415,7 +3433,7 @@ MyResponseType:
...
+which means the payload _MUST_, by validation, match exactly one of the schemas described by `Cat`, `Dog`, or `Lizard`. Deserialization of a `oneOf` can be a costly operation, as it requires determining which schema matches the payload and thus should be used in deserialization. This problem also exists for `anyOf` schemas. A `discriminator` MAY be used as a "hint" to improve the efficiency of selection of the matching schema. The Discriminator Object cannot change the validation result of the `oneOf`, it can only help make the deserialization more efficient and provide better error messaging. We can specify the exact field that tells us which schema is expected to match the instance:
```
jeremyfiel, line 3436:
```suggestion
which means the payload _MUST_, by validation, match exactly one of the schemas described by `Cat`, `Dog`, or `Lizard`. Deserialization of a `oneOf` can be a costly operation, as it requires determining which schema matches the payload and thus should be used in deserialization. This problem also exists for `anyOf` schemas. A `discriminator` MAY be used as a "hint" to improve the efficiency of selection of the matching schema. The [Discriminator Object](#discriminator-object) cannot change the validation result of the `oneOf`, it can only help make the deserialization more efficient and provide better error messaging. We can specify the exact field that tells us which schema is expected to match the instance:
```

src/oas.md
```diff
@@ -3458,6 +3476,28 @@ Here the discriminating value of `dog` will map to the schema `#/components/sche
 
 When used in conjunction with the `anyOf` construct, the use of the discriminator can avoid ambiguity for serializers/deserializers where multiple schemas may satisfy a single payload.
 
+When the discriminating property is defined as optional, the Discriminator Object must include a `defaultMapping` field that specifies a schema of the `anyOf` or `oneOf` is expected to validate the structure of the model when the discriminating property is not present in the payload. This allows the schema to still be validated correctly even if the discriminator property is missing.
```
jeremyfiel, line 3479:
```suggestion
When the discriminating property is defined as optional, the [Discriminator Object](#discriminator-object) must include a `defaultMapping` field that specifies a schema of the `anyOf` or `oneOf` is expected to validate the structure of the model when the discriminating property is not present in the payload. This allows the schema to still be validated correctly even if the discriminator property is missing.
```

src/oas.md
```diff
@@ -3372,8 +3372,9 @@ Note that `discriminator` MUST NOT change the validation outcome of the schema.
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
-| <a name="property-name"></a>propertyName | `string` | **REQUIRED**. The name of the property in the payload that will hold the discriminating value. This property SHOULD be required in the payload schema, as the behavior when the property is absent is undefined. |
+| <a name="property-name"></a>propertyName | `string` | **REQUIRED**. The name of the discriminating property in the payload that will hold the discriminating value. The discriminating property may be defined as required or optional, but when defined as optional the Discriminator Object must include a `defaultMapping` field that specifies which schema is expected to validate the structure of the model when no discriminating property is present. |
```
jeremyfiel, line 3375:
```suggestion
| <a name="property-name"></a>propertyName | `string` | **REQUIRED**. The name of the discriminating property in the payload that will hold the discriminating value. The discriminating property may be defined as required or optional, but when defined as optional the [Discriminator Object](#discriminator-object) must include a `defaultMapping` field that specifies which schema is expected to validate the structure of the model when no discriminating property is present. |
```

src/oas.md
```diff
@@ -3362,7 +3360,9 @@ components:
 
 #### Discriminator Object
 
-When request bodies or response payloads may be one of a number of different schemas, a Discriminator Object gives a hint about the expected schema of the document.
+When request bodies or response payloads may be one of a number of different schemas, these should use the JSON Schema `anyOf` or `oneOf` keywords to describe the possible schemas (see [Composition and Inheritance](#composition-and-inheritance-polymorphism)).
+
+A polymorphic schema MAY include a Discriminator Object, which defines the name of the property that may be used as a hint for which schema of the `anyOf` or `oneOf` is expected to validate the structure of the model.
```
jeremyfiel, line 3365:
```suggestion
A polymorphic schema MAY include a [Discriminator Object](#discriminator-object), which defines the name of the property that may be used as a hint for which schema of the `anyOf` or `oneOf` is expected to validate the structure of the model.
```

src/oas.md
```diff
@@ -2870,10 +2870,16 @@ The OpenAPI Specification allows combining and extending model definitions using
...
-When used, the `discriminator` indicates the name of the property that hints which schema definition is expected to validate the structure of the model.
-As such, the `discriminator` field MUST be a required field.
-There are two ways to define the value of a discriminator for an inheriting instance.
+
+JSON Schema also provides the `anyOf` and `oneOf` keywords, which allow defining multiple schemas where at least one or exactly one of them must be valid, respectively.
+As is the case with `allOf`, the schemas are validated _independently_.
+These keywords can be used to describe polymorphism, where a single field can accept multiple types of values.
+
+The OpenAPI specification extends the JSON Schema support for polymorphism by adding the [`discriminator`](#schema-discriminator) field whose value is a [Discriminator Object](#discriminator-object).
+When used, the Discriminator Object indicates the name of the property that hints which schema of an `anyOf` or `oneOf` is expected to validate the structure of the model.
```
jeremyfiel, line 2879:
```suggestion
When used, the [Discriminator Object](#discriminator-object) indicates the name of the property that hints which schema of an `anyOf` or `oneOf` is expected to validate the structure of the model.
```

jeremyfiel (approved):
This may be overkill to link all of the anchors, but I think that's how the spec usually does it.  Feel free to ignore 

There is one typo with the `defaultMapping` naming that should be corrected if you choose to ignore the anchor links.

src/oas.md
```diff
@@ -2870,10 +2870,16 @@ The OpenAPI Specification allows combining and extending model definitions using
...
+The discriminating property may be defined as required or optional, but when defined as an optional property the Discriminator Object must include a `defaultMapping` field that specifies which schema of the `anyOf` or `oneOf` is expected to validate the structure of the model when the discriminating property is not present.
```
handrews, line 2880:
```suggestion
The discriminating property may be defined as required or optional, but when defined as an optional property the Discriminator Object MUST include a `defaultMapping` field that specifies which schema of the `anyOf` or `oneOf`, or which schema that references the current schema in an `allOf`, is expected to validate the structure of the model when the discriminating property is not present.
```

src/oas.md
handrews, line 3365: Probably need some `allOf` language here, but I don't want to conflict with @jeremyfiel 's proposed edit.

src/oas.md
handrews: MAY and MUST need to be all-caps here.

src/oas.md
```diff
@@ -3395,12 +3396,29 @@ The behavior of any configuration of `oneOf`, `anyOf`, `allOf` and `discriminato
...
-The behavior of a `mapping` value that is both a valid schema name and a valid relative URI reference is implementation-defined, but it is RECOMMENDED that it be treated as a schema name.
+The behavior of a `mapping` value or `default` value that is both a valid schema name and a valid relative URI reference is implementation-defined, but it is RECOMMENDED that it be treated as a schema name.
 To ensure that an ambiguous value (e.g. `"foo"`) is treated as a relative URI reference by all implementations, authors MUST prefix it with the `"."` path segment (e.g. `"./foo"`).
 
 Mapping keys MUST be string values, but tooling MAY convert response values to strings for comparison.
 However, the exact nature of such conversions are implementation-defined.
 
+##### Optional discriminating property
+
+When the discriminating property is defined as optional, the [Discriminator Object](#discriminator-object) must include a `defaultMapping` field that specifies a schema that is expected to validate the structure of the model when the discriminating property is not present in the payload. This allows the schema to still be validated correctly even if the discriminating property is missing.
```
handrews, line 3407:
```suggestion
When the discriminating property is defined as optional, the [Discriminator Object](#discriminator-object) MUST include a `defaultMapping` field that specifies a schema that is expected to validate the structure of the model when the discriminating property is not present in the payload. This allows the schema to still be validated correctly even if the discriminating property is missing.
```

src/oas.md
handrews: must -> MUST again.

src/oas.md
```diff
@@ -3372,8 +3372,9 @@ Note that `discriminator` MUST NOT change the validation outcome of the schema.
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
-| <a name="property-name"></a>propertyName | `string` | **REQUIRED**. The name of the property in the payload that will hold the discriminating value. This property SHOULD be required in the payload schema, as the behavior when the property is absent is undefined. |
+| <a name="property-name"></a>propertyName | `string` | **REQUIRED**. The name of the discriminating property in the payload that will hold the discriminating value. The discriminating property may be defined as required or optional, but when defined as optional the Discriminator Object must include a `defaultMapping` field that specifies which schema is expected to validate the structure of the model when no discriminating property is present. |
 | <a name="discriminator-mapping"></a> mapping | Map[`string`, `string`] | An object to hold mappings between payload values and schema names or URI references. |
+| <a name="default"></a> defaultMapping | `string` | The schema name or URI reference to a schema that is expected to validate the structure of the model when the discriminating property is not present in the payload. |
```
handrews, line 3377: The name `defaultMapping` suggests to me that this is the default whenever there isn't a defined mapping.  So not just when the property is missing, but when the property has a value that cannot otherwise be resolved.  Is this the intention?  If not, I'd suggest a name tying to absence rather than "default."

handrews (dismissed):
As problematic as the `allOf` form is, I'm not sure I'm OK with removing the example entirely.  People do use it, and we can't remove it in a minor version, and it's extremely hard to understand even with an example.

src/oas.md
mikekistler: That was not the intention. But it is an interesting idea. This could be accomplished with a slightly different schema for the defaultMapping. Instead of `not: required: ['petType']` it could be:

```
    properties:
      petType:
        not:
          enum: ['dog', 'cat']
```

This might get unwieldy for a schema with many variants, but there is some attractiveness to this.

Thoughts?

mikekistler:
I pushed some commits that address most of the outstanding comments.

I made musts and mays uppercase. I added the language @handrews suggested about how discriminating values are used for `allOf` polymorphism. And I added back the example of polymorphism with allOf. It is a fair point that this pattern is confusing even with the example, so dropping it would only make matters worse.

I fixed the instance where `default` should have been `defaultMapping` -- thanks @jeremyfiel for catching this.

I did not make every instance of "Discriminator Object" into a link. When there were multiple occurrences of Discriminator Object in a section, I made the first one a link and left the rest as plain text. I think other "Object" types are handled similarly -- for example only 13 of the 20 occurrences of "Path Item Object" are links, the rest are plain text.

I left open the question about the name `defaultMapping` and whether we should let this apply when there is no other matching mapping element. Curious for folks opinions on this.

src/oas.md
```diff
@@ -3135,71 +3141,114 @@ components:
...
+          description: the size of the pack the dog is from
+          default: 0
+          minimum: 0
+      required:
+        - packSize
+```
+
+###### Models with Polymorphism Support and a Discriminator Object
+
+The following example extends the example of the previous section by adding a [Discriminator Object](#discriminator-object) to the `Pet` schema. Note that the Discriminator Object is only a hint to the consumer of the API, and does not change the validation outcome of the schema.
```
ralfhandl, line 3194:
```suggestion
The following example extends the example of the previous section by adding a [Discriminator Object](#discriminator-object) to the `Pet` schema. Note that the Discriminator Object is only a hint to the consumer of the API and does not change the validation outcome of the schema.
```

src/oas.md
```diff
@@ -3395,12 +3447,29 @@ The behavior of any configuration of `oneOf`, `anyOf`, `allOf` and `discriminato
...
+```yaml
+OtherPet:
+  type: object
+  not:
+    required: ['petType']
+```
+
+This will prevent the default mapping schema from validating a payload that includes the discriminating property, which would cause a validation of the payload to fail when polymorphism is described using the `oneOf` JSON schema keyword.
+
 ##### Examples
```
ralfhandl, line 3473: This section contains requirements (MUST) in addition to examples, so the section title is misleading.

src/oas.md
ralfhandl: This is in the section that defines the Discriminator Object, so no link.

src/oas.md
handrews: @ralfhandl hmm... that's a good point.  It might be that we don't do the RFC2119 capitalization in "informative" sections such as examples.  We would just need to make sure that every requirement is present with correct capitalization in the "normative" sections.  This is probably something that should go in the style guide.

src/oas.md
jeremyfiel: apologies for not checking the style guide first. Thanks @ralfhandl for the reminder

src/oas.md
```diff
@@ -3433,7 +3488,7 @@ MyResponseType:
...
+which means the payload _MUST_, by validation, match exactly one of the schemas described by `Cat`, `Dog`, or `Lizard`. Deserialization of a `oneOf` can be a costly operation, as it requires determining which schema matches the payload and thus should be used in deserialization. This problem also exists for `anyOf` schemas. A `discriminator` MAY be used as a "hint" to improve the efficiency of selection of the matching schema. The [Discriminator Object](#discriminator-object) cannot change the validation result of the `oneOf`, it can only help make the deserialization more efficient and provide better error messaging. We can specify the exact field that tells us which schema is expected to match the instance:
```
ralfhandl, line 3491:
```suggestion
which means the payload _must_, by validation, match exactly one of the schemas described by `Cat`, `Dog`, or `Lizard`. Deserialization of a `oneOf` can be a costly operation, as it requires determining which schema matches the payload and thus should be used in deserialization. This problem also exists for `anyOf` schemas. A `discriminator` may be used as a "hint" to improve the efficiency of selection of the matching schema. The [Discriminator Object](#discriminator-object) cannot change the validation result of the `oneOf`, it can only help make the deserialization more efficient and provide better error messaging. We can specify the exact field that tells us which schema is expected to match the instance:
```

src/oas.md
```diff
@@ -3438,7 +3511,7 @@ The expectation now is that a property with name `petType` _MUST_ be present in
 
 will indicate that the `Cat` schema is expected to match this payload.
 
-In scenarios where the value of the `discriminator` field does not match the schema name or implicit mapping is not possible, an optional `mapping` definition MAY be used:
+In scenarios where the value of the discriminating property does not match the schema name or implicit mapping is not possible, an optional `mapping` definition MAY be used:
```
ralfhandl, line 3514:
```suggestion
In scenarios where the value of the discriminating property does not match the schema name or implicit mapping is not possible, an optional `mapping` definition can be used:
```

src/oas.md
```diff
@@ -3458,6 +3531,30 @@ Here the discriminating value of `dog` will map to the schema `#/components/sche
 
 When used in conjunction with the `anyOf` construct, the use of the discriminator can avoid ambiguity for serializers/deserializers where multiple schemas may satisfy a single payload.
 
+When the discriminating property is defined as optional, the Discriminator Object MUST include a `defaultMapping` field that specifies a schema of the `anyOf` or `oneOf` is expected to validate the structure of the model when the discriminating property is not present in the payload. This allows the schema to still be validated correctly even if the discriminator property is missing.
```
ralfhandl, line 3534:
```suggestion
When the discriminating property is defined as optional, the Discriminator Object has to include a `defaultMapping` field that specifies a schema of the `anyOf` or `oneOf` is expected to validate the structure of the model when the discriminating property is not present in the payload. This allows the schema to still be validated correctly even if the discriminator property is missing.
```

ralfhandl (changes requested):
Avoid "MUST" and "MAY" in Examples section

src/oas.md
ralfhandl: Done:
* #4382

lornajane:
We'll wait until we can get input from @handrews on this one.

handrews (approved):
Looks good!  Thanks for the refinements.

handrews:
I think this was ready to go except waiting on me so I'll go ahead and merge it.

src/oas.md
```diff
@@ -3135,71 +3141,114 @@ components:
...
+          default: 0
+          minimum: 0
+      required:
+        - petType
+        - packSize
+```
+
+###### Models with Polymorphism Support using allOf and a Discriminator Object
+
+It is also possible to describe polymorphic models using `allOf`. The following example uses `allOf` with a [Discriminator Object](#discriminator-object) to describe a polymorphic `Pet` model.
```
hfhbd, line 3250: Is it valid (but duplicate) or illegal to also add a `oneOf` in Pet too?

mikekistler: This is covered in the example just above.

hfhbd: I do mean the combination of `oneOf` in `Pet` and `allOf` in `Cat`:
```yaml
components:
  schemas:
    Pet:
      type: object
      discriminator:
        propertyName: petType
      properties:
        name:
          type: string
        petType:
          type: string
      required:
        - name
        - petType
       oneOf:
    Cat: # "Cat" will be used as the discriminating value
      description: A representation of a cat
      allOf:
        - $ref: '#/components/schemas/Pet'
        - type: object
          properties:
            huntingSkill:
              type: string
              description: The measured skill for hunting
              enum:
                - clueless
                - lazy
                - adventurous
                - aggressive
          required:
            - huntingSkill
    Dog: # "Dog" will be used as the discriminating value
      description: A representation of a dog
      allOf:
        - $ref: '#/components/schemas/Pet'
        - type: object
          properties:
            packSize:
              type: integer
              format: int32
              description: the size of the pack the dog is from
              default: 0
              minimum: 0
          required:
            - packSize
```

handrews: @hfhbd no, that is not valid in terms of JSON Schema (and therefore also invalid OAS).

The reason is that this creates a cyclic reference _that cannot be disambiguated by runtime evaluation._

You can see this because `allOf` and `oneOf` are examples of an _in-place applicator_:  They apply one or more sub-schemas to the _same instance location._  JSON Schema implementations are required to detect infinite loops that occur when you revisit the same schema object twice without having changed instance location, which is what happens here.  Let me know if that is not clear enough and I can explain further.

hfhbd: Thank you for your helpful answer and clarification!

jeremyfiel:
@mikekistler  believe the OAS 3.2 `meta` schema needs to be updated with the new discriminator keyword.  
https://github.com/OAI/OpenAPI-Specification/blob/1f3782204954c4d4046309f6156fa33b58f0fb62/oas/3.2/meta/2025-05-14#L27-L45

PR #4340, opened 2025-02-06
build: Build accepts version argument

duncanbeevers:
### 🗣️ Conversation

The current `npm run build` script is not capable of one-off documentation builds.
The build script contains essential information and compilation steps necessary for previewing builds.

`npm run build` continues to build all versions
`npm run build 3.0.0` builds a specific version
`npm run build latest` builds from the latest spec

### DIFF
--- a/scripts/md2html/build.sh
+++ b/scripts/md2html/build.sh
@@ -15,8 +15,24 @@ cp -p ../../node_modules/respec/builds/respec-w3c.* ../../deploy/js/
 latest=`git describe --abbrev=0 --tags`
 latestCopied=none
 lastMinor="-"
-for filename in $(ls -1 ../../versions/[23456789].*.md | sort -r) ; do
-  if [[ ${filename} == *-editors.md ]];then
+
+# If $1 is not provided, process all versions
+# If $1 is "latest", process the last-tagged version
+# Otherwise, process the specified version
+if [ -z "$1" ]; then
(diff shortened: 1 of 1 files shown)


ralfhandl (approved):
@duncanbeevers Could you please explain your use case for this?

Background: we don't change the already published specs, and if we change the build scripts, we'll need to rebuild all published versions anyway to check whether they all are affected only in the expected ways.

The work-in-progress version `src/oas-md` in a `*dev` branch can already be built separately.

duncanbeevers:
I have been experimenting with generating variants for the examples included in the schema, rather than having the variants explicitly inlined. eg; schema authors write YAML, and the JSON example gets auto-generated.

In order to accomplish this, I've needed to make some changes to the build script, so I have to run it a lot, and I simply wanted an easier way to run it in a more-focused fashion.

I'd be happy to adjust my workflow if there's a better-isolated way to run one-off builds, though I don't quite understand your comment.

> The work-in-progress version src/oas-md in a *dev branch can already be built separately

I'll experiment in the `dev` branch then, and see whether that suffices.

duncanbeevers:
@ralfhandl Can you explain a little how the development process works? I'm happy to work in `dev`, but it appears that branch is missing some changes present in `main` (eg; dark mode support).

duncanbeevers:
Closing in favor of #4341

PR #4337, opened 2025-02-06
respec.yaml: use older Ubuntu

ralfhandl:
Respec depends on puppeteer which has a problem with the latest Ubuntu

Tick one of the following options:

- [ ] schema changes are included in this pull request
- [ ] schema changes are needed for this pull request but not done yet
- [x] no schema changes are needed for this pull request

PR #4358, opened 2025-02-13
Run respec without sandbox

ralfhandl:
Alternative to
* #4337

Respec depends on puppeteer which does not support running in a sandbox with the latest Ubuntu. Official recommendation:

> **[What if I don‘t have root access to the machine and can’t install anything?](https://chromium.googlesource.com/chromium/src/+/main/docs/security/apparmor-userns-restrictions.md#what-if-i-dont-have-root-access-to-the-machine-and-cant-install-anything)**
>
> You will need to run developer builds with the `--no-sandbox` command line flag, but be aware that this disables critical security features of Chromium and should never be used when browsing the open web.

I don't want to mess with the GitHub Ubuntu image, we are not browsing the open web, and we only tell puppeteer to access local HTML files that we generated ourselves, so this should be safe enough.

### DIFF
--- a/scripts/md2html/build.sh
+++ b/scripts/md2html/build.sh
@@ -54,7 +54,7 @@ for specification in $specifications; do
   echo === Building $version to $destination
 
   node scripts/md2html/md2html.js --maintainers $maintainers $specification > $tempfile
-  npx respec --use-local --src $tempfile --out $destination
+  npx respec --no-sandbox --use-local --src $tempfile --out $destination
   rm $tempfile
 
   echo === Built $destination


duncanbeevers:
I don't know how far down the rabbit hole it's worth going on this, but I don't think it's unreasonable to set up an AppArmor profile for puppeteer.

There's an `aa-genprof` tool we can use during a build to determine what permissions are necessary and to automatically generate a minimal AppArmor profile, which we can then commit and enforce.

ralfhandl:
I do not want to go anywhere near that rabbit hole.

There are two easy ways to get out spec build process running again:
- use the last working Github Action runner version that worked (#4337)
- use a respec option to make it work (this PR)

Or we can leave the build process broken until the involved tools dtart working together again in a future version.

lornajane (approved):
I think this is a good solution for the time being. Thank you!

PR #4365, opened 2025-02-17
gh-pages: show OAS 2.0 spec

ralfhandl:
It got lost with the introduction of minor-version symlinks

- index page: add bullet point for OAS 2.0 - `specification-version-list` include intentionally ignores files without patch version
- `specification-version-list` include: remove unwanted whitespace at end of list - messes up spacing of bullet list if items are added on index page (typo that escaped notice so far)
- `.gitignore`: add local folders needed by build scripts on `main` and `dev` branches

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 _site/
+coverage/
 deploy/
+deploy-preview/
 node_modules/
 scripts/
 package-lock.json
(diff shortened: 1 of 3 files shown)


handrews (approved):
Thanks!

Random trivia from my searches through the history while working on the patch releaes:  The documents released as Swagger 2.0 and as OAS 2.0 have numerous small differences.  There were effectively like 3 patch releases of 2.0, we just didn't mark them in any visible way.

lornajane (approved):

Issue #4152, opened 2024-10-22
Make the "latest" schema accessible programmatically


handrews:
We don't want schema URIs with `latest` in place of the date because schema contents should not change unexpectedly.  However, we need some way for folks to access the latest schemas, preferably without pulling all of the dates and sorting them.

Unfortunately, GitHub pages does not support HTTP redirects (or setting media types independent of file extensions) so we need to figure out a workarond.  Or move to Netlify or Cloudflare Pages or something else with actual proper HTTP support.

LasneF:
another alternative would be to push it to a code repo like npm that handle this kind of versionning policy

cboitel:
From a user perspective, i would expect to be able to access the latest schema for a specific OpenAPI version so i can validate an OpenAPI specification document from it. 

E.g. :
1. open document, extract version
2. download schema for version (/vX.Y.Z/schema)
3. validate doc with downloaded schema

NPM targets only Javascript ecopsystems: what about rust/golang/c++/... ?

Having an OpenAPI documented API which describes all operations available at specs.openapis.org would also be a must-have.

Hope it helps.

seriousme:
Hi,

If you use a symbolic link in the GH-pages branch (say from `latest` to `2024‑11‑14` then toolmakers that poll `latest` will notice that there is a new version and can work from there, also via HTTPS.

Currently my [tooling](https://github.com/seriousme/openapi-schema-validator/blob/master/test/check-versions.js) just checks out the GH-pages branch once a week and compares what I have to what is there to find updates. Every now and then I need to tune the mechanism a bit to accomodate for slight changes in the branch structure. 

For me it would be brilliant if:
- schema files explicitly have the .json extension, or are not surrounded by other , non-schema files. (e.g. I now need to filter out HTML files)
- there would be a single folder with all the latest patch versions per minor version (e.g. like https://github.com/seriousme/openapi-schema-validator/tree/master/schemas)  with a either a distinct folder or clear naming convention between base and full schema.
- archived versions are in a separate folder.

Even better would indeed be an NPM/Python/Go package, but I can see that this could become messy for the OAI project. So you could argue that packaging it up should be left to the community of the specific language.
I'd be happy to publish the schema's on NPM, separated from the rest of my tooling.

Alternatively there's sites like:  https://www.schemastore.org/ which just publish the schema. 

Hope this helps!

Kind regards,
Hans

ralfhandl:
@seriousme Hi Hans,

[...]

PR #4376, opened 2025-02-20
Update schema-publish.sh

### LINKED ISSUE #4223: Create dev build infrastructure

duncanbeevers:
Fixes #4223

## 🗣️ Discussion

This PR brings the schema publishing workflow into closer alignment with the specification publishing workflow.

### 🔧 Under-the-hood changes

* Configure deploydir based on branch name
* Check branch name for conformance to vX.Y-dev structure
  * build to `/deploy/oas/$version` if branch name is conformant
  * build to `/deploy-preview` if branch name is non-conformant
* Perform only one traversal of schemas structure
* Remove `datesHash` data structure
* Accumulate sed sub-commands & invoke with sed single -e argument

### 🧑 Contributor-facing Changes

* <details><summary>Schema publishing process works out-of-the-box on macOS</summary>A side-effect of removing intermediate data structures is that we avoid bash-4-isms, enabling out-of-the-box builds on modern macOS versions which ship an old 3.2.57 bash version.</details>
* Schema publishing process from ad-hoc branch names produces sensible artifacts

## 🚗 Demo

This screenshot demonstrates running the `scripts/schema-publish.sh` script from different branch names.
It also includes an interrogation of a generated schema file in order to validate that all of its placeholders are replaced correctly. (Thanks [@ralfhandl](https://github.com/OAI/OpenAPI-Specification/pull/4376#discussion_r1964152528)!)

Screenshot demonstrating running the schemas build process in different branch names, showing how the deploy directory changes, and also showing how multiple different placeholders are replaced with their correct values.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
   "license": "Apache-2.0",
   "scripts": {
     "build": "bash ./scripts/md2html/build.sh",
-    "build-src": "npm run validate-markdown && bash ./scripts/md2html/build.sh src",
+    "build-src": "npm run validate-markdown && bash ./scripts/md2html/build.sh src && bash ./scripts/schema-publish.sh src",
     "test": "c8 --100 vitest --watch=false && bash scripts/schema-test-coverage.sh",
     "format-markdown": "bash ./scripts/format-markdown.sh ./src/oas.md",
     "validate-markdown": "npx mdv src/oas.md && npx markdownlint-cli src/oas.md"
--- a/scripts/schema-publish.sh
+++ b/scripts/schema-publish.sh
@@ -6,15 +6,61 @@
 
 schemaDir="src/schemas/validation"
 branch=$(git branch --show-current)
-version=${branch:1:3}
-echo === Building schemas into ./deploy/oas/$version
+
+
+if [ -z "$1" ]; then
+  if [[ $branch =~ ^v([0-9]+\.[0-9]+)-dev$ ]]; then
+    deploydir="./deploy/oas/${BASH_REMATCH[1]}"
+  else
+    echo "Unable to determine version from branch name; should be vMAJOR.MINOR.PATCH-dev"
+    exit 1
+  fi
+elif [ $1 = "src" ]; then
+  deploydir="./deploy-preview"
+else
+  echo "Unrecognized argument"
+  exit 1
+fi
+
+# create the date-stamped schemas
+publish_schema() {
+  local schema="$1"
+  local date="$2"
+  local sedCmd="$3"
+
+  local base=$(basename $schema '.yaml')
+  local target=$deploydir/$base/$date
+
+  mkdir -p $deploydir/$base
+  # replace the WORK-IN-PROGRESS placeholders
+  sed -e $sedCmd $schemaDir/$schema > $target.yaml
+
+  node scripts/yaml2json/yaml2json.js "$target.yaml"
+  rm "$target.yaml"
+  mv "$target.json" "$target"
+
+  # Find the jekyll lander markdown file for this iteration.
+  local jekyllLander=$(find "$deploydir/$base" -maxdepth 1 -name "*.md")
+
+  # Move the jekyll lander markdown for this iteration to the deploy destination.
+  # The lander files only exist in the gh-pages branch.
+  if [ ! -z "$jekyllLander" ]; then
+    mv $jekyllLander $target.md
+    echo " * $newestCommitDate: $schema & jekyll lander $(basename $jekyllLander)"
+  else
+    echo " * $newestCommitDate: $schema"
+  fi
+
+}
+
+echo === Building schemas into $deploydir
 
 # list of schemas to process, dependent schemas come first
 schemas=(meta.yaml dialect.yaml schema.yaml schema-base.yaml)
 
-# find the newest commit date for each schema
+# publish each schema using its or any of its dependencies newest commit date.
 maxDate=""
-declare -A datesHash
+sedCmds=()
 for schema in "${schemas[@]}"; do
   if [ -f  "$schemaDir/$schema" ]; then
     newestCommitDate=$(git log -1 --format="%ad" --date=short "$schemaDir/$schema")
@@ -23,31 +69,13 @@ for schema in "${schemas[@]}"; do
     if [ "$newestCommitDate" \> "$maxDate" ]; then
       maxDate=$newestCommitDate
     fi
-    datesHash["$schema"]=$maxDate
-    echo $schema changed at $newestCommitDate
+
+    base=$(basename $schema '.yaml')
+    # Add the replacement for this schema's placeholder to list of sed commands.
+    sedCmds+=("s/${base}\/WORK-IN-PROGRESS/${base}\/${maxDate}/g")
+
+    publish_schema "$schema" "$maxDate" $(printf '%s;' "${sedCmds[@]}")
   fi
-done
-
-# construct sed command
-sedCmd=()
-for schema in "${!datesHash[@]}"; do
-  base=$(basename "$schema" .yaml)
-  sedCmd+=("-e s/$base\/WORK-IN-PROGRESS/$base\/${datesHash[$schema]}/g")
-done
-
-# create the date-stamped schemas
-for schema in "${!datesHash[@]}"; do
-  base=$(basename "$schema" .yaml)
-  target=deploy/oas/$version/$base/${datesHash[$schema]}
-
-  mkdir -p "deploy/oas/$version/$base"
-
-  sed ${sedCmd[@]} $schemaDir/$schema > $target.yaml
-  node scripts/yaml2json/yaml2json.js $target.yaml
-  rm $target.yaml
-  mv $target.json $target
-
-  mv deploy/oas/$version/$base/*.md $target.md
 done
 
 echo === Built


scripts/schema-publish.sh
```diff
@@ -6,15 +6,38 @@
...
+publish_schema() {
+  local schema="$1"
+  local date="$2"
+
+  local base=$(basename $schema '.yaml')
+  local target=$deploydir/$base/$date
+
+  mkdir -p $deploydir/$base
+  # replace the WORK-IN-PROGRESS placeholder
+  sed -e "s/${base}\\/WORK-IN-PROGRESS/${base}\\/${date}/g" $schemaDir/$schema > $target.yaml
```
ralfhandl, line 26: This is insufficient as the four schema files reference each other, and these references also need to be updated with the date for the corresponding referenced schema file, for example in 

https://github.com/OAI/OpenAPI-Specification/blob/9fd4e2ebb49625f407e176ab507e2ca693789a5c/src/schemas/validation/dialect.yaml#L21

ralfhandl (changes requested):

scripts/schema-publish.sh
duncanbeevers: Thanks. I pushed a change to address this by accumulating sed sub-commands in an array and serializing the full sed script as an argument to `publish_schema`.

scripts/schema-publish.sh
```diff
@@ -23,31 +48,13 @@ for schema in "${schemas[@]}"; do
...
+    sedCmds+=("s/${base}\/WORK-IN-PROGRESS/${base}\/${maxDate}/g")
 
-  mkdir -p "deploy/oas/$version/$base"
-
-  sed ${sedCmd[@]} $schemaDir/$schema > $target.yaml
-  node scripts/yaml2json/yaml2json.js $target.yaml
-  rm $target.yaml
-  mv $target.json $target
-
-  mv deploy/oas/$version/$base/*.md $target.md
```
ralfhandl, line 50: This command seems to have been lost. It is necessary when building to `deploy/` for the

> * view schema/2024-11-14

links on https://spec.openapis.org to work.

ralfhandl (changes requested):

scripts/schema-publish.sh
duncanbeevers: I saw that command, but couldn't reason what it was doing, since no `.md` files are generated in `/deploy/**`.

Can you point to me where the link-updating bit happens?

ralfhandl: The `*.md` files only exist in the `gh-pages` branch, for example in https://github.com/OAI/OpenAPI-Specification/tree/gh-pages/oas/3.1/schema.

The `title` of each file is specific to that file, and I didn't want to hard-code them in the build script to minimize the differences between the OpenAPI, Overlay, and Arazzo copies of the build script.

duncanbeevers: I'll re-add this command, but I think it indicates a weird space in the abstraction. It seems contradictory to me to both `mkdir -p deploy/oas/3.1.1` *and* to expect that there are already markdown files present there.

I also think it's a bad pattern to do a `mv` where the input could be multiple files, but the destination is a single file.

It looks like the file we expect to move is actually named the same as $date. Short of [restoring `latest`](https://github.com/OAI/OpenAPI-Specification/pull/4133), could I just move that one `$date.md` file?

Alternatively, we could go back to having a `latest.md`, but on publish, rename it to match the latest iteration of the schema. This would preserve the publishing behavior (not creating a `latest` link or schema), while still removing step of manually renaming the `.md` file with each revision *and* being able to access a stable filename for copying in this script.

duncanbeevers: I've pushed a change which replicates the current `*.md` behavior, as well as introduces some commented-out code probing the pain points mentioned in [my comment above](https://github.com/OAI/OpenAPI-Specification/pull/4376#discussion_r1964517681).

* 😐 One implementation supposes the `.md` file's name always matches the iteration date, which matches the [current setup](https://github.com/OAI/OpenAPI-Specification/tree/gh-pages/oas/3.1/schema).

* 😄 One implementation supposes the `.md` file is always `latest.md`. This used to be the setup, but it was [abandoned](https://github.com/OAI/OpenAPI-Specification/pull/4133). I think this approach is best, as it minimizes the steps necessary to begin iterating on a new version of the schemas.

* 😞 One implementation mirrors the current behavior, but selects only the "first" markdown file encountered. This aligns better with the sentiment _A `mv` command whose destination is a file should have a file as its first argument_, but in a kind of nonsense way.

* 😞 The actual implementation mirrors the current behavior, attempting to copy *all* the markdown files in the deploy directory to $target.md. If there are multiple markdown files present this results in an error later in the script. The actual implementation mirrors the previous one with the `head -n 1` removed, to highlight its similarities to that approach.

scripts/schema-publish.sh
```diff
@@ -6,15 +6,69 @@
 
 schemaDir="src/schemas/validation"
 branch=$(git branch --show-current)
-version=${branch:1:3}
-echo === Building schemas into ./deploy/oas/$version
+
+if [[ $branch =~ ^v([0-9]+\.[0-9]+\.[0-9]+) ]]; then
```
ralfhandl, line 10:
```suggestion
if [[ $branch =~ ^v([0-9]+\.[0-9]+)-dev$ ]]; then
```
The deploy/oas subfolders are named 3.0, 3.1, etc., only MAJOR.MINOR.

scripts/schema-publish.sh
```diff
@@ -6,15 +6,69 @@
...
+  
+  # Define jekyllLander by finding the first markdown file already-present in the
+  # deploy directory.
+  # local jekyllLander=$(find "$deploydir/$base" -maxdepth 1 -name "*.md" | head -n 1)
+
+  # Define jekyllLander by finding all the markdown files already-present in the
+  # deploy directory. If multiple files are present, this script will fail as it
+  # attempts to move these multiple files to a single $target.md
+  # This behavior is the closest match to that of the previous version of the
+  # schema-publish script.
```
ralfhandl, line 50:
```suggestion
```
Let's not confuse the next maintainers with possible future implementation alternatives.

Let's instead discuss the future implementation in #4152.

duncanbeevers: Absolutely. This was really intended as placeholder for discussion. Seems well-enough captured here now to be referenced in #4152. I'll pull all this cruft.

scripts/schema-publish.sh
duncanbeevers: I adjusted the capture group to only grab MAJOR MINOR, while still demanding the presence of PATCH.
Hence, dev branches are still expected to be named `vX.Y.Z-whatever`.

scripts/schema-publish.sh
```diff
@@ -6,15 +6,52 @@
 
 schemaDir="src/schemas/validation"
 branch=$(git branch --show-current)
-version=${branch:1:3}
-echo === Building schemas into ./deploy/oas/$version
+
+if [[ $branch =~ ^v([0-9]+\.[0-9]+)\.[0-9]+ ]]; then
```
ralfhandl, line 10: The branches are named
* v3.1-dev
* v3.2-dev

so the pattern really needs to be
```suggestion
if [[ $branch =~ ^v([0-9]+\.[0-9]+)-dev$ ]]; then
```

package.json
```diff
@@ -14,7 +14,7 @@
   "license": "Apache-2.0",
   "scripts": {
     "build": "bash ./scripts/md2html/build.sh",
-    "build-src": "npm run validate-markdown && bash ./scripts/md2html/build.sh src",
+    "build-src": "npm run validate-markdown && bash ./scripts/md2html/build.sh src && ./scripts/schema-publish.sh",
```
duncanbeevers, line 17: > [Could you extend the build-src npm script](https://github.com/OAI/OpenAPI-Specification/pull/4376#pullrequestreview-2632281517)

I made this change, but I dislike how it makes the branch name an implicit input to the `build-src` task; eg it may, or may not build entirely to `/deploy-preview`, depending…

In `build.sh` the behavior is driven by an explicit argument.

I'm going to change the `schema-publish.sh` behavior to match, requiring `src` as an input in order to build to `/deploy-preview`, and I'll change the branch name check to a validation instead of an implicit input.

ralfhandl (changes requested):
Almost there. And please trust me, I've tested it on a `v9.9-dev` branch 😄

ralfhandl (approved):
Thanks!

scripts/schema-publish.sh
```diff
@@ -7,10 +7,19 @@
...
+    deploydir="./deploy/oas/${BASH_REMATCH[1]}"
+  else
+    echo "Unable to determine version from branch name; should be vMAJOR.MINOR.PATCH-dev"
+    exit 1
+  fi
+elif [ $1 = "src" ]; then
   deploydir="./deploy-preview"
+else
+  echo "Unrecognized argument"
+  exit 1
```
ralfhandl, line 22: This change causes the npm script `build-src` to fail.

The "old" logic of building to `deploy` in `vX.Y-dev` branches and to `deploy-preview` in all other branches seems to cover all reasonable use cases.

Why would I ever want to build to `deploy-preview` in a `vX.Y-dev` branch?

ralfhandl (changes requested):

scripts/schema-publish.sh
duncanbeevers: I imagined that while developing schema versions, running `build-src` would create a predictable set of artifacts in `deploy-preview`, including the specification and its current schema.

ralfhandl (approved):

duncanbeevers:
I'm going squash this to a single commit to avoid [another `fixup!`](https://github.com/OAI/OpenAPI-Specification/commit/639c6cadcbaab6df23e2cdf496680458694d87ce) commit hitting `main`.

PR #4382, opened 2025-02-25
style guide: no requirements in "Examples" sections

ralfhandl:
Follow-up to https://github.com/OAI/OpenAPI-Specification/pull/4339#discussion_r1958587868.

PR #4339 added text to an "Examples" section that used requirements language with "MUST" and "MAY". In this case the text was just copied verbatim from "non-example" sections of the spec, and no new requirements were added.

Better style is to
* not to add requirements in a section called "Examples" or similar because readers may be confused whether these are really normative requirements or just explanations to the examples
* not to use requirements language in explanations to examples

### DIFF
--- a/style-guide.md
+++ b/style-guide.md
@@ -21,7 +21,7 @@ The following additional rules should be followed but currently are not enforced
 9. Use [Oxford commas](https://en.wikipedia.org/wiki/Serial_comma), avoid [Shatner commas](https://www.latimes.com/archives/blogs/jacket-copy/story/2011-06-30/goodbye-oxford-comma-hello-shatner-comma).
 10. Use `<span id="thing"></span>` for link anchors. The `<a name="thing"></a>` format has been deprecated.
 11. Headings use [title case](https://en.wikipedia.org/wiki/Title_case) and are followed by a blank line.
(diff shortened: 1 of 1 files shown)


style-guide.md
```diff
@@ -21,6 +21,7 @@ The following additional rules should be followed but currently are not enforced
 9. Use [Oxford commas](https://en.wikipedia.org/wiki/Serial_comma), avoid [Shatner commas](https://www.latimes.com/archives/blogs/jacket-copy/story/2011-06-30/goodbye-oxford-comma-hello-shatner-comma).
 10. Use `<span id="thing"></span>` for link anchors. The `<a name="thing"></a>` format has been deprecated.
 11. Headings use [title case](https://en.wikipedia.org/wiki/Title_case) and are followed by a blank line.
+12. Do not add requirements in "Examples" sections, and avoid [RFC2119 key words (MUST, MAY, ...)](https://datatracker.ietf.org/doc/html/rfc2119) in these sections.
```
lornajane, line 24: What are requirements here? Is that the same thing as using a MUST/MAY keyword. Would it be accurate to say that this sort of content needs to be in a different section?

lornajane:
I tried to follow the thread of the other pull request, some comments, and what caused this change, but I'm not really clear, sorry. Could you add a note to the pull request description about what problem this solves or what we need to do or not do? I think it would help to inform the reviewer (me!)

lornajane (approved):
Thanks for the update and the description, feel much more confident that I know what I'm approving this time!

PR #4386, opened 2025-02-27
Remove yaml2json

duncanbeevers:
The `yaml2json` script duplicates functionality already present in the `yaml` module / CLI.
Rather than generating and juggling tempfiles,

This PR generates the schema JSON files by piping and redirecting…
…instead of generating-then-transforming-then-deleting-then-moving


scripts/schema-publish.sh
```diff
@@ -32,12 +32,9 @@ publish_schema() {
...
 
   mkdir -p $deploydir/$base
-  # replace the WORK-IN-PROGRESS placeholders
-  sed -e $sedCmd $schemaDir/$schema > $target.yaml
 
-  node scripts/yaml2json/yaml2json.js "$target.yaml"
-  rm "$target.yaml"
-  mv "$target.json" "$target"
+  # replace the WORK-IN-PROGRESS placeholders
+  sed ${sedCmd[@]} $schemaDir/$schema | npx yaml --json --indent 2 --single > $target
```
duncanbeevers, line 37: The `--json` flag transforms the streamed yaml to JSON.
The `--indent 2` flag specifies the [indentation level of the JSON](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#space)
The `--single` flag indicates we are processing a single file and not a stream, which prevents the output from being wrapped in an extra `[]` pair.

ralfhandl (approved):

PR #4388, opened 2025-02-28
Allow URIs for Security Schemes

### LINKED ISSUE #3776: Use URI references for Security Requirements in 3.2

handrews:
Fixes #3776

This allows Security Requirement Objects to reference Security Scheme Objects by URI instead of implicit component name. Without this ability, it is difficult to share Security Schemes in a way that is consistent with re-usable component documents.

This approach provides parity with how the Discriminator Object's mapping field works.

Also add a note about the complexity of these rules to the Security Considerations section.

Note that since the existing name syntax is still valid, and no constraints are present on names, there are no meaningful schema updates for this change.

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -214,8 +214,8 @@ This allows Security Scheme Objects and Tag Objects to be defined next to the AP
 
 The interface approach can also work for Discriminator Objects and Schema Objects, but it is also possible to keep the Discriminator Object's behavior within a single document using the relative URI-reference syntax of `mapping`.
 
-There are no URI-based alternatives for the Security Requirement Object or for the Operation Object's `tags` field.
-These limitations are expected to be addressed in a future release.
+There are no URI-based alternatives for the Operation Object's `tags` field.
+This limitation is expected to be addressed in a future release.
 
 See [Appendix F: Resolving Security Requirements in a Referenced Document](#appendix-f-resolving-security-requirements-in-a-referenced-document) for an example of the possible resolutions, including which one is recommended by this section.
 The behavior for Discrimator Object non-URI mappings and for the Operation Object's `tags` field operate on the same principles.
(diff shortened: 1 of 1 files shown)


mikekistler (dismissed):
Looks good! 👍

src/oas.md
```diff
@@ -4069,7 +4069,12 @@ flows:
 #### Security Requirement Object
 
 Lists the required security schemes to execute this operation.
-The name used for each property MUST correspond to a security scheme declared in the [Security Schemes](#security-scheme-object) under the [Components Object](#components-object).
+
+The name used for each property MUST either correspond to a security scheme declared in the [Security Schemes](#security-scheme-object) under the [Components Object](#components-object), or be the URI of a Security Scheme Object.
+Property names that match the syntax of a component name under the Components Object MUST be treated as a component name.
```
ralfhandl, line 4074: I'd prefer
```suggestion
Property names that are the name of a security scheme declared under the Components Object MUST be treated as a component name.
```
"Match the syntax" can be misunderstood as match the regular expression `^[a-zA-Z0-9\.\-_]+$` irrespective of the existence of a security scheme with that name.

src/oas.md
```diff
@@ -4082,8 +4087,8 @@ An empty Security Requirement Object (`{}`) indicates anonymous access is suppor
...
-| ---- | :----: | ---- |
-| <a name="security-requirements-name"></a>{name} | [`string`] | Each name MUST correspond to a security scheme which is declared in the [Security Schemes](#security-scheme-object) under the [Components Object](#components-object). If the security scheme is of type `"oauth2"` or `"openIdConnect"`, then the value is a list of scope names required for the execution, and the list MAY be empty if authorization does not require a specified scope. For other security scheme types, the array MAY contain a list of role names which are required for the execution, but are not otherwise defined or exchanged in-band. |
+| --- | :---: | --- |
+| <a name="securityRequirementsName"></a>{name} | [`string`] | Each name or URI MUST correspond to a security scheme as described above. If the security scheme is of type `"oauth2"` or `"openIdConnect"`, then the value is a list of scope names required for the execution, and the list MAY be empty if authorization does not require a specified scope. For other security scheme types, the array MAY contain a list of role names which are required for the execution, but are not otherwise defined or exchanged in-band. |
```
ralfhandl, line 4091: Why change the anchor name? The dash style is consistent with other anchor names for fixed fields.

handrews: @ralfhandl it was unintentional, thanks for catching.  This  was ported from a much earlier PR against the old v3.2.0-dev branch.  I'll check to maker sure there's nothing else ported wrong and then update.

handrews:
@ralfhandl this should now have your comments addressed.

ralfhandl (approved):

PR #4389, opened 2025-02-28
Add `$self` for self-identifying documents

handrews:
See [the proposal](https://github.com/OAI/OpenAPI-Specification/blob/main/proposals/2024-08-01-Self-Identification.md) for background information.

This is a rather minimal approach, as @karenetheridge and I plan to work together on a more thorough revamp of the document parsing / reference resolution sections.

This adds `$self` as a way for a document to define its own URI for use in reference targets, and as the base URI for relative URI references in the document.

This does not impact the resolution of relative API URLs. _[**NOTE:** I'm not entirely sure about this, but it seems more useful this way to allow multiple deployed _locations_ of an OAD to correspond to multiple deployments of the API.)_

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -310,8 +310,10 @@ Unless specified otherwise, all fields that are URIs MAY be relative references 
 
 Relative references in [Schema Objects](#schema-object), including any that appear as `$id` values, use the nearest parent `$id` as a Base URI, as described by [JSON Schema Specification Draft 2020-12](https://tools.ietf.org/html/draft-bhutton-json-schema-00#section-8.2).
 
-Relative URI references in other Objects, and in Schema Objects where no parent schema contains an `$id`, MUST be resolved using the referring document's base URI, which is determined in accordance with [[RFC3986]] [Section 5.1.2 – 5.1.4](https://tools.ietf.org/html/rfc3986#section-5.1.2).
+Relative URI references in other Objects, and in Schema Objects where no parent schema contains an `$id`, MUST be resolved using the `$self` field of the [OpenAPI Object](#openapi-object).
+If no `$self` field is present relative URI references MUST be resolved using the referring document's base URI, which is determined in accordance with [[RFC3986]] [Section 5.1.2 – 5.1.4](https://tools.ietf.org/html/rfc3986#section-5.1.2).
 In practice, this is usually the retrieval URI of the document, which MAY be determined based on either its current actual location or a user-supplied expected location.
+The document's base URI MUST also be used to resolve a relative `$self` URI reference.
 
 If a URI contains a fragment identifier, then the fragment should be resolved per the fragment resolution mechanism of the referenced document. If the representation of the referenced document is JSON or YAML, then the fragment identifier SHOULD be interpreted as a JSON-Pointer as per [RFC6901](https://tools.ietf.org/html/rfc6901).
 
@@ -322,7 +324,7 @@ Relative references in CommonMark hyperlinks are resolved in their rendered cont
 API endpoints are by definition accessed as locations, and are described by this specification as **_URLs_**.
 
 Unless specified otherwise, all fields that are URLs MAY be relative references as defined by [RFC3986](https://tools.ietf.org/html/rfc3986#section-4.2).
-Unless specified otherwise, relative references are resolved using the URLs defined in the [Server Object](#server-object) as a Base URL. Note that these themselves MAY be relative to the referring document.
+Unless specified otherwise, relative references are resolved using the URLs defined in the [Server Object](#server-object) as a Base URL. Note that these themselves MAY be relative to the referring document (**NOT** the [OpenAPI Object's](#openapi-object) `$self` field).
 
 ### Schema
 
@@ -342,6 +344,7 @@ This is the root object of the [OpenAPI Description](#openapi-description).
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
 | <a name="oas-version"></a>openapi | `string` | **REQUIRED**. This string MUST be the [version number](#versions) of the OpenAPI Specification that the OpenAPI Document uses. The `openapi` field SHOULD be used by tooling to interpret the OpenAPI Document. This is _not_ related to the API [`info.version`](#info-version) string. |
+| <a name-"oas-self"></a>$self | `URI-reference` (without a fragment) | Sets the URI of this document, which also serves as its base URI in accordance with [RFC 3986 §5.1.1](https://www.rfc-editor.org/rfc/rfc3986#section-5.1.1); the value MUST NOT be the empty string and MUST NOT contain a fragment (even if the fragment is empty).  Implementations MUST support referencing a document by the resolved URI defined by this field. |
 | <a name="oas-info"></a>info | [Info Object](#info-object) | **REQUIRED**. Provides metadata about the API. The metadata MAY be used by tooling as required. |
 | <a name="oas-json-schema-dialect"></a> jsonSchemaDialect | `string` | The default value for the `$schema` keyword within [Schema Objects](#schema-object) contained within this OAS document. This MUST be in the form of a URI. |
 | <a name="oas-servers"></a>servers | [[Server Object](#server-object)] | An array of Server Objects, which provide connectivity information to a target server. If the `servers` field is not provided, or is an empty array, the default value would be a [Server Object](#server-object) with a [url](#server-url) value of `/`. |
@@ -475,7 +478,7 @@ An object representing a Server.
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
-| <a name="server-url"></a>url | `string` | **REQUIRED**. A URL to the target host. This URL supports Server Variables and MAY be relative, to indicate that the host location is relative to the location where the document containing the Server Object is being served. Variable substitutions will be made when a variable is named in `{`braces`}`. |
+| <a name="server-url"></a>url | `string` | **REQUIRED**. A URL to the target host. This URL supports Server Variables and MAY be relative, to indicate that the host location is relative to the location where the document containing the Server Object is being served. Variable substitutions will be made when a variable is named in `{`braces`}`.   Note that the [OpenAPI Object's](#openapi-object) `$self` field is **NOT** used for relative URL reference resolution. |
 | <a name="server-description"></a>description | `string` | An optional string describing the host designated by the URL. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation. |
 | <a name="server-variables"></a>variables | Map[`string`, [Server Variable Object](#server-variable-object)] | A map between a variable name and its value. The value is used for substitution in the server's URL template. |
 
--- a/src/schemas/validation/schema.yaml
+++ b/src/schemas/validation/schema.yaml
@@ -8,6 +8,9 @@ properties:
   openapi:
     type: string
     pattern: '^3\.1\.\d+(-.+)?$'
+  $self:
+    type: string
+    format: uri-reference
   info:
     $ref: '#/$defs/info'
   jsonSchemaDialect:


lornajane (commented):
This looks pretty solid, but could we introduce use of $self in one or two examples so people realise it's there?

ralfhandl:
Looks good. An example would be nice.

handrews:
@ralfhandl I thought I did include one but apparently I did that on an earlier attempt on a different branch.  I'll port it over in the morning.

src/oas.md
```diff
@@ -310,8 +310,10 @@ Unless specified otherwise, all fields that are URIs MAY be relative references
...
 
-Relative URI references in other Objects, and in Schema Objects where no parent schema contains an `$id`, MUST be resolved using the referring document's base URI, which is determined in accordance with [[RFC3986]] [Section 5.1.2 – 5.1.4](https://tools.ietf.org/html/rfc3986#section-5.1.2).
+Relative URI references in other Objects, and in Schema Objects where no parent schema contains an `$id`, MUST be resolved using the `$self` field of the [OpenAPI Object](#openapi-object).
+If no `$self` field is present relative URI references MUST be resolved using the referring document's base URI, which is determined in accordance with [[RFC3986]] [Section 5.1.2 – 5.1.4](https://tools.ietf.org/html/rfc3986#section-5.1.2).
 In practice, this is usually the retrieval URI of the document, which MAY be determined based on either its current actual location or a user-supplied expected location.
+The document's base URI MUST also be used to resolve a relative `$self` URI reference.
```
karenetheridge, line 316: add "...which is then used to resolve relative URI references in other Objects, as above" ?

An example to use lower down could be: the document has `"$self": "/api"` in it, but at runtime the base URI is provided as as "https://dev.example.com" -- therefore the effective value of $self for this document (the URI to use as the base for all future relative resolutions) is "https://dev.example.com/api".

src/oas.md
```diff
@@ -322,7 +324,7 @@ Relative references in CommonMark hyperlinks are resolved in their rendered cont
 API endpoints are by definition accessed as locations, and are described by this specification as **_URLs_**.
 
 Unless specified otherwise, all fields that are URLs MAY be relative references as defined by [RFC3986](https://tools.ietf.org/html/rfc3986#section-4.2).
-Unless specified otherwise, relative references are resolved using the URLs defined in the [Server Object](#server-object) as a Base URL. Note that these themselves MAY be relative to the referring document.
+Unless specified otherwise, relative references are resolved using the URLs defined in the [Server Object](#server-object) as a Base URL. Note that these themselves MAY be relative to the referring document (**NOT** the [OpenAPI Object's](#openapi-object) `$self` field).
```
karenetheridge, line 327: add "..., which may itself be relative)" to the end of this sentence?

src/oas.md
```diff
@@ -342,6 +344,7 @@ This is the root object of the [OpenAPI Description](#openapi-description).
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
 | <a name="oas-version"></a>openapi | `string` | **REQUIRED**. This string MUST be the [version number](#versions) of the OpenAPI Specification that the OpenAPI Document uses. The `openapi` field SHOULD be used by tooling to interpret the OpenAPI Document. This is _not_ related to the API [`info.version`](#info-version) string. |
+| <a name-"oas-self"></a>$self | `URI-reference` (without a fragment) | Sets the URI of this document, which also serves as its base URI in accordance with [RFC 3986 §5.1.1](https://www.rfc-editor.org/rfc/rfc3986#section-5.1.1); the value MUST NOT be the empty string and MUST NOT contain a fragment (even if the fragment is empty).  Implementations MUST support referencing a document by the resolved URI defined by this field. |
```
karenetheridge, line 347: yay explicit mention of no fragment :)

Using a regex to explicitly prohibit `#` in these uri-references in schemas would be good (I'll add that to my list of schema todos).

src/oas.md
```diff
@@ -475,7 +478,7 @@ An object representing a Server.
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
-| <a name="server-url"></a>url | `string` | **REQUIRED**. A URL to the target host. This URL supports Server Variables and MAY be relative, to indicate that the host location is relative to the location where the document containing the Server Object is being served. Variable substitutions will be made when a variable is named in `{`braces`}`. |
+| <a name="server-url"></a>url | `string` | **REQUIRED**. A URL to the target host. This URL supports Server Variables and MAY be relative, to indicate that the host location is relative to the location where the document containing the Server Object is being served. Variable substitutions will be made when a variable is named in `{`braces`}`.   Note that the [OpenAPI Object's](#openapi-object) `$self` field is **NOT** used for relative URL reference resolution. |
```
karenetheridge, line 481: But we do use the document's URI itself -- so should we mention that explicitly here?

src/schemas/validation/schema.yaml
```diff
@@ -8,6 +8,9 @@ properties:
   openapi:
     type: string
     pattern: '^3\.1\.\d+(-.+)?$'
+  $self:
+    type: string
+    format: uri-reference
```
karenetheridge, line 13: we could add `regex: '^[^#]+$'` here, but I think this would be better done as a definition so the other uri-references can also benefit (will todo as a separate item).

karenetheridge (commented):
A few comments/questions throughout.

src/oas.md
karenetheridge: I'm not sure if what you're thinking is the same as what I'm thinking. Can we come up with some examples?

handrews: @karenetheridge yes, see other comments on how I did write examples, they just got lost on another branch.  I will update as soon as the house-moving tasks let up, which will probably be next week sometime.

src/oas.md
lornajane: I'm confused about this addition since in the description of the pull request there's a mention that the `$self` value does NOT affect how relative URLs are resolved. They should still be relative to wherever-you-found-this, as I understood it.

handrews:
@lornajane @ralfhandl @karenetheridge I ended up reworking this a bit as I added examples.

src/oas.md
```diff
@@ -308,21 +308,95 @@ Note that some URI fields are named `url` for historical reasons, but the descri
...
+    Foo:
+      $id: ./schemas/foo
+      properties:
+        bar:
+          $ref: bar
+    Bar:
+      $id: ./schemas/bar
+```
+
+In this example, both Schema Objects use `https://example.com/openapi` as their base URI for resolving their relative `$id` values to `https://example.com/foo` and `https://example.com/bar`.  The `$ref` under `properties` is resolved against the `$id`-provided base URI `https://example.com/foo`, producing `https://example.com/bar`, which is the `$id`-assigned URI of the Bar schema component.
```
ralfhandl, line 364:
```suggestion
In this example, both Schema Objects use `https://example.com/openapi` as their base URI for resolving their relative `$id` values to `https://example.com/schemas/foo` and `https://example.com/schemas/bar`.  The `$ref` under `properties` is resolved against the `$id`-provided base URI `https://example.com/schemas/foo`, producing `https://example.com/schemas/bar`, which is the `$id`-assigned URI of the Bar schema component.
```

ralfhandl (changes requested):
+1, after correcting the resolved URLs in the example.

src/oas.md
handrews: @karenetheridge I'm not sure what you mean, we already say that the Server Object URLs may be relative, and the point of the parenthetical aside is that `$self` URI is irrelevant here, so adding more information about `$self` would just be a distraction and make it feel relevant.

src/schemas/validation/schema.yaml
handrews: Added to this PR.

ralfhandl (dismissed):

ralfhandl (approved):

handrews:
@ralfhandl please don't merge things into my PRs on my behalf.  I prefer rebasing and make a point to do so.  I seem to have forgotten this time, but just ask and I will do it.

ralfhandl:
Fair point, will do.

We seem to use git and GitHub somewhat differently. I for example dislike force-pushes to PRs that I already have reviewed. Which is why I usually only review PRs that are set to „ready for review“ unless explicitly asked to review a draft PR.

handrews:
@ralfhandl I never use force-pushes to change existing commits, just to rebase.  Any further changes are always new commits.

handrews:
@ralfhandl also I'm not sure what "ready to review" has to do with this situation.  The PR was ready for review, then I got feedback and added more commits to address it.

handrews:
@ralfhandl I should explain this:

> sometimes rework that structure which is extremely hard to do when there are merges

I don't do it often on this project, but sometimes if the set of commits is lengthy and complex I will `git rebase-i` and clean it up before merging.  I've also just been working this way for years, including on the OAS, and suddenly having my git usage patterns changed on me without warning is disconcerting.

src/oas.md
karenetheridge: "wherever-you-found-this" _is_ the URI in $self, is it not?

handrews: @lornajane I think there might be some terminology muddiness here- URIs (meaning OpenAPI Description URIs) and URLs (meaning API URLs) are resolved differently, and have been since OAS v3.1.0.

@karenetheridge no, "wherever-you-found-this" is the retrieval URI per RFC3986 §5.1.3.  `$self` is the URI defined in content per RFC3986 §5.1.1.  Since `$self` is a URI, it may not be a sort of URI that indicates any sort of "where", while the retrieval URI is more-or-less by definition a URL (the "more-or-less" is because there are some URI schemes that do not inherently indicate a location, but have a well-defined process to map to a location that might change over time, e.g. the `doi:` scheme).

src/oas.md
```diff
@@ -308,21 +308,95 @@ Note that some URI fields are named `url` for historical reasons, but the descri
...
+      properties:
+        bar:
+          $ref: bar
+    Bar:
+      $id: ./schemas/bar
+```
+
+In this example, both Schema Objects use `https://example.com/openapi` as their base URI for resolving their relative `$id` values to `https://example.com/schemas/foo` and `https://example.com/schemas/bar`.  The `$ref` under `properties` is resolved against the `$id`-provided base URI `https://example.com/schemas/foo`, producing `https://example.com/schemas/bar`, which is the `$id`-assigned URI of the Bar schema component.
+
+Note that using embedded `$id` keywords prevents using `$ref: "#/components/schemas/Bar"` in the `properties` field's `$ref` because the base URI for such fragments is set by the `$id`.  Therefore, a `$ref: "#/components/schemas/Bar"` would resolve to `"https://example.com/schemas/foo#/components/schemas/Bar"`, which is not useful.
```
karenetheridge, line 366:
```suggestion
Note that using embedded `$id` keywords would prevent using `$ref: "#/components/schemas/Bar"` in the `properties` field's `$ref` because the base URI for such fragments is set by the `$id`.  Therefore, a `$ref: "#/components/schemas/Bar"` would resolve to `"https://example.com/schemas/foo#/components/schemas/Bar"`, which is not the presumably-intended location.
```

src/oas.md
```diff
@@ -308,21 +308,95 @@ Note that some URI fields are named `url` for historical reasons, but the descri
...
+info:
+  version: 1.0
+components:
+  pathItems:
+    Foo:
+      $ref: "./shared#/components/pathItems/Foo"
+```
+
+1. The base URI used to resolve `$self` is the retrieval URI, resulting in an effective `$self` value of `https://example.com/openapi` (OADs that use HTTP content negotiation to provide both JSON and YAML representations typically do not use a file extension in their self-assigned URI, to avoid needing to change URIs based on the format).
+2. The base URI for the `$ref` comes from `$self` (`https://example.com/openapi`), producing a resolved URI of `https://example.com/shared#/components/pathItems/Foo`
```
karenetheridge, line 342:
```suggestion
2. The base URI for the `$ref` comes from `$self` (`https://example.com/openapi`), producing a resolved URI of `https://example.com/shared#/components/pathItems/Foo`.
```

src/oas.md
karenetheridge: Okay. I originally actually overlooked the "$self is not used here" part entirely as it wasn't what I was expecting.

That means that the referring URL of the document (if there is one) holds relevance to more than just $self - other things are resolved against it as well.

I'll take up the rest of this in your new examples.

src/oas.md
```diff
@@ -308,21 +308,95 @@ Note that some URI fields are named `url` for historical reasons, but the descri
...
+info:
+  version: 1.0
+servers:
+- url: .
+  description: The production API on this device
+- url: ./test
+  description: The test API on this device
+```
+
+For API URLs, the `$self` field, which identifies the OpenAPI Document, is ignored, and the retrieval URL is used instead.  This produces a normalized production URL of `https://device1.example.com`, and a normalized test URL of `https://device1.example.com/test`.
```
karenetheridge, line 399: Thank you for this paragraph -- this actually provides a bit of clarity regarding how server urls are handled, which I had been meaning to raise in an issue as I was running into conflicts when implementing them as part of request matching.

src/oas.md
karenetheridge: > This indicates that we use the document's URL, but not its $self-defined URI. I'm uncertain if a.) this is the right approach, or b.) it is explained well.

(b) was indeed the case above. Your examples helped, as per my new comments now.

src/oas.md
```diff
@@ -342,6 +416,7 @@ This is the root object of the [OpenAPI Description](#openapi-description).
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
 | <a name="oas-version"></a>openapi | `string` | **REQUIRED**. This string MUST be the [version number](#versions) of the OpenAPI Specification that the OpenAPI Document uses. The `openapi` field SHOULD be used by tooling to interpret the OpenAPI Document. This is _not_ related to the API [`info.version`](#info-version) string. |
+| <a name-"oas-self"></a>$self | `URI-reference` (without a fragment) | Sets the URI of this document, which also serves as its base URI in accordance with [RFC 3986 §5.1.1](https://www.rfc-editor.org/rfc/rfc3986#section-5.1.1); the value MUST NOT be the empty string and MUST NOT contain a fragment (even if the fragment is empty).  Implementations MUST support referencing a document by the resolved URI defined by this field. |
```
karenetheridge, line 419: I'm not sure of the exact wording, but can we somehow make it clear that "referencing" here means "via a $ref keyword", not "Locatable through an HTTP request at this location"?

src/oas.md
```diff
@@ -308,21 +308,95 @@ Note that some URI fields are named `url` for historical reasons, but the descri
...
+RFC3986 Section 5.1.1 requires determining the base URI from within a resource's contents, which for the OAS means the `$self` field of the [OpenAPI Object](#openapi-object) for an [OpenAPI Document](#openapi-document), or the `$id` JSON Schema keyword in [Schema Objects](#schema-object).
+Within an OpenAPI Document, a Schema Object that does not have its base URI set by `$id` uses `$self` the same as any other Object, treating the OpenAPI Document as the "encapsulating entity" in accordance with RFC3986 Section 5.1.2.
+See [JSON Schema draft 2020-12 Section 8.2](https://www.ietf.org/archive/id/draft-bhutton-json-schema-01.html#section-8.2) for more information about base URIs in Schema Objects.
+
+The most common base URI source in the absence of `$self` or `$id` is the retrieval URI, in accordance with RFC3986 Section 5.1.3.
+Since not all tools support direct retrieval, and because direct retrieval is sometimes prevented by network conditions or security policies, tools SHOULD allow users to provide the intended retrieval URI along with the document.
```
karenetheridge, line 319: Do we need to say here that if such a retrieval URI is provided by the user, that the document MUST actually be reachable at that location (i.e. making this a URL not just a URI)?

However, that would mean that $refs within the document are only actually reachable at their resolved URI if and only if the (resolved) $self URI is reachable.

karenetheridge (commented):
I think that's all my comments for this pass.  They will read a bit confusingly though because the order in which they appear in the document is nearly the reverse of that in which I wrote them -- as I had to get down to your examples before I really understood what you're going for, and then I had to go back and drop in some questions higher up where the concepts are first mentioned! (sorry!)

handrews:
Thanks, @karenetheridge , I have tried to address your remaining comments in the most recent commit!  (the force-push prior to that is just a rebase, with identical commits and no conflicts).

ralfhandl:
@handrews This is how I usually do a follow-up review of changes after my last review:

image

And this is what I see here:

image

Which is why I dislike force-pushes to PR branches, and favor squash-merges of PRs. I usually stop reviewing a PR when I see this.

lornajane:
Could we split the git workflow discussion out to another location, please? It's making it hard to follow the content of the change.

src/oas.md
karenetheridge: > the $self value does NOT affect how relative URLs are resolved. They should still be relative to wherever-you-found-this, as I understood it.

As per the added examples (thank you!):
- the retrieval URI is the "wherever you found it" URI (or some other URI that is provided to the implementation out-of-band): e.g. "here is an OAS document that you shall load, and I shall call it Squishy"
- relative `$self` URIs are also resolved using the retrieval URI
- relative `$ref`s within the document (OAS sections and JSON Schemas) and relative `$id`s within schemas within the document are resolved using the closest identifier (which is either an `$id` in a schema, or the (resolved form of the) document's `$self` URI
- server urls are resolved using the retrieval URI and NOT the `$self` URI.
- when matching HTTP requests against entries in /paths, we find the closest `servers` `url` (operation, path-item, global), and prepend the resolved form of it to the path template (e.g. for path-item `/paths/foo/{foo_id}`, the template is `/foo/{foo_id}`.

src/oas.md
karenetheridge: Addressing this will probably want to change a few other locations in the document too (probably in the URI-vs-URL section, and when we expect to find something on the network vs just access something in memory that is known to have a particular URI), and that can be done in a review pass before 3.2.

karenetheridge (approved):
(modulo Ralf's formatting edits) I am happy to sign off on this!

src/oas.md
handrews: @karenetheridge If you substitute "externally determined URI per RFC3986 §5.1.2–5.1.4" in place of "retrieval URI" then this is correct.  It just happens that the retrieval URI (RFC3986 §5.1.3) is by far the most likely of these, and the one worth discussing.  But you'll notice I mention the other possibilities as well.

For example, if you are sent an `application/openapi+yaml` document as one part in a `multipart/related` archive per [RFC2557](https://www.rfc-editor.org/rfc/rfc2557.html), then the relevant URI would _not_ be the retrieval URI (which probably doesn't meaningfully exist) but _would_ be the `Content-Location` header of the part containing the `application/openapi+yaml` document, per RFC3986 §5.1.2 (base URI from encapsulating entity).

But going through the details of all of this is not desirable because a.) it's covered by RFC3986 which we normatively cite, and b.) most people will never care.

src/oas.md
handrews: @karenetheridge I believe my latest commit addressed this here in this fixed field row.

src/oas.md
```diff
@@ -308,21 +308,101 @@ Note that some URI fields are named `url` for historical reasons, but the descri
...
+    Foo:
+      $id: ./schemas/foo
+      properties:
+        bar:
+          $ref: bar
+    Bar:
+      $id: ./schemas/bar
+```
+
+In this example, both Schema Objects use `https://example.com/openapi` as their base URI for resolving their relative `$id` values to `https://example.com/schemas/foo` and `https://example.com/schemas/bar`. The `$ref` under `properties` is resolved against the `$id`-provided base URI `https://example.com/schemas/foo`, producing `https://example.com/schemas/bar`, which is the `$id`-assigned URI of the Bar schema component.
```
ralfhandl, line 370:
```suggestion
In this example, both Schema Objects use `https://example.com/openapi` as their base URI for resolving their relative `$id` values to `https://example.com/schemas/foo` and `https://example.com/schemas/bar`. The `$ref` under `properties` is resolved against the `$id`-provided base URI `https://example.com/schemas/foo`, producing `https://example.com/schemas/bar`, which is the `$id`-assigned URI of the `Bar` schema component.
```

src/oas.md
karenetheridge: which commit is that? the latest commits all seem to be stylistic changes only.

handrews: @karenetheridge [this commit, on this line](https://github.com/OAI/OpenAPI-Specification/pull/4389/commits/67012b50e92c98759ea1943879851c2526d0af50#diff-a662b9afa95a60c2a67cc48014a8d5e00fa6ab2bd5ba2a54531674a5dd398c73R425).  It does not specifically talk about individual fields because it applies to all API Description URIs.  But the section to which it links talks about which fields are API Description URIs.

ralfhandl (dismissed):

lornajane:
General consensus from the technical meeting today is that this pull request is close, if anyone can help with final reviews, we'd appreciate it!

src/oas.md
```diff
@@ -308,21 +308,101 @@ Note that some URI fields are named `url` for historical reasons, but the descri
...
+    Foo:
+      $ref: "./shared#/components/pathItems/Foo"
+```
+
+1. The base URI used to resolve `$self` is the retrieval URI, resulting in an effective `$self` value of `https://example.com/openapi` (OADs that use HTTP content negotiation to provide both JSON and YAML representations typically do not use a file extension in their self-assigned URI, to avoid needing to change URIs based on the format).
+2. The base URI for the `$ref` comes from `$self` (`https://example.com/openapi`), producing a resolved URI of `https://example.com/shared#/components/pathItems/Foo`.
+
+Relative `$self` values are often used for APIs deployed in multiple locations, such as a device management API that is hosted on each device.
+
+In the next example, the retrieval URI is irrelevant, because `$self` is already a full URI:
```
notEthan, line 352: /full URI/absolute URI/ ?

handrews: @notEthan it is an absolute URI which is a subset of full URIs, but the important part is that it is full and therefore does not need resolving.  The absolute-ness is incidental in this particular case, and I wanted to emphasize the part that is most relevant.

src/oas.md
```diff
@@ -308,21 +308,101 @@ Note that some URI fields are named `url` for historical reasons, but the descri
...
+info:
+  version: 1.0
+servers:
+- url: .
+  description: The production API on this device
+- url: ./test
+  description: The test API on this device
+```
+
+For API URLs, the `$self` field, which identifies the OpenAPI Document, is ignored, and the retrieval URL is used instead. This produces a normalized production URL of `https://device1.example.com`, and a normalized test URL of `https://device1.example.com/test`.
```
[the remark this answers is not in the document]
handrews: API description URIs and API endpoint (or other) URLs are just different things.  `$self` makes sense for API description URIs.  I honestly don't know what makes the most sense for API endpoint URLs, but I _do_ know that tying them to `$self` is problematic because an API description has a single most-relevant URI (`$self` if it is present, then RFC3986 §5.1.2-5.1.4 in order if not), while _many_ API instances, each deployed at a different location, can exist while being described by the same OAD.

So it does not make sense to combine something that has a singular identity with something that can have arbitrarily many instances.  This PR does not change anything about how servers are handled, so if you want to see something different with servers, that should be a separate discussion.

src/oas.md
```diff
@@ -308,21 +308,101 @@ Note that some URI fields are named `url` for historical reasons, but the descri
...
 
-Relative URI references in other Objects, and in Schema Objects where no parent schema contains an `$id`, MUST be resolved using the referring document's base URI, which is determined in accordance with [[RFC3986]] [Section 5.1.2 – 5.1.4](https://tools.ietf.org/html/rfc3986#section-5.1.2).
-In practice, this is usually the retrieval URI of the document, which MAY be determined based on either its current actual location or a user-supplied expected location.
+Relative URI references are resolved using the appropriate base URI, which MUST be determined in accordance with [[RFC3986]] [Section 5.1.1 – 5.1.4](https://tools.ietf.org/html/rfc3986#section-5.1.1).
+RFC3986 Section 5.1.1 requires determining the base URI from within a resource's contents, which for the OAS means the `$self` field of the [OpenAPI Object](#openapi-object) for an [OpenAPI Document](#openapi-document), or the `$id` JSON Schema keyword in [Schema Objects](#schema-object).
+Within an OpenAPI Document, a Schema Object that does not have its base URI set by `$id` uses `$self` the same as any other Object, treating the OpenAPI Document as the "encapsulating entity" in accordance with RFC3986 Section 5.1.2.
```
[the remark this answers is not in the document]
handrews: @notEthan I see your point here, thanks for the suggestion.  Let me think a bit more on this wording.

src/oas.md
```diff
@@ -308,21 +308,101 @@ Note that some URI fields are named `url` for historical reasons, but the descri
...
+
+##### Examples of Base URI Determination and Reference Resolution
+
+Given a retrieval URI of `https://example.com/foo/bar/openapi.yaml`, and the following OpenAPI Document:
+
+```YAML
+openapi: 3.2.0
+$self: /openapi
+info:
+  version: 1.0
```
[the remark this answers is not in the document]
handrews: UGH.  Why is the Info Object like it is?

ralfhandl (approved):

handrews:
The latest force-push is just a rebase that resolves conflicts with recent merged PRs.

src/oas.md
notEthan: I thought I was suggesting a more precise term, but it sounds like 'full URI' is a precise term, but one I am not familiar with. I am looking but not finding quite what a full URI is or how it relates to absolute, could you point me to that information?

src/oas.md
```diff
@@ -308,21 +308,101 @@ Note that some URI fields are named `url` for historical reasons, but the descri
...
+
+Note that these rules mean that different fields in the document, particularly those in Schema Objects vs other kinds of Objects, can have different base URIs.
+Relative URI references in `$self` or `$id` are resolved against the base URI appropriate for their location in the document.
+
+Implementations SHOULD allow users to provide the intended retrieval URI along with each document, as this is necessary in situations including:
+
+* Implementations that do not support direct retrieval, which requires additional dependencies and security considerations
+* Network configurations or conditions that prevent direct retrieval
+* Test configurations that need to simulate the document being hosted in a production location
+* Documents that exist only in-memory and have no readily identifiable location (although for a single document without references to other documents, an application-specific default base URI in accordance with RFC3986 Section 5.1.4 would also be suitable)
```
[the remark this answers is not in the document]
notEthan: I don't mean to suggest it be removed, but that the listed cases kind of eclipse what seems to me the broader, more general point that an implementation always needs to know a base URI if $self/other URIs need to be resolved.
I'm good if this is considered fine as is, not trying to push further for a change, just to clarify my meaning.

src/oas.md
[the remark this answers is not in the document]
notEthan: I certainly agree/understand that the URIs of the API description and objects within are quite different than (or unrelated to) API server/endpoint URLs. My hesitation is on the idea that the retrieval URI for the OAD is likely to be more closely related to the server URL than $self (I've generally seen the OAD served from one place accompanying documentation, not with each deployment, and aren't server variables the better mechanism for deployment-specific URLs?).

Or rather, my hesitation as an implementer of tooling is on whether that difference is worth implementing - I have one base URI for objects in the OAD (including `servers`), which will be changing to inherit from $self, and no place to put another base URI. It's not _difficult_ to implement but it adds some complexity in a place I have doubts on the real utility of it.

> This PR does not change anything about how servers are handled, so if you want to see something different with servers, that should be a separate discussion.

Well, it changes the base URI of everything except `servers`, I think discussing that exception belongs here.

Having said that / been heard on the above, I put my skepticism aside and accept this is worth having `servers` be an exception and use the document retrieval URI. In the case of each deployment serving the OAD, although I haven't encountered that, it is more correct.

src/oas.md
handrews: The precise term is "URI", which is not very useful as most people think "URI" means "URI-reference."  There is no more precise term than that.

handrews: And in particular, the OAS uses "URI" to mean "URI-reference" so we're left without proper terminology.

src/oas.md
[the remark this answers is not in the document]
handrews: > I've generally seen the OAD served from one place accompanying documentation, not with each deployment, and aren't server variables the better mechanism for deployment-specific URLs?

There are other use cases, such as APIs for device management that are deployed on each device, often with the OAD because the larger internet might not be accessible from the network containing the device.

I find that a lot of people think in terms of consumer-facing APIs and not enterprise APIs that are more likely to be deployed in complex network configurations.  But complex network configurations are real use cases, too.

src/oas.md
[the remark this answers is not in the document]
handrews: @notEthan if people want to understand base URIs better, they can read RFC3986.  They've got a nice ASCII art precedence diagram :-)

src/oas.md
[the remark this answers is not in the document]
handrews: @notEthan I tried some new wording here (in [this commit](https://github.com/OAI/OpenAPI-Specification/pull/4389/commits/f9f90bde4727bc4fb7fc6812f32670d34dc762a9))

src/oas.md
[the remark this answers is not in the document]
handrews: @notEthan perhaps more seriously, so much of the OAS needs a full rewrite for flow and clarity.  It's extremely hard to do little changes here and there and get the big-picture flow you want, but we don't have the time/budget to do that in 3.x (we discussed it for data modeling in a recent TDC call and opted against it).  This sort of thing should be presented more clearly in Moonwalk, but I just don't have the time to do the level of rewriting that would be needed to truly unravel this.

src/oas.md
handrews: @notEthan I have kept thinking about this.  I've long been frustrated at the lack of anything more precise than "URI" for what we want.  It plagued us with the JSON Schema spec as well, and that's where "full URI" came from.  Although at least with JSON Schema we used "URI-reference" where appropriate.  OAS further overloading "URI" makes this harder.

idk, maybe absolute URI would be better as it is also accurate and not overloaded?  Meh.

src/oas.md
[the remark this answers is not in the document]
handrews: @notEthan BTW I would be perfectly happy to continue the discussion of how to handle server URLs for a potential follow-on change.  I just opted to do the least amount of change there, in part because it's not an area in which I feel exceptionally confident.  I do want to support enterprise multi-deployment use cases, but I don't want to do that at the expense of tons of commentary.

Anyway, I just wanted to encourage further discussion- I know I pushed back on several of your comments here a bit hard, but that's more about me being distracted by other Life Stuff™ than anything else, and I do apologize for not having the bandwidth to reply with more consideration.

src/oas.md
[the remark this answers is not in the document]
handrews: @notEthan I started trying to rework things for better flow, and (as with the data modeling) it would be a couple of weeks of work to do it properly.  So I think the best thing to do here is make sure that this wording is clear enough, and we can keep working at it in the future.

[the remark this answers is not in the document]
karenetheridge: Perhaps we can simplify this by simply strongly recommending that `$self` is an absolute URI (i.e. fully resolved; resolving it against another base URI is a no-op), and point to the RFC to explain why.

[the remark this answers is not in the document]
karenetheridge: We can also offer caution that if server urls are not absolute, then the retrieval URI will be used to resolve them, which might not be desirable because if the scheme+host of the retrieval URI may not actually be the same scheme+host as where the APIs are being hosted, the result will be that any matching of API HTTP requests against path-items in the OpenAPI document will fail.  

This caveat is not new with these changes, but a user might think that an absolute `$self` is sufficient to ward off this problem, which it isn't.

(I've waited until I had fully implemented `$self` and proper server url matching myself until I said this, to be sure of what I'm saying, and written test cases to demonstrate the issue and potentially be able to detect it sufficiently that it can be communicated to the user as a warning.)

src/oas.md
karenetheridge: Is that the sound of a glossary being born? :D

src/oas.md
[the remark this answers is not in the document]
handrews: @karenetheridge there is no reason to prevent people from making `$self` relative to a document location (or a `Content-Location` multipart header, or whatever).  It should behave just like a document-root-level `$id`.  That's the analogy, and we shouldn't deviate from it.

> We can also offer caution that if server urls are not absolute, then the retrieval URI will be used to resolve them

We already say this.

> because if the scheme+host of the retrieval URI may not actually be the same scheme+host as where the APIs are being hosted, the result will be that any matching of API HTTP requests against path-items in the OpenAPI document will fail.

as you note...

> This caveat is not new with these changes

So this idea needs to go in a separate issue and PR and not further complicate this one.

handrews:
@notEthan and others, after talking with @karenetheridge and the rest of the TDC call today, I am going to split out all of the things about further clarification of Server Object behavior into another discussion or issue, and mark things resolved or (if that is not available) off-topic.  This does *NOT* mean that the comment was off-topic in a bad way, just that there was a request on the TDC call to make this PR easier to read and focus on the remaining issues needed to get this across the finish line.

src/oas.md
handrews: @notEthan I [reworked it a bit](https://github.com/OAI/OpenAPI-Specification/pull/4389/commits/34fb6d61c1b40688fed84b90a51c36387b3c96af) to contrast "fully resolved URI" with "relative URI-reference", see if that works for you?

ralfhandl (approved):

src/oas.md
notEthan: I like that, I think it reads significantly more clearly.

src/oas.md
```diff
@@ -308,21 +308,104 @@ Note that some URI fields are named `url` for historical reasons, but the descri
...
+components:
+  pathItems:
+    Foo:
+      $ref: "./shared#/components/pathItems/Foo"
+```
+
+1. The base URI used to resolve `$self` is the retrieval URI, resulting in an effective `$self` value of `https://example.com/openapi` (OADs that use HTTP content negotiation to provide both JSON and YAML representations typically do not use a file extension in their self-assigned URI, to avoid needing to change URIs based on the format).
+2. The base URI for the `$ref` comes from `$self` (`https://example.com/openapi`), producing a resolved URI of `https://example.com/shared#/components/pathItems/Foo`.
+
+Relative `$self` values are often used for APIs deployed in multiple locations, such as a device management API that is hosted on each device.
```
mikekistler, line 351: I'm going to reveal my ignorance here, but this seems like the exact situation where you want to use absolute $self value. I thought the idea of distinguishing URIs -- identifiers -- from URLs -- locators -- is that you could have the same document in multiple locations and would want to make clear that it is the same document -- with a single identifier -- even though it could be retrieved from multiple locations.

src/oas.md
```diff
@@ -308,21 +308,104 @@ Note that some URI fields are named `url` for historical reasons, but the descri
...
+  title: Example API
+  version: 1.0
+servers:
+- url: .
+  description: The production API on this device
+- url: ./test
+  description: The test API on this device
+```
+
+For API URLs, the `$self` field, which identifies the OpenAPI Document, is ignored, and the retrieval URL is used instead. This produces a normalized production URL of `https://device1.example.com`, and a normalized test URL of `https://device1.example.com/test`.
```
mikekistler, line 408: This may be a nit but here we use "retrieval URL" but above (line 323) we use "retrieval URI". I'm guessing that these are referring to the same thing, and if so I think we should try to use consistent terminology.

src/oas.md
```diff
@@ -342,6 +425,7 @@ This is the root object of the [OpenAPI Description](#openapi-description).
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
 | <a name="oas-version"></a>openapi | `string` | **REQUIRED**. This string MUST be the [version number](#versions) of the OpenAPI Specification that the OpenAPI Document uses. The `openapi` field SHOULD be used by tooling to interpret the OpenAPI Document. This is _not_ related to the API [`info.version`](#info-version) string. |
+| <a name="oas-self"></a>$self | `string` | This string MUST be in the form of a URI-reference, MUST NOT be an empty string, and MUST NOT contain a fragment (empty or otherwise). The `$self` field provides the self-assigned URI of this document, which also serves as its base URI in accordance with [[RFC3986]] [Section 5.1.1](https://www.rfc-editor.org/rfc/rfc3986#section-5.1.1). Implementations MUST support identifying the targets of [API description URIs](#relative-references-in-api-description-uris) using the resolved URI defined by this field, as shown under [Examples of Base URI Determination and Reference Resolution](#examples-of-base-uri-determination-and-reference-resolution). |
```
mikekistler, line 428: I think it would be good to explain -- or point to the explanation of -- the expected behavior when this field is not present.

mikekistler (dismissed):
Looks good. 👍 

I left a few minor comments I hope can be addressed in this PR, but do no object to deferring these to a follow-on PR.

src/oas.md
handrews: @mikekistler I just did `$self` by analogy with JSON Schema's `$id`.  If folks want to deviate from that (and as you note there are reasons to do so for simplicity) then we can, but let's decide that intentionally.  I think @karenetheridge brought up this in some way as well?  I admit I've somewhat lost track at this point.

I do not have a use case where you would want the identity to change based on the location, although I'm sure I could construct one where it's _somehow_ beneficial.  But maybe not so beneficial as to be worth discarding the simplicity of requiring an absolute URI.

Notably, the newly proposed JSON Schema replacement [JSON Structure](https://json-structure.github.io/core/draft-vasters-json-structure-core.html) requires an absolute `$id` (although they also say it SHOULD be a URL which I need to reply to because that is missing the point of `$id` entirely in my view and guiding people to the wrong behavior).

src/oas.md
handrews: @mikekistler if that's covered under "Examples of Base URI Detrermination.." is that sufficient?

handrews:
I have produce a substantially different approach to this in PR #4556.

PR #4400, opened 2025-03-01
main: copy schema-publish script & workflow from dev

ralfhandl:
New schema iterations are published directly from the corresponding `vX.Y-dev` branches. 

Copying the updated scripts to `main` to prevent accidental loss.


ralfhandl:
Eventually `dev` flows to `v3.1-dev` from which `v3.1.2-rel` branches off and flows back into `main`.

This PR just speeds up that flow and prevents script changes on `main` to interfere with what needs to be on the dev branches.

For the future I propose that scripts are always changed on `main` and then flow on so that we do not disturb our own circles.

It also helps my tests of how to produce a release branch if there are fewer "noise" differences.

Issue #3432, opened 2023-11-02
Add automation to prevent PRs from changing specification files that need to be immutable.


darrelmiller:
It is very hard for people to know which specification file to update when making changes.  If people do mistakenly try and change the wrong file we should prevent that and provide guidance as to which is the correct file to edit.

kevinswiber:
I can help with this.  Which file paths would we like to restrict?  Would we want an override mechanism?

Bellangelo:
This issue needs some extra clarifications. For example, what files should we forbid? Should we forbid the files that we auto-generate, like the produced JSON spec files? Also, if we need to forbid extra files, do we need to forbid them based on the branch that you are about to merge? For example, you can update 3.1.0 spec only when you are about to merge your PR in the `3.1.0-dev` branch.

PR #4402, opened 2025-03-02
main: status check for changes to `versions` and `schemas` folder

ralfhandl:
Part of #3432

This PR adds a workflow that checks for changes to files in the `schemas` and `versions` folder.

This workflow SHOULD NOT become a required check because
* we still need to add future spec versions to the `versions` folder
* we want to propagate new spec versions from `main` to `dev` and `vX.Y-dev` etc.

So this will be a "red flag" that can be disregarded, which is better than no warning, and may help to prevent accidents such as #4328 targeting the wrong file.

### DIFF
--- a/.github/workflows/check-restricted-files.yaml
+++ b/.github/workflows/check-restricted-files.yaml
@@ -1,6 +1,6 @@
 name: check-restricted-files
 
-# Autor: @ralfhandl
+# Author: @ralfhandl
 # Issue: https://github.com/OAI/OpenAPI-Specification/issues/3432
 
 # This workflow checks for changes of restricted files in a pull request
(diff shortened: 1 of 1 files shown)


.github/workflows/check-restricted-files.yaml
```diff
@@ -0,0 +1,30 @@
+name: check-restricted-files
+
+# Autor: @ralfhandl
```
lornajane, line 3:
```suggestion
# Author: @ralfhandl
```

lornajane:
Mostly because I'm curious: why do it this way rather than with a [path filter](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) that would fire when we changed anything in the matching paths?

ralfhandl:
Thanks, let's try that 😄

ralfhandl:
Closed in favor of
* #4407

PR #4403, opened 2025-03-04
Repair Google Tag Manager

ralfhandl:
Apparently the respec build interferes with how Google Tag Manager (GTM) works. This is an attempt to fix that:
- move GTM script up as far as possible following [Google's recommendation](https://developers.google.com/tag-platform/gtagjs)
- remove GTM stuff added by respec build

Preview of output
* https://ralfhandl.github.io/OpenAPI-Specification/oas/v3.1.1.html


lornajane:
Can we provide sample output so that people can compare with the currently-broken version? I think it would be easier to get feedback from the outreach people if we did.

Issue #1590, opened 2018-05-23
Can a schema parameter's 'required' attribute be overridden?


cmeeren:
I have a common object defined in `schemas` referenced in a couple of endpoints. However, the specific set of required fields differ between the references. For example, one endpoint requires properties `foo` and `bar`, while another endpoint requires properties `foo` and `qux`.

Is it possible to override the `required` list when referencing a schema? Is there another fairly DRY way to accomplish this? (The only workaround I can think of is inlining the schema everywhere, which leads to duplication and possibility of inconsistencies.)

If it's not possible, could this be implemented?

handrews:
Yes, since `foo` is required in both cases, have `"required": ["foo"]` in your re-usable schema under `components`, and then use `allOf` to reference the re-usable schema and also add `"required": ["bar"]` or `"required": ["qux"]` or whatever.

https://tools.ietf.org/html/draft-handrews-json-schema-validation-01#section-6.7.1

cmeeren:
Brilliant, thanks! I'll have to read up some more on the possibilities of `allOf`/`anyOf`.

emvau:
Even if the issue is closed, my question would have the same or similar title, so I decided to append it here.

I got a JSON schema which refers to other JSON schemas, which also refer to other JSON schemas and so on... so it's pretty nested.

=> Now, if I want to override the "required" attribute of one of the nested JSON schemas, would the following structure work, i.e. be valid?
And if not, is there another lean way to express what I want within the OpenAPI definition?

```json
"allOf": [
    {
        "$ref": "https://git.mercedes-benz.com/mavonge/schemas/offer.json"
    },
    {
        "properties": {
            "vehicleOffers": {
                "items": {
                    "properties": {
                        "vehicles": {
                            "items": {
                                "required": [
                                    "options"
                                ]
                            }
                        }
                    }
                }
            }
        }
    }
]
```

handrews:
@emvau the schemas in an `allOf` are each evaluated separately, and must all pass.  There is no "overriding" here, you just add constraints in each branch of the `allOf`.  In practical terms for `required`, this just means that you can't use `allOf` to _remove_ a requirement.

As for nesting, the 2nd schema under `allOf` in your example reads as:

* if the instance is an object, then...
* if it has a `"vehicleOffers"` property which is an array, then...
* if an entry in the array is an object with a `"vehicles"` property which is itself an array, then...
[...]

PR #4423, opened 2025-03-08
Disallow fragments in uri-references that point elsewhere in the OpenAPI description

karenetheridge:
As referenced in https://github.com/OAI/OpenAPI-Specification/pull/4389

### DIFF
--- a/schemas/v3.1/schema.yaml
+++ b/schemas/v3.1/schema.yaml
@@ -11,8 +11,7 @@ properties:
   info:
     $ref: '#/$defs/info'
   jsonSchemaDialect:
-    type: string
-    format: uri-reference
+    $ref: '#/$defs/uri-reference-no-fragment'
     default: 'https://spec.openapis.org/oas/3.1/dialect/WORK-IN-PROGRESS'
   servers:
     type: array
@@ -972,3 +971,9 @@ $defs:
       properties:
         explode:
           default: false
+
+  uri-reference-no-fragment:
+    $comment: a uri-reference used to refer to an entire document, not a position in an OpenAPI description
+    type: string
+    format: uri-reference
+    pattern: '^[^#]*$'


handrews (changes requested):
The definition is fine, but AFAICT the only place where it would be used would be the new `$self` field.  I thought there were more but I can't think of any now.

miqui:
finally got all caught up:  proposal-->pr 4389. @karenetheridge , @rahulsom  are we missing anything?

karenetheridge:
Looking again.. I think I was on autopilot when I wrote this ;)

karenetheridge:
Fixed. This only applies (for now) to jsonSchemaDialect, as a metaschema cannot be embedded inside another document e.g. as a definition.

handrews:
@karenetheridge

> a metaschema cannot be embedded inside another document e.g. as a definition.

Can you provide a citation for this?  [§8.1.1 The "$schema" keyword](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#section-8.1.1) states:

> The "$schema" keyword is both used as a JSON Schema dialect identifier and as the identifier of a resource which is itself a JSON Schema, which describes the set of valid schemas written for this particular dialect.
>
>   The value of this keyword MUST be a URI [[RFC3986](https://datatracker.ietf.org/doc/html/rfc3986)] (containing a scheme) and this URI MUST be normalized.  The current schema MUST be valid against the meta-schema identified by this URI.
>
> If this URI identifies a retrievable resource, that resource SHOULD be of media type "application/schema+json".

It goes on to forbid _using_ `$schema` in a non-resource-root, but I don't see anything forbidding it from _pointing to_ such a schema.

>   The "$schema" keyword SHOULD be used in the document root schema object, and MAY be used in the root schema objects of embedded schema resources.  It MUST NOT appear in non-resource root schema objects. If absent from the document root schema, the resulting behavior is implementation-defined.

(honestly I'm a bit surprised, as I also thought it was forbidden, but I can't find any requirement forbidding it?)

karenetheridge:
Ok, a `$schema` (and `$vocabulary`) keyword can only exist at a schema resource, which is not necessarily the root.

However, `$id` must also be present (which is what makes the location a schema resource). `$id` cannot have a fragment.

Is it legal to refer to a metaschema by a non-canonical identifier (using a json pointer fragment from a higher location)?

I created issue #1590 in the json-schema-spec repo to get some clarity (and we should really tighten this spec language).

handrews:
> Ok, a `$schema` (and `$vocabulary`) keyword can only exist at a schema resource, which is not necessarily the root.

@karenetheridge what you're talking about here is where `$schema` _is used._  This does not impact where it _points_, AFAICT.  I don't see any language that requires the _target_ of `$schema` to be a schema resource.  AFAICT it's legal to say `"$schema": "#/$defs/foo"`.  I can't find language that forbids that (which, again, is a surprise to me).

handrews:
@karenetheridge I guess part of this is that we typically assume that any metaschema will itself have a `$schema`.  Which of course would be the proper way to do it.  And therefore it would have a `$id` (although I have no answer for you about the non-canonical URI usage).  But I also don't see any clear, enforceable requirement that a metaschema use `$schema`.

karenetheridge:
> what you're talking about here is where $schema is used. This does not impact where it points, AFAICT.

Where it's used has relevance as to what its address is (whether canonical or non-canonical).  If it can only be used at a document root, then it would never have a json pointer fragment.

I did find a place in the spec that says that `$vocabulary` MUST only be at a document root, but not all metaschemas need use `$vocabulary`.

handrews:
@karenetheridge I'm still confused.  Yes, `$schema` MUST be in a resource root (document root or adacent to `$id`), but this is not relevant to the _value_ of `$schema`.  I cannot find a requirement that forbids this:

```YAML
$schema: "#/$defs/metaschema"
$defs:
   $ref: "https://json-schema/draft/2020-12"
   unevaluatedProperties: false
```

Now, this is not that useful since `$vocabulary` is not transitive (and can only appear in a resource root, and so cannot be added here and preserve the scenario), so technically this would not recognize any non-core keywords.  But I can't find anything that says it's _invalid_.  It's just not very useful.

karenetheridge:
@handrews I don't think that last example can ever work because the metaschema is contained inside the schema -- and in order to parse the entire document we need to know what the metaschema is first, which hasn't been parsed yet.

handrews:
@karenetheridge but what is the requirement that says so?  I just don't think we can forbid a syntax without a requirement that forbids it.  There are a lot of things in JSON Schema (and the OAS) that don't "work" but aren't forbidden.

handrews:
> and in order to parse the entire document we need to know what the metaschema is first, which hasn't been parsed yet.

@karenetheridge Actually that is not entirely true.  The example only relies on the core vocabulary, which MUST be assumed in order to bootstrap anything.  If you take out the `unevaluatedProperties` keyword, the example would _only_ use the core vocabulary (and be even less useful, but "useful" is not the relevant metric here).

karenetheridge:
I'm just saying that using an example that cannot mathematically work is not a useful example :)

handrews:
@karenetheridge OK in that case `jsonSchemaDialect` needs to be changed back to `format: uri-reference` because these are the vaidation schemas and can only fail on outright invalid OADs (not just ones that can't be useful).

karenetheridge:
There's nothing to do here; the definition can be added to the `$self` addition if it is relevant there.

baywet:
@handrews to pick up the relevant pieces from this and follow up with another PR.

PR #4453, opened 2025-03-18
main: update branch diagram in CONTRIBUTING.md

ralfhandl:
Adjust branch diagram to refined release process
- we can't merge `vX.Y-dev` back into `dev`: this would mess up schemas and schema tests because
  - the `openapi` field requires pattern X.Y.Z
  - added fixed fields are reflected in schemas and schema tests and can't go into `vX.(Y-1)-dev` etc.
- add sync from `main` to `dev` and `dev` to `vX.Y-dev`


handrews (changes requested):
Thanks for keeping this up to date!

I've added a suggestion to clarify what is going on with the new lines.  If accepted, this would need to be done for the other change groups in the diagram as well.

handrews (approved):

PR #4455, opened 2025-03-18
dev: schema tests output reason if "pass" test fixture fails

ralfhandl:
This is a quality-of-life improvement for writing "pass" test fixtures: output validation errors if a "pass" test fixture fails validation. 


duncanbeevers (approved):
This is a fine change, but I think the PR would be easier to review if you showed exactly what a failure looked like before, and what it looks like with this change applied.

miqui (approved):

PR #4458, opened 2025-03-19
Validate fragments are allowed in path keys

duncanbeevers:
## 🗣️ Discussion

Related to [Slack Discussion](https://open-api.slack.com/archives/C1137F8HF/p1741624538482899) about valid vs lint-happy schemas.

This PR adds a test case asserting that paths entries including a URL fragment are valid, despite the fragment being functionally meaningless.

PRs which enforce rules forbidding such paths would not be compliant with the spec, so this test case makes explicit the answer to that question. 

If this PR is merged, future PRs proposing lint rules forbidding fragments in `paths` keys will fail, and be forced to address the fact that their proposal represents opinion rather than conformance to the specification.

### DIFF
--- a/tests/schema/pass/path-key-fragment.yaml
+++ b/tests/schema/pass/path-key-fragment.yaml
@@ -1,0 +1,7 @@
+openapi: 3.1.0
+info:
+  title: API
+  version: 1.0.0
+paths:
+  /foo#fragment:
+    get: {}


handrews:
@duncanbeevers apologies for not thinking this example through in the slack discussion- I wasn't treating it as a PR-ready proposal.

This one is actually tricky.  The specification for patterned fields here [reads](https://spec.openapis.org/oas/v3.1.1#paths-object):

> A relative path to an individual endpoint. The field name MUST begin with a forward slash (/). The path is appended (no relative URL resolution) to the expanded URL from the [Server Object](https://spec.openapis.org/oas/v3.1.1#server-object)’s url field in order to construct the full URL. [Path templating](https://spec.openapis.org/oas/v3.1.1#path-templating) is allowed. When matching URLs, concrete (non-templated) paths would be matched before their templated counterparts. Templated paths with the same hierarchy but different templated names MUST NOT exist as they are identical. In case of ambiguous matching, it’s up to the tooling to decide which one to use.

This clearly requires the value to be a _path_ as defined by RFC3986, modified by the [path templating syntax](https://spec.openapis.org/oas/v3.1.1#path-templating), which reads in part:

> The value for these path parameters MUST NOT contain any unescaped “generic syntax” characters described by [[RFC3986](https://spec.openapis.org/oas/v3.1.1#bib-rfc3986)] [Section 3](https://datatracker.ietf.org/doc/html/rfc3986#section-3): forward slashes (/), question marks (?), or hashes (#).

Now, that path templating bit apples to the _path parameter values_ and not the path template, but it demonstrates that this is all about paths (as including `?` or `#` would end the path and cause the remaining characters to be interpreted differently.

I would argue that the refernce to paths is sufficiently clear, and the meaning of "path" sufficienly well-defined across a very, very large number of specifications, that entries in the Paths Object are in fact forbidden from including `?` or `#`.  So in this case, I would actually support adding such validation to the schema... except it's even more complex because of the lack of constraints on template variable _names_, so writing a regex that really captures this would be non-trivial.

This is another reason I dislike getting too precise with the schemas.  At some point you really hit diminishing returns.

handrews:
I'm going to close this because I'm confident that this specific example is actually not allowed, and the general point is better suited to a discussion than a PR or issue.

duncanbeevers:
@handrews Thanks for the PR feedback.

I guess there are two discussion to be had…

* Are [path fragments](https://github.com/OAI/OpenAPI-Specification/pull/4458) allowed in `paths` keys? (@handrews Your answer seems pretty definitive here)
* Should we introduce test cases asserting the validity of anti-patterns?

duncanbeevers:
Kinda pushing back on this…

If this test case should fail then maybe it should be put into the `fail` corpus, (for the same reasons discussed above ☝️.

handrews:
Of course it's hard to tell what to open where given the historical state of the repo, but we have actually recently [added guidance on this to CONTRIBUTING.md](https://github.com/OAI/OpenAPI-Specification/blob/main/CONTRIBUTING.md#how-to-contribute)

PR #4472, opened 2025-03-20
Gather data modeling sections into a "Modeling Data" top-level section

handrews:
I went to add streaming JSON support to 3.2 and was struck again at how scattered various bits of guidance on data modeling are.

This mostly just moves sections (big sections in the first commit, then some examples in the second) and then does a tiny bit of renaming, re-indenting, and linking to make it flow.  It is a draft because (assuming we go this route), it is not clear where to draw the line, particularly with examples.

There is significant overlap between the examples moved from the Schema Object and the examples still in the Discriminator Object section.  I also left the XML Object Examples where they were, which makes the Modeling XML section rather perfunctory.

If folks like this general idea, we can work out how to balance what goes where.

Note that I also slightly moved "Rich Text Formatting" as taking the data modeling sections out of the introduction left it sandwiched awkwardly between the Document Structure section and the Resolving URIs/URLs sections, which really ought to be next to each other anyway.  (***Please*** do not discuss re-organizing other sections.  In particular, @karenetheridge and I are already planning to reorganize Document Structure and the Resolving sections, so please wait for that effort to open that topic).

-----

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -1548,6 +1548,8 @@ application/json:
       $ref: '#/components/examples/frog-example'
 ```
 
+For additional techniques and examples, see [Modeling Non-JSON Data](#modeling-non-json-data).
+
 #### Encoding Object
 
 A single encoding definition applied to a single schema property.
@@ -1598,6 +1600,8 @@ See also [Appendix C: Using RFC6570 Implementations](#appendix-c-using-rfc6570-b
 
 Note that the presence of at least one of `style`, `explode`, or `allowReserved` with an explicit value is equivalent to using `schema` with `in: "query"` Parameter Objects.
 The absence of all three of those fields is the equivalent of using `content`, but with the media type specified in `contentType` rather than through a Media Type Object.
+
+For additional techniques and examples, see [Modeling Non-JSON Data](#modeling-non-json-data).
 
 #### Responses Object
 
@@ -2512,6 +2516,8 @@ JSON Schema implementations MAY choose to treat keywords defined by the OpenAPI 
 
 This object MAY be extended with [Specification Extensions](#specification-extensions), though as noted, additional properties MAY omit the `x-` prefix within this object.
 
+For additional techniques and examples, see [Schema Modeling Techniques](#schema-modeling-techniques).
+
 ##### Extended Validation with Annotations
 
 JSON Schema Draft 2020-12 supports [collecting annotations](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#section-7.7.1), including [treating unrecognized keywords as annotations](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#section-6.5).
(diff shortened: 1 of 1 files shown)


src/oas.md
```diff
@@ -4012,257 +3285,996 @@ This object MAY be extended with [Specification Extensions](#specification-exten
...
+
+| `format` | JSON Data Type | Comments |
+| ---- | ---- | ---- |
+| `int32` | number | signed 32 bits |
+| `int64` | number | signed 64 bits (a.k.a long) |
+| `float` | number | |
+| `double` | number | |
+| `password` | string | A hint to obscure the value. |
+
+As noted under [Data Type](#data-types), both `type: number` and `type: integer` are considered to be numbers in the data model.
```
ralfhandl, line 3573:
```suggestion
As noted under [Data Types](#data-types), both `type: number` and `type: integer` are considered to be numbers in the data model.
```

ralfhandl (commented):
++1, with two nits

ralfhandl (dismissed):

src/oas.md
```diff
@@ -4012,257 +3285,996 @@ This object MAY be extended with [Specification Extensions](#specification-exten
...
 ```
 
-#### OAuth Flows Object
-
-Allows configuration of the supported OAuth Flows.
+### Modeling Non-JSON Data
 
-##### Fixed Fields
+This section describes how to model non-JSON formats, either by mapping the format into the
+[JSON Schema data model](https://www.ietf.org/archive/id/draft-bhutton-json-schema-01.html#name-instance-data-model)
```
ralfhandl, line 4020: The preferred URL pattern seems to be
```suggestion
[JSON Schema data model](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-01.html#name-instance-data-model)
```

src/oas.md
handrews: @ralfhandl I wish you had not committed this as it is the original text and I was making an effort *not* to change that text for the most part.  In any event, this draft will not be merged I just wanted feedback on the direction.

src/oas.md
handrews: @ralfhandl I copied it from elsewhere in the text, we can audit it separately.  This PR will not be merged so fixing these things is not relevant here.

ralfhandl (dismissed):

handrews:
@ralfhandl I've done a bit more work and incorporated your changes that were still relevant, including picking a preferred JSON Schema URL which we can change further elsewhere (since this is still just a "get feedback on direction" PR)

ralfhandl:
I like the direction of keeping the Objects sections simple & straightforward and collecting the nitty gritty special cases further down.

lornajane:
Just wanted to add a note after seeing this in the meeting, we discussed whether using some of this content on the learn site might make for more maintainable/navigable content and a shorter/more readable specification.

PR #4480, opened 2025-03-20
Proposal for referencing Tag Objects with URIs.

handrews:
This proposal replaces draft PR #4390 based on today's TDC call.

### DIFF
--- a/proposals/2025-03-20-URIs-for-Tags.md
+++ b/proposals/2025-03-20-URIs-for-Tags.md
@@ -7,8 +7,8 @@
 |---- | ---------------- |
 |Proposal |[2025-03-20-URIs-for-Tags](https://github.com/OAI/OpenAPI-Specification/tree/main/proposals/{2025-03-20-URIs-for-Tags.md})|
 |Authors|[Henry Andrews](https://github.com/handrews)|
-|Review Manager | TBD |
-|Status |Proposal|
+|Review Manager | [Lorna Mitchell](https://github.com/lornajane) |
+|Status | rejected |
 |Implementations | n/a |
 |Issues |[#2905 Allow the use of $ref and json pointer in tags](https://github.com/OAI/OpenAPI-Specification/issues/2905), consolidated into [#3853 Consolidated $ref-to-Some Object feature request](https://github.com/OAI/OpenAPI-Specification/issues/3853)|
 |Previous Revisions |n/a|
@@ -18,6 +18,7 @@
 |Date |Responsible Party |Description |
 |---- | ---------------- | ---------- |
 |2025-03-20 | @handrews | Initial publication |
+|2025-03-26 | @handrews | Document rejection |
 
 ## Introduction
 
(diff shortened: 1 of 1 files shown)


lornajane:
Thanks for putting this together, good structure and good starting point for getting input on the idea!

ralfhandl (dismissed):

lornajane (dismissed):
I am not in favour of these additions for the 3.x branch. I wish that we'd implemented tags differently in the first place, and I'm sure that all the constructive discussion around the tags feature will help us a lot in future major releases.

I believe that the limitations of the current tag situation can be overcome with helper tooling and that this change (while solving a narrow but valid use case) adds complexity to the specification that is unnecessary and does not benefit the majority of users. As custodians of a widely-used standard, we have a responsibility to maintain something that is appropriate for its audience, and we should be "reluctant" in all our changes unless we see that they are really needed.

I propose that users would be equally well served by leaving the requirement to resolve tags from the entry document. Organisations can either maintain an extensive list of tags in all OpenAPI documents, and then remove any that aren't used before publishing ([tooling exists](https://github.com/Redocly/redocly-cli-cookbook/tree/main/custom-plugin-decorators/remove-unused-tags) for this use case), or alternatively if a tool wants to include tags found in the wider context of referenced OpenAPI documents by adding them to the top-level tags array during processing, that would work well too.

The tags array is a list of strings. It isn't an ID like the Operation uses, and it's not a named entry like the security schemes, so it is appropriate to approach the limitations of it differently. My proposal is to offer some advice or documentation on approaching this problem, but not to bring it in scope of the specification for 3.x since other options are available.

(I've marked this as "request changes" so that my review isn't missed/ignored - but please feel free to dismiss it if we reach consensus as a group that this proposal should go forward anyway).

handrews:
@lornajane @ralfhandl I have updated the proposal to document the rejected status so we can point to this in issues where the feature is requested.

ralfhandl (approved):

lornajane:
I'd like to leave this open a week or two longer if that's okay - a proposal isn't a minor thing and I don't want to move so quickly we don't hear input from others if they want to give it.

proposals/2025-03-20-URIs-for-Tags.md
```diff
@@ -0,0 +1,138 @@
...
+This makes clear that they expect tags to be resolved from the _current document_, which may be a [referenced document rather than the entry document](https://spec.openapis.org/oas/v3.1.1#openapi-description).
+
+However, in [3.0.4](https://spec.openapis.org/oas/v3.0.4#resolving-implicit-connections) and [3.1.1](https://spec.openapis.org/oas/v3.1.1#resolving-implicit-connections) we RECOMMEND (== SHOULD) that tag names are resolved to Tag Objects in the _entry document_.
+This means that there is no way to resolve them from the current document, which is the mirror image of the problem as that encountered by @SandroG.
+
+In today's TDC call, @lornajane stated that she expects tag names to be resolved from the entry document, and @kevinswiber expressed doubt that anyone implements anything else (sadly @SandroG does not mention their tool, which presumably resolves from the current document or else they would not have explained the issue in this way).
```
mikekistler, line 85: I'm sorry I missed this discussion in the TDC call. I can easily imagine scenarios where tags should be resolved to the current document. Consider an API that is actually implemented as a collection of microservices that each implement the CRUD operations for one or more resources. The API description might be structured as an entry document that $ref's in groups of path items from separate documents for each microservice. Now suppose we want one tag per resource -- these tags would naturally be defined in the document for a micro service and not the entry document.

mikekistler (approved):
I concur with rejecting this for 3.x. I think the rationale given is sound.

proposals/2025-03-20-URIs-for-Tags.md
handrews: @mikekistler yes, that's exactly the sort of scenario I feel we should support using the same mechanisms we use for everything else.

I'm a little curious why you see this scenario but reject solving it?

handrews:
@lornajane That's fine with me, if somehow things change it's still valuable to document your rationale- under "alternatives" if we end up finding a version of this to accept.

proposals/2025-03-20-URIs-for-Tags.md
handrews: @mikekistler got it.  yeah in 4.0 we need to have just one mechanism.  Or at least one mechanism per use case (e.g. one magic "look in the entry document" mechanism, and one explicit "look exactly where I say" mechanism).  Note that in JSON Schema, `$dynamicRef` is the (more flexible but also more complex) equivalent of the magic entry document mechanism.

karenetheridge:
> [@lornajane said] I propose that users would be equally well served by leaving the requirement to resolve tags from the entry document.

I disagree with this; I think there is value in pulling in definitions from multiple documents. The Motivations section in this document has many compelling sections.

I believe we can solve the problem of sharing tags across documents by revising the section at https://spec.openapis.org/oas/v3.1.1#resolving-implicit-connections....

> [From the proposal]  A new field or fields would be added to identify, with a URI, the OpenAPI Document from which a tag MUST be resolved.

I had a similar thought a few days ago, but as a more universal solution of where to resolve all sorts of things from, not just tags.  I intend to write this up as a proposal (I may have to implement first on my end to think through some of the edge cases).

> [@lornajane said] The tags array is a list of strings. It isn't an ID like the Operation uses, and it's not a named entry like the security schemes

I think it would make sense to restructure /tags as a named entry, like webhooks and security schemes. There is no need for this list to be ordered (if anyone has heard of a tool that prefers the document order rather than alphabetical, that would be useful to know), and if we use a named entry, we still allow for the future possibility of referring to a tag by URI (which would contain the name): e.g. https://example.com/other_api#/tags/foo rather than https://example.com/other_api#/tags/42.  However, that's a moonwalk-scoped change unless we wanted to allow /tags to be polymorphic (allow either array-based or object-based), which I believe others have said would cause major inconveniences in some tooling implementations.

lornajane:
Thanks @karenetheridge and I do agree that these are really excellent ideas and we should definitely get them inked in to the moonwalk plans!

MurdoqZodoq (commented):
go

PR #4494, opened 2025-03-24
main: CONTRIBUTING note on `buid-src` for Windows users

ralfhandl:
Note callouts apparently can't be within bullet lists.

Moving it to the end of the section.

> [!NOTE] 
> Note for VS Code users: the local preview nicely renders Note callouts in bullet lists and allows changing the callout title to something other than "Note". This doesn't work in Github.


lornajane (commented):
I realise I'm nitpicking, but can we put the note at the start of the section, before the user needs the information? Also it looks odd in context where we now have a note, then a heading, then another note right after that.

lornajane (approved):
Thanks!

ralfhandl:
> can we put the note at the start of the section, before the user needs the information?

@lornajane I was wondering that myself, thanks for confirming my doubts!

PR #4495, opened 2025-03-24
main: run `respec` workflow only on changes to relevant files

ralfhandl:
Getting tired of re-reviewing spec publishing PRs on every change to unrelated files on `main`:
- only run the respec workflow if itself, one of its scripts, or one of its inputs change
- filter out "random" changes to the produced HTML
- filter out ReSpec minor & patch version part - it changes more often than we publish specs

### DIFF
--- a/scripts/md2html/build.sh
+++ b/scripts/md2html/build.sh
@@ -58,9 +58,11 @@ for specification in $specifications; do
 
   node scripts/md2html/md2html.js --maintainers $maintainers $specification "$allVersions" > $tempfile
   npx respec --no-sandbox --use-local --src $tempfile --out $tempfile2
-  # remove unwanted Google Tag Manager and Google Analytics scripts
+  # remove unwanted Google Tag Manager and Google Analytics scripts, non-deterministic changes, and ReSpec minor version
   sed -e 's/<script type="text\/javascript" async="" src="https:\/\/www.google-analytics.com\/analytics.js"><\/script>//' \
       -e 's/<script type="text\/javascript" async="" src="https:\/\/www.googletagmanager.com\/gtag\/js?id=G-[^"]*"><\/script>//' \
+      -e 's/ toc-inline//' -e 's/ darkmode//' \
+      -e 's/"ReSpec \([0-9]*\)[.0-9]*"/"ReSpec \1"/' \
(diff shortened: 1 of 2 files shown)


duncanbeevers:
@ralfhandl Would you link to one of these spurious spec publishing PRs?
I would expect unrelated changes to be a no-op, and if they triggered a PR that it would be free of any changes.
Are there non-deterministic outputs from the respec workflow?

ralfhandl:
@duncanbeevers For example #4412.

Opened three weeks ago, still unmerged because it did not get a second approval, refreshed several times due to merges into `main`, almost every refresh dismissed one of my approvals.

Now I'll wait for someone else to approve first and hope that I can approve second and immediately merge before the next refresh dismisses the first review.

ralfhandl:
> Are there non-deterministic outputs from the respec workflow?

Unfortunately there are seemingly random flip-flop changes that have no visible effect that I can spot.

Some seem to be related to the time-of-day on the server and dark mode. 

Another is whether the table of contents is inlined, which should only depend on the width of the browser window, and that width should be deterministic for the head-less puppeteer browser.

ralfhandl:
@duncanbeevers Thanks for asking, it got me thinking: we anyway tweak the respec output to avoid breaking Google Analytics, so why not also filter out the known non-deterministic things.

duncanbeevers:
I see there are changes in here to make the output more stable. I think those changes should be landed first / independently.

scripts/md2html/build.sh
```diff
@@ -58,9 +58,11 @@ for specification in $specifications; do
 
   node scripts/md2html/md2html.js --maintainers $maintainers $specification "$allVersions" > $tempfile
   npx respec --no-sandbox --use-local --src $tempfile --out $tempfile2
```
duncanbeevers, line 60: Consider using the `TZ` env var while running `npx respec` to ensure more-consistent output, regardless of where the workflow / script is run.

ralfhandl: Interesting idea. Did you test whether it has the desired effect for the fluctuating `darkmode`?

ralfhandl:
Closed in favor of
* #4531

PR #4504, opened 2025-03-27
main: add guidance on bullet lists to style guide

### LINKED ISSUE #4502: Style Guide: Periods at the end of bullet lists?
I struggle to write bulleted lists consistently, but when I remember to do so I usually _avoid_ a period at the end of lists where each bullet is one sentence (or phrase or shorter), but _include_ them if each bullet is multiple sentences (more-or-less a paragraph).

We should offer guidance in the style guide.  I've included my personal preference as an example, but don't have strong feelings as to what the rule should be.

This is brought to you by #4389 (specifically [here](https://github.com/OAI/OpenAPI-Specification/pull/4389/files/ff315fa3ae0cd7a58d54731bc4aa9c66c959fcea#diff-a662b9afa95a60c2a67cc48014a8d5e00fa6ab2bd5ba2a54531674a5dd398c73R341)), where I was inconsistent and @karenetheridge suggested adding a period to the bullet without one, but I'd slightly prefer to take it off of the one with a period.

But really I just want a rule so we don't have to think about it.

ralfhandl:
Fixes #4502

Checked the first ten search results on "bullet lists and periods" and came up with the included recommendation.

The checked guidelines agree on the first rule (full sentences) and second rule (words and short phrases).

The third rule (complete a stem sentence) is the majority recommendation. Another mentioned style is to end each completing bullet list item except the last two with a semicolon, end the second to last with a semicolon followed by the word "and" or "or", depending on which kind of logic the list expresses, and end the last bullet list item with a period. I find this style as difficult to read as it is to describe, so I opted against it 😎 

### DIFF
--- a/style-guide.md
+++ b/style-guide.md
@@ -22,6 +22,12 @@ The following additional rules should be followed but currently are not enforced
 10. Use `<span id="thing"></span>` for link anchors. The `<a name="thing"></a>` format has been deprecated.
 11. Headings use [title case](https://en.wikipedia.org/wiki/Title_case) and are followed by a blank line.
 12. Do not use [RFC2119 key words (MUST, MAY, ...)](https://datatracker.ietf.org/doc/html/rfc2119) in "Examples" sections or when explaining examples, and state requirements only in sections that are clearly normative.
+13. Bullet lists and punctuation:
+    - If a bullet list item is a complete sentence or paragraph, start it with an uppercase letter and end it with a period.
+    - If a bullet list item is a word or short phrase, start it with an uppercase letter and do not end it with a punctuation character.
+    - If a bullet list item completes a stem sentence immediately preceding the bullet list, start it with a lowercase letter and end it with a period.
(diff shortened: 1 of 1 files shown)


lornajane (commented):
I suggest that we automate the checks for this and any other aspects of the style guide that we actually care about. Since we don't have editorial staff, it's unlikely that these guidelines will be observed without some automated assistance.

handrews:
Thanks for the research and write-up, @ralfhandl !  That's going above and beyond 🙂 

@lornajane this is still tremendously valuable to me when making changes and dsicussing reviews.  What happens is that one person writes it one way, and another suggests it be changed to a different way.  With this update, we can quickly resolve the discrepancy without having to debate it.

I honestly don't care if these style guide things are ever automated.  I care that they help us stop wasting time discussing them more than once.  As someone who writes a lot of specification text, this stuff is a major factor for me when you add it all up.

handrews (approved):

PR #4505, opened 2025-03-27
3.2: Remove promise of URIs for tags, note alternatives

handrews:
Since #4480 has been rejected, replace the promise of a future URI-based alternative with advice to look outside of the OAS.

@lornajane feel free to propose or submit an alternative, this is my interpretation of your intent from the proposal rejection but may not be accurate.  Also, it is not my intent for this to come across as some sort of passive-aggressive sulking but I realize it might and very much want pushback if that is the case- I really do just want to point people towards the recommende solution.  Which would seem to be Overlays?


ralfhandl (approved):

lornajane (approved):
Thank you!

PR #4514, opened 2025-03-28
feat: adds additional operations for path item object

### LINKED ISSUE #1747: Support all HTTP Methods
# Feature Request
## [RFC 2616](https://tools.ietf.org/html/rfc2616) updated by [RFC 7231](https://tools.ietf.org/html/rfc7231) ([errata list for 7231](https://www.rfc-editor.org/errata_search.php?rfc=7231))

- [X] OPTIONS
- [X] GET
- [X] HEAD
- [X] POST
- [X] PUT
- [X] DELETE
- [X] TRACE
- [ ] CONNECT

## [RFC 2518](https://tools.ietf.org/html/rfc2518) updated by [RFC 4918](https://tools.ietf.org/html/rfc4918) and [RFC 5689 for MKCOL](https://tools.ietf.org/html/rfc5689)

- [ ] PROPFIND
- [ ] PROPPATCH
- [ ] MKCOL
- [ ] COPY
- [ ] MOVE
- [ ] LOCK
- [ ] UNLOCK

## [RFC 3253](https://tools.ietf.org/html/rfc3253) ([errata list for 3253](http://www.webdav.org/deltav/protocol/rfc3253-issues-list.htm))

- [ ] VERSION-CONTROL
- [ ] REPORT
- [ ] CHECKOUT
- [ ] CHECKIN
- [ ] UNCHECKOUT
- [ ] MKWORKSPACE
- [ ] UPDATE
- [ ] LABEL
- [ ] MERGE
- [ ] BASELINE-CONTROL
- [ ] MKACTIVITY

## [RFC 3648](https://tools.ietf.org/html/rfc3648)

- [ ] ORDERPATCH

## [RFC 3744](https://tools.ietf.org/html/rfc3744) ([errata list for 3744](https://greenbytes.de/tech/webdav/draft-reschke-rfc3744bis-issues.html))

- [ ] ACL

## [RFC 5323](https://tools.ietf.org/html/rfc5323)

- [ ] SEARCH

## [draft ietf httpbis safe method w body 03](https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-03.html)

- [ ] QUERY

## [draft snell link method 01](https://tools.ietf.org/id/draft-snell-link-method-01.html)

- [ ] LINK
- [ ] UNLINK

# Related Issues
#658
#1306
#480
https://github.com/swagger-api/swagger-core/issues/1760

# Iana link
Another link with methods and their RFC : [iana](https://www.iana.org/assignments/http-methods/http-methods.xhtml)

### This post will be re-edited soon to add a small description for all methods

Regards.

baywet:
fixes #1747


tests/schema/pass/path-item-object-example.yaml
```diff
@@ -32,4 +32,24 @@ paths:
           type: array
           items:
             type: string
-        style: simple
\ No newline at end of file
+        style: simple
+    additionalOperations:
+      query:
```
ralfhandl, line 37:
```suggestion
      QUERY:
```
We want the original spelling.

src/oas.md
```diff
@@ -924,6 +924,7 @@ The path itself is still exposed to the documentation viewer but they will not k
...
 | <a name="path-item-summary"></a>summary | `string` | An optional string summary, intended to apply to all operations in this path. |
 | <a name="path-item-description"></a>description | `string` | An optional string description, intended to apply to all operations in this path. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation. |
+| <a name="path-item-additional-operations"></a>additionalOperations | Map[`string`, [Operation Object](#operation-object)] | A map of additional operations on this path. This map MUST NOT contain any entry matching (case-insensitive) any operation that can be defined on the parent path item object. |
```
ralfhandl, line 927: I'd place this after the last explicit method name `trace` to make clear what it is "additional" to.

src/oas.md
```diff
@@ -985,7 +986,39 @@ This object MAY be extended with [Specification Extensions](#specification-exten
       },
       "style": "simple"
     }
-  ]
+  ],
+  "additionalOperations": {
+    "query": {
```
ralfhandl, line 991:
```suggestion
    "QUERY": {
```

src/oas.md
```diff
@@ -1019,6 +1052,26 @@ parameters:
       items:
         type: string
     style: simple
+additionalOperations:
+  query:
```
ralfhandl, line 1056:
```suggestion
  QUERY:
```

ralfhandl (commented):
Could you please add a failing test case trying to use "GET" as an additional operation?

src/oas.md
baywet: done. It kind of always bothers me that properties are not in alphabetical order on this spec, but now I know there's some kind of "logical order" going on.

baywet:
> Could you please add a failing test case trying to use "GET" as an additional operation?

done. I added POST instead to make it clear that it's not just a collision with what's defined in the document, but rather a matter of "where things should go".

tests/schema/fail/path-item-object-example.yaml
ralfhandl, line 1: Please let the filename express why this is failing

tests/schema/fail/path-item-object-example.yaml
```diff
@@ -0,0 +1,64 @@
...
+          type: array
+          items:
+            type: string
+        style: simple
+    additionalOperations:
+      POST:
+        description: Returns pets based on ID
+        summary: Find pets by ID
+        operationId: postPetsById
+        request-body:
```
ralfhandl, line 41:
```suggestion
        requestBody:
```
`request-body` is not a fixed field of the Operation Object, which is why this file fails validation.

With correct spelling it passes validation because the schema only checks for "not post".

src/schemas/validation/schema.yaml
```diff
@@ -230,6 +230,52 @@ $defs:
...
+            - properties:
+                patch:
+                  $ref: '#/$defs/operation'
+              required:
+                - patch
+            - properties:
+                trace:
+                  $ref: '#/$defs/operation'
+              required:
+                - trace
```
ralfhandl, line 278: This only catches additional methods in lowercase spelling.

Maybe @handrews or @karenetheridge know how to express this case-insensitively.

ralfhandl (changes requested):

tests/schema/fail/path-item-object-example.yaml
baywet: "le LLM m'a tuer" ;-)

tests/schema/fail/path-item-object-example.yaml
baywet: what do you think about path-item-object-conflicting-additional-operation? or something along those lines?

src/schemas/validation/schema.yaml
notEthan, line 278: Does the `$ref: '#/$defs/operation'` on each `not/anyOf/*/properties/*` serve a purpose?

```yaml
not:
  anyOf:
    - properties:
        get:
          $ref: '#/$defs/operation'
      required:
        - get
```

Reads as: additionalOperations must not have a `get` property with a value that is a `#/$defs/operation` - so it _may_ have a `get` property that is not a valid operation object.
That additionalOperations object would still be invalid because the additionalOperations schema's `additionalProperties: {$ref: '#/$defs/operation'}` will fail.
I think it is simpler and validates the same instances to just have the `required` in each `not/anyOf/*` and drop the `properties`.

ralfhandl, line 278: How about using something like this:
```suggestion
        propertyNames:
          not:
            pattern: '^(get|put|[pP][oO][sS][tT]|delete|options|head|patch|trace)$'
```

Of course adding the lower/Upper pattern for all verbs in addition to `POST`.

(Being lazy I only added it for POST to make the new "fail" testcase actually fail.)

handrews:
@baywet @kevinswiber I did not see a reply in the issue to my idea that we include `query` among the special-cased fields rather than relegating it to `additionalOperations`, since it is on the verge of becoming an IANA-registered general-purpose (non-WebDAV) HTTP method.  I would prefer that approach.

src/schemas/validation/schema.yaml
handrews, line 278: All that this needs is

```YAML
not:
  required:
  - GET
  - PUT
  - POST
  - DELETE
  - OPTIONS
  - HEAD
  - TRACE
  - PATCH
```

(and if my proposal to add `query` to the special-case operation fields is accepted, add `  - QUERY` to that array).

We only want to forbid all-caps variations, regardless of thier value, as those would duplicate the lower-cased special fields.  However, we don't want to forbid additional operations using methods `get`, `Get`, etc. as these are distinct from `GET` which is the special case.

I would be _very_ skeptical of anyone actually wanting to use a differently-cased variation of a standard HTTP method, but it is legal to do.  `GET`, `get`, `Get`, `gEt`, etc. are all distinct HTTP methods per RFC9110, and our special `get` field only produces the `GET` method.  The other capitalizations are properly considered additional operations.

src/oas.md
```diff
@@ -932,6 +932,7 @@ The path itself is still exposed to the documentation viewer but they will not k
 | <a name="path-item-head"></a>head | [Operation Object](#operation-object) | A definition of a HEAD operation on this path. |
 | <a name="path-item-patch"></a>patch | [Operation Object](#operation-object) | A definition of a PATCH operation on this path. |
 | <a name="path-item-trace"></a>trace | [Operation Object](#operation-object) | A definition of a TRACE operation on this path. |
+| <a name="path-item-additional-operations"></a>additionalOperations | Map[`string`, [Operation Object](#operation-object)] | A map of additional operations on this path. This map MUST NOT contain any entry matching (case-insensitive) any operation that can be defined on the parent path item object. |
```
handrews, line 935:
```suggestion
| <a name="path-item-additional-operations"></a>additionalOperations | Map[`string`, [Operation Object](#operation-object)] | A map of additional operations on this path. This map MUST NOT contain any entry for the methods that can be defined by other Operation Object fields (e.g. no `POST` entry, as the Operation Object field `post` is used for this method). |
```

handrews (changes requested):

src/schemas/validation/schema.yaml
notEthan: I think this still needs `anyOf`, as `required` functions like `allOf`. e.g.

```yaml
not:
  anyOf:
    - required: [GET]
    - required: [PUT]
    - required: [POST]
    ...
```

handrews: @notEthan yeah you're right, we want the effect of

```YAML
allOf:
- not:
  required: [GET]
- not:
  required: [PUT]
...
```
so we need either that or the `anyOf` form you wrote.  Whichever folks feel is more readable.

src/oas.md
ralfhandl: We should also call out that the method name is case-sensitive and that clients are expected to send the exact map key as the method.

src/schemas/validation/schema.yaml
ralfhandl: Neither of these variants with `not` and `required` fails the test case with `POST` in `additionalOperations`.

Which is a consolation because I didn't understand why they should work 😎 

My `propertyNames` proposals both detect the forbidden method name `POST`.

ralfhandl: That one works.

I still prefer

~~~yaml
        propertyNames:
          not:
            enum:
              - GET
              - PUT
              - POST
              - DELETE
              - OPTIONS
              - HEAD
              - PATCH
              - TRACE
~~~

which also avoids regular expressions.

tests/schema/fail/path-item-object-example.yaml
baywet: fixed

src/schemas/validation/schema.yaml
baywet: resolving this since we have another thread with a potential solution

baywet: updated to the enum version, with query, push soon

baywet: not really. Thanks for calling this out. Resolving since we settled on the enum avenue to exclude those properties

tests/schema/pass/path-item-object-example.yaml
ralfhandl, line 1: Please add a `query` operation to this example.

ralfhandl (changes requested):

src/oas.md
```diff
@@ -932,6 +932,7 @@ The path itself is still exposed to the documentation viewer but they will not k
 | <a name="path-item-head"></a>head | [Operation Object](#operation-object) | A definition of a HEAD operation on this path. |
 | <a name="path-item-patch"></a>patch | [Operation Object](#operation-object) | A definition of a PATCH operation on this path. |
 | <a name="path-item-trace"></a>trace | [Operation Object](#operation-object) | A definition of a TRACE operation on this path. |
+| <a name="path-item-additional-operations"></a>additionalOperations | Map[`string`, [Operation Object](#operation-object)] | A map of additional operations on this path. This map MUST NOT contain any entry for the methods that can be defined by other Operation Object fields (e.g. no `POST` entry, as the Operation Object field `post` is used for this method). |
```
ralfhandl, line 935:
```suggestion
| <a name="path-item-additional-operations"></a>additionalOperations | Map[`string`, [Operation Object](#operation-object)] | A map of additional operations on this path. The map key is the HTTP method, with the same capitalization that is to be sent in the request. This map MUST NOT contain any entry for the methods that can be defined by other Operation Object fields (e.g. no `POST` entry, as the Operation Object field `post` is used for this method). |
```

ralfhandl: Please add a clarification similar to this.

ralfhandl (dismissed):

src/oas.md
```diff
@@ -932,6 +932,7 @@ The path itself is still exposed to the documentation viewer but they will not k
 | <a name="path-item-head"></a>head | [Operation Object](#operation-object) | A definition of a HEAD operation on this path. |
 | <a name="path-item-patch"></a>patch | [Operation Object](#operation-object) | A definition of a PATCH operation on this path. |
 | <a name="path-item-trace"></a>trace | [Operation Object](#operation-object) | A definition of a TRACE operation on this path. |
+| <a name="path-item-additional-operations"></a>additionalOperations | Map[`string`, [Operation Object](#operation-object)] | A map of additional operations on this path. The map key is the HTTP method with the same capitalization that is to be sent in the request. This map MUST NOT contain any entry for the methods that can be defined by other Operation Object fields (e.g. no `POST` entry, as the Operation Object field `post` is used for this method). |
```
handrews, line 935:
```suggestion
| <a name="path-item-trace"></a>trace | [Operation Object](#operation-object) | A definition of a TRACE operation on this path. |
| <a nmae="path-item-query"></a>query | [Operation Object](#operation-object) | A definition of a QUERY operation, as defined in the most recent IETF draft ([draft-ietf-httpbis-safe-method-w-body-08](https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-08.html) as of this writing) or its RFC successor, on this path. |
| <a name="path-item-additional-operations"></a>additionalOperations | Map[`string`, [Operation Object](#operation-object)] | A map of additional operations on this path. The map key is the HTTP method with the same capitalization that is to be sent in the request. This map MUST NOT contain any entry for the methods that can be defined by other Operation Object fields (e.g. no `POST` entry, as the Operation Object field `post` is used for this method). |
```

handrews (dismissed):
Thanks for catching and fixing my copy-paste error!

src/schemas/validation/schema.yaml
```diff
@@ -230,6 +230,24 @@ $defs:
         type: array
         items:
           $ref: '#/$defs/parameter-or-reference'
+      additionalOperations:
+        type: object
+        additionalProperties:
+          $ref: '#/$defs/operation'
+        propertyNames:
+          $comment: RFC9110 restricts methods to "1*tchar" in ABNF
+          pattern: "^[a-zA-Z0-9!#$%&'*+.-^`|~]+$"
```
notEthan, line 239: The character class will currently want to interpret `.-^` as a character range - the `-` should be at the end or beginning. `\`-escaping is also an option but I think putting it at an end is more universal.

src/schemas/validation/schema.yaml
```diff
@@ -230,6 +230,24 @@ $defs:
...
+            enum:
+              - GET
+              - PUT
+              - POST
+              - DELETE
+              - OPTIONS
+              - HEAD
+              - PATCH
+              - TRACE
+              - QUERY
```
notEthan, line 250: I agree this is a good form for this schema. I'm confused that these are upper case, though, when the property names of the `path-item` object are lower-case. Neither `path-item` nor `additionalOperations` can contain `GET` but both may contain `get`.

src/schemas/validation/schema.yaml
handrews: @notEthan I thought I'd had it at the end... yeah this needs fixing.
```suggestion
          pattern: "^[a-zA-Z0-9!#$%&'*+.^`|~]+$-"
```

src/schemas/validation/schema.yaml
handrews: @notEthan BTW this is part of why I argued that you _either_ used the existing fields _or_ used a new `operations` field that would handle all methods with the same casing rules (exact case match).  Because I think it's confusing to use a mixture of fields that are the "wrong" case for the method they produce plus a supplementary object that uses the exact case as-is.  But I lost that argument.

notEthan: Ah, this was said on #1747 which I didn't read or recall sufficiently. Thanks.

src/schemas/validation/schema.yaml
notEthan:
```suggestion
          pattern: "^[a-zA-Z0-9!#$%&'*+.^`|~-]+$"
```

src/schemas/validation/schema.yaml
```diff
@@ -230,6 +230,24 @@ $defs:
         type: array
         items:
           $ref: '#/$defs/parameter-or-reference'
+      additionalOperations:
+        type: object
+        additionalProperties:
+          $ref: '#/$defs/operation'
+        propertyNames:
+          $comment: RFC9110 restricts methods to "1*tchar" in ABNF
+          pattern: "^[a-zA-Z0-9!#$%&'*+.^`|~-]+$"
```
handrews, line 239: I thought I committed this but apparently not?  Trying again.

```suggestion
          pattern: "^[a-zA-Z0-9!#$%&'*+.^`|~]+$-"
```

notEthan: You changed it once to move the hyphen the end of the character class (correct) but then moved it to the end of the regex (not correct)

handrews: :facepalm:

src/schemas/validation/schema.yaml
```diff
@@ -230,6 +230,24 @@ $defs:
         type: array
         items:
           $ref: '#/$defs/parameter-or-reference'
+      additionalOperations:
+        type: object
+        additionalProperties:
+          $ref: '#/$defs/operation'
+        propertyNames:
+          $comment: RFC9110 restricts methods to "1*tchar" in ABNF
+          pattern: "^[a-zA-Z0-9!#$%&'*+.^`|~]+$-"
```
handrews, line 239:
```suggestion
          pattern: "^[a-zA-Z0-9!#$%&'*+.^`|~-]+$"
```

handrews (dismissed):

notEthan (approved):

src/oas.md
```diff
@@ -932,6 +932,8 @@ The path itself is still exposed to the documentation viewer but they will not k
 | <a name="path-item-head"></a>head | [Operation Object](#operation-object) | A definition of a HEAD operation on this path. |
 | <a name="path-item-patch"></a>patch | [Operation Object](#operation-object) | A definition of a PATCH operation on this path. |
 | <a name="path-item-trace"></a>trace | [Operation Object](#operation-object) | A definition of a TRACE operation on this path. |
+| <a nmae="path-item-query"></a>query | [Operation Object](#operation-object) | A definition of a QUERY operation, as defined in the most recent IETF draft ([draft-ietf-httpbis-safe-method-w-body-08](https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-08.html) as of this writing) or its RFC successor, on this path. |
```
ralfhandl, line 935:
```suggestion
| <a name="path-item-query"></a>query | [Operation Object](#operation-object) | A definition of a QUERY operation, as defined in the most recent IETF draft ([draft-ietf-httpbis-safe-method-w-body-08](https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-08.html) as of this writing) or its RFC successor, on this path. |
```

ralfhandl (changes requested):
+1, just one typo and one omission

ralfhandl (approved):

handrews (approved):

PR #4515, opened 2025-03-28
Improve Markdown linting and use it on more files

### LINKED ISSUE #4249: validate-markdown: include markdown files in project root
... and clean them up to pass validation

lornajane:
Fixes #4249 which asks for Markdown linting to be applied to the markdown files at the root level of the repo.

Points to look for:

- I added a second markdown lint configuration, I expect the standards for the spec and for our everyday markdown files to diverge so while they're not all that different now, I still think it's worth the overhead.
- We no longer use `mdv`, it is outdated and we can do most of what's needed with markdownlint. My only concern is that we're missing some formatting of json/yaml/abnf code examples, so we should consider any options on that (and be careful with them in the meantime!)
- I also removed prettier. We had a script which ran prettier and then did a `sed` manouvre to fix it. I don't think the benefits of using prettier are worth having this sort of hackery in the project so we'll use the markdownlint "fix" feature instead and I removed the script.

I realise that `src/oas.md` isn't present on this branch but the same configuration seemed to be okay with what's currently on both `v3.1-dev` and `v3.2-dev`.


package.json
ralfhandl, line 1: 1. `mdv` is still listed in `dependencies`: can we remove it?
2. `markdownlint-cli2` is not declared as a dependency, `devDependencies` still lists the traditional `markdownlint-cli` that we used before

ralfhandl:
The `validate-markdown.yaml` workflow still calls `mdv` - should that call be removed?

It also calls `markdownlint-cli --config .markdownlint.yaml` on the spec files:
- old cli instead of cli2
- uses config file that is now intended for non-specs on the spec files

TOB.md
```diff
@@ -1,14 +1,16 @@
 # Technical Oversight Board ("TOB")
 
-## Description: 
+## Description
+>
```
ralfhandl, line 4:
```suggestion

```

ralfhandl (changes requested):
Good direction, let's just remove `mdv` everywhere.

lornajane:
Good catch on the dependencies! Using `npx` doesn't need the packages pre-installed so they don't need to be in the dependencies list.

lornajane:
Hmm. Those GitHub workflows seem to check the old versions (which should be immutable!), this work is to improve the standards for the in-progress spec versions on the two dev branches, and all `*.md` files at the root of the project in all branches. Although now I've removed the libraries from the dependencies, maybe we should do something with those as well. At least for the ones that are 3.0.4 or 3.1.1 and later since they should be at a standard that works. Or stop validating the version files because everything shoud have been in a valid `src/oas.md` anyway before it got baked into a release?

ralfhandl:
> stop validating the version files because everything shoud have been in a valid `src/oas.md` anyway before it got baked into a release?

That would be my way forward. I've sketched what I'd do in
* https://github.com/lornajane/OpenAPI-Specification/pull/4

@lornajane The above mentioned PR is in your repo and will update this PR's head branch. Feel free to review, merge (in your repo), do something similar in your head branch, ... 😎

lornajane:
Thanks @ralfhandl , I merged your changes (with a minor edit so we only validate and don't fix in the workflow), could you take another look please?

ralfhandl (approved):

ralfhandl:
Note: changed branch protection rule for `main` and made the `lint` job a required status check instead of its old name `mdv`.

The same change will be needed for other branch protection rules.

PR #4518, opened 2025-03-29
Sequential (Streaming) media types and link to registry

handrews:
This adds a link to the forthcoming media type registry (PR #4517), and also adds support for various sequential media types:

* `application/json-seq`
* `application/jsonl`
* `application/x-ndjson`
* `text/event-stream`

Given how various modeling and encoding techniques are scattered throughout the specification, the Media Types section seemed like the best place to add these, preceded by a link to he new Media Type Registry which will essentially be a catalog of where to find the existing guidance for various media types.

Also paging @robertlagrant and @disintegrator

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -84,6 +84,130 @@ Some examples of possible media type definitions:
   application/vnd.github.v3.patch
 ```
 
+#### Media Type Registry
+
+While the [Schema Object](#schema-object) is designed to describe and validate JSON, several other media types are commonly used in APIs.
+Requirements regarding support for other media types are documented in this Media Types section and in several Object sections later in this specification.
+For convenience and future extensibility, these are cataloged in the OpenAPI Initiative's [Media Type Registry](https://spec.openapis.org/registry/media-type/), which indicates where in this specification the relevant requirements can be found.
+
+#### Sequential Media Types
+
+Several media types exist to transport a sequence of values, separated by some delimiter, either as a single document or as multiple documents representing chunks of a logical stream.
+Depending on the media type, the values could either be in another existing format such as JSON, or in a custom format specific to the sequential media type.
+
+Implementations MUST support modeling sequential media types with the [Schema Object](#schema-object) by treating the sequence as an array with the same items and ordering as the sequence.
+
+##### Working With Indefinite-Length Streams
+
+In addition to regular document-style use, sequential media types can be used to represent some portion of a stream that may not have a well-defined beginning or end.
+In such use cases, either the client or server makes a decision to work with one or more elements in the sequence at a time, but this subsequence is not a complete array in the sense of normal JSON arrays.
+
+OpenAPI Description authors are responsible for avoiding the use of JSON Schema keywords such as `prefixItems`, `minItems`, `maxItems`, `contains`, `minContains`, or `maxContains` that rely on a beginning (for relative positioning) or an ending (to determine if a threshold has been reached or a limit has been exceeded) when the sequence is intended to represent a subsequence of a larger stream.
+If such keywords are used, their behavior remains well-defined but may be counter-intuitive for users that expect them to apply to the stream as a whole rather than each subsequence as it is processed.
+
+##### Sequential JSON
+
+For any media type where the items in the sequence are JSON values, no conversion beyond treating the sequence as an array is required.
+JSON Text Sequences (`application/json-seq` and the `+json-seq` suffix, [[?RFC7464]]), JSON Lines (`application/jsonl`), and NDJSON (`application/x-ndjson`) are all in this category.
+Note that the media types for JSON Lines and NDJSON are not registered with the IANA, but are in common use.
+
+The following example, which uses `application/json-seq` but would be identical aside from the media type for either `application/jsonl` or `application/ndjson`, models a finite stream consisting of a single metadata document followed by an indefinite number of data documents consisting of numeric measurements with units:
+
+```YAML
+content:
+  application/json-seq:
+    schema:
+      type: array
+      prefixItems:
+      - $comment: Metadata for all subsequent data documents
+        type: object
+        required:
+        - subject
+        - dateCollected
+        properties:
+          subject:
+            type: string
+          dateCollected:
+            type: string
+            format: date-time
+      items:
+        $comment: A JSON document holding data
+        type: object
+        required:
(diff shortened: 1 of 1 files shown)


handrews:
BTW I don't know that under "Media Types" is the right place for the sequential media type requirements.  I could see it going as a new subsection under "Data Types", _maybe_?  Or next to it?

It really does not feel like it should go under the Schema Object, as it's more about what you do or don't use _with_ the Schema Object rather than how the Schema Object works in general.  Once you map from the sequential format to the JSON Schema data model, the Schema Object behaves normally.

src/oas.md
```diff
@@ -84,6 +84,130 @@ Some examples of possible media type definitions:
...
+While the [Schema Object](#schema-object) is designed to describe and validate JSON, several other media types are commonly used in APIs.
+Requirements regarding support for other media types are documented in this Media Types section and in several Object sections later in this specification.
+For convenience and future extensibility, these are cataloged in the OpenAPI Initiative's [Media Type Registry](https://spec.openapis.org/registry/media-type/), which indicates where in this specification the relevant requirements can be found.
+
+#### Sequential Media Types
+
+Several media types exist to transport a sequence of values, separated by some delimiter, either as a single document or as multiple documents representing chunks of a logical stream.
+Depending on the media type, the values could either be in another existing format such as JSON, or in a custom format specific to the sequential media type.
+
+Implementations MUST support modeling sequential media types with the [Schema Object](#schema-object) by treating the sequence as an array with the same items and ordering as the sequence.
```
duncanbeevers, line 98: This wording is confusing to me, and doesn't seem to reflect the requirement that the Schema Object modeling the sequence must itself be of `type: array`.

handrews: There is no requirement that the Schema Object include `type: array`, although it would be a good practice.

What we're talking about here is not so much what to put in the Schema Object, but what data structure to convert the document to in order to use that data structure with the document.

Implementations don't get that from the Schema Object, they get that from these requirements, so it would be an error on the part of the implementation to pass anything but an array here.  Of course, it's good practice to put the `type: array` in, and if you have other tools that depend on the `type` keyword and aren't paying attention to the  media type with which the Schema Object is used, then you have to do that.  But there's no requirement for it to be in the Schema Object.

handrews:
The force-push just rebases the unchanged commit in order to get the syntax highlighting for `text/event-stream`.

src/oas.md
handrews: @duncanbeevers my most recent commit (after a force-push that was a re-base of the unchanged original commit) added some clarification here, please see if that helps!

src/oas.md
```diff
@@ -84,6 +84,131 @@ Some examples of possible media type definitions:
...
+
+##### Working With Indefinite-Length Streams
+
+In addition to regular document-style use, sequential media types can be used to represent some portion of a stream that may not have a well-defined beginning or end.
+In such use cases, either the client or server makes a decision to work with one or more elements in the sequence at a time, but this subsequence is not a complete array in the sense of normal JSON arrays.
+
+OpenAPI Description authors are responsible for avoiding the use of JSON Schema keywords such as `prefixItems`, `minItems`, `maxItems`, `contains`, `minContains`, or `maxContains` that rely on a beginning (for relative positioning) or an ending (to determine if a threshold has been reached or a limit has been exceeded) when the sequence is intended to represent a subsequence of a larger stream.
+If such keywords are used, their behavior remains well-defined but may be counter-intuitive for users that expect them to apply to the stream as a whole rather than each subsequence as it is processed.
```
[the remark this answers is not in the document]
handrews: @ThomasRooney first, let me apologize for not tagging you in the original PR comment- I knew I was missing someone!

I'm going to take a while to think through this further, and also tag @gregsdennis who asked about this direction on Slack.

For now, I'll just state a few important principles that are guiding me here:
* We model media types, and not protocols implemented on top of media types.  There's nothing wrong with modeling protocols, but it can't be done by repurposing the media type layer.  It would need a new mechanism, and that's too big of a change for 3.2, which needs to ship by this summer.  Really, that would be better as a companion specification as it is beyond the current scope of the OAS.
* The challenge here is that there's nothing in any of the JSON media types that says that every entry MUST be in the same format.  If that were the case, then yes, the natural modeling would be to just model the single entry type.  But we need to work with the media types as written, not as would make them more convenient.  `text/event-stream` will tend to be more uniform, but there's no guarantee that someone won't use it in an unexpected way.
* I feel like you're focusing on the response use case, but there are request use cases where the JSONL being sent is closer to a normal document.
* I'm not that fixated on `prefixItems`, and in fact I think a more common relevant use would be to use `maxItems` as a way to limit the chunk size, although I do not know if that is ever actually done.
* The Encoding Object is problematic for far too many reasons to get into here, and is due for a re-think in 3.3

handrews:
@karenetheridge I largely agree with you here.  Attempting to manage chunks has been unsatisfactory.  And on a walk back from the store just now I more-or-less arrived at the same sort of `itemSchema` solution, although I might have some of the details a bit different.  I'll write more on this later, although everyone should feel free to keep developing @karenetheridge 's idea without waiting for me.

handrews:
OK I'm going to close this in favor of starting over based on @karenetheridge 's idea.  Karen, I'll follow up with you about whether you want to write a PR or be credited on a rewrite that I post (I said I'd gotten to the same point, but my idea was at the stage of "um... we probably need a new field" while you provided a complete write-up and explanation, and worked out many details I hadn't even considered yet).

mikekistler:
I had a long discussion on this topic with @hudlow today, and I now understand the situation much better. I agree that a new keyword seems like the right approach. Unfortunately, I think that means that there is no good solution to describing streaming media types prior to V3.2.

hudlow:
I'm also good with @karenetheridge's proposed direction although I _also_ sort of wonder about a `chunkSchema` that _would_ be modeled as an array schema and that you could use if you wanted to guarantee a whole number of items would be in a chunk and make further assertions like minimum or maximum items in each chunk.

handrews:
@mikekistler @hudlow @karenetheridge etc. please continue this discussion at https://github.com/OAI/OpenAPI-Specification/discussions/4171#discussioncomment-12796636 (my apologies for not noting this earlier- you were not expected to guess that I had transferred Karen's proposal over there!)

PR #4553, opened 2025-04-27
highlight.js: support for JSONL, JSON-Seq, basic multipart, improve text/event-stream

handrews:
Improve text/event-stream `highlight.js` support with comment support and generally better configuration (one rule wasn't doing anything).

Add support for JSONL/NDJSON (basically JSON without the check for values separated only by whitespace) and JSON Text Sequences (which requires using `0x1E` as the record separator, and highlighting it separately).

Multipart support is very basic, and only supports `Content-` part headers (which is what you generally use anyway).

Examples from the updated test fixtures:

Screen Shot 2025-04-28 at 8 56 37 PM

### DIFF
--- a/scripts/md2html/md2html.js
+++ b/scripts/md2html/md2html.js
@@ -52,18 +52,70 @@ hljs.registerLanguage('eventstream', function() {
     return {
       contains: [
           {
-              scope: "attr",
-              begin: /^/,
-              end: ":",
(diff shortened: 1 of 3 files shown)


handrews:
Last commit added some very basic multipart support.

ralfhandl (approved):

PR #4554, opened 2025-04-28
v3.2: Sequential/streaming media types (take 2)

handrews:
@ThomasRooney @disintegrator @gregsdennis @philsturgeon @hudlow here is a new attempt at supporting sequential and/or streaming media types.

The fundamental approach in this change was proposed by @karenetheridge as a comment on the previous attempt at solving this problem.

This adds `itemSchema` as a schema to apply to each entry in a sequential media type instance.  It also defines how to map sequential media tyes for use with `schema`, and explains that `schema` applies to complete content only.

JSON-based and text/event-stream media types are included.

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -1661,7 +1661,7 @@ See also the [Media Type Registry](#media-type-registry).
 
 The `schema` field MUST be applied to the complete content, as defined by the media type and the context ([Request Body Object](#request-body-object), [Response Object](#response-object), [Parameter Object](#parameter-object), or [Header Object](#header-object).
 Unless some sort of streaming JSON Schema processor is available, this requires loading the entire content into memory.
-This poses a challenge for streamed media, particularly streams where the client is intended to choose when to stop reading as there is no well-defined end to the stream.
+This poses a challenge for streamed content, particularly streams where the client is intended to choose when to stop reading as there is no well-defined end to the stream.
 
 ###### Binary Streams
 
--- a/src/oas.md
+++ b/src/oas.md
@@ -1690,21 +1690,19 @@ Note that `contentSchema` is [not automatically validated by default](https://ww
 The following Schema Object is a generic schema for the `text/event-stream` media type as documented by the HTML specification as of the time of this writing:
 
 ```YAML
-type: array
-items:
-  type: object
-  required:
-  - data
-  properties:
-    data:
-      type: string
-    event:
-      type: string
-    id:
-      type: string
-    retry:
-      type: integer
-      minimum: 0
+type: object
+required:
+- data
+properties:
+  data:
+    type: string
+  event:
+    type: string
+  id:
+    type: string
+  retry:
+    type: integer
+    minimum: 0
 ```
 
 ##### Encoding Usage and Restrictions
--- a/src/oas.md
+++ b/src/oas.md
@@ -1665,8 +1665,8 @@ This poses a challenge for streamed media, particularly streams where the client
 
 ###### Binary Streams
 
-The `maxLength` keyword MAY be used to set an expected upper bound on the length of a streaming payload.
-The keyword can be applied to either string data, including encoded binary data, or to unencoded binary data. For unencoded binary, the length is the number of octets.
+The `maxLength` keyword MAY be used to set an expected upper bound on the length of a streaming payload that consists of either string data, including encoded binary data, or to unencoded binary data.
+For unencoded binary, the length is the number of octets.
 For this use case, `maxLength` MAY be implemented outside of regular JSON Schema evaluation as JSON Schema does not directly apply to binary data, and an encoded binary stream may be impractical to store in memory in its entirety.
 
 ###### Streaming Sequential Media Types
--- a/src/oas.md
+++ b/src/oas.md
@@ -1665,7 +1665,7 @@ Use cases where client is intended to choose when to stop reading are particular
 
 ###### Binary Streams
 
-The `maxLength` keyword MAY be used to set an expected upper bound on the length of a streaming payload that consists of either string data, including encoded binary data, or to unencoded binary data.
+The `maxLength` keyword MAY be used to set an expected upper bound on the length of a streaming payload that consists of either string data, including encoded binary data, or unencoded binary data.
 For unencoded binary, the length is the number of octets.
 For this use case, `maxLength` MAY be implemented outside of regular JSON Schema evaluation as JSON Schema does not directly apply to binary data, and an encoded binary stream may be impractical to store in memory in its entirety.
 
--- a/src/oas.md
+++ b/src/oas.md
@@ -1671,7 +1671,7 @@ For this use case, `maxLength` MAY be implemented outside of regular JSON Schema
 
 ###### Streaming Sequential Media Types
 
-The `itemSchema` field is provided to support streaming use case for sequential media types.
+The `itemSchema` field is provided to support streaming use cases for sequential media types.
 Unlike `schema`, which is applied to the complete content (treated as an array as described in the [sequential media types](#sequential-media-types) section), `itemSchema` MUST be applied to each item in the stream independently, which supports processing each item as it is read from the stream.
 
 Both `schema` and `itemSchema` MAY be used in the same Media Type Object, although doing so is unlikely to have significant advantages over using the `items` keyword within the `schema` field.
--- a/src/oas.md
+++ b/src/oas.md
@@ -1648,7 +1648,7 @@ See [Working With Examples](#working-with-examples) for further guidance regardi
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
 | <a name="media-type-schema"></a>schema | [Schema Object](#schema-object) | A schema describing the complete content of the request, response, parameter, or header. |
-| <a name="media-type-item-schema"></a>itemSchema> | [Schema Object](#schema-object) | A schema describing each item within a [sequential media type](#sequential-media-types). |
+| <a name="media-type-item-schema"></a>itemSchema | [Schema Object](#schema-object) | A schema describing each item within a [sequential media type](#sequential-media-types). |
 | <a name="media-type-example"></a>example | Any | Example of the media type; see [Working With Examples](#working-with-examples). |
 | <a name="media-type-examples"></a>examples | Map[ `string`, [Example Object](#example-object) \| [Reference Object](#reference-object)] | Examples of the media type; see [Working With Examples](#working-with-examples). |
 | <a name="media-type-encoding"></a>encoding | Map[`string`, [Encoding Object](#encoding-object)] | A map between a property name and its encoding information, as defined under [Encoding Usage and Restrictions](#encoding-usage-and-restrictions).  The `encoding` field SHALL only apply when the media type is `multipart` or `application/x-www-form-urlencoded`. If no Encoding Object is provided for a property, the behavior is determined by the default values documented for the Encoding Object. |
--- a/src/oas.md
+++ b/src/oas.md
@@ -1680,7 +1680,7 @@ For example, if partial content is read from a stream and then passed with the `
 
 ##### Special Considerations for `text/event-stream` Content
 
-For `text/event-stream`, each item in the array MUST be treated as if it were a JSON object with property names taken from the left side of the `:` (or the enter non-empty line if no ":" is present), property values from the right side, and consecutive lines with the same name treated as a single property, with the value combined in accordance with the [`text/event-stream` specification](https://html.spec.whatwg.org/multipage/iana.html#text/event-stream).
+For `text/event-stream`, each item in the array MUST be treated as if it were a JSON object with property names taken from the left side of the `:` (or the enter non-empty line if no ":" is present), property values from the right side, and consecutive lines with the same name treated as a single property, with the value combined in accordance with the [`text/event-stream` specification](https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream).
 
 Field names can be repeated within an item to allow splitting the value across multiple lines; such split values MUST be treated the same as if they were a single field, with newlines added as required by the [`text/event-stream` specification](https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream).
 Similarly, when a field name appears without a value, the value MUST be treated as an empty string.
--- a/src/oas.md
+++ b/src/oas.md
@@ -91,7 +91,7 @@ Each mapping is addressed where the relevant media type is discussed in this sec
 
 #### Sequential Media Types
 
-Within this specification, a _sequential media type_ is defined as any media type that consists of a repeating structure, separated by some delimiter, without any sort of header, footer, envelope, or other metadata in addition to the sequence.
+Within this specification, a _sequential media type_ is defined as any media type that consists of a repeating structure, without any sort of header, footer, envelope, or other metadata in addition to the sequence.
 
 Some examples of sequential media types (including some that are not IANA-registered but are in common use) are:
 
@@ -1660,8 +1660,8 @@ See also the [Media Type Registry](#media-type-registry).
 ##### Complete vs Streaming Content
 
 The `schema` field MUST be applied to the complete content, as defined by the media type and the context ([Request Body Object](#request-body-object), [Response Object](#response-object), [Parameter Object](#parameter-object), or [Header Object](#header-object).
-Unless some sort of streaming JSON Schema processor is available, this requires loading the entire content into memory.
-This poses a challenge for streamed content, particularly streams where the client is intended to choose when to stop reading as there is no well-defined end to the stream.
+Because this requires loading the content into memory in its entirety, it poses a challenge for streamed content.
+Use cases where client is intended to choose when to stop reading are particularly challenging as there is no well-defined end to the stream.
 
 ###### Binary Streams
 
@@ -1680,13 +1680,7 @@ For example, if partial content is read from a stream and then passed with the `
 
 ##### Special Considerations for `text/event-stream` Content
 
-For `text/event-stream`, each item in the array MUST be treated as if it were a JSON object with property names taken from the left side of the `:` (or the enter non-empty line if no ":" is present), property values from the right side, and consecutive lines with the same name treated as a single property, with the value combined in accordance with the [`text/event-stream` specification](https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream).
-
-Field names can be repeated within an item to allow splitting the value across multiple lines; such split values MUST be treated the same as if they were a single field, with newlines added as required by the [`text/event-stream` specification](https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream).
-Similarly, when a field name appears without a value, the value MUST be treated as an empty string.
-
-The `text/event-stream` specification requires that fields with Unknown names, as well as `id` fields where the value contains `U+0000 NULL` and `retry` fields with characters other than ASCII digits be ignored.
-These fields SHOULD NOT be present in the data used with the Schema Object.
+For `text/event-stream`, implementations MUST work with event data after it has been parsed according to the [`text/event-stream` specification](https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream), including all guidance on ignoring certain fields (including comments) and/or values, and on combining values split across multiple lines.
 
 Field value types MUST be handled as specified by the `text/event-stream` specification (e.g. the `retry` field value is modeled as a JSON number that is expected to be of JSON Schema `type: integer`), and fields not given an explicit value type MUST be handled as strings.
 
--- a/src/oas.md
+++ b/src/oas.md
@@ -1674,9 +1674,8 @@ For this use case, `maxLength` MAY be implemented outside of regular JSON Schema
 The `itemSchema` field is provided to support streaming use cases for sequential media types.
 Unlike `schema`, which is applied to the complete content (treated as an array as described in the [sequential media types](#sequential-media-types) section), `itemSchema` MUST be applied to each item in the stream independently, which supports processing each item as it is read from the stream.
 
-Both `schema` and `itemSchema` MAY be used in the same Media Type Object, although doing so is unlikely to have significant advantages over using the `items` keyword within the `schema` field.
-OpenAPI Description authors are responsible for avoiding the use of the `schema` in any situation where tooling may not be able to discern when the content is complete.
-For example, if partial content is read from a stream and then passed with the `schema` value to a schema evaluator that is unaware of the stream context, the results will not be meaningful as there may be additional items in the stream that are necessary to or prohibitive of successful validation.
+Both `schema` and `itemSchema` MAY be used in the same Media Type Object.
+However, doing so is unlikely to have significant advantages over using the `items` keyword within the `schema` field.
 
 ##### Special Considerations for `text/event-stream` Content
 
--- a/src/oas.md
+++ b/src/oas.md
@@ -1680,11 +1680,12 @@ For example, if partial content is read from a stream and then passed with the `
 
 ##### Special Considerations for `text/event-stream` Content
 
-For `text/event-stream`, each item in the array MUST be treated as if it were a JSON object with property names taken from the left side of the `:`, property values from the right side, and consecutive lines with the same name treated as a single property, with the value combined in accordance with the [`text/event-stream` specification](https://html.spec.whatwg.org/multipage/iana.html#text/event-stream).
+For `text/event-stream`, each item in the array MUST be treated as if it were a JSON object with property names taken from the left side of the `:` (or the enter non-empty line if no ":" is present), property values from the right side, and consecutive lines with the same name treated as a single property, with the value combined in accordance with the [`text/event-stream` specification](https://html.spec.whatwg.org/multipage/iana.html#text/event-stream).
 
 Field names can be repeated within an item to allow splitting the value across multiple lines; such split values MUST be treated the same as if they were a single field, with newlines added as required by the [`text/event-stream` specification](https://html.spec.whatwg.org/multipage/iana.html#text/event-stream).
+Similarly, when a field name appears without a value, the value MUST be treated as an empty string.
 
-The `text/event-stream` specification requires that fields with Unknown names, as well as `id` fields where the value contains `U+0000 NULL` be ignored.
+The `text/event-stream` specification requires that fields with Unknown names, as well as `id` fields where the value contains `U+0000 NULL` and `retry` fields with characters other than ASCII digits be ignored.
 These fields SHOULD NOT be present in the data used with the Schema Object.
 
 Field value types MUST be handled as specified by the `text/event-stream` specification (e.g. the `retry` field value is modeled as a JSON number that is expected to be of JSON Schema `type: integer`), and fields not given an explicit value type MUST be handled as strings.
@@ -1710,6 +1711,7 @@ items:
       type: string
     retry:
       type: integer
+      minimum: 0
 ```
 
 ##### Encoding Usage and Restrictions
--- a/src/oas.md
+++ b/src/oas.md
@@ -1676,7 +1676,7 @@ Unlike `schema`, which is applied to the complete content (treated as an array a
 
 Both `schema` and `itemSchema` MAY be used in the same Media Type Object, although doing so is unlikely to have significant advantages over using the `items` keyword within the `schema` field.
 OpenAPI Description authors are responsible for avoiding the use of the `schema` in any situation where tooling may not be able to discern when the content is complete.
-For example, if partial content is read from a stream and then passed with the `schema` value to a schema evaluator that is unaware of the stream context, the results are well-defined but will not be meaningful in terms of validating a keyword like `maxItems` as there may be additional items in the stream that are unknown to the schema evaluator.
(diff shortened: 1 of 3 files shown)


handrews:
@ThomasRooney this does not address your ideas around the Encoding Object and sentinel events, mostly to avoid further complexity.  I'm happy to discuss that idea further if you want to start a new thread on the discussion (#4171).  We are also talking there about how streaming `multipart/mixed` would work, which would involve the Encoding Object, so I'll take on Encoding Object-related changes separately.  I'm not entirely sold that a "sentinel event" is an encoding, but I might be convinced of it.

src/oas.md
```diff
@@ -1612,50 +1647,80 @@ See [Working With Examples](#working-with-examples) for further guidance regardi
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
-| <a name="media-type-schema"></a>schema | [Schema Object](#schema-object) | The schema defining the content of the request, response, parameter, or header. |
+| <a name="media-type-schema"></a>schema | [Schema Object](#schema-object) | A schema describing the complete content of the request, response, parameter, or header. |
+| <a name="media-type-item-schema"></a>itemSchema> | [Schema Object](#schema-object) | A schema describing each item within a [sequential media type](#sequential-media-types). |
```
gregsdennis, line 1651: I think there's an extra `>` after `itemSchema`.
```suggestion
| <a name="media-type-item-schema"></a>itemSchema | [Schema Object](#schema-object) | A schema describing each item within a [sequential media type](#sequential-media-types). |
```

src/oas.md
```diff
@@ -1612,50 +1647,80 @@ See [Working With Examples](#working-with-examples) for further guidance regardi
...
-}
+##### Complete vs Streaming Content
+
+The `schema` field MUST be applied to the complete content, as defined by the media type and the context ([Request Body Object](#request-body-object), [Response Object](#response-object), [Parameter Object](#parameter-object), or [Header Object](#header-object).
+Unless some sort of streaming JSON Schema processor is available, this requires loading the entire content into memory.
+This poses a challenge for streamed media, particularly streams where the client is intended to choose when to stop reading as there is no well-defined end to the stream.
+
+###### Binary Streams
+
+The `maxLength` keyword MAY be used to set an expected upper bound on the length of a streaming payload.
```
gregsdennis, line 1668: Does "length" here refer to the number of items in the "array" or to the lexical length of the stream?

src/oas.md
```diff
@@ -1612,50 +1647,80 @@ See [Working With Examples](#working-with-examples) for further guidance regardi
...
+For example, if partial content is read from a stream and then passed with the `schema` value to a schema evaluator that is unaware of the stream context, the results are well-defined but will not be meaningful in terms of validating a keyword like `maxItems` as there may be additional items in the stream that are unknown to the schema evaluator.
+
+##### Special Considerations for `text/event-stream` Content
+
+For `text/event-stream`, each item in the array MUST be treated as if it were a JSON object with property names taken from the left side of the `:`, property values from the right side, and consecutive lines with the same name treated as a single property, with the value combined in accordance with the [`text/event-stream` specification](https://html.spec.whatwg.org/multipage/iana.html#text/event-stream).
+
+Field names can be repeated within an item to allow splitting the value across multiple lines; such split values MUST be treated the same as if they were a single field, with newlines added as required by the [`text/event-stream` specification](https://html.spec.whatwg.org/multipage/iana.html#text/event-stream).
```
ThomasRooney, line 1685: Might be slightly better to use this link as it defines the ABNF

```suggestion
Field names can be repeated within an item to allow splitting the value across multiple lines; such split values MUST be treated the same as if they were a single field, with newlines added as required by the [`text/event-stream` specification](https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream).
```

src/oas.md
handrews: This is in the binary streams section so it really only covers those streams- the answer is in the next paragraph or two.  I did not change this section, btw, just moved it down from the data format section on binary data.

src/oas.md
handrews: Yeah probably.  I find WHATWG specs baffling- I don't look for "parsing a..." as a format definition.  Although I should give them credit for at least providing an ABNF for once.

I'll need to update some other things based on this, like handling lines without colons.

handrews:
@ThomasRooney added more details in this most recent commit from reading that part of the WHATWG spec, including a `minimum: 0` on `retry` because of the ignore-unless-ASCII digits restriction, which I had not previously noticed.

src/oas.md
handrews: @gregsdennis I tightened up the wording a bit to make the scope of this techinque more clear.

src/oas.md
```diff
@@ -84,6 +84,40 @@ Some examples of possible media type definitions:
   application/vnd.github.v3.patch
 ```
 
+JSON-based and JSON-compatible YAML-based media types can make direct use of the [Schema Object](#schema-object) as the Object uses JSON Schema.
+The use of the Schema Object with other media types is handled by mapping them into the JSON Schema [instance data model](https://www.ietf.org/archive/id/draft-bhutton-json-schema-01.html#name-instance-data-model).
+These mappings may be implicit based on the media type, or explicit based on the values of particular fields.
+Each mapping is addressed where the relevant media type is discussed in this section or under the [Media Type Object](#media-type-object) or [Encoding Object](#encoding-object)
+
```
hudlow, line 91:
```suggestion
To optimize for either time or space complexity, an implementation applying a schema to content of a mapped media type MAY forgo serializing the data into JSON and running a standard JSON Schema validator provided that the optimized validation logic is manifestly equivalent for all inputs.

```

src/oas.md
```diff
@@ -1612,50 +1647,82 @@ See [Working With Examples](#working-with-examples) for further guidance regardi
...
+The `schema` field MUST be applied to the complete content, as defined by the media type and the context ([Request Body Object](#request-body-object), [Response Object](#response-object), [Parameter Object](#parameter-object), or [Header Object](#header-object).
+Unless some sort of streaming JSON Schema processor is available, this requires loading the entire content into memory.
+This poses a challenge for streamed media, particularly streams where the client is intended to choose when to stop reading as there is no well-defined end to the stream.
+
+###### Binary Streams
+
+The `maxLength` keyword MAY be used to set an expected upper bound on the length of a streaming payload that consists of either string data, including encoded binary data, or to unencoded binary data.
+For unencoded binary, the length is the number of octets.
+For this use case, `maxLength` MAY be implemented outside of regular JSON Schema evaluation as JSON Schema does not directly apply to binary data, and an encoded binary stream may be impractical to store in memory in its entirety.
```
hudlow, line 1670:
```suggestion
The `maxLength` keyword MAY be used to set an expected upper bound on the length of a streaming payload that consists of string data, including encoded binary data.

When `maxLength` is given for a schema applied to unencoded binary data, its value SHOULD be interpreted as the number of octets.

**Deprecated:** The use of the `maxLength` property to describe the maximum number of octets for unencoded binary data has been deprecated in favor of the `maxItems` property. Later versions of this specification may cease to recommend that implementations support its use in this context.

The `maxItems` keyword MAY be used to set an expected upper bound on the length of a streaming payload that consists of unencoded binary data, which is mapped to the JSON schema instance data model as an array of (0 ≤ n < 256) integers.
```

No clue what toes I'm stepping on here but, to my mind, it should be possible to _actually_ represent content of mapped media types in the instance data model, and in my experience an idiomatic representation would be the JSON equivalent of an array of unsigned 8-bit integers.

hudlow (commented):
Great work on this! I only had time for a partial review — it will take more time to get to the rest of it. Feel free to shoot down suggestions — I'm making them to put options on the table. 🙂

src/oas.md
```diff
@@ -1612,50 +1647,82 @@ See [Working With Examples](#working-with-examples) for further guidance regardi
...
-      "frog": {
-        "$ref": "#/components/examples/frog-example"
-      }
-    }
-  }
-}
+##### Complete vs Streaming Content
+
+The `schema` field MUST be applied to the complete content, as defined by the media type and the context ([Request Body Object](#request-body-object), [Response Object](#response-object), [Parameter Object](#parameter-object), or [Header Object](#header-object).
+Unless some sort of streaming JSON Schema processor is available, this requires loading the entire content into memory.
```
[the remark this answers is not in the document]
handrews: Yeah this is a bit awkward.  I only know of one attempt at a [streaming json schema parser/validator](https://github.com/awwright/jsonschemaparse) so it's probably not worth calling out.  I feel like the suggestion is a bit too much passive voice (ironic given how much I overuse passive voice myself), but I'll think on this as well.

src/oas.md
handrews: @hudlow OK so _I_ know what this means, but I feel like it's a bit too abstract for most folks based on past feedback.  I feel like addressing just the main case at hand is going to work better, although maybe I'm overcorrecting.   idk.

src/oas.md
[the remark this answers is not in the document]
handrews: Well... _technically_ binary data is outside of the data model but the JSON Schema specification allows handling things outside of the data model.  It just doesn't specify _how_ to do so.  I might reconsider this after some thought, though.

handrews: @hudlow this would need a new issue/discussion.   I did not add any of this, I just moved it from where it used to be since it makes more sense in the streaming section now.  And I'd prefer to keep the binary and sequential cases separate, not mix `maxItems` in here.

I actually try to do the minimal change required.  I would love for people to do a pass on making sure things are well-organized and have the most effective language, but that's a separate thing than getting the new work into the spec.

handrews: (OK I clarified about how `maxLength` could be implemented, but it was always implicit in the old text)

ThomasRooney (approved):
I really like this, and highly appreciate the second iteration. 

Structured streaming of data is rapidly growing in popularity (primarily due to the AI API wave), and I think the examples you put in here are going to help a lot of vendors grok and build tooling around these formats.

src/oas.md
```diff
@@ -84,6 +84,40 @@ Some examples of possible media type definitions:
...
 ```
 
+JSON-based and JSON-compatible YAML-based media types can make direct use of the [Schema Object](#schema-object) as the Object uses JSON Schema.
+The use of the Schema Object with other media types is handled by mapping them into the JSON Schema [instance data model](https://www.ietf.org/archive/id/draft-bhutton-json-schema-01.html#name-instance-data-model).
+These mappings may be implicit based on the media type, or explicit based on the values of particular fields.
+Each mapping is addressed where the relevant media type is discussed in this section or under the [Media Type Object](#media-type-object) or [Encoding Object](#encoding-object)
+
+#### Sequential Media Types
+
+Within this specification, a _sequential media type_ is defined as any media type that consists of a repeating structure, separated by some delimiter, without any sort of header, footer, envelope, or other metadata in addition to the sequence.
```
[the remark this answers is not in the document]
hudlow: Feel free to resolve this comment whenever you are satisfied.

src/oas.md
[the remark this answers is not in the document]
hudlow: Feel free to resolve this comment whenever you are satisfied.

src/oas.md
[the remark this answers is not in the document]
hudlow: Feel free to resolve this comment whenever you are satisfied.

src/oas.md
hudlow: Understood, whatever you think.

src/oas.md
hudlow: Feel free to resolve this comment.

src/oas.md
```diff
@@ -1612,50 +1647,82 @@ See [Working With Examples](#working-with-examples) for further guidance regardi
...
+
+###### Binary Streams
+
+The `maxLength` keyword MAY be used to set an expected upper bound on the length of a streaming payload that consists of either string data, including encoded binary data, or to unencoded binary data.
+For unencoded binary, the length is the number of octets.
+For this use case, `maxLength` MAY be implemented outside of regular JSON Schema evaluation as JSON Schema does not directly apply to binary data, and an encoded binary stream may be impractical to store in memory in its entirety.
+
+###### Streaming Sequential Media Types
+
+The `itemSchema` field is provided to support streaming use case for sequential media types.
```
hudlow, line 1674:
```suggestion
The `itemSchema` field is provided to support streaming use cases for sequential media types.
```

I think?

src/oas.md
```diff
@@ -1612,50 +1647,82 @@ See [Working With Examples](#working-with-examples) for further guidance regardi
...
+Unlike `schema`, which is applied to the complete content (treated as an array as described in the [sequential media types](#sequential-media-types) section), `itemSchema` MUST be applied to each item in the stream independently, which supports processing each item as it is read from the stream.
+
+Both `schema` and `itemSchema` MAY be used in the same Media Type Object, although doing so is unlikely to have significant advantages over using the `items` keyword within the `schema` field.
+OpenAPI Description authors are responsible for avoiding the use of the `schema` in any situation where tooling may not be able to discern when the content is complete.
```
hudlow, line 1678: I'm not sure I _quite_ follow this. It seems to leap from authors of the description all the way to (client-side?) runtime tooling but... can't the runtime tooling know when it's handling an unterminated stream? Isn't _it_ responsible for knowing that it may not be meaningful to apply `schema` to the content from an unterminated stream?

src/oas.md
```diff
@@ -1612,50 +1647,82 @@ See [Working With Examples](#working-with-examples) for further guidance regardi
...
+For example, if partial content is read from a stream and then passed with the `schema` value to a schema evaluator that is unaware of the stream context, the results are well-defined but will not be meaningful in terms of validating a keyword like `maxItems` as there may be additional items in the stream that are unknown to the schema evaluator.
```
hudlow, line 1679: I think I know what you mean by "well-defined" in that it's pretty easy for everyone to agree on how to serialize the partial content into a JSON array and give it and the schema to a compliant validator, and any two competent implementations will yield the same result, but I think the caveat here is too weak. I'd suggest something more like:

```suggestion
For example, if partial content is read from a stream and then passed with the `schema` value to a schema evaluator that is unaware of the stream context, the results will not be meaningful as there may be additional items in the stream that are necessary to or prohibitive of successful validation.
```

`minItems` and `maxItems` are simple, easy-to-grasp examples of when attempting to validate part of the content could lead to a false negative or a false positive, but when you also consider `items` arrays, `contains`, `uniqueItems`, and all of the array keywords combined with in-place applicators, there is a dizzying combination of ways to get meaningless results.

src/oas.md
```diff
@@ -1612,50 +1647,82 @@ See [Working With Examples](#working-with-examples) for further guidance regardi
...
+
+##### Special Considerations for `text/event-stream` Content
+
+For `text/event-stream`, each item in the array MUST be treated as if it were a JSON object with property names taken from the left side of the `:` (or the enter non-empty line if no ":" is present), property values from the right side, and consecutive lines with the same name treated as a single property, with the value combined in accordance with the [`text/event-stream` specification](https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream).
```
hudlow, line 1683: > property values from the right side

trimming a single leading space if present?

hudlow, line 1683: > or the enter non-empty line if no ":" is present

I couldn't parse this

src/oas.md
```diff
@@ -1612,50 +1647,82 @@ See [Working With Examples](#working-with-examples) for further guidance regardi
...
+These fields SHOULD NOT be present in the data used with the Schema Object.
+
+Field value types MUST be handled as specified by the `text/event-stream` specification (e.g. the `retry` field value is modeled as a JSON number that is expected to be of JSON Schema `type: integer`), and fields not given an explicit value type MUST be handled as strings.
+
+Some users of `text/event-stream` use a format such as JSON for field values, particularly the `data` field.
+Use JSON Schema's keywords for working with the [contents of string-encoded data](https://www.ietf.org/archive/id/draft-bhutton-json-schema-validation-01.html#name-a-vocabulary-for-the-conten), particularly `contentMediaType` and `contentSchema`, to describe and validate such fields with more detail than string-related validation keywords such as `pattern` can support.
+Note that `contentSchema` is [not automatically validated by default](https://www.ietf.org/archive/id/draft-bhutton-json-schema-validation-01.html#name-implementation-requirements-2) (see also the [Non-validating constraint keywords](#non-validating-constraint-keywords) section of this specification).
```
hudlow, line 1695: I have no experience with `text/event-stream` and, as is always the case for WHATWG content, reading the specification is a special kind of torture and I was not able to get terribly far in parsing it — so with that huge caveat — this seems to me to strike a really uncomfortable compromise.

I actually would have _started_ with the assumption that the JSON schema instance data model for items (events?) in a `text/event-stream` would be constructed from the representation of each item _after_ it had been fully-parsed in a manner compliant with the specification, which would imply things _such as_:

1. You don't have to talk about colons and lines because those are serialization artifacts.
2. You don't have to consider retries because, again, that's lower in the abstraction stack than where you're operating.

_Alternatively_ I could imagine implementing some _very narrow_ parts of the initial parse process, such as a raw array of lines or _maybe_ an array of field-value tuples...

But what we have is something low-level enough that it's concerned with `retry` fields, but high-level enough that it knows how to stitch `data` fields together and that just _feels_ like a weird mixture to me.

(But for the record, the rationale for the compromise you're proposing is not _entirely_ mysterious to me — I think I sort of see it; the result just makes me a bit uncomfortable.)

src/oas.md
```diff
@@ -1681,6 +1748,214 @@ application/json:
...
+      type: array
+      items:
+        $ref: "#/components/schemas/LogEntry"
+      maxItems: 100
+  examples:
+    LogJSONSeq:
+      summary: Log entries in application/json-seq
+      # JSON Text Sequences require an unprintable character
+      # that cannot be escaped in a YAML string, and therefore
+      # must be placed in an external document shown below
```
hudlow, line 1813: What's wrong with `\x1E` `\u001E`?

src/oas.md
handrews: @hudlow that's because it's gibberish 🤦   I'll fix it tomorrow.

src/oas.md
```diff
@@ -1681,6 +1748,214 @@ application/json:
...
+      externalValue: examples/log.json-seq
+    LogJSONPerLine:
+      summary: Log entries in application/jsonl or application/x-ndjson
+      description: JSONL and NDJSON are identical for this example
+      # Note that the value must be written as a string with newlines,
+      # as JSONL and NDJSON are not valid YAML
```
hudlow, line 1819: Seems like we _could_ say that examples may be structured as a JSON array, in the same way that we use an array for the JSON schema instance data model... but the obvious reason not to do this is that the example value may be useful _for_ its formatting and not merely for its semantic content.

src/oas.md
handrews: Yeah, I do take your point.  Which is further illustrated by the fact that I'm not trying to (re-)specifythat process, so it's not meant to be a "compromise."  I'll figure out a way to re-work this, because you're right that it duplicates a lot of things (which it does in part because of that special kind of torture inherent in reading WHATWG if you are doing anything other than implementing a browser).  I actually mostly worked from MDN's description TBH.  It's much more clear.

src/oas.md
hudlow: Oh... those don't work in _multiline_ strings. But we don't have to use multiline strings, do we?

handrews: Those [don't work in block scalars, only in double-quoted strings](https://yaml.org/spec/1.2-old/spec.html#id2787109).  And making that example a one-line double-quoted string would be unreadable and useless as an example.

handrews: @hudlow (I know, I couldn't' believe it either)

src/oas.md
handrews: No, examples are supposed to be _of the media type_, not of the data modeled for the media type.

handrews: (I think we have something that discusses this distinction somewhere, but I should probably double-check in the morning)

hudlow: I get it — basically the implicit YAML-as-JSON examples are an exception and definitely as far as we want to go. Feel free to resolve.

handrews:
@hudlow I merged a couple of your suggestions and will sort out the other updates I mentioned soon.

src/oas.md
hudlow: Empirically, this also seems to work but I'm not sure it's better than the single-line treatment. 😬 

```yaml
example: "\x1E{ \
    \"timestamp\": \"1985-04-12T23:20:50.52Z\", \
    \"level\": 1, \
    \"message\": \"Hi!\" \
  }\x1E{\
    \"timestamp\": \"1985-04-12T23:20:51.37Z\", \
    \"level\": 1, \
    \"message\": \"Bye!\" \
  }"
```

src/oas.md
handrews: @hudlow it's certainly _possible_ to design tools that do this, but OAS tools aren't monolithic things, and it's not clear to me that that's a reasonable requirement.

@ThomasRooney @disintegrator do either of you have thoughts on this?  @hudlow is correct that it would be nicer if we could make it a requirement.

src/oas.md
handrews: @hudlow I applaud your creativity and dedication!  That said... I'm gonna go with "nope" on this one.  But it's an apperciative "nope"!

A bit more seriously, I really like having this show both inline (`value`) and external (`externalValue`) Example Objects.  And I really like being able to apply `application/json-seq`-specific syntax highlighting to help people understand what is going on with the unprintable character, which is not possible to do in an embedded YAML string (of any format).  Or at least it is way beyond my highlightjs skills.

src/oas.md
handrews: @hudlow after thinking on it more I just removed the caveat.  The requirements as listed (`schema` MUST apply to complete content) already require the error checking you want.  The caveat came from the previous iteration of this PR where the use of `schema` was muddled and ambiguous.

While it's possible that some tool might not handle this correctly, I think we can make that a compliance issue now that I've had a chance to think through it more.  The most likely case where it would go wrong would be some application that, rather than using an actual implementation of OAS just sort of applies bits and pieces of it here and there.  But that's a... creative.. use of the spec that is outside what we need to be concerned about, I think.

hudlow (approved):

src/oas.md
```diff
@@ -1612,50 +1647,75 @@ See [Working With Examples](#working-with-examples) for further guidance regardi
...
+###### Binary Streams
+
+The `maxLength` keyword MAY be used to set an expected upper bound on the length of a streaming payload that consists of either string data, including encoded binary data, or to unencoded binary data.
+For unencoded binary, the length is the number of octets.
+For this use case, `maxLength` MAY be implemented outside of regular JSON Schema evaluation as JSON Schema does not directly apply to binary data, and an encoded binary stream may be impractical to store in memory in its entirety.
+
+###### Streaming Sequential Media Types
+
+The `itemSchema` field is provided to support streaming use cases for sequential media types.
+Unlike `schema`, which is applied to the complete content (treated as an array as described in the [sequential media types](#sequential-media-types) section), `itemSchema` MUST be applied to each item in the stream independently, which supports processing each item as it is read from the stream.
```
mikekistler, line 1675: I feel like there is still something important missing here. Maybe it is just missing in my mental model. I don't think that using `schema` _prevents_ processing content as it comes in -- I believe there are streaming JSON parsers that already do this. And I'm not aware (but will be happy to be informed) of a _requirement_ that if the "complete content" does not satisfy the `schema` that none of it may be accepted/processed.

So from that perspective (which may be misinformed, and if so I'm happy to be corrected), the purpose of `itemSchema` is to signal that, not only is it "possible" to process items as they are read, but it is "expected" that each item is completely independent and should be processed regardless of whether past or future items satisfy the `itemSchema`.

mikekistler (commented):
I found one minor wording issue that should be easy to fix.

And I left a few comments that may just reveal my lack of understanding, but I thought I'd make them anyway in case they are helpful.

src/oas.md
handrews: @hudlow this is why I had that awkward caveat about a streaming JSON Schema processor in here.

Every time someone wants me to take something out, someone else wants me to put it in.  idk what to do here.  We can't put every caveat in the text is so long as it is.

There are streaming JSON parsers, but streaming validation isn't really a thing (I know of one partial implementation).  So you don't have to load the whole content to parse it, and _in theory_ you might be able to validate it, but _in practice_ you need to load the whole thing to validate it.

How much do we really need to explain here _in the text_?

handrews: Also some JSON Schema keywords (e.g `maxProperties` on a root object) do require that you parse the whole document, although at least in theory you can do those without holding the whole document in memory.  Although with `uniqueItems` on a root array, you might end up needing to hold it all in memory, or at least hold a hash of each item... but we are getting _extremely deep_ in the weeds here.

@mikekistler it's not clear to me what the problem is with the current wording.  The wording doesn't stop someone from implementing streaming JSON Schema validation, it just says that `itemSchema` supports processing item-by-item.  This is not true of `schema` because in general JSON documents are not composed of "items" (unless they are arrays, but... do we really need to call that out?  It would take a paragraph or two even to set up the distinction)

hudlow: @mikekistler Yeah, I'm also confused. Do you actually object to anything in this paragraph or do you object to something on lines 1662 or 1663?

src/oas.md
```diff
@@ -1612,50 +1647,75 @@ See [Working With Examples](#working-with-examples) for further guidance regardi
...
+  properties:
+    data:
+      type: string
+    event:
+      type: string
+    id:
+      type: string
+    retry:
+      type: integer
+      minimum: 0
```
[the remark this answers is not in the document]
mikekistler: I don't understand. Do we expect each item in a `text/event-stream` to be an array?

src/oas.md
mikekistler: I'm not objecting to anything currently here. There is nothing wrong with the current wording except that I think it does not explain the purpose if `itemSchema` fully. To say that `itemSchema` "supports processing each item as it is read from the stream" is accurate but does not distinguish it from `schema`. I think the essential difference that that `itemSchema` indicates that items are _expected_ to be processed as they are received and the validity of any item, preceding or following, does not change that.

hudlow: > Unlike `schema`, which is applied to the **complete content**...

Under this proposal, OpenAPI:
1. Defines what it means for `schema` to be applied to the whole of the content in a message (represented as a JSON array).
2. Defines what it means for `itemSchema` to be applied to any individual item in a sequence within a message (represented in JSON without an array wrapper representing the sequence itself).

And **does not**:
1. Define what it means for a `schema` to be applied to a subset of the items in a sequence within a message (represented as a JSON array or any other way).
2. Define what it means for _some part_ of `schema` to an individual item or any subset of the items in a sequence within a message (represented in JSON with or without an array wrapper representing the sequence itself or any other way).

If an implementor wants to unpack `schema` and some kind of reasoning to it to determine that it knows how to safely apply part of it (say, its `items` keyword) to a part of a sequence (e.g., an individual item), this proposal doesn't really go out of its way to stop them from doing that but it certainly doesn't suggest it's a good idea or give any hints as to how to do it.

`itemSchema` is a way for OpenAPI to give implementors a clear, safe way to apply a schema to individual items in a sequence.

src/oas.md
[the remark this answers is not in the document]
handrews: 🤦 No.  That was left over from the previous iteration of this PR where it would have been correct.  I'll fix it.

src/oas.md
handrews: @mikekistler if adding the word "expected" is what's holding up approval, sure, go ahead and add it where you think it helps.  I don't understand the distinction you're making but I don't see any problem with putting that word in somewhere if that's what you're trying to get at.  (I'm still not sure TBH).

handrews: @mikekistler coming back to this, I still do not understand your objection.  You say:

> To say that `itemSchema` "supports processing each item as it is read from the stream" is accurate but does not distinguish it from `schema`. 

`schema` does not support processing as a stream.  That is _the entire reason this PR exists at all._  If `schema` supported this, we wouldn't need a change here.

handrews: @mikekistler we don't want to require processing each item individually for `itemSchema`.  I expect that will be common, and is the easiest option.  But implementations might choose to process them in small batches.  Or even not as a stream (although why they would do that I don't know). The point is that `itemSchema` gives implementations options that they do not have with `schema`.

handrews:
The force-push is to resolve conflicts with #4562.  The only conflicts involved git not knowing what order to put the added sections in, as both PRs added in the same place.  No text within any commit on this PR was changed.

mikekistler (approved):
Looks good. 👍

handrews:
From the TDC call today, @mikekistler and @hudlow have concerns about how much of an expectation for users the OAS sets, which @hudlow notes is a larger topic than just this one PR.  They will follow-on with a discussion (#4570) and additional PR as needed.  The consensus of the call was that this PR as written does not prevent applications from offering various additional guarantees, it just does not mandate any of them.  #4570 will address whether an additional mandate is required here.

handrews:
Had to fix the `format` example (new commit) as @hudlow noted that the format registry does not apply `float` to strings, and I don't have the bandwidth to debate changing it.  Use `int64` instead.

The subsequent force-push just rebases onto v3.2-dev with the latest syntax highlighting included, which ensures the build succeeds without warnings.

@mikekistler my apologies but you'll have to hit approve again.

miqui (approved):

handrews:
OK since the change that reset @mikekistler 's approval was trivial (and easily fixed in the unlikely event that I somehow messed it up), I'm going to go ahead and merge this so I can build on it in subsequent PRs.

PR #4556, opened 2025-04-29
v3.2: `$self` field (Alternative Approach)

handrews:
This is an alternative to PR #4389.  The only functional difference is that `$self` is required to be absolute (a change that @karenetheridge and @mikekistler are debating).

The main difference is that I drastically reduced the normative specification text and instead gave examples of all of the different ways to establish a base URI.  This is probably more understandable than trying to explain the options in prose.

@notEthan I also dumped a lot of the use cases as I think I figured out a more concise way to get the point across.

This adds `$self` as a way for a document to define its own URI for use in reference targets, and as the base URI for relative URI references in the document.

This does not impact the resolution of relative API URLs.


handrews:
After staring at the giant examples section a lot, I decided it works better as an appendix.

lornajane:
This is the most readable of the lot, and I think I "get" the whole thing a lot more now. Thank you for doing this work and all the iterations!

Minor points/suggestions from me (I'm not ready to approve but I've only got small questions):
 - in the proposal we called it `self` but it looks like now it is `$self`. We have other URI fields with plain-text names, why does this one need the `$`? Since it's being added at the top level, we can be confident of not having name clashes. (why isn't it in `info` with the other document information like license? Am I asking too many questions?)
 - I'd like to propose we move the appendix to the learn site and link to it. Is the learn site really good/mature/stable enough to link to from a spec file? I think it is, and the information would be better there (but needs to be very well signposted since we haven't done this for other content).

lornajane:
Good discussion in the meeting:
 - content does need to go to the learn site, but we're not going to block this change for it
 - `$self` makes sense because it's like `$ref` in being "special" and related to how references work

karenetheridge:
I am in agreement with the shuffling of the examples to the learn site rather than in the main spec, but not with the change from relative to absolute uri.

mikeschinkel:
@handrews — A big +1 from me for including examples in the spec. I would prefer them to be in-line so a reader does not have to jump around to follow them, but if those examples need to be in an appendix that is better than not having any or having a link to someplace that might eventually 404. 

My biggest pet peeve about specs in general is that they do not include examples which make most specs all-but-inaccessible to those like me who need examples to understand the prose. 

P.S. Moving to the learn site only works IMO if there is a guarantee that the learn site will never change those examples nor take them down which, I believe, is impossible to guarantee. FTW.

ralfhandl:
Relative vs. absolute `$self` doesn't seem to make much of a difference. 

- A relative `$ref` is now relative to the retrieval URL. Making it relative to a relative `$self` which is itself relative to the retrieval URL only adds the convenience of making the `$ref` value shorter. It does not add a feature. 

- A relative `$ref` that is relative to an absolute `$self` also only adds the convenience of making the `$ref` value shorter (and relative instead of absolute). It does not add a feature.

Seems the only feature added by `$self` is an in-document identifier for an OpenAPI description, and that feature seems to benefit from an absolute URI value. 

What am I missing?

src/oas.md
```diff
@@ -4886,3 +4920,169 @@ components:
...
+        bar:
+          $ref: bar
+    Bar:
+      $id: https://example.com/api/schemas/bar
+      type: string
+```
+
+In this example, the retrieval URIs are irrelevant because both documents define `$self`.
+
+For the relative `$ref` in the first document, it is resolved against `$self` to produce `https://example.com/shared/foo#/components/requestBodies/Foo`.
```
mikekistler, line 4974: This did not look right to me, but I am far from an expert on URI resolution, so I had GitHub Copilot create a Jupyter notebook to implement the URI resolution logic in RFC 3986 Section 5.2. According to this program, the target URI is actually "https://example.com/api/shared#/components/requestBodies/Foo".

Here's a link to the code if you want to have a look.

https://github.com/mikekistler/UriResolution

The notebook includes 42 tests that it extracted from the RFC and these all pass, I have some confidence that it is working correctly.

mikekistler (commented):
I think there may be some errors in the examples of URI resolution in Appendix G. I pointed out two and I think the rest should probably be checked for accuracy before we merge this.

src/oas.md
```diff
@@ -4886,3 +4920,169 @@ components:
...
+The portion of that URI before the '#' matches the `$self` of the second document, so the reference target is resolved to `#/components/requestBodies/Foo` in that second document.
+
+In that document, the `$ref` in the Request Body Object is resolved using that document's `$self` as the base URI, producing `https://example.com/schemas/foo`.
```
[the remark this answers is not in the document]
handrews: @mikekistler should be fixed.  Really, "these look wrong" would have been sufficient here.  The errors were just because I reworked the various paths at some point to make the examles cover more cases, and some things didn't get updated post-rework.  Nothing obscure was going on.

Although it sounds like you produced something useful beyond this which is cool!

karenetheridge:
Another reason to allow `$self` to be relative is that it is consistent with JSON Schema, where `$id` can be relative, so it would be easier to reason through how `$ref` resolution works in a document and apply RFC3986 consistently if both keywords worked the same way.

I see no gain from forcing `$self` to be absolute; it just restricts what users and implementations can do with the keyword.  We could just as easily say that `$self` is not needed at all, because the user can always supply the base URI via the retrieval URI.  Our goal should be to open up usecases, not restrict them.

src/oas.md
ralfhandl: Using `node` interactively produces
~~~pwsh
> new URL("./shared/foo#/components/requestBodies/Foo","https://example.com/openapi")
URL {
  href: 'https://example.com/shared/foo#/components/requestBodies/Foo',
  origin: 'https://example.com',
  protocol: 'https:',
  username: '',
  password: '',
  host: 'example.com',
  hostname: 'example.com',
  port: '',
  pathname: '/shared/foo',
  search: '',
  searchParams: URLSearchParams {},
  hash: '#/components/requestBodies/Foo'
}
~~~
which is exactly what the current PR text says.

Abstaining from debugging AI-generated code...

ralfhandl:
> `$self` is not needed at all, because the user can always supply the base URI via the retrieval URI

The main feature `$self` adds is an in-document identifier allowing to determine whether documents retrieved from different URIs or locally stored without their retrieval URI are actually the same and not just textually identical by chance.

Using `$self` with a relative URI means opting out its main feature and can be considered an abuse of `$self`, which would be prevented by requiring an absolute URI value.

src/oas.md
mikekistler: It looks like I passed the wrong values into my UriResolution tool. Doing my best to be careful now:

image

- "the relative `$ref` in the first document" is "shared/foo#/components/requestBodies/Foo", and
- `$self` (in the first document) is "https://example.com/api/openapi"

Using the correct values, I now get

image

I did check this against `node` and I think it gives the same result.

src/oas.md
```diff
@@ -354,6 +384,8 @@ This is the root object of the [OpenAPI Description](#openapi-description).
 
 This object MAY be extended with [Specification Extensions](#specification-extensions).
 
+Implementations MAY choose to support referencing OpenAPI Documents that contain `$self` by another URI such as the retrieval URI, however this behavior is not interoperable and relying on it is NOT RECOMMENDED.
```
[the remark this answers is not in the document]
handrews: @ralfhandl As stated in RFC2119, "SHOULD" and "RECOMMENDED" are equivalent, as are their negations.  So anything that is NOT RECOMMENDED is something that you SHOULD NOT do, and therefore you SHOULD do the other thing (in this case, use the `$self` URI as your reference target).

I can add a sentence making that explicit if you think it is needed.

-----

This came from a discussion with @karenetheridge about other ambiguities with `$id`.  Let's say that you have enabled auto-retrieval of references, treating them as URLs (because you have some reason to believe that will work, like the OAD author said so, out-of-band).  If you successfully retrieve a referenced document from X because one `$ref` used X, and it has a `$self` URI of Y... you've already retrieved it from X.  Can you keep referencing it as X or do you _have_ to reference it as Y, even though you know X works because you just got it there?

This is a weird case.  With JSON Schema, we made a mistake and didn't talk about it, as @karenetheridge pointed out.  But intuitively, most people seem to think that if you can use X successfully, and particularly if you _already_ used X as the retrieval UR***L*** (actually retrieved, not just supplied as the "yeah if you retrieved this you woudl have gotten it from X" simulated retrieval UR***I***) once and it worked, then X should keep working.

You can come up with a similar situation where the "other" URI is from an encapsulated entity (`multipart/related` example) or even an application-supplied default.  These get harder and harder to contrive, but it's possible.

I don't want to require (MUST) that implementations support these messy situations.  Implementations really SHOULD NOT perform unrestricted retrieval automatically as it's a security nightmare.  To say nothing of the weird contortions needed to justify the other possible sources.

But if you are supporting retrieval, you can easily get into this situation where someone uses a retrieval URL that does not match `$self`, and it works, and you have to decide whether to keep letting it work or not.

This sentence allows implementations to keep letting that work if they want to, without requiring it.  Given the history of `$ref`, particularly in 2.0 (and arguably 3.0) where it is strictly a URL, many tools almost certainly behave this way already.  This sentence gives them the flexibility to leave that in place (and keep compatibility for their existing 3.1 customers) while adding the new 3.2 functionality.

Does that help?  Again, I can tack on "Therefore, for interoperability, OAD authors SHOULD _(or even MUST?  Techncially it's not interoperable otherwise?)_ ensure their reference target URIs align with `$self`." or something to that effect if you think it is needed.

karenetheridge:
> Using $self with a relative URI means opting out its main feature and can be considered an abuse of $self

Is there an example (for the learn site, appendix or wherever) that you can construct that would demonstrate this issue, as a way of cautioning against it?

lornajane:
After many discussions about relative vs absolute, we agreed in TDC this week that supporting both would make a more open feature and since the retrieval URL is already used in other places, the additional burden for tools maintainers is not huge.

handrews:
I have updated this with the outcome of today's TDC call.  I have also:

* @ralfhandl further clarified the consequences of using alternative URIs
* Added an example (a random UUID URN) for the application-specific default option
* Added an example of relative `$self` and `$id`

The new commits were added in one push, and then I force-pushed the rebased commits unchanged to pick up the latest syntax highlighting that supports this change building without warnings.

handrews:
hmm... I thought I was only adding two new commits, looks like maybe I had not pushed some other local fixes?  My apologies for any confusion, just be aware that I have never edited commits once pushed, so if a commit looks familiar it's the same as it was.

src/oas.md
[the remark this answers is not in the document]
ralfhandl: Thanks for the explanation, that is what I _guessed_ the text wants to express. Please add the explicit positive instruction "OAD authors SHOULD...".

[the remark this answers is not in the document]
handrews: @ralfhandl already done https://github.com/OAI/OpenAPI-Specification/pull/4556/files#diff-a662b9afa95a60c2a67cc48014a8d5e00fa6ab2bd5ba2a54531674a5dd398c73R390

handrews:
One more new bugfix commit plus an unmodified rebase force-push.

src/oas.md
```diff
@@ -5161,3 +5198,221 @@ components:
...
+The portion of that URI before the '#' matches the `$self` of the second document, so the reference target is resolved to `#/components/requestBodies/Foo` in that second document.
+
+In that document, the `$ref` in the Request Body Object is resolved using that document's `$self` as the base URI, producing `https://example.com/api/schemas/foo`.
+This matches the `$id` at `#/components/schemas/Foo/$id` so it points to that Schema Object.
+That Schema Object has a subschema with `$ref: bar`, which is resolved against the `$id` to produce `https://example.com/api/schemas/bar`, which matches the `$id` at `#/components/schemas/Bar/$id`.
+
+Note that referring to a schema with a JSON Pointer that crosses a Schema Object with a `$id` [is not interoperable](https://www.ietf.org/archive/id/draft-bhutton-json-schema-01.html#name-json-pointer-fragments-and-).
+The JSON Schema specification does not address the case of using a pointer _to_ a Schema Object containing an `$id` without crossing into that Schema Object.
+Therefore it is RECOMMENDED that OAD authors use `$id` values to reference such schemas rather than JSON Pointers.
```
ralfhandl, line 5263: Try to formulate this as a positive requirement:

```suggestion
Note that a Schema Object with a `$id` SHOULD be referenced by its `$id` value because the JSON Schema specification does not address the case of using a [JSON Pointer to a Schema Object containing an `$id`](https://www.ietf.org/archive/id/draft-bhutton-json-schema-01.html#name-json-pointer-fragments-and-).
```

src/oas.md
```diff
@@ -5161,3 +5198,221 @@ components:
...
+
+Note also that it is impossible for the reference at `#/components/schemas/Foo/properties/bar/$ref` to reference the schema at `#/components/schemas/Bar` using a JSON Pointer, as the JSON Pointer would be resolved relative to `https://example.com/api/schemas/foo`, not to the OpenAPI Document's base URI from `$self`.
+
```
ralfhandl, line 5266: Is this something that a lot of people try to do despite it being impossible? Maybe remove this paragraph.

```suggestion
```

src/oas.md
```diff
@@ -5161,3 +5198,221 @@ components:
...
+        application/json:
+          schema:
+            $ref: "#/components/api/schemas/Foo"
+  schemas:
+    Foo:
+      properties:
+        bar:
+          $ref: schemas/bar
+--boundary-example
+Content-Type: application/schema+json; schema=https://spec.openapis.org/oas/3.2/schema-base/YYYY-MM-DD
```
ralfhandl, line 5301: The `schema=...` media type parameter is surprising and doesn't seem to be defined by [JSON Schema](https://json-schema.org/draft/2020-12/json-schema-core#name-application-schemajson) - can we please remove it?

src/oas.md
```diff
@@ -5161,3 +5198,221 @@ components:
...
+The resulting reference to `https://example.com/schemas/bar` matches the `Content-Location` header of the second part, which according to RFC2557 allows the reference target to be located within the multipart archive.
+
+Similarly, the `url` field of the [External Documentation Object](#external-documentation-object) is resolved against the base URI from `Content-Location`, producing `https://example.com/api/docs.html` which matches the `Content-Location` of the third part.
+
+### Base URI From the Retrieval URI
+
+If no base URI is provided from either of the previous sources, the next source is the retrieval URI (RFC 3986 Section 5.1.3).
+
+For this example, assume that the YAML OpenAPI Document was retrieved from `https://example.com/api/openapis.yaml` and the JSON Schema document from `https://example.com/api/schemas/foo`
+
```
ralfhandl, line 5333: This paragraph seems to be completely redundant.
```suggestion
```

src/oas.md
```diff
@@ -5161,3 +5198,221 @@ components:
...
+      $id: /api/schemas/foo
+      properties:
+        bar:
+          $ref: bar
+    Bar:
+      $id: /api/schemas/bar
+      type: string
+```
+
+In this example, All of the `$self` and `$id` values are relative URI-references consisting of an absolute path.
```
ralfhandl, line 5416:
```suggestion
In this example, all of the `$self` and `$id` values are relative URI-references consisting of an absolute path.
```

src/oas.md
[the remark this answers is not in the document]
handrews: When you actually retrieved something, it's the retrieval URL.  When you supply it as a simulation of having retrieved it, it's more properly a retrieval URI, which is how RFC3986 states it.

So if we only use one, it should be retrieval URI, not retrieval URL.

src/oas.md
handrews: Yes, many people don't understand that the `$id` creates what's essentially a new scope, so I would prefer not to remove this.

src/oas.md
handrews: @ralfhandl that paraeter has always been part of JSON Schema.  I believe the situation in that draft is the result of an incomplete removal of the media type registration, which moved to [draft-ietf-httpapi-rest-api-mediatypes](https://www.ietf.org/archive/id/draft-ietf-httpapi-rest-api-mediatypes-04.html#name-identifying-a-schema-via-a-).  However, after draft-ietf-httpapi-rest-api-mediatypes-04, JSON Schema was split out, and due to the current JSON Schema team's abandonment of the IETF, has not be placed in a new draft.

But the `schema` media type parameter has been supported for over a decade and is correct, and this is the correct usage.

src/oas.md
```diff
@@ -388,6 +420,9 @@ This is the root object of the [OpenAPI Description](#openapi-description).
 
 This object MAY be extended with [Specification Extensions](#specification-extensions).
 
+Implementations MAY choose to support referencing OpenAPI Documents that contain `$self` by another URI such as the retrieval URI, however this behavior is not interoperable and relying on it is NOT RECOMMENDED.
```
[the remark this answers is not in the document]
handrews: I see your point but this reworking doesn't quite feel right, and also as noted elsewhere, we should standardize on retrieval URI rather than URL.  I will rework it.

src/oas.md
handrews: I don't think your restatement quite captures everything, but I will look at it and see what I can do.

handrews:
@ralfhandl I have tried to address everything that I did not accept as a commit.

* Everything is now "retrieval URI"
* I reworked the crossing-`$id` text and dropped the mention of the unclear "point too but not across `$id`" case; I think @karenetheridge had wanted that included?  But it's tricky to word without just being more confusing, and idk how much we should try to compensate for JSON Schema's shortcomings here
* I added "only" to the weird JSON Pointer case that you asked if people actually do, because I agree it read strangely, and I think the new wording is much more clear

Regarding the `schema` media type parameter, while that usage is the most correct usage, it is a bit confusing because of the need to put YYYY-MM-DD in the URI, so I might be willing to drop it on those grounds.  Otherwise, we'll need to update that to the correct date so people don't copy-paste the placeholder.

handrews:
@ralfhandl went ahead and removed the parameter because of the YYYY-MM-DD problem.  Which makes me sad because I would rather promote the long-established correct use of the media type.

src/oas.md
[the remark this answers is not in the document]
ralfhandl: Fine with me, as long as it is consistent 😎

karenetheridge:
> because of the YYYY-MM-DD problem

I think we've talked about the possibility of having `.../latest` aliases in the spec tree?  Not as a schema with `/latest` in its $id, but just a URL that happens to map to the latest version.  I think that would still be a valid URI in a media type? it's not canonical but the URL does resolve to a real schema.

handrews:
@karenetheridge Yeah that's issue #4152, but the GitHub Pages/Jekyll/Liquid tech stack is in the way.  So I don't want to hold this PR up waiting for that.  If/when we fix that mess we can always add the parameter back.  Or argue about adding it back ;-P

@karenetheridge are you good with the PR as it stands now?

src/oas.md
```diff
@@ -5161,3 +5198,218 @@ components:
...
+This matches the `$id` at `#/components/schemas/Foo/$id` so it points to that Schema Object.
+That Schema Object has a subschema with `$ref: bar`, which is resolved against the `$id` to produce `https://example.com/api/schemas/bar`, which matches the `$id` at `#/components/schemas/Bar/$id`.
+
+To guarantee interoperability, Schema Objects containing an `$id`, or that are under a schema containing an `$id`, MUST be referenced by the nearest such `$id` for the non-fragment part of the reference.
+As the JSON Schema specification notes, using a base URI other than the nearest `$id` and crossing that `$id` with a JSON Pointer fragment [is not interoperable](https://www.ietf.org/archive/id/draft-bhutton-json-schema-01.html#name-json-pointer-fragments-and-).
+
+Note also that it is impossible for the reference at `#/components/schemas/Foo/properties/bar/$ref` to reference the schema at `#/components/schemas/Bar` using _only_ a JSON Pointer fragment, as the JSON Pointer would be resolved relative to `https://example.com/api/schemas/foo`, not to the OpenAPI Document's base URI from `$self`.
```
lornajane, line 5264: Wouldn't this cause quite a lot of 3.1 descriptions to be broken in 3.2? Or only if the `$self` field is actually used?

handrews: @lornajane this is not new- `$id` says that the Schema Object that contains it is the root of a resource with the given base URI, which means that fragment references stay within that resource (resolved relative to that base, and not to the base of the overall document- the distinction between document and resource is discussed in-depth in the JSON Schema specification.

karenetheridge (approved):

mikekistler (approved):
Looks good. 👍

PR #4559, opened 2025-04-29
v3.2 Respect `encoding` in all Media Type Objects

handrews:
There is no reason to restrict use of the `encoding` field of the Media Type Object to Request Body Objects only.  Not only does this exclude legitimate use cases of returning multipart content (particularly as we fix the support for `multipart/mixed`), it also complicates implementation by requiring the handling of the Media Type Object to be context-aware.

This is a rare change that both simplifies implementation and adds functionality.


lornajane (approved):
This definitely needs to make it into the release notes! Tools will have support for encoding, but may have actively removed support for that outside of request bodies.

mikekistler (approved):
Looks good! 👍

handrews:
For posterity, the use case for this most recently came up in an API that provides a [streaming multipart json response](https://github.com/OAI/OpenAPI-Specification/discussions/4171#discussioncomment-12938067).  Somewhere I know this "request body only" restriction came up in an older issue but I cannot find it now.

PR #4573, opened 2025-05-11
v3.2: Port of 3.1.2 changes to 3.2

handrews:
This gathers all of the commits from 3.1.2 that need to go into 3.2.0, including some typos and errors that were getting confusing to keep running across on the 3.2 branch.  Many of these changes came from @char0n .

The only notable merge conflicts were that I had to rework the changes based on @mkistler 's feedback about the Encoding Object examples to fit with the recent changes around Encoding Objects.  Otherwise things either applied cleanly or had obvious merge fixes.

One fix that looks like it could apply (changing an ABNF rule name from `char` to `CHAR`) does not actually apply and was left out, because we updated the RFC reference in 3.2 and the newer RFC uses `char`.


ralfhandl (approved):

mikekistler (approved):
Looks good! 👍

PR #4576, opened 2025-05-13
v3.1.2: Root XML element name comes from component name

### LINKED ISSUE #1435: xml property and root schemas
This is a question about the specification, and the intent of the following description in the [Schema Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#fixed-fields-20), relating to the `xml` property:

> This MAY be used only on properties schemas. It has no effect on root schemas. Adds additional metadata to describe the XML representation of this property.

This has the effect of banning the construct

```yaml
components:
  schemas:
    mySchema:
      type: object
      properties:
        a:
          type: string
      xml:
        name: xmlSchema
```

As a point-of-reference only, both `swagger-ui` and `swagger-editor` appear to honour this (useful) construct.

Is it possible to align the specification with real-world use for this case, and possibly move the restriction on `xml` properties on root schemas to those sub-properties (such as `attribute` and `wrapped`) where it seems to make sense?

handrews:
Fixes #1435 (as much as can be in a patch release)

 Clarifies that the name of the root XML element comes from the component name, which was shown in an example but was unclear due to the use of the obsolete OAS 2.0 terminology "model."

This does not change the restriction (in the `xml` field of the Schema Object) that the `xml` field only applies to property schemas (and not root schemas).

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -3447,7 +3447,7 @@ See examples for expected behavior.
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
-| <a name="xml-name"></a>name | `string` | Replaces the name of the element/attribute used for the described schema property. When defined within `items`, it will affect the name of the individual XML elements within the list. When defined alongside `type` being `"array"` (outside the `items`), it will affect the wrapping element if and only if `wrapped` is `true`. If `wrapped` is `false`, it will be ignored. |
+| <a name="xml-name"></a>name | `string` | Replaces the name of the element/attribute used for the described schema property. For the root XML element, the name comes from the [schema component](#components-schemas) name; for other elements or attributes, the name comes from the property name.  When defined within `items`, it will affect the name of the individual XML elements within the list. When defined alongside `type` being `"array"` (outside the `items`), it will affect the wrapping element if and only if `wrapped` is `true`. If `wrapped` is `false`, it will be ignored. |
 | <a name="xml-namespace"></a>namespace | `string` | The URI of the namespace definition. Value MUST be in the form of a non-relative URI. |
 | <a name="xml-prefix"></a>prefix | `string` | The prefix to be used for the [name](#xml-name). |
 | <a name="xml-attribute"></a>attribute | `boolean` | Declares whether the property definition translates to an attribute instead of an element. Default value is `false`. |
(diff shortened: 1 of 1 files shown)


src/oas.md
```diff
@@ -3447,7 +3447,7 @@ See examples for expected behavior.
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
-| <a name="xml-name"></a>name | `string` | Replaces the name of the element/attribute used for the described schema property. When defined within `items`, it will affect the name of the individual XML elements within the list. When defined alongside `type` being `"array"` (outside the `items`), it will affect the wrapping element if and only if `wrapped` is `true`. If `wrapped` is `false`, it will be ignored. |
+| <a name="xml-name"></a>name | `string` | Replaces the name of the element/attribute used for the described schema property. For the root XML element, the name comes from the [schema component](#components-schemas) name; for other elements or attributes, the name comes from the property name.  When defined within `items`, it will affect the name of the individual XML elements within the list. When defined alongside `type` being `"array"` (outside the `items`), it will affect the wrapping element if and only if `wrapped` is `true`. If `wrapped` is `false`, it will be ignored. |
```
ralfhandl, line 3450: Some implementations allow the `xml` fixed field in any Schema Object, including the Schema Object for the root XML element, and use its `name` value for the XML element name. If absent, the name of the XML element (or attribute) is the name of the name/value pair whose value is the Schema Object.

Which is straight-forward and what I would have expected, without the funny restriction to "property schemas", a term that is nowhere defined.

I'd not reinforce the surprising restriction to "property schemas" further, or at least add that implementations MAY allow it in any Schema Object, including the one for the XML root element.

ralfhandl (changes requested):

src/oas.md
handrews: When we port this to 3.2 I was planning on removing the "property schemas" restriction, as I assumed the `xml` field in a "root schema" should behave exactly as you say.

I'm not sure how to incorporate that here in 3.1.2, though.  If @OAI/tsc is OK with relaxing that (as a MAY) in a patch release, I'm happy to do so.

For now, let me try to reword this short of adding that MAY (we can add the MAY in a later PR if we can't sort that out immediately- I'd love to see it, TBH).

lornajane (approved):

PR #4577, opened 2025-05-13
v3.2: Only use YAML for examples

handrews:
IIRC, we discussed a while ago that we don't need both JSON and YAML examples.  And there are many examples now that are only shown in YAML.  This standardizes on YAML for all OAD examples.  In addition to cutting down on the ever-growing length of the spec, I find this more readable when there are a bunch of examples in a row, as you're not constantly skipping over duplicates.

We still use JSON for example payloads, and there is one example in Appendix F that discusses retrieving OADs by content type, so those continue to show JSON.

Also standardize on lowercase language names for fenced code blocks, as that was more common (I think the uppercase ones were mostly added by me).

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -152,14 +152,7 @@ Occasionally, non-backwards compatible changes may be made in `minor` versions o
 ### Format
 
 An OpenAPI Document that conforms to the OpenAPI Specification is itself a JSON object, which may be represented either in [[RFC8259|JSON]] or [[YAML|YAML]] format.
-
-For example, if a field has an array value, the JSON array representation will be used:
-
-```json
-{
-  "field": [1, 2, 3]
-}
-```
+Examples in this specification will be shown in YAML for brevity.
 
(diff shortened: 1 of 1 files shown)


ralfhandl:
> we don't need both JSON and YAML examples

Correct, the same example in both JSON and YAML is redundant.

However I'd like to have some examples in JSON only, and some in YAML only, to avoid the impression that YAML is preferred or even mandatory.

ralfhandl (changes requested):
> we don't need both JSON and YAML examples

Correct, the same example in both JSON and YAML is redundant.

However I'd like to have some examples in JSON only, and some in YAML only, to avoid the impression that YAML is preferred or even mandatory.

handrews:
@ralfhandl if later on you want to switch some to JSON to balance it out, I definitely will not object.  But I appreciate being able to reduce the duplication for now as a first step. 🙏

mikekistler (approved):
Looks good! 👍

PR #4579, opened 2025-05-14
Allow IRIs for XML namespaces

handrews:
XML allows IRIs for namespaces, and compares them without any encoding or decoding.  This means that our requirement that namespaces are URIs will cause comparisons with other tools that use IRIs (without mapping them down to URIs) to fail.

This change relaxes the description and allows the use of unencoded IRIs, which allows us to remove a cumbersome warning.


ralfhandl (approved):

mikekistler (approved):
Looks good! 👍

PR #4585, opened 2025-05-15
feat: adds uint16, 32 and 64 formats

### LINKED ISSUE #4564: Format extension registry: unsigned integer types (`uint16`, `uint32`, `uint64`)
I’m proposing to add the following unsigned integer types:

value | description | JSON data type
--  | -- | --
`uint16` | unsigned 16-bit integer | number
`uint32` | unsigned 32-bit integer | number
`uint64` | unsigned 64-bit integer | number, string (same caveats as `int64`)

The `uint32` and `uint64` formats are already in use by `gnostic`.

This is effectively a repeat of https://github.com/OAI/OpenAPI-Specification/issues/2617, now that there is an extension registry, as it looks like only `uint8` was added when that registry was created.

baywet:
fixes #4564

### DIFF
--- a/registries/_format/uint16.md
+++ b/registries/_format/uint16.md
@@ -1,0 +1,15 @@
+---
+owner: baywet
+issue: 4564
+description: unsigned 16-bit integer
+base_type: number
+layout: default
+source: https://spec.openapis.org/oas/latest.html#data-types
+source_label: OAS
+---
+
+{% capture summary %}
+The `{{page.slug}}` format represents an unsigned 16-bit integer, with the range 0 to 65535.
+{% endcapture %}
+
+{% include format-entry.md summary=summary %}
--- a/registries/_format/uint32.md
+++ b/registries/_format/uint32.md
@@ -1,0 +1,15 @@
+---
+owner: baywet
+issue: 4564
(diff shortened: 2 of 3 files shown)


mikekistler (approved):
Looks good! 👍

lornajane (approved):
Thanks for adding these

karenetheridge:
This was created and merged so quickly I didn't have a chance to flag an issue with this...

There is no need for an integer type to be represented as a string in JSON (or YAML, or whatever other serialization format is being used).  If the language/implementation/architecture cannot interpret the value as a native integer, then it can decode it into whatever makes sense for that architecture, and the format validation can then know that it is intended to represent a very large number for the purpose of its validation.

Allowing this value to be a string, for the purpose of encoding into JSON/YAML/etc, is not the right level of abstraction.

baywet:
So then we should remove the string JSON data type from:
- uint64
- int64
- decimal
- decimal128

?

karenetheridge:
Yes, I don't see why we need it, and it only adds complication.

karenetheridge:
After a good conversation with @hudlow I have realized I was wrong here! I will open a new PR to correct this, with some more explanation.

Issue #3725, opened 2024-04-20
Support `multipart/byteranges` for 206 responses


handrews:
[RFC 9110 §14.6](https://www.rfc-editor.org/rfc/rfc9110.html#name-media-type-multipart-bytera) defines the `multipart/byteranges` media type for use in 206 (Partial Content) responses.  We should see what, if anything, is needed to support it.  Probably something involving the Encoding Object.

Some clients apparently used a not-entirely-compatible `multipart/x-byteranges`, but I'd argue that RFC 9110 carries enough weight that we should support `multiparte/byteranges` even if the old `x-` type is currently more broadly used.

lornajane:
This is implemented in 3.2, so I'll close the issue.

Issue #3721, opened 2024-04-19
Support for `multipart/mixed` (and possibly better `multipart/*` support in general)


handrews:
We occasionally see folks trying to use OAS with `multipart/mixed`, which does not support it.  But it would not be hard to define a mapping along the lines of `multipart/form-data` but with more of an array model.  See also discussion #2599.  I think @jeremyfiel also has experience with this.

_[EDIT: See also [`multipart/x-mixed-replace`](https://html.spec.whatwg.org/multipage/iana.html#multipart/x-mixed-replace) which AFAICT does not have a usable specification as WHATWG explicitly states in that section that there spec "describes processing rules for web browsers" and relegates everything else to `multipart/mixed`... I guess non-web-browsers aren't supposed to process it? 😒 ]_

handrews:
It's also worth noting that when uploading multiple files for a form field with `multipart/form-data`, RFC 7578 requires having [multiple parts with the same name](https://datatracker.ietf.org/doc/html/rfc7578#section-4.3), which can't be modeled by the current "treat it as a JSON object" approach.

jeremyfiel:
Another multipart option missing is the ability to send multiple files in a single form field. 

It should be a form-data payload with a nested multipart body with individual parts for each file. There is no way to define a nested multipart payload. 


This is as far as I can get with it.  But no way to define the nested body parts if I were to send multiple file types  

```yaml
requestBody:
  content:
    'multipart/form-data':
      schema:
       name:
         type: string
       file:
         type: string
         format: binary
      encoding:
        name:
          headers:
            content-disposition:
              schema:
                type: string
          contentType: 'application/json'
        file:
          headers:
            content-disposition:
              schema:
                type: string
          contentType: 'multipart/form-data'
```

handrews:
@jeremyfiel I know that `multipart/mixed` and similar MIME-derived formats explicitly allow nesting.

I noticed that the most recent `multipart/form-data` RFC, RFC 7578, [deprecates the nesting in favor of multiple parts with the same name](https://datatracker.ietf.org/doc/html/rfc7578#section-4.3).  According to the RFC's appendix, [the nested approach wasn't known to be implemented](https://datatracker.ietf.org/doc/html/rfc7578#appendix-A) but I assume you have an implementation that does it.

jeremyfiel:
A big missing piece from the `multipart/mixed` RFC is the ability to define the body parts in the same way form-data uses `content-disposition` header.  

[...]

PR #4589, opened 2025-05-15
v3.2: Support ordered multipart including streaming

handrews:
Fixes:
* #3721 (`multipart/mixed` in general)
* #3725 (`multipart/byteranges`)
* https://github.com/OAI/OpenAPI-Specification/discussions/4171#discussioncomment-12938067 (streaming `application/json` with `multipart/mixed`)

This adds support for all `multipart` media types that do not have named parts, including support for streaming such media types. Note that `multipart/mixed` defines the basic processing rules for all `multipart` types, and implementations that encounter unrecognized `multipart` subtypes are required to process them as `multipart/mixed`.  Therefore support for `multipart/mixed` addresses all other subtypes to some degree.

This builds on the recent support for sequential media types:

* `multipart/mixed` and similar meet the definition for a sequential media type, requiring it to be modeled as an array.  This does use an expansive definition of "repeating the same structure", where the structure is literally any content with a media type.
* As a sequential media type, it also supports `itemSchema`
* Adding a parallel `itemEncoding` is the obvious solution to `multipart/mixed` streams requiring an Encoding Object
* We have regularly received requests to support truly mixed `multipart/mixed` payloads, and previously claimed such support from 3.0.0 onwards, without actually supporting it. Adding `prefixEncoding` along with `itemEncoding` supports this use case with a clear parallel to `prefixItems`, which is the schema construct needed to support this case.
* There is no need for a `prefixSchema` field because the streaming use case requires a repetition of the same schema for each item. Therefore all mixed use cases can use `schema` and `prefixItems`

We do not seem to run tests on the 3.2 schemas, and I couldn't quickly figure out how to add that, so we should do that separately and include coverage for this and other new fields.

Also paging @thecheatah, @jeremyfiel

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -107,7 +107,7 @@ Some examples of sequential media types (including some that are not IANA-regist
 In the first three above, the repeating structure is any [JSON value](https://tools.ietf.org/html/rfc8259#section-3).
 The fourth repeats `application/geo+json`-structured values, while `text/event-stream` repeats a custom text format related to Server-Sent Events.
 The final media type listed above, `multipart/mixed`, provides an ordered list of documents of any media type, and is sometimes streamed.
-Note that while `multipart` formats technically allow a preamble and an epilogue, the RFC directs that they are to be ignored, making the effectively comments, and this specification does not model them.
+Note that while `multipart` formats technically allow a preamble and an epilogue, the RFC directs that they are to be ignored, making them effectively comments, and this specification does not model them.
 
 Implementations MUST support mapping sequential media types into the JSON Schema data model by treating them as if the values were in an array in the same order.
 
--- a/src/oas.md
+++ b/src/oas.md
@@ -101,14 +101,17 @@ Some examples of sequential media types (including some that are not IANA-regist
   application/json-seq
   application/geo+json-seq
   text/event-stream
+  multipart/mixed
 ```
 
 In the first three above, the repeating structure is any [JSON value](https://tools.ietf.org/html/rfc8259#section-3).
-The fourth repeats `application/geo+json`-structured values, while the last repeats a custom text format related to Server-Sent Events.
+The fourth repeats `application/geo+json`-structured values, while `text/event-stream` repeats a custom text format related to Server-Sent Events.
+The final media type listed above, `multipart/mixed`, provides an ordered list of documents of any media type, and is sometimes streamed.
 
 Implementations MUST support mapping sequential media types into the JSON Schema data model by treating them as if the values were in an array in the same order.
 
 See [Complete vs Streaming Content](#complete-vs-streaming-content) for more information on handling sequential media types in a streaming context, including special considerations for `text/event-stream` content.
+For `multipart` types, see also [Encoding By Position](#encoding-by-position).
 
 #### Media Type Registry
 
@@ -1253,7 +1256,9 @@ See [Working With Examples](#working-with-examples) for further guidance regardi
 | <a name="media-type-item-schema"></a>itemSchema | [Schema Object](#schema-object) | A schema describing each item within a [sequential media type](#sequential-media-types). |
 | <a name="media-type-example"></a>example | Any | Example of the media type; see [Working With Examples](#working-with-examples). |
 | <a name="media-type-examples"></a>examples | Map[ `string`, [Example Object](#example-object) \| [Reference Object](#reference-object)] | Examples of the media type; see [Working With Examples](#working-with-examples). |
-| <a name="media-type-encoding"></a>encoding | Map[`string`, [Encoding Object](#encoding-object)] | A map between a property name and its encoding information, as defined under [Encoding Usage and Restrictions](#encoding-usage-and-restrictions).  The `encoding` field SHALL only apply when the media type is `multipart` or `application/x-www-form-urlencoded`. If no Encoding Object is provided for a property, the behavior is determined by the default values documented for the Encoding Object. |
+| <a name="media-type-encoding"></a>encoding | Map[`string`, [Encoding Object](#encoding-object)] | A map between a property name and its encoding information, as defined under [Encoding By Name](#encoding-by-name).  The `encoding` field SHALL only apply when the media type is `multipart` or `application/x-www-form-urlencoded`. If no Encoding Object is provided for a property, the behavior is determined by the default values documented for the Encoding Object. This field MUST NOT be present if `prefixEncoding` or `itemEncoding` are present. |
+| <a name="media-type-prefix-encoding"></a>prefixEncoding | [[Encoding Object](#encoding-object)] | An array of positional encoding information, as defined under [Encoding By Position](#encoding-by-position).  The `prefixEncoding` field SHALL only apply when the media type is `multipart`. If no Encoding Object is provided for a property, the behavior is determined by the default values documented for the Encoding Object. This field MUST NOT be present if `encoding` is present. |
+| <a name="media-type-item-encoding"></a>itemEncoding | [Encoding Object](#encoding-object) | A single Encoding Object that provides encoding information for multiple array items, as defined under [Encoding By Position](#encoding-by-position). The `itemEncoding` field SHALL only apply when the media type is `multipart`. If no Encoding Object is provided for a property, the behavior is determined by the default values documented for the Encoding Object. This field MUST NOT be present if `encoding` is present. |
 
 This object MAY be extended with [Specification Extensions](#specification-extensions).
 
@@ -1273,7 +1278,8 @@ For this use case, `maxLength` MAY be implemented outside of regular JSON Schema
 
 ###### Streaming Sequential Media Types
 
-The `itemSchema` field is provided to support streaming use cases for sequential media types.
+The `itemSchema` field is provided to support streaming use cases for sequential media types, with `itemEncoding` as a corresponding encoding mechanism for streaming [positional `multipart` media types](#encoding-by-position).
+
 Unlike `schema`, which is applied to the complete content (treated as an array as described in the [sequential media types](#sequential-media-types) section), `itemSchema` MUST be applied to each item in the stream independently, which supports processing each item as it is read from the stream.
 
 Both `schema` and `itemSchema` MAY be used in the same Media Type Object.
@@ -1309,13 +1315,16 @@ properties:
 
 ##### Encoding Usage and Restrictions
 
-The `encoding` field defines how to map each [Encoding Object](#encoding-object) to a specific value in the data.
+The three encoding fields define how to map each [Encoding Object](#encoding object) to a specific value in the data.
+Each field has its own set of media types with which it can be used; for all other media types all three fields SHALL be ignored.
 
-To use the `encoding` field, a `schema` MUST exist, and the `encoding` field's  keys MUST exist in the schema as properties.
-Array properties MUST be handled by applying the given Encoding Object to one part per array item, each with the same `name`, as is recommended by [[?RFC7578]] [Section 4.3](https://www.rfc-editor.org/rfc/rfc7578.html#section-4.3) for supplying multiple values per form field.
-For all other value types for both top-level non-array properties and for values, including array values, within a top-level array, the Encoding Object MUST be applied to the entire value.
+###### Encoding By Name
 
 The behavior of the `encoding` field is designed to support web forms, and is therefore only defined for media types structured as name-value pairs that allow repeat values, most notably `application/x-www-form-urlencoded` and `multipart/form-data`.
+
+To use the `encoding` field, each key under the field MUST exist in the `schema` as a property.
+Array properties MUST be handled by applying the given Encoding Object to produce one encoded value per array item, each with the same `name`, as is recommended by [[?RFC7578]] [Section 4.3](https://www.rfc-editor.org/rfc/rfc7578.html#section-4.3) for supplying multiple values per form field.
+For all other value types for both top-level non-array properties and for values, including array values, within a top-level array, the Encoding Object MUST be applied to the entire value.
 The order of these name-value pairs in the target media type is implementation-defined.
 
 For `application/x-www-form-urlencoded`, the encoding keys MUST map to parameter names, with the values produced according to the rules of the [Encoding Object](#encoding-object).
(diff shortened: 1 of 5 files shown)


handrews:
@jeremyfiel aww... I was hoping no one would bring up nested multipart... 😵‍💫 

I think it would be hard to do that, because there isn't anywhere to put the nested Encoding Object.  I think we'd have to add `encoding`, `prefixEncoding`, and `itemEncoding` to the Encoding Object as well as the Media Type Object.  I'm a bit hesitant to do that, but we could talk about it at the Thursday call and I could submit it as a follow-up if it gains traction.

Alternatively, we could recommend trying that as an extension given that it adds significant complexity and is a rare case that is deprecated by the current RFC (I know that's small consolation when you're the "rare case" and built things in good faith using older RFCs when they were current).

The complexity is not just the recursive structure, but also that you are now correlating two separate trees of structure.

jeremyfiel:
I'm not entirely sure this is a correct statement to include `multipart/mixed`.  It is registered in the [IANA registry](https://www.iana.org/assignments/media-types/media-types.xhtml#multipart) and it does technically have an `envelope` with the `boundary` parameter. 

> #### Sequential Media Types
> Within this specification, a _sequential media type_ is defined as any media type that consists of a repeating structure, without any sort of header, footer, envelope, or other metadata in addition to the sequence.
> Some examples of sequential media types (including some that are not IANA-registered but are in common use) are:
> ```text
>   application/jsonl
>   application/x-ndjson
>   application/json-seq
>   application/geo+json-seq
>   text/event-stream
>   multipart/mixed

handrews:
@jeremyfiel the problem is that instead of just re-using the Media Type Object, we came up with the `contentType` field :-(

handrews:
@jeremyfiel That statement only says that some of the listed types are not registered.  `application/json-seq`, `application/geo+json-seq`, and `multipart/mixed` are all registered.

I decided not to get into the preamble and postamble of `multipart` because AFAICT they're supposed to be ignored and are there for historical purposes.  Media type parameters are not part of the actual media type content, and the boundaries in the content are no more (or  less) significant than the various differences in the three sequential JSON media type delimiters.

handrews:
@jeremyfiel I added some clarifications about the envelope/preamble/epilogue and the lack of nesting support.

src/oas.md
```diff
@@ -101,14 +101,18 @@ Some examples of sequential media types (including some that are not IANA-regist
...
   application/geo+json-seq
   text/event-stream
+  multipart/mixed
 ```
 
 In the first three above, the repeating structure is any [JSON value](https://tools.ietf.org/html/rfc8259#section-3).
-The fourth repeats `application/geo+json`-structured values, while the last repeats a custom text format related to Server-Sent Events.
+The fourth repeats `application/geo+json`-structured values, while `text/event-stream` repeats a custom text format related to Server-Sent Events.
+The final media type listed above, `multipart/mixed`, provides an ordered list of documents of any media type, and is sometimes streamed.
+Note that while `multipart` formats technically allow a preamble and an epilogue, the RFC directs that they are to be ignored, making the effectively comments, and this specification does not model them.
```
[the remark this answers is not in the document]
handrews: @thecheatah thanks, just added a commit to fix this.

src/oas.md
```diff
@@ -1324,15 +1334,29 @@ See [Encoding the `x-www-form-urlencoded` Media Type](#encoding-the-x-www-form-u
...
+
+The `itemEncoding` field can also be used with `itemSchema` to support streaming `multipart` content.
+
+###### Additional Encoding Approaches
+
+The `prefixEncoding` field can be used with any `multipart` content to require a fixed part order.
+This includes `multipart/form-data`, for which the Encoding Object's `headers` field MUST be used to provide the `Content-Disposition` and part name, as no property names exist to provide the names automatically.
+
+Prior versions of this specifications advised using the `name` [`Content-Disposition` parameter](https://www.iana.org/assignments/cont-disp/cont-disp.xhtml#cont-disp-2) of the `form-data` [`Content-Disposition` value](https://www.iana.org/assignments/cont-disp/cont-disp.xhtml#cont-disp-1) with `multipart` media types other than `multipart/form-data` in order to work around the limitations of the `encoding` field.
```
jeremyfiel, line 1354:
```suggestion
Prior versions of this specification advised using the `name` [`Content-Disposition` parameter](https://www.iana.org/assignments/cont-disp/cont-disp.xhtml#cont-disp-2) of the `form-data` [`Content-Disposition` value](https://www.iana.org/assignments/cont-disp/cont-disp.xhtml#cont-disp-1) with `multipart` media types other than `multipart/form-data` in order to work around the limitations of the `encoding` field.
```

src/oas.md
```diff
@@ -1869,6 +1896,97 @@ requestBody:
...
+###### Example: Ordered, Unnamed Multipart
+
+A `multipart/mixed` payload consisting of a JSON metadata document followed by an image which the metadata describes:
+
+```yaml
+multipart/mixed:
+  schema:
+    prefixItems:
+    - # default content type for objects
+      # is `application/json`type: object
```
jeremyfiel, line 1908:
```suggestion
      # is `application/json`
      type: object
```

jeremyfiel (commented):
LGTM!

src/schemas/validation/schema.yaml
```diff
@@ -533,9 +533,20 @@ $defs:
...
+        type: array
+        items:
+          $ref: '#/$defs/encoding'
+      itemEncoding:
+        $ref: '#/$defs/encoding'
     allOf:
-      - $ref: '#/$defs/specification-extensions'
       - $ref: '#/$defs/examples'
+      - $ref: '#/$defs/specification-extensions'
+      - dependentSchema:
```
jeremyfiel:
```suggestion
      - dependentSchemas:
```

typo

src/oas.md
```diff
@@ -1324,15 +1334,29 @@ See [Encoding the `x-www-form-urlencoded` Media Type](#encoding-the-x-www-form-u
...
-Implementations MAY choose to support the a `Conent-Disposition` of `form-data` with a `name` parameter in other `multipart` media types in order to use the `encoding` field with them, but this usage is unlikely to be supported by generic `multipart` implementations.
-
 See [Encoding `multipart` Media Types](#encoding-multipart-media-types) for further guidance and examples, both with and without the `encoding` field.
 
+###### Encoding By Position
+
+Most `multipart` media types, including `multipart/mixed` which defines the underlying rules for parsing all `multipart` types, do not have named parts.
+Data for these media types are modeled as an array, with one item per part, in order.
+
+To use the `prefixEncoding` and/or `itemEncoding` fields, either an array `schema` or `itemSchema` MUST be present.
```
```suggestion
To use the `prefixEncoding` and/or `itemEncoding` fields, either  `itemSchema` or an array `schema` MUST be present.
```
Reorder to make clear that „array“ only applies to `schema`?

src/oas.md
```diff
@@ -1661,6 +1686,8 @@ Determining how to handle a `type` value of `null` depends on how `null` values
 If `null` values are entirely omitted, then the `contentType` is irrelevant.
 See [Appendix B](#appendix-b-data-type-conversion) for a discussion of data type conversion options.
 
+It is not currently possible to model nested `multipart` media types.
```
ralfhandl, line 1734: Pity, my only multipart use case is that the multipart request consists of parts that are either of
- media type `application/http`, or
- media type `multipart/mixed` with parts of media type `application/http`,

see [Multipart Batch Format](https://docs.oasis-open.org/odata/odata/v4.02/csd01/part1-protocol/odata-v4.02-csd01-part1-protocol.html#MultipartBatchFormat).

I had assumed I can model that with an `itemSchema` that is `oneOf` an array of HTTP requests or a single HTTP request, but the corresponding Encoding Object may be difficult.

handrews: @ralfhandl It's telling that the two people to give the most comprehensive reviews of this PR to date both need nested `multipart`.  I do have thoughts on how to do this, and would be more than happy to consider it part of this [release-blocking functionality](https://github.com/OAI/OpenAPI-Specification/issues/4595) for 3.2.  It would involve adding `encoding`, `prefixEncoding`, and `itemEncoding` to the Encoding Object, but there are challenging edge cases and I don't want to drag this PR down with them.  I'd rather get this merged, submit a follow-on PR (which is impractical without merging this first), and then we can decide if the follow-on works.

Regarding `application/http`, the problem (to me) is not "how to use an Encoding Object with `application/http`" but "how to model `application/http` at all."  Can you point me to how you would model an `application/http` payload if it were _not_ in a `multipart` part?  I realize that would be a rather strange payload outside of `multipart`, but please humor me- feel free to open a new discussion on it if there's no clear answer.  If we know how to model it in general, we can look at how to model it as a `multipart` part.  Although that might have to wait for 3.3.

src/oas.md
handrews: I'll update this- you have extra spaces in your suggestion so I can't merge it as-is.

ralfhandl: Fixed 😊

handrews:
@ralfhandl I have fixed the sentence ordering, and also added a new section, **Encoding and `type`** _[see commits below- initial push failed and I didn't notice at first]_, that clarifies how to handle detecting the "schema type" that gets mentioned in many places but is never explained.  I suspect that originally the expectation was that the Schema Object under the `schema` field in the Media Type Object (adjacent to the Encoding Objects' parent `encoding` field) would be an inline schema with inline `properties` subschemas, or at most a single `$ref` directly under `schema`.

This is not realistic, so I think that setting some boundaries on whatis expected is required.  I stuck with requiring (MUST) only the most unambiguous scenario, although I considered including the search-order support for multi-valued `type` keywords as a MUST.  Or at least if it is two values and the second value is `"null"`.  But you can get into some weird corner cases when you are doing that so it felt better to banish it under the "MAY choose to implement more complex things."  Although I also considered pulling that part out of the MAY and making it a SHOULD on its own.  _Opinions here would be much appreciated._

This, btw, is a prerequisite for supporitng nested `multipart/mixed` as the problem becomes _much_ more complex the further down you go in nested objects and arrays.  But this is really needed already, whether we support nesting or not, so I decided to add it to this PR.

handrews:
@ralfhandl oops, push had failed and I hadn't noticed.  The section mentioned in my last comment is now actually added, sorry about that.

src/oas.md
```diff
@@ -1309,13 +1316,61 @@ properties:
...
+    Thing:
+      schema:
+        $ref: "#/components/Schemas/Thing"
+      encoding:
+        foo:
+          # The default `contentType` is `text/plain`
+```
+
+Implementations MAY attempt to handle more complex schema arrangements, in which case they MUST document what is handled and with what behavior.
+If they do, then `type` keywords that contain multiple values (e.g. `type: ["number", "nul"]`) SHOULD be handled by attempting to parse according to each type in the order provided, falling back to the next type until the list is exhausted.
```
```suggestion
If they do, then `type` keywords that contain multiple values (e.g. `type: ["number", "null"]`) SHOULD be handled by attempting to parse according to each type in the order provided.
```

Typo, and I am confused by the "falling back" part.

handrews:
@ralfhandl added more tests, merged your suggestions (and yeah "falling back" is probably too idiomatic to be used here)

handrews:
Also rebased to pick up the test-running changes.  No modifications in the rebase force-push

src/oas.md
jeremyfiel: @ralfhandl   is my understanding you have either a single `application/http` payload or a `multipart/mixed` payload?  But the multipart/mixed does not have a nested multipart document? 

```http
GET /pages HTTP/1.1
accept: application/http

GET /page/{page-id} HTTP/1.1
accept: application/json

{
  "pages": [{}, {}]
}
```

```http
POST /pages HTTP/1.1
content-type: multipart/mixed; boundary='my-boundary'

--my-boundary
content-type: application/http

GET /page/{page-id} HTTP/1.1
accept: application/json

{
  "pages": [{}, {}]
}
--my-boundary
content-type: application/http

PATCH /pages/{page-id} HTTP/1.1
content-type: application/json

{
   "pageTitle": "a new title"
}
--my-boundary
content-type: text/plain

this is a text file describing the two different http messages included in this request
--my-boundary--

```

if the multipart example is true, you can define it with OAS 3.2.x like this: 

```yaml
openapi: 3.2.0
info:
  title: multipart application/http
  version: 0.0.1
paths:
  '/pages':
     summary: a list of http pages
     requestBody:
      content:
        'multipart/mixed':
          schema:
            prefixItems:  // these can be in any order you define
            - type: string
            - type: string
            - type: string
            prefixEncoding:   // these can be in any order you define
            - contentType: application/http
            - contentType: application/http
            - contentType: text/plain

Comment (jeremyfiel) [reply src/oas.md:1734]: my example shows mixed GET and PATCH, this appears to be forbidden by [RFC9112](https://datatracker.ietf.org/doc/html/rfc9112#section-10.2-1)

but the OAS definition still applies.

Comment (handrews) [reply src/oas.md:1734]: @jeremyfiel I think @ralfhandl is talking about [OData's `multipart` batch format](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_MultipartBatchRequestBody).

Comment (jeremyfiel) [reply src/oas.md:1734]: ah ok. i see it has two distinct $batch options, which he eluded to. 

the second is indeed a nested multipart.  That's a bummer

Comment (handrews) [reply src/oas.md:1734]: @jeremyfiel I think I have figured out how to support nested `multipart`... we'll go through this PR on Thursday's call, and if there's time I can introduce the nested proposal.

Comment (handrews) [reply src/oas.md:1734]: **[EDIT: This is not as bad as I thought, see [next comment](https://github.com/OAI/OpenAPI-Specification/pull/4589#discussion_r2151429909)]**

@jeremyfiel @ralfhandl I'm realizing that there is a substantial limitation with ordered `multipart` where something like the OData batch format (ignoring the nested-ness for now) can have arbitrarily many parts of a few types, where the types can appear in any order.  Or even in a fixed order but not a fixed number of each media type.

Another example would be an RFC2557 web page archive, which might want to allow any number of HTML, CSS, JavaScript and/or image parts.  What we _could_ do is allow `contentMediaType` to be used to specify this.  We say that `contentMediaType` SHALL be ignored if it _conflicts_ with `contentType` (even the implicit default value of `contentType`), but we do not say anything about `contentMediaType` _narrowing_ the media type from `contentType`.  Which, if we set to `*/*`, would effectively delegate to `contentMediaType`.  It would look like this:

```yaml
multipart/related:
  schema:
    items:
      anyOf:
      - type: string
        contentMediaType: text/html
      - type: string
        contentMediaType: text/css
      - type: string
        contentMediaType: text/javascript
      - contentMediaType: image/*
  itemEncoding:
    contentType: */*
    headers:
      ContentType:
        $ref: '#/components/headers/requiredContentType'
```

The only wrinkle here is that now implementations would have to detect patterns like this `anyOf`, at least if (as with the `type` detection rules included in this PR) they are reasonably find-able from the relevant Schema Object.

This seems like a big limitation that we ought to address, and a plausible way to address it.  But how reasonable is it?  Could we include it and put some boundaries around just how much tools are expected to handle?  Should we instead add another keyword or mechanism of some sort?  Or just let it be for now?

handrews: Wait... maybe this isn't as bad as I thought, because `contentType` takes a comma-separated list.  So you could do:

```yaml
multipart/related:
  schema:
    items:
      anyOf:
      - type: string
      - $comment: Must allow anything to allow binary
  itemEncoding:
    contentType: text/html, text/css, text/javascript, image/*
    headers:
      ContentType:
        $ref: '#/components/headers/requiredContentType'
```

handrews:
The most recent commit improves one example based on [this comment](https://github.com/OAI/OpenAPI-Specification/pull/4589#discussion_r2151429909), and also adds some Security Considerations related to it.

src/oas.md
jeremyfiel: This sounds like it may require some way to identify the parts and which `contentType` should be applied to the  `string` value

The other thing is my initial impression is the complexity to understand and write this schema is not very user friendly to non-experts and there aren't many guardrails around defining this type of schema or an invalid variation of it.

handrews: @jeremyfiel Well, it's possible to do whether we say anything about it or not, so better to explain it (although I'm open to there being a better example if someone wants to propose one).  I did start writing up something about mapping schemas to things, but that is also not actually a new problem and I kind of ran out of gas.

When parsing, the per-part `Content-Type` field will make clear which to use.  When serializing, presumably you know the data you're sending, but if you do not, then that is why I added the bit about the security considerations of content-sniffing.  See if what I put in the most recent commit helps, and also lmk if you have ideas for guardrails.

We have not added any new rules to `contentType`, but what has changed is that `items` applies to many parts, and is the only way to work with variable numbers of parts of a particular type.  I'm open to other ways ot discuss that use case, but if we just try to ignore it, people will fill in the gaps their own way and experience shows us that usually isn't good for interoperability.

handrews: @jeremyfiel I expect to spend tomorrow's TDC call on this PR, probably in a few phases- the basics, the simpler examples, and then complex examples and the part about finding the relevant `type` keyword (which I would have made a separate PR but it all overlaps in difficult ways).

handrews:
As this has grown with more guidance and examples, plus many comment threads, it has gotten unwieldy.  I have split it into the following PRs to replace this one:

* #4743 (an improved approach to analyzing schemas for `type` and other things, as well as handling mixed JSON/binary data in schema evaluation, which is more closely related than it sounds)
* #4744 (guidance for when `contentType` is a list, since as @jeremyfiel noted, that brings up some complexities that we have never addressed, and it will come up more with ordered `multipart` as seen in one of the examples
* #4745 (all of the content of this PR that is neither an example nor part of the two previous PRs in this list- ***no changes*** other than splitting it up)
* #4746 (the examples, which are provoking involved enough discussions that they should be separate, plus they will be affected by pending Example Object changes so it's easier to manage them separately- includes the recently revised `multipart/related` example that was particularly provocative)

Issue #3084, opened 2022-11-29
Proposal: dotDelimited style for query serialization.


maarlo:
Regarding ``[`` and ``]`` must be urlencoded, I propose the usage of dot notation to serialize arrays and objects at query parameters.

Having the following array:
```json
{
  "x": ["b","d","f"]
}
```

It could be serailized as:
```
?x.0=b&x.1=d&x.2=f 
```

Having the following object:
```json
{
  "x": {
    "a": "b",
    "c": "d",
    "e": "f"
  }
}
```

It could be serailized as
```
?x.a=b&x.c=d&x.e=f
```

And mixing both of them:
```json
{
  "x": [
    {
      "a": "b"
    },
    {
      "a": "d"
    }
  ]
}
```

It could be serailized as
```
?x.0.a=b&x.1.a=d
```

jeremyfiel:
you can find serialization examples here, multiple types are supported

https://spec.openapis.org/oas/v3.0.3#fixed-fields-9

maarlo:
Yes, but the proposed style is not defined.

handrews:
@OAI/tsc review request:  Can we decide whether to add this or not?  My sense is that we don't want to add further serialization options/complexity, but if we do it is straightforward.  This issue has been open for a year and a half without attracting further support or examples of usage in existing APIs.

karenetheridge:
Previous serialization styles have been defined in terms of uri-template syntax (although only the swagger documentation -- https://swagger.io/docs/specification/serialization/ -- explicitly provides those templates - it's not even at https://learn.openapis.org/specification/parameters.html).

If we want to restrict ourselves to this, is it possible to define the newly proposed dotDelimited format in terms of uri-templates?

maarlo:
I think so, just like happens with deepObject, but instead of using brackets using points, and with the addition of also allowing the serialization of arrays

ralfhandl:
> is it possible to define the newly proposed dotDelimited format in terms of uri-templates?

I couldn't find anything in [RFC6570](https://datatracker.ietf.org/doc/html/rfc6570)

handrews:
I'm leaning towards advocating that we support #1502 (Support for arbitrary query strings) in 3.2, which could allow folks to experiment with new/custom serializations without us having to add more complexity to the mandatory options.

mikekistler:
[...]

Issue #2347, opened 2020-09-11
Supporting of one separate 'required' option for query parameters


Morgunov-Vitaly:
Hi!
For example, there are several paths that include similar parameters, but with different required options: in some paths parameters are required, in others they are optional.

If we can specify the required option separately, we can reuse parameter descriptions. Like this:

`/api/v1/rootname:
    get:
      summary: 'Path example'
      parameters:
        - $ref: '#/components/parameters/queryFiltersDateFrom'
        - $ref: '#/components/parameters/queryFiltersDateTo'
        - $ref: '#/components/parameters/queryGroups'
        - $ref: '#/components/parameters/queryOrderByColumn'
        - $ref: '#/components/parameters/queryOrderByDirection'
      required:
        - 'filters[date_from]'
        - 'filters[date_to]'
        - 'groups'
        - 'order_by_column'
        - 'order_by_direction'`

hkosova:
Related (or duplicate): #2026

handrews:
This is supported for query parameters in 3.2 through PR #4590 by modeling all query parameters in a single schema, which lets you use the regular `required` for this.  We will consider further work in future releases.

Issue #2298, opened 2020-07-29
Aliasing query parameters


o-felixz:
Consider the query parameter `query` in the following request:
`https://my.api/api/endpoint?query=some%20query`

I could document this parameter like so:

```yaml
parameters:
  - name: query
    in: query
    description: Search term
    schema:
      type: string
    required: true
```

Now, our api supports a short alias for the `query` parameter, `q`:
```https://my.api/api/endpoint?q=some%20query```

From my research I am unable to find an explicit example of how to document this properly in OAS 3.0.
Is there already a solution to this in the spec that I have missed?

Otherwise I would suggest introducing an `aliases` keyword for parameters, like so:

```yaml
parameters:
  - name: query
    aliases:
      - q
    in: query
    description: Search term
    schema:
      type: string
    required: true
```

I expect the `name` would be a "primary" alias and that it, together with `aliases` would form a list, perhaps in descending order of importance if necessary, of mutually exclusive names/aliases for the given parameter.

MikeRalphson:
Although aliasing parameter names doesn't seem to be common (I could quickly only find one definite example - https://api.open.fec.gov/swagger/ - where the aliased parameters were duplicated) that could be because of the lack of existing support, and duplicating a parameter doesn't help if it should be `required: true`.

I personally see some value in this for describing existing APIs, though it would add complexity to implementing the specification around detecting duplicate and overriding parameters, as well as path-parameter substitution.

mschechter-bellese:
This would be very useful for content properties as well. I use Java, and [Jackson](https://github.com/FasterXML/jackson) allows you to create [multiple aliases for a field](https://www.baeldung.com/json-multiple-fields-single-java-field) to ensure backward-compatibility when re-working an API.

It would be useful to provide that information at the contract level so that it's not isolated to the implementation side.

SampsonCrowley:
@MikeRalphson it's definitely less common because of lack of support

kscheirer:
I am evaluating this as a possible use case for overlays, https://github.com/OAI/Overlay-Specification/discussions/9.

I don't think overlays would be useful here. There is no copy operation in overlays to duplicate a parameter. An overlay could add a second parameter, but as noted this fails when "required" is true.

handrews:
With query parameters now (optionally) modeled by one schema, this (along with the comment about aliasing for "content properties" is now a JSON Schema concern.  I'm pretty sure there's an issue there for such things, although idk if it's open or closed.

PR #4590, opened 2025-05-15
Add `in: "querystring"` parameter location treating the entire query string as one chunk of content

handrews:
**NOTE:** This change was tentativley approved for 3.2 a long time ago, then punted to 3.3 when we thought we'd need to defer larger data modeling changes to that release.  However, several things have changed that makes this suitable for 3.2 now:

* All this really does is allow using the Encoding Object in a new place, which we are already doing with PRs #4559 and #4589.  _We should really include all of the "use the Encoding Object in more places" changes in one release!_
* We already decided to add the media type registry in 3.2, when at one point we weren't going to do that until 3.3
* I realized that limiting `in: querystring` to using `content` avoids all of the confusing problems of how to deal with `style`/`explode`/etc. in an environment where they do not make sense

-----

* Fixes:
    * #1502 (arbitrary query strings)
    * #2347 (re-using parameter schemas independent of required-ness)
    * #2511 (disallow unknown query parameters, could be done with `deepObject` but now can be done while also using `style`/`explode` for individual parameters)
* Fixes as much as we can fix without someting much more drastic:
    * #2026 (re-using with differing constraints- this is as much as we can solve this without going to Moonwalk)
    * #2298 (aliasing- no special aliasing construct, but using `$ref`s comes pretty close)
* Enables fix through registering a media type or defining an out-of-band format for use with `in: querystring, content: {text/plain: {...}`:
    * #1501 
    * #1504 
    * #1706 
    * #3084 

-----

This allows treating the entire query string as a single parameter, which for `application/x-www-form-urlencoded` results in it being handled exactly as request bodies of that media type are handled.

Only one `in: "querystring"` is allowed per operation, and if it is present, no `in: "query"` parameters are allowed.

The `content` field MUST be used for `in: "querystring"`, as the `style` and related fields are handled in the Encoding Object.

This also includes a recommendation that implementors design media types for complex query string formats such as those defined by various frameworks, and register a procedure for supporting them in our media type registry.  This is intended to address a slow but steady trickle of requests to support various complex and often contradictory query string formats.

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -905,7 +905,8 @@ See [Appendix E](#appendix-e-percent-encoding-and-form-media-types) for a detail
 There are four possible parameter locations specified by the `in` field:
 
 * path - Used together with [Path Templating](#path-templating), where the parameter value is actually part of the operation's URL. This does not include the host or base path of the API. For example, in `/items/{itemId}`, the path parameter is `itemId`.
-* query - Parameters that are appended to the URL. For example, in `/items?id=###`, the query parameter is `id`.
+* query - Parameters that are appended to the URL. For example, in `/items?id=###`, the query parameter is `id`; MUST NOT appear in the same operation as an `in: "querystring"` parameter.
+* querystring - A parameter that treats the entire URL query string as a value which MUST be specified using the `content` field, most often with media type `application/x-www-form-urlencoded` using [Encoding Objects](#encoding-object) in the same way as with request bodies of that media type; MUST NOT appear more than once, and MUST NOT appear in the same operation with any `in: "query"` parameters.
 * header - Custom headers that are expected as part of the request. Note that [RFC9110](https://www.rfc-editor.org/rfc/rfc9110.html#section-5.1) states header names are case insensitive.
 * cookie - Used to pass a specific cookie value to the API.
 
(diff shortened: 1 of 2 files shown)


davebelais:
@handrews : I just want to confirm my understanding—so that I can plan for related tooling: when we have a parameter which has a `style="querystring"`—that operation shouldn't have any other parameters which have a `style="query"`, right?

handrews:
@davebelais replace `style` with `in` and then yes.  Because `querystring` consumes the entire part of the URL between `?` and `#` (or the end if no `#` is present).  Trying to come up with rules for how `in: querystring` and `in: query` would share that space is way too complex.  Plus `in: querystring` plus the Encoding Object can do everything that a set of `in: query` parameters can do and more, I believe.

lornajane (dismissed):
I LOVE this addition!

handrews:
This latest commit replaces "no defined mapping" with "no registered mapping" (linking to the section on the registry) in the JSONPath example.  It also adds a serialized form to the JSON `in: querystring` example.  (@lornajane I figured your approval got cleared anyway and needs re-adding, so I might as well tweak these).

miqui:
> This latest commit replaces "no defined mapping" with "no registered mapping" (linking to the section on the registry) in the JSONPath example. It also adds a serialized form to the JSON `in: querystring` example. (@lornajane I figured your approval got cleared anyway and needs re-adding, so I might as well tweak these).

@handrews Hi. When  I follow the click path with respect to the media registry, I end up with this URL: https://spec.openapis.org/registry/media-type/    which is broken (404).

handrews:
@miqui yeah we still need to put the registry up.  It will be published along with 3.2 and 3.1.2, so it's expected to be broken for now.

karenetheridge (approved):
LGTM!

handrews:
@lornajane @miqui can you re-approve after these fixes of things that @karenetheridge found?

miqui (dismissed):

handrews:
@karenetheridge now that I am thinking more about this, I think the change to `example` was incorrect, as JSON/YAML examples do not need to be encoded.  And in fact just look like strings if they are.  It would be shown as a string if we could use the `example` field next to `content` one level up.

I'm going to change that one part back.  There's a lot of additional debate around examples right now. so we may need to do a follow-up.  I'd like to not hold this up on our example confusion if at all possible, as that applies to many things, and if we need to fix it, we can fix it here along with everywhere else if this PR is already merged.

miqui (approved):

Issue #3959, opened 2024-07-16
Formalize how to express null value in xml


ahmednfwela:
Since XML has no concept of `null`, how can we handle validating a null value (both as an attribute and as an element) ?

consider the following 3.0 schema :
```yaml
person:
  type: object
  required:
    - name
    - attrName
  properties:
    name:
      type: string
      nullable: true
    attrName:
      type: string
      nullable: true
      xml:
        attribute: true
```
notice how `required` here prevents us from removing the attribute/element

I have thought about this, and here are some of the approaches I came up with:
### Approach 1: self closing tags
```xml
<person>
  <name />
</person>
```
**Pros:** Makes sense to whoever reads it
**Cons:** Nothing i can think of, but maybe some xml parsers can consider a self-closing tag equivalent to empty string and don't distinguish between them, which means they don't survive round tripping:
e.g.
`<name />` gets represented on the way back to: `<name></name>`

### Approach 2: empty string
```xml
<person>
  <name></name>
</person>
```
**Cons:** if the property is of type string, it's not possible to distinguish between a non-null empty string and a null.
**Workaround:** force strings to be wrapped around double quotes, e.g.
- this is null:
```xml
<name></name>
```
- this is empty string:
```xml
<name>""</name>
```
- this is valid string:
```xml
<name>"hello"</name>
```
- this is non valid string
```xml
<name>hello</name>
```
Ofc this workaround is very problematic and not good since most parsers consider `""` A valid 2 character string.

### Approach 3: special marker attribute
```xml
<name xsi:nil="true"></name>
<name xsi:nil="true"/>
```
**Pros:** Can represent nulls consistently without having to check the contents of the element, this is also [how xml schema does it](https://www.w3.org/TR/xmlschema-1/#xsi_nil).
**Cons:** Size overhead of having to use `xsi:nil="true"` everywhere null is used.

### Approach 1: empty string
```xml
<person attrName="" />
```
### Approach 2: disallow nullable attributes altogether

make it that `xml.attribute: true` and `nullable: true` are mutually exclusive

ralfhandl:
@ahmednfwela Thanks for reporting this, and for the detailed research and references.

#### Preliminary analysis for elements

[...]

Issue #630, opened 2016-04-07
How to represent XML elements with attributes


micovery:
Suppose I have the following XML:

``` xml
<location>
  <coordinate format="lat-lon">a-b</coordinate>
</location>
```

I this the correct way to represent this using Swagger Schema objects?

``` yaml

---
  type: "object"
  xml: 
    name: "location"
  properties: 
    coordinate: 
      type: "string"
      properties: 
        format: 
          type: "string"
          xml: 
            attribute: true
```

It seems weird for the "coordinate" property ... which itself is of type "string" to contain a "properties" field. Coming from JSON-Schema-Draft-V4 background, you would not expect a "string" field to have properties, because there is no such thing in JSON.

But then again, this is a "Swagger Schema Object" ... not a JSON-Schema-Draft-V4. I just want confirmation whether or not this is the correct way to do it.

webron:
When it comes to pure json schema, that's actually not illegal, but the properties, when it comes to json, would be effectively ignored.

As for the answer to your actual question - there's pretty much no way to represent that currently with 2.0. The attributes can pretty much be applied only to objects.

ePaul:
Nice idea.

I guess Swagger/OpenAPI can not describe many types of XML, just those that have an obvious 1:1 mapping to JSON which can be described by OpenAPI itself. As there is no JSON equivalent to your XML fragment, it can't be described with Swagger either.

micovery:
Thanks @webron and @ePaul for the clarifications.

That's correct, my sample XML has no way of being represented as JSON ... unless you adopt some convention like BadgerFish (http://www.sklar.com/badgerfish/)

where it would be represented like this:
```json
{
  "location": {
    "coordinate": {
      "@format": "lat-lon",
      "$": "a-b",
    }
  }
}
```

(which itself could be described purely by JSON Schema without resorting to using the `xml` field that Swagger came up with).

But, putting that aside ... since it's not illegal (as @webron pointed out) in JSON-Schema-Draft-V4 for a string to have properties ... OpenAPI could accept it as a valid schema to represent XML ... just a matter of convention.

eminence:
This is primarily a problem when you are trying to use swagger to document an already existing API (that is, an API that you can't change).   Which is a pretty big bummer!

ralfhandl:
This could be solved by adding the convention to OpenAPI 3.0 that the text node of an XML element is represented as a property whose name is the empty string:

```yaml
  type: "object"
  xml: 
[...]

PR #4592, opened 2025-05-16
v3.2: Support all common XML node types (element, attribute, text, cdata)

handrews:
* Fixes:
    * #630 (elements with attributes and content)
    * #1435 (root schemas — This fix incorporates the spirit of the 3.1.2 PR #4576)
* As of the latest commit (28 May 2025 10:45AM Pacific time) also fixes:
    * #3959 (handling `null` values) 

-----

This change removes the restriction on where the `xml` field and XML Object can appear, and adds a `nodeType` field to support the four most commonly used XML node types: `element`, `attribute`, `text`, and `cdata`. A fifth nodetype, `none`, is used to prevent a Schema Object from producing a node.

It also adds many examples of how to model different XML scenarios, including using `prefixItems` for complex ordered nodes.

While this could have been made much simpler by always defaulting `nodeType` to `element`, a more complex requirement was needed to preserve compatibility with the existing (but now deprecated - see also PR #4591) `attribute` and `wrapped` fields.  This way, the behavior of an XML object that is empty or uses only `name`, `prefix`, and/or `namespace` behaves the same whether you look at defaults for `nodeType` or for `attribute` and `wrapped`.

Finally, this PR ports the guidance on handling `null` attributes and elements to 3.2.

-----

I expect the deprecation of `attribute` and `wrapped` to be controversial, but there are several reasons for doing it:

* The solution proposed in #630 was to add another boolean field, `text`, to indicate text nodes.  But if we are going to support text nodes, we definitely also want to support CDATA sections, which would require three fields (`attribute`, `text`, and `cdata`) that were all mutually exclusive.  One boolean is fine, but three to manage a single state is more design smell than I can handle.
* The `wrapped` field is a very specific case of a more general problem, which is Schema Objects that do not correspond to XML nodes.  The old wording only acknowledged "property" and "root" schemas, and ignored the many other sorts of Schema Objects that might appear, and the complex ways in which they might map (or not) to XML nodes.  The `nodeType: none` construct allows controlling this in a generalized way.
* It becomes much more straightforward to define the `name` behavior, as certain types have names and others do not.  This is much more intuitive than logic involving `wrapped`.
* We might want to support additional node types in the future.  I just don't know enough about any of them to know what might be worth supporting, much less how to model them, and no one has asked for those (yet).
* The `null`-handling problems in #3959 are tricky, and this allows us to (if we want) define clear behavior for `nodeType: attribute` without breaking existing support for `attribute: true`.  _[NOTE: I ended up not trying to put restrictions on `null` and `nodeType: attribute` because there isn't really any good option but "don't use `null` with this" and forbidding that rather than just discouraging it seems problematic.]_

I came up with several other possible caveats to call out or clarify, but I wanted to keep this as minimal as possible.  If the related concerns come up in review I can add them, otherwise I will do it as a follow-on if it seems useful.  But let's see how this lands first.

-----

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -2582,7 +2582,7 @@ JSON Schema implementations MAY choose to treat keywords defined by the OpenAPI 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
 | <a name="schema-discriminator"></a>discriminator | [Discriminator Object](#discriminator-object) | The discriminator provides a "hint" for which of a set of schemas a payload is expected to satisfy. See [Composition and Inheritance](#composition-and-inheritance-polymorphism) for more details. |
-| <a name="schema-xml"></a>xml | [XML Object](#xml-object) | This MAY be used only on property schemas. It has no effect on root schemas. Adds additional metadata to describe the XML representation of this property. |
+| <a name="schema-xml"></a>xml | [XML Object](#xml-object) | Adds additional metadata to describe the XML representation of this schema. |
 | <a name="schema-external-docs"></a>externalDocs | [External Documentation Object](#external-documentation-object) | Additional external documentation for this schema. |
 | <a name="schema-example"></a>example | Any | A free-form field to include an example of an instance for this schema. To represent examples that cannot be naturally represented in JSON or YAML, a string value can be used to contain the example with escaping where necessary.<br><br>**Deprecated:** The `example` field has been deprecated in favor of the JSON Schema `examples` keyword. Use of `example` is discouraged, and later versions of this specification may remove it. |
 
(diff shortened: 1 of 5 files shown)


src/oas.md
```diff
@@ -3077,52 +3077,100 @@ will map to `#/components/schemas/Dog` because the `dog` entry in the `mapping`
...
-When using arrays, XML element names are _not_ inferred (for singular/plural forms) and the `name` field SHOULD be used to add that information.
-See examples for expected behavior.
+When using a Schema Object with XML, if no XML Object is present, the behavior is determined by the XML Object's default field values.
 
 ##### Fixed Fields
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
-| <a name="xml-name"></a>name | `string` | Replaces the name of the element/attribute used for the described schema property. When defined within `items`, it will affect the name of the individual XML elements within the list. When defined alongside `type` being `"array"` (outside the `items`), it will affect the wrapping element if and only if `wrapped` is `true`. If `wrapped` is `false`, it will be ignored. |
+| <a name="xml-node-type"></a>nodeType | `string` | One of `element`, `attribute`, `text`, `cdata`, or `none`, as explained under [XML Node Types](#xml-node-types).  The default value is `none` if `$ref`, `$dynamicRef`, or `type: "array"` is present in the [Schema Object](#schema-object) containing the XML Object, and `element` otherwise. |
```
handrews, line 3086: I'm not 100% the bits about `$ref` and `$dynamicRef` are backwards-compatible, because I'm not sure what, if anything, the behavior was in those cases before, as "property schema" was not well-defined.  If we mean "literal inline schema directly under a name under `properties` without references", then there was no defined behavior and we can do what we want.

Otherwise, it might not be, and we should let schemas containing `$ref` or `$dynamicRef` have a default `nodeType` of `element`, and folks can explicitly set them to `none`. I don't think that change would impact anything else in this PR, and might be the more conservative compatibility choice.

ralfhandl (commented):
Reviewed up to "XML Object Examples"

src/oas.md
```diff
@@ -3077,52 +3077,100 @@ will map to `#/components/schemas/Dog` because the `dog` entry in the `mapping`
...
+| <a name="xml-attribute"></a>attribute | `boolean` | Declares whether the property definition translates to an attribute instead of an element. Default value is `false`. If `nodeType` is present, this field MUST NOT be present.<br /><br />**Deprecated:** Use `nodeType: "attribute"` in place of `attribute: true` |
+| <a name="xml-wrapped"></a>wrapped | `boolean` | MAY be used only for an array definition. Signifies whether the array is wrapped (for example, `<books><book/><book/></books>`) or unwrapped (`<book/><book/>`). Default value is `false`. The definition takes effect only when defined alongside `type` being `"array"` (outside the `items`). If `nodeType` is present, this field MUST NOT be present.<br /><br />**Deprecated:** Set `nodeType: "element"` explicitly in place of `wrapped: true` |
+
+Note that when generating an XML document from object data, the order of the nodes is undefined.
+Use `prefixItems` to control node ordering.
```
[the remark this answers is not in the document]
handrews: @ralfhandl there is an example... didn't you say you hadn't read the examples section yet? ;-)

In all seriousness, I'd love better examples once you get there, I struggle to come up with them.  But there is a `prefixItems` example.

src/oas.md
```diff
@@ -3077,52 +3077,100 @@ will map to `#/components/schemas/Dog` because the `dog` entry in the `mapping`
...
+
+See [Appendix B](#appendix-b-data-type-conversion) for a discussion of converting values of various types to string representations.
 
 This object MAY be extended with [Specification Extensions](#specification-extensions).
 
+##### XML Node Types
+
+Each Schema Object describes a particular type of XML [node](https://dom.spec.whatwg.org/#interface-node) which is specified by the `nodeType` field, which has the following possible values.
```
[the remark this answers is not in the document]
handrews: @ralfhandl I'd be happy to- what do I need to do for that?  If this is just standard respec I can read about in its docs, just lmk.

src/oas.md
```diff
@@ -3077,52 +3077,100 @@ will map to `#/components/schemas/Dog` because the `dog` entry in the `mapping`
...
+Except for the special value `none`, these values have numeric equivalents in the DOM [specification](https://dom.spec.whatwg.org/#interface-node) which are given in parentheses after the name:
+
+* `element` (1): The schema represents an element and describes its contents
+* `attribute` (2): The schema represents an attribute and describes its value
+* `text` (3): The schema represents a text node (parsed character data)
+* `cdata` (4): The schema represents a CDATA section
+* `none`: The schema does not correspond to any node in the XML document, and its contents are included directly under the parent schema's node
```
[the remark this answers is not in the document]
handrews: @ralfhandl Perhaps "the nodes corresponding to its subschema(s) are included directly under its parent schema's node"?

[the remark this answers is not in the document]
handrews: I've added that wording, see if that works for you.

src/oas.md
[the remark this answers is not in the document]
handrews: @ralfhandl done- I made it normative so that the definitions (particularly the numbers which clarify which meaning of "text" is intended) are part of the spec.

src/oas.md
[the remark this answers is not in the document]
handrews: @ralfhandl I expanded those parts of the examples and added a link from this sentence so that it's less cryptic.

src/oas.md
```diff
@@ -3284,6 +3341,229 @@ animals:
...
+    xml:
+      nodeType: element
+      name: animals
+    items:
+      properties:
+        kind:
+          type: string
+          xml:
+            nodeType: attribute
+            name: animal
```
ralfhandl, line 3359:
```suggestion
    items:
      xml:
        name: animal
      properties:
        kind:
          type: string
          xml:
            nodeType: attribute
```
I think the items schema needs an xml field to set the name of the XML elements corresponding to the items.

handrews:
Now with `null`-handling advice ported from 3.1.2, updated to use `nodeType: attribute` (as the first added commit) and with an added bullet point for `text` and `cdata` nodes (as the second).

I decided that outright forbidding `null` with `nodeType: attribute` is probably more trouble than it's worth.  This keeps our compatibility and migration story from 3.1.2 simple (meaning that `attribute: true` and `nodeType: attribute` have exactly the same behavior).

ralfhandl (approved):

dinamic (approved):

src/oas.md
```diff
@@ -3146,22 +3211,24 @@ animals:
...
+components:
+  schemas:
+    Person:
+      type: object
+      properties:
+        id:
+          type: integer
+          format: int32
+          xml:
+            attribute: true
```
hudlow: @handrews just a reminder that you were thinking about changing this on the call.

src/oas.md
```diff
@@ -3284,6 +3357,280 @@ animals:
...
+  /docs:
+    get:
+      responses:
+        "200":
+          content:
+            application/xml:
+              xml:
+                nodeType: element
+                name: StoredDocument
+              $ref: "#/components/schemas/Documentation"
```
```suggestion
            application/xml:
              schema:
                xml:
                  nodeType: element
                  name: StoredDocument
                $ref: "#/components/schemas/Documentation"
```

src/oas.md
```diff
@@ -3284,6 +3357,280 @@ animals:
...
+    put:
+      requestBody:
+        required: true
+        content:
+          application/xml:
+            xml:
+              nodeType: element
+              name: UpdatedDocument
+            $ref: "#/components/schemas/Documentation"
```
```suggestion
          application/xml:
            schema:
              xml:
                nodeType: element
                name: UpdatedDocument
              $ref: "#/components/schemas/Documentation"
```

src/oas.md
```diff
@@ -3284,6 +3357,280 @@ animals:
...
+
+```yaml
+paths:
+  /docs:
+    get:
+      responses:
+        "200":
+          content:
+            application/xml:
+              $ref: "#/components/schemas/Documentation"
```
```suggestion
            application/xml:
              schema:
                $ref: "#/components/schemas/Documentation"
```

src/oas.md
```diff
@@ -3077,52 +3077,116 @@ will map to `#/components/schemas/Dog` because the `dog` entry in the `mapping`
...
+Explicit `text` nodes are necessary if an element has both attributes and content.
+
+Note that placing two `text` nodes adjacent to each other is ambiguous for parsing, and the resulting behavior is implementation-defined.
+
+##### XML Node Names
+
+The `element` and `attribute` node types require a name, which MUST be inferred from the schema as follows, unless overridden by the `name` field:
+
+* For schemas directly under the [Components Object's](#components-object) `schemas` field, the component name is the inferred name.
+* For property schemas, and for array item schemas under a property schema, the property name is the inferred name
```
hudlow, line 3242: Do these two bullets need to be transposed to avoid an incompatible change? 

I assume that in the existing behavior the inferred node name for the `<Farm>.animals` property would be `animal`:

```yaml
components:
  schemas:
    Alien:
      type: string
    Farm:
      type: object
      properties:
        animal:
          $ref: "#/components/schemas/Alien"
```

src/oas.md
```diff
@@ -3284,6 +3357,280 @@ animals:
...
+  "Some preamble text."
+  42,
+  "Some postamble text."
+]
+```
+
+###### XML With `null` Values
+
+Recall that the schema validates the in-memory data, not the XML document itself.
+The properties of the `"metadata"` element are omitted for brevity as it is here to show how the `null` value is represented.
```
```suggestion
The properties of the `"related"` element object are omitted for brevity as it is here to show how the `null` value is represented.
```

src/oas.md
```suggestion
            nodeType: attribute
```

handrews:
@ralfhandl I can't update tests here until PR #4704 is merged *and then* PR #4701 is merged and propagated through to v3.2-dev.  Otherwise the negative tests fail because the schema does not actually perform validation.

handrews:
Since we previously had 3 TSC approvals in GitHub and/or yesteday's TDC call (@ralfhandl @mikekistler @miqui ), and the new commits are test-only which usually require only one approval, I am going to merge this.  Thanks to everyone involved in getting it to the finish line!

PR #4600, opened 2025-05-16
Draft release notes for 3.2.0 and 3.1.2

lornajane:


miqui:
@lornajane So far looks good to me.

jeremyfiel:
should add more detailed info on new keywords in MediaType object.

lornajane:
@jeremyfiel can you say something about what's missing? (I obviously don't know or it wouldn't be missing ...) I did just update with some of the changes from the last couple of weeks, but I don't think there is anything here that would match what you're asking for.

draft-release-notes.md
```diff
@@ -0,0 +1,101 @@
...
+### Reference resolution
+
+  - additional top-level `$self` to be used as a base URI for resolving references in the OpenAPI description. If not present, the existing/earlier behaviour of using the retrieval URL as a base applies.
+
+### **ABNF** (Augmented Backus–Naur Form) formalised for path templating, server variables, and runtime expressions in the Links object.
+
+### Improvements for APIs using XML as a content format
+
+  - XML namespaces can be IRIs (rather than URIs)
+  - Explanation and example on how to handle `null` in XML
```
handrews:
```suggestion
  - New `nodeType` field allows maping schemas to all common XML node types (elements, attributes, text, or cdata) or to nothing
  - `attribute: true` deprecated in favor of `nodeType: attribute`
  - `wrapped: true` deprecated in favor of `nodeType: element` (as `nodeType` defaults to `none` for arrays to preserve compatibility)
  - The `xml` keyword can be used in any Schema Object
  - XML namespaces can be IRIs (rather than URIs)
  - Explanation and example on how to handle `null` in XML
```

draft-release-notes.md
```diff
@@ -0,0 +1,101 @@
...
+  - Responses can be a repeating data structure, and are treated as if they are an array of schema objects.
+  - Use `itemSchema` in a mediatype entry to describe each item
+  - Related: a new media types registry is published to give more context for each of the media types
+  - Also a "Complete vs Streaming Content" section for guidance on streaming binary payloads
+
+### Parameters
+
+  - additional parameter location `querystring`, to allow parsing the entire query string as a single field similar to the way a request body is handled, using the `content` field. Useful for complex or unconventional query data. Mutually exclusive with the `query` field.
+  - parameters can therefore be `in` the `querystring` as an alternative to the existing `header`, `cookie`, `query` and `path` values.
+  - `allowReserved` field is now permitted on parameters with any value of `in` (however this many not be a Good Idea (TM) in some scenarios)
```
handrews:
```suggestion
### Parameters and Headers

  - additional parameter location `querystring`, to allow parsing the entire query string as a single field similar to the way a request body is handled, using the `content` field. Useful for complex or unconventional query data. Mutually exclusive with the `query` field.
  - parameters can therefore be `in` the `querystring` as an alternative to the existing `header`, `cookie`, `query` and `path` values.
  - `allowReserved` field is now permitted on parameters and headers with any value of `in` (however this many not be a Good Idea (TM) in some scenarios)
```

draft-release-notes.md
```diff
@@ -0,0 +1,101 @@
...
+    
+### Responses
+
+  - `description` field for responses are now optional (they used to be required but they could be empty)
+  - additional `summary` field for responses, useful when displaying responses in a list context
+
+### Minor edits that are worth a mention:
+
+  - Streamlined to YAML examples (unless something specific to another format) to try to make it easier to follow
+  - `allowReserved` is now supported for any parameter or header, regardless of `in` location
```
handrews: This is now covered under "Parameters and Headers"

draft-release-notes.md
```diff
@@ -0,0 +1,101 @@
...
+
+  - Update to https://www.ietf.org/archive/id/draft-bhutton-json-schema-01.html of JSON Schema Specification
+  - Update to https://www.ietf.org/archive/id/draft-bhutton-json-schema-validation-01.html of JSON Schema Validation Specification
+  - Use [RFC8529](https://tools.ietf.org/html/rfc8259) for JSON
+  - Use [RFC9110](https://tools.ietf.org/html/rfc9110) for HTTP
+
+### Editorial changes:
+
+  - Extensive additions around media types, encoding, sequential media types, SSE examples, working with binary data,
+  - Clarification that Example Objects can be used in Header Objects.
```
handrews: I don't think we changed anything here?  PR #4648 allows additional use of `example`/`examples` for both Parameter and Header fields, but that has not merged yet?

draft-release-notes.md
```diff
@@ -0,0 +1,101 @@
...
+  - clarify that server URLs should not include fragment or query.
+  - support new `name` field alongside `description`, `url` and `variables`.
+  - formal path templating support for variable substitution in server urls.
+
+### Discriminator
+
+  - no change from previous versions: use `discriminator` to hint which entry in `anyOf` or `oneOf` is expected.
+  - no change from previous versions: use `mapping` to link the discriminator property value to the Schema name if they aren't an exact match.
+  - the discriminator `propertyName` can now be an optional field.
+  - additional `defaultMapping` field to indicate which schema to use if the `propertyName` is not set, or if the value is unrecognized.
```
handrews: [deleted original suggestion due to typo pointed out by Jeremy below — see following comment for updated suggestion]

draft-release-notes.md
```diff
@@ -0,0 +1,101 @@
...
+  - Clarification that Example Objects can be used in Header Objects.
+  - Better explanation and examples for using Encoding.
+  - Clarify that Request Body Objects need to specify at least one media type to be meaningful
+  - How to more clearly indicate that responses will not have a body
+
+## 3.1 Updates
+
+Version 3.1.2 has no material changes but does contain editorial fixes.
+
+- Clarification that Example Objects can be used in Header Objects.
```
handrews: Same comment as under 3.2

handrews (changes requested):

draft-release-notes.md
jeremyfiel: `Code Generation`

draft-release-notes.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Improvements for APIs using XML as a content format
+
+  - XML namespaces can be IRIs (rather than URIs)
+  - Explanation and example on how to handle `null` in XML
+    Clarify that the root schema of an XML object should use the component name.
+
+### Sequential media type
+
+  - Support for sequential mediatypes such as `text/event-stream` for server-sent events (SSE) and `application/jsonl`, `application/json-seq` and others for sequential data.
```
karenetheridge, line 53: "media types" or "media-types"

draft-release-notes.md
handrews:
```suggestion
### Code Generation

  - the discriminator `propertyName` can now be an optional field.
  - additional `defaultMapping` field to indicate which schema to use if the `propertyName` is not set, or if the value is unrecognized.
  - no change from previous versions: use `discriminator` to hint which entry in `anyOf` or `oneOf` is expected.
  - no change from previous versions: use `mapping` to link the discriminator property value to the Schema name if they aren't an exact match.
  - Implementations now SHOULD (rather than MAY) support templates/generics using `$dynamicRef`
```

(thanks @jeremyfiel )

handrews:
I'd like to propose a slightly different ordering to improve the storytelling around this release:

1. Our marquee features are ***improved tagging*** and ***sequential/streaming support**, and the latter of those is part of a pretty comprehensive update to data modeling and representation.
2. Expanded HTTP method support, self-identifying documents, and expanded OAuth support are our other major changes.
3. Improving the details and rigor of servers, paths, runtime expressions, and responses are "quality-of-life" improvements, to quote a recent Slack conversation.

This would give us a flow more like the following:

## 3.2 Updates

### Tags

### Data Modeling and Representation

OAS v3.2 makes significant progress towards more consistent, modular, and extensible media type and parameter support, while also expanding the set of media types supported in response to both emerging and legacy use cases.  In addition, we have removed ambiguity regarding how to present examples in a variety of complex scenarios.

#### Media Type Registry

#### Sequential and Streaming Media Types

#### Parameters and Headers

#### Multipart Media Types

#### XML

#### Code Generation

#### Example Object

### Additional Major Additions

_Note: This subsection ordering within this section makes sense to me, but I do not feel strongly about it._

#### Methods

#### Document Identification and Reference Resolution

#### Security

### Smaller Improvements

#### Servers, Paths, and Runtime Expressions

#### Responses

#### Updated to New Versions of Other Standards

#### etc.

lornajane:
There's about 50% more spec change since I wrote this draft, re ordering definitely in our future! Thanks for the outline , I'll be trying to catch up this week

draft-release-notes.md
lornajane: Thanks, updated.

draft-release-notes.md
```diff
@@ -0,0 +1,106 @@
...
+  - Responses can be a repeating data structure, and are treated as if they are an array of schema objects.
+  - Use `itemSchema` in a mediatype entry to describe each item
+  - Related: a new media types registry is published to give more context for each of the media types
+  - Also a "Complete vs Streaming Content" section for guidance on streaming binary payloads
+
+### Parameters and Headers
+
+  - additional parameter location `querystring`, to allow parsing the entire query string as a single field similar to the way a request body is handled, using the `content` field. Useful for complex or unconventional query data. Mutually exclusive with the `query` field.
+  - parameters can therefore be `in` the `querystring` as an alternative to the existing `header`, `cookie`, `query` and `path` values.
+  - `allowReserved` field is now permitted on parameters and headers with any value of `in` (however this many not be a Good Idea (TM) in some scenarios)
```
lornajane: Can you suggest some situations where it might be unwise? I'm not sure what's implied.

handrews: @lornajane let me make a suggestion without that line.  It's addressed in the spec- it's more that percent-encoding (which is mandated by `schema`+`style` and is reduced but not eliminated by `allowReserved`) is not suitable for all parameter types.  I'm not sure why I thought it should be mentioned in the notes.

```suggestion
  - `allowReserved` field is now permitted on parameters and headers with any value of `in`
```

draft-release-notes.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Flexible response metadata fields
+
+- `description` field for responses are now optional (they used to be required but they could be empty).
+- Additional `summary` field for responses, useful when displaying responses in a list context.
+
+### Minor edits that are worth a mention
+
+- Streamlined to YAML examples (unless something specific to another format) to try to make it easier to follow.
+
```
handrews:
```suggestion
- Streamlined to YAML examples (unless something specific to another format) to try to make it easier to follow.
- Non-Schema examples no longer "override" Schema examples; tools are free to use the most appropriate example for any given use case
```

ralfhandl (commented):
Nice!

draft-release-notes.md
lornajane: Updated

draft-release-notes.md
lornajane: Ah, it made more sense when I reviewed those changes this time. Good catch

draft-release-notes.md
lornajane: removed from the 3.1 section, I don't think we changed anything here

lornajane:
Started adding the suggested structure (in 3 parts: digestible high level features, the masterpieces of data modeling, and then everything else), but I note that we don't have all the sections you mentioned @handrews , in particular:
 - media types registry, these are the release notes on the specification and the media types registry isn't in this document so they're not included. Hold that thought for the blog content though
 - code gen, I'm not sure what should go here
 - multipart handling, how did I get this far without adding it? I have no idea but I've put a placeholder for now!

draft-release-notes.md
```diff
@@ -0,0 +1,147 @@
...
+### Headline features
+
+A summary of the biggest changes.
+
+#### Nested, multipurpose tags
+
+- New `summary` field to allow short descriptions, used when displaying lists of tags. If you use the `x-displayName` extension, you can now use `summary` instead.
+- `parent` field to point to the tag that this tag is nested under. If you use `x-tagGroups`, adjust to use a tags hierarchy.
+- `kind` to allow multiple categories of tag. The `kind` field is free-form text, however there are some expected/conventional values such as `nav` (in line with the most common current usage as grouping for documentation output).
```
ralfhandl:
```suggestion
- `kind` to allow multiple categories of tags. The `kind` field is free-form text, however there are some expected/conventional values such as `nav` (in line with the most common current usage as grouping for documentation output).
```

draft-release-notes.md
```diff
@@ -0,0 +1,147 @@
...
+- A [registry](https://spec.openapis.org/registry/tag-kind/index.html) to establish conventions for values used in `kind`.
+
+#### Support additional HTTP methods
+
+- Support the new `query` method alongside the existing `get`/`post`/`put`/`delete`/`options`/`head`/`patch`/`trace`.
+- Under an `additionalOperations` entry in a Path, use any other methods not listed as keys using the correct capitalization. e.g. do NOT add HEAD under this, use the existing sibling `head`.
```
ralfhandl:
```suggestion
- Under an `additionalOperations` entry in a Path, use any other methods not listed as keys using the correct capitalization, e.g. `LINK`. Do NOT add `HEAD` under this, use the existing sibling `head`.
```

draft-release-notes.md
```diff
@@ -0,0 +1,147 @@
...
+
+- New `nodeType` field allows mapping schemas to common XML node types: `element`, `attribute`, `text`, `cdata`, or `none`.
+- `attribute: true` deprecated in favor of `nodeType: attribute`.
+- `wrapped: true` deprecated in favor of `nodeType: element` (as `nodeType` defaults to `none` for arrays to preserve compatibility).
+- The `xml` keyword can be used in any Schema Object.
+- XML namespaces can be IRIs (rather than URIs).
+- Explanation and examples for many use cases including handling `null`, handling arrays, replacing the name, and handling ordered elements.
+- Clarify that the root schema of an XML object should use the component name.
+
+#### Show examples in both structure and serialized forms
```
ralfhandl:
```suggestion
#### Show examples in both structured and serialized forms
```

draft-release-notes.md
```diff
@@ -0,0 +1,147 @@
...
+
+#### Editorial changes
+
+- Extensive additions around media types, encoding, sequential media types, SSE examples, working with binary data.
+- Better explanation and examples for using Encoding.
+- Clarify that Request Body Objects need to specify at least one media type to be meaningful.
+- How to more clearly indicate that responses will not have a body.
+- Explanation and examples of headers including `Link` and `Set-Cookie`.
+- No change but extensive additional notes on parsing and serializing JSON and non-JSON data formats.
+  Particularly if you are buildling OpenAPI tooling, this section gives much better guidance on some of those tricky edge cases.
```
ralfhandl:
```suggestion
  Particularly if you are building OpenAPI tooling, this section gives much better guidance on some of those tricky edge cases.
```

draft-release-notes.md
```diff
@@ -0,0 +1,147 @@
...
+- Under an `additionalOperations` entry in a Path, use any other methods not listed as keys using the correct capitalization. e.g. do NOT add HEAD under this, use the existing sibling `head`.
+
+#### Document identity and URL resolution
+
+- Additional top-level field `$self` is added to allow users to define the base URI of the document, used to resolve relative references.
+  If no `$self` field is defined, then the retrieval URI is used - just as it was in previous versions of OpenAPI.
+- Other URL/URI handling does not change between 3.1 and 3.2 (but bears a recap in case you're wondering how it all goes together):
+    - Other URLs, such as to external documentation or a license, are resolved against the base URI.
+    - Relative links inside `description` fields are resolved relative to their rendered context, such as your published API documentation page.
+    - API endpoints are resolved against the URL in the Server Object, which itself might be relative and resolved against the base URI.
```
ralfhandl:
```suggestion
  - Other URLs, such as to external documentation or a license, are resolved against the base URI.
  - Relative links inside `description` fields are resolved relative to their rendered context, such as your published API documentation page.
  - API endpoints are resolved against the URL in the Server Object, which itself might be relative and resolved against the base URI.
```
To satisfy our markdown-lint rules

draft-release-notes.md
```diff
@@ -0,0 +1,223 @@
...
+
+All the changes in the OpenAPI 3.2 release.
+
+### Headline features
+
+A summary of the biggest changes.
+
+#### Nested, multipurpose tags
+
+- New `summary` field to allow short descriptions, used when displaying lists of tags. If you use the `x-displayName` extension, you can now use `summary` instead.
```
karenetheridge: should we identify which tool(s) this exension is for?

lornajane: I think that an exhaustive list would be very difficult - I certainly couldn't get this correct without a lot of research. And the tools don't matter too much - I tried to word it so that you know what to look for in your OpenAPI description and how to update that.

draft-release-notes.md
lornajane: I would need an English grammar citation to make this change, it reads correctly as it is but I note that American English does use surprising plurals sometimes so maybe it's just that?

ralfhandl: Found this one: https://english.stackexchange.com/questions/29843/types-of-followed-by-singular-or-plural.

Nice rule of thumb: turn the sentence around and check how it sounds:
1. to allow tag of multiple categories
2. to allow tags of multiple categories

Only the second variant makes sense to me.

lornajane: I think maybe it needs an article, I'll try rewording

lornajane:
@handrews Could you take a look at the multipart media types section I just added? We had it as `**TODO**` still!

miqui:
@lornajane Nice job!! Looks GREAT to me.

draft-release-notes.md
ralfhandl: Suggestion now aligned with the new example.

Alternatively, replace period before "e.g." with a comma.

lornajane:
This is chaos, I'll be restructuring to make much slimmer release notes, an upgrade guide for the learn site and an announcement blog post (with the option to drill in with blog posts on more topics). I don't have a lot of time for the next week or so but I'll push updates when I can - since I'll mostly be moving content around, feedback on this current version is still very helpful if you have any

summary-release-notes.md
```diff
@@ -0,0 +1,133 @@
...
+
+    - New `nodeType` field allows mapping schemas to common XML node types: `element`, `attribute`, `text`, `cdata`, or `none`.
+    - `attribute: true` is now deprecated in favor of `nodeType: attribute`.
+    - `wrapped: true` is now deprecated in favor of `nodeType: element` (as `nodeType` defaults to `none` for arrays to preserve compatibility).
+    - The `xml` keyword can be used in any Schema Object.
+    - XML namespaces can be IRIs (rather than URIs).
+    - Explanation and examples for many use cases including handling `null`, handling arrays, replacing the name, and handling ordered elements.
+    - Clarify that the root schema of an XML object should use the component name.
+
+- Examples as either structured or serialised values
```
ralfhandl:
```suggestion
- Examples as either structured or serialized values
```
If we consistently want to use US English.

summary-release-notes.md
```diff
@@ -0,0 +1,133 @@
...
+### Additional features
+
+- Updated security schemes
+
+    - Support for [OAuth2 Device Authorization flow](https://datatracker.ietf.org/doc/html/rfc8628) with additional `deviceAuthorization` field in the `flows` object and for the individual flow, a new field `deviceAuthorizationUrl` alongside `tokenUrl`.
+    - Additional security scheme field: `oauth2MetadataUrl` URL for auth server metadata, as described by the [OAuth2 Server Metadata Standard](https://datatracker.ietf.org/doc/html/rfc8414).
+    - Additional `deprecated` field for security schemes (indicating that the scheme may still be supported, but that it should not be used).
+    - Ability to reference a security scheme by URI rather than declaring it in components.
+
+- Improvements to the Servers object
```
ralfhandl:
```suggestion
- Improvements to the Server Object
```

summary-release-notes.md
```diff
@@ -0,0 +1,133 @@
...
+- Better polymorphic support
+
+    - The discriminator `propertyName` can now be an optional field.
+    - New field `defaultMapping` to indicate which schema to use if the `propertyName` is not set, or if the value is unrecognized.
+
+- Templates with formal syntax
+
+    - The specification now includes **ABNF** (Augmented Backus–Naur Form) for path templating, server variables, and runtime expressions in the Links object.
+
+- Flexible metadata fields in the Response object
```
ralfhandl:
```suggestion
- Flexible metadata fields in the Response Object
```

summary-release-notes.md
```diff
@@ -0,0 +1,133 @@
...
+    - Formal ABNF syntax for the allowed variable substitution in server urls, alongside guidance that each variable can only be used once in a URL.
+
+- Better polymorphic support
+
+    - The discriminator `propertyName` can now be an optional field.
+    - New field `defaultMapping` to indicate which schema to use if the `propertyName` is not set, or if the value is unrecognized.
+
+- Templates with formal syntax
+
+    - The specification now includes **ABNF** (Augmented Backus–Naur Form) for path templating, server variables, and runtime expressions in the Links object.
```
ralfhandl:
```suggestion
    - The specification now includes **ABNF** (Augmented Backus–Naur Form) for path templating, server variables, and runtime expressions in the Link Object.
```

summary-release-notes.md
```diff
@@ -0,0 +1,133 @@
...
+
+Version 3.1.2 has no material changes but does contain editorial fixes, additional examples, and clarifications.
+
+- Clarify that `$ref` in a Schema Object is a JSONSchema `$ref` keyword.
+- Detailed explanation of percent-encoding and -decoding, updated examples and references to match, including the Style Examples table.
+- Better explanation and examples for using Encoding and Serialization, and a note not to apply percent-encoding to headers.
+- Clarify that Request Body Objects need to specify at least one media type to be meaningful.
+- How to more clearly indicate that no response will have a body.
+- How to handle `null` in XML as an advisory note; since the functionality cannot be changed it is implementation-defined for 3.1 tooling.
+- Clarify that the root schema of an XML object should use the component name.
```
ralfhandl, line 130:
```suggestion
- Clarify that the root schema of an XML Object should use the component name.
```

summary-release-notes.md
```diff
@@ -0,0 +1,133 @@
...
+
+    - Additional top-level field `$self` is added to allow users to define the base URI of the document, used to resolve relative references.
+    - More explanation and examples regarding URL resolution.
+
+### Data modeling and representation
+
+- Streaming support: sequential media types including SSE
+
+    - Support for sequential media types such as `text/event-stream` for server-sent events (SSE) and `multipart/mixed`, `application/jsonl`, `application/json-seq` and others for sequential data.
+    - Responses can be a repeating data structure, and are treated as if they are an array of schema objects.
```
ralfhandl:
```suggestion
    - Responses can be a repeating data structure, and are treated as if they are an array of Schema Objects.
```

summary-release-notes.md
```diff
@@ -0,0 +1,133 @@
...
+
+- Improvements for APIs using XML as a content format
+
+    - New `nodeType` field allows mapping schemas to common XML node types: `element`, `attribute`, `text`, `cdata`, or `none`.
+    - `attribute: true` is now deprecated in favor of `nodeType: attribute`.
+    - `wrapped: true` is now deprecated in favor of `nodeType: element` (as `nodeType` defaults to `none` for arrays to preserve compatibility).
+    - The `xml` keyword can be used in any Schema Object.
+    - XML namespaces can be IRIs (rather than URIs).
+    - Explanation and examples for many use cases including handling `null`, handling arrays, replacing the name, and handling ordered elements.
+    - Clarify that the root schema of an XML object should use the component name.
```
ralfhandl:
```suggestion
    - Clarify that the root schema of an XML Object should use the component name.
```

summary-release-notes.md
```diff
@@ -0,0 +1,133 @@
...
+    - The discriminator `propertyName` can now be an optional field.
+    - New field `defaultMapping` to indicate which schema to use if the `propertyName` is not set, or if the value is unrecognized.
+
+- Templates with formal syntax
+
+    - The specification now includes **ABNF** (Augmented Backus–Naur Form) for path templating, server variables, and runtime expressions in the Links object.
+
+- Flexible metadata fields in the Response object
+
+    - `description` field for responses are now optional.
```
ralfhandl:
```suggestion
    - `description` field for responses is now optional.
```

draft-release-notes.md
```diff
@@ -161,10 +161,15 @@ There is also a new Media Types Registry, to provide further resources for worki
 
 - Additional parameter location `querystring`, to allow parsing the entire query string as a single field similar to the way a request body is handled, using the `content` field. Useful for complex or unconventional query data. Mutually exclusive with the `query` field.
 - Parameters can therefore be `in` the `querystring` as an alternative to the existing `header`, `cookie`, `query` and `path` values.
-- `allowReserved` field is now permitted on headers and on parameters with any value of `in`.
+- `allowReserved` field is now permitted on headers and on parameters with any value of `in`, and applies where the combination of `in` and `style` automatically percent-encode the value.
 - Remove incorrect mention of Reference Object in the header `schema` field. The JSONSchema ref would be the correct thing to use in this context.
```
ralfhandl:
```suggestion
- Remove incorrect mention of Reference Object in the header `schema` field. The JSON Schema `$ref` would be the correct thing to use in this context.
```

draft-release-notes.md
```diff
@@ -336,8 +341,9 @@ Particularly if you are building OpenAPI tooling, these sections give much bette
 
 ## 3.1 Updates
 
-Version 3.1.2 has no material changes but does contain editorial fixes.
+Version 3.1.2 has no material changes but does contain editorial fixes, additional examples, and clarifications.
 
+- Clarify that `$ref` in a Schema Object is a JSONSchema `$ref` keyword.
```
ralfhandl:
```suggestion
- Clarify that `$ref` in a Schema Object is a JSON Schema `$ref` keyword.
```

summary-release-notes.md
```diff
@@ -0,0 +1,133 @@
...
+    - How to more clearly indicate that responses will not have a body.
+    - Explanation and examples of headers including `Link` and `Set-Cookie`.
+    - Detailed explanation of percent-encoding and -decoding, updated examples and references to match, including the Style Examples table.
+    - Extensive additional notes on parsing and serializing JSON and non-JSON data formats.
+
+## 3.1 Updates
+
+Version 3.1.2 has no material changes but does contain editorial fixes, additional examples, and clarifications.
+
+- Clarify that `$ref` in a Schema Object is a JSONSchema `$ref` keyword.
```
ralfhandl:
```suggestion
- Clarify that `$ref` in a Schema Object is a JSON Schema `$ref` keyword.
```

ralfhandl (commented):
Nice! Only minor suggestions.

summary-release-notes.md
lornajane: I think in this instance it's an object in the XML rather than a proper noun that we should Frighteningly Capitalise.

draft-release-notes.md
```diff
@@ -0,0 +1,354 @@
...
+```
+
+#### Document identity and URL resolution
+
+- Additional top-level field `$self` is added to allow users to define the base URI of the document, used to resolve relative references.
+  If no `$self` field is defined, then the retrieval URI is used - just as it was in previous versions of OpenAPI.
+- Other URL/URI handling does not change between 3.1 and 3.2 (but bears a recap in case you're wondering how it all goes together):
+  - Other URLs, such as to external documentation or a license, are resolved against the base URI.
+  - Relative links inside `description` fields are resolved relative to their rendered context, such as your published API documentation page.
+  - API endpoints are resolved against the URL in the Server Object, which itself might be relative and resolved against the base URI.
```
karenetheridge:
```suggestion
  - API endpoints (such as /paths entries, and authentication URLs) are resolved against the URL in the Server Object, which itself might be relative and resolved against the base URI.
```

draft-release-notes.md
```diff
@@ -0,0 +1,354 @@
...
+- Use `itemSchema` in a mediatype entry to describe each item.
+- Related: a new media types registry is published to give more context for each of the media types.
+- Also a "Complete vs Streaming Content" section for guidance on streaming binary payloads.
+
+#### Parameter and header changes
+
+- Additional parameter location `querystring`, to allow parsing the entire query string as a single field similar to the way a request body is handled, using the `content` field. Useful for complex or unconventional query data. Mutually exclusive with the `query` field.
+- Parameters can therefore be `in` the `querystring` as an alternative to the existing `header`, `cookie`, `query` and `path` values.
+- `allowReserved` field is now permitted on headers and on parameters with any value of `in`, and applies where the combination of `in` and `style` automatically percent-encode the value.
+- Remove incorrect mention of Reference Object in the header `schema` field. The JSON Schema `$ref` would be the correct thing to use in this context.
```
karenetheridge:
```suggestion
- Remove incorrect mention of Reference Object in the header `schema` field, which is redundant with the `$ref` coming from JSON Schema itself.
```

draft-release-notes.md
```diff
@@ -0,0 +1,354 @@
...
+
+OpenAPI specification v3.2 brings consistent, modular, and extensible media type and parameter support, while also expanding the set of media types supported in response to both emerging and legacy use cases. In addition, the ambiguity regarding how to present examples in a variety of complex scenarios is reduced.
+
+There is also a new Media Types Registry, to provide further resources for working with different media types.
+
+#### Support for sequential media types
+
+- Support for sequential media types such as `text/event-stream` for server-sent events (SSE) and `multipart/mixed`, `application/jsonl`, `application/json-seq` and others for sequential data.
+- Responses can be a repeating data structure, and are treated as if they are an array of schema objects.
+- Use `itemSchema` in a mediatype entry to describe each item.
```
karenetheridge:
```suggestion
- Use `itemSchema` in a media-type entry to describe each item.
```

draft-release-notes.md
```diff
@@ -0,0 +1,354 @@
...
+- Additional parameter location `querystring`, to allow parsing the entire query string as a single field similar to the way a request body is handled, using the `content` field. Useful for complex or unconventional query data. Mutually exclusive with the `query` field.
+- Parameters can therefore be `in` the `querystring` as an alternative to the existing `header`, `cookie`, `query` and `path` values.
+- `allowReserved` field is now permitted on headers and on parameters with any value of `in`, and applies where the combination of `in` and `style` automatically percent-encode the value.
+- Remove incorrect mention of Reference Object in the header `schema` field. The JSON Schema `$ref` would be the correct thing to use in this context.
+- The `examples` (and older `example`) field is now supported with `content`.
+
+#### Specific `style` option for cookies
+
+- Additional `style` option `cookie` for content in a cookie, which uses semicolon as a separator and does not encode data values.
+  The `form` style is a close fit, but you may wish to switch to `cookie` for more correct description.
```
karenetheridge:
```suggestion
  The `form` style is a close fit, but you may wish to switch to `cookie` for a more correct description.
```

draft-release-notes.md
```diff
@@ -0,0 +1,354 @@
...
+- `wrapped: true` deprecated in favor of `nodeType: element` (as `nodeType` defaults to `none` for arrays to preserve compatibility).
+- The `xml` keyword can be used in any Schema Object.
+- XML namespaces can be IRIs (rather than URIs).
+- Explanation and examples for many use cases including handling `null`, handling arrays, replacing the name, and handling ordered elements.
+- Clarify that the root schema of an XML object should use the component name.
+
+#### Describe examples in both structured and serialized forms
+
+- The Example Object (used in `examples` fields) gets two new fields: `dataValue` and `serializedValue`.
+- `dataValue` describes the example in structured format.
```
karenetheridge:
```suggestion
- `dataValue` describes the example in structured format (deserialized into the JSON document model, suitable for evaluating with a `schema`).
```

draft-release-notes.md
```diff
@@ -0,0 +1,354 @@
...
+            application/json:
+              schema:
+                type: array
+                items:
+                  type: string
+```
+
+#### Minor updates that are worth a mention
+
+- Non-Schema examples no longer "override" Schema examples; tools are free to use the most appropriate example for any given use case.
```
handrews: Please let's not include this.  No one can remember what "override" meant, and given the widespread incorrect implementation of the thing usually doing the "override", this is hard to even explain.  Just let that language quietly disappear.  As best we can tell, it had no practical implications but was instead a somewhat philosophical statement about which examples are closest to the wire format.

karenetheridge: I think this is what is intended? This is like "James while John had had had had had had had had had had had a better effect on the teacher"

```suggestion
- Non-Schema `example`s no longer "override" Schema `example`s; tools are free to use the most appropriate `example` for any given use case.
```

summary-release-notes.md
```diff
@@ -0,0 +1,133 @@
...
+    - The specification now includes **ABNF** (Augmented Backus–Naur Form) for path templating, server variables, and runtime expressions in the Link Object.
+
+- Flexible metadata fields in the Response Object
+
+    - `description` field for responses is now optional.
+    - Additional `summary` field for responses.
+
+- Additional updates
+
+    - Non-Schema examples no longer "override" Schema examples; tools are free to use the most appropriate example for any given use case.
```
karenetheridge:
```suggestion
    - Non-Schema `example`s no longer "override" Schema `example`s; tools are free to use the most appropriate `example` for any given use case.
```

handrews: I still say just drop this from the changelog entirely.  The suggested change is not quite correct as there are also two different `examples`es as well as two different `example`s.  And all of them interact in different ways.  This is just removing confusing language that doesn't do anything.  It shouldn't be announced any more than announcing we fixed some spelling or something.

draft-release-notes.md
```diff
@@ -0,0 +1,354 @@
...
+    kind: badge
+    externalDocs:
+      description: Seasonal menu planning guide
+      url: https://docs.cakestore.com/seasonal-planning
+```
+
+#### Support additional HTTP methods
+
+- Support the new `query` method alongside the existing `get`/`post`/`put`/`delete`/`options`/`head`/`patch`/`trace`.
+  Query is a new method in the HTTP standard, designed to allow complex filtering query data to be sent in the body of a request when it doesn't fit in the query string.
```
handrews: QUERY should definitely be all-caps, but if you don't like the rest of my suggestion below it's fine to ignore the other parts.

```suggestion
  QUERY is a [draft standard](https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-11.html) new HTTP method nearing final publication, which is designed to allow complex filtering query data to be sent in the body of a request when it doesn't fit in the query string.
```

draft-release-notes.md
```diff
@@ -0,0 +1,354 @@
...
+- Under an `additionalOperations` entry in a Path, use any other methods not listed as keys using the correct capitalization, e.g. `LINK`. Do NOT add `HEAD` under this, use the existing sibling `head`.
```
handrews: Suggestion for flow, I do not feel at all strongly about it.

```suggestion
- Under an `additionalOperations` entry in a Path, use any other methods not listed as keys using the correct capitalization, e.g. you can add `LINK` under this field, but do NOT add `HEAD`; use the existing Operation Object `head` field.
```

draft-release-notes.md
```diff
@@ -0,0 +1,354 @@
...
+
+The following example describes both `query` and `link` methods for the `cakes/` endpoint:
```
handrews: Let's use either the HTTP form of the names (all-caps, no backtics) or use the field names (`query` and `LINK`); I find `link` confusing because it is neither the HTTP form nor the field form.  Since the goal is to describe HTTP methods, the suggestion below uses that form.

```suggestion
The following example describes both QUERY and LINK methods for the `cakes/` endpoint:
```

draft-release-notes.md
```diff
@@ -0,0 +1,354 @@
...
+                    example: "recommended-with"
+        responses:
+          '204':
+            description: Link created successfully
+```
+
+#### Document identity and URL resolution
+
+- Additional top-level field `$self` is added to allow users to define the base URI of the document, used to resolve relative references.
+  If no `$self` field is defined, then the retrieval URI is used - just as it was in previous versions of OpenAPI.
```
handrews: The retrieval URI is not always correct.

```suggestion
  If no `$self` field is defined, then just as in previous versions of the OAS, the base URI according to RFC3986 Section 5.1.2 – 5.1.4 is used.  This is most often the retrieval URI, but see the new Appendix G for examples of all possible sources.
```

draft-release-notes.md
```diff
@@ -0,0 +1,354 @@
...
+- Other URL/URI handling does not change between 3.1 and 3.2 (but bears a recap in case you're wondering how it all goes together):
+  - Other URLs, such as to external documentation or a license, are resolved against the base URI.
```
handrews: It is important to call these URIs as they are not necessarily locators.  And in the text (ignoring field names we can't change), we are careful to call these URIs and only API endpoints are called URLs.

```suggestion
  - Other URIs, such as to external documentation or a license, are resolved against the base URI.
```

draft-release-notes.md
handrews: I'd suggest wording slightly different from @karenetheridge, primarily because "resolve" means a specific thing in RFC3986 and `/paths` do _not_ "resolve".  They are appended.  Also, the server URLs are not resolved against the base URI, but against the retrieval document (we specifically say that server URL resolution _does not_ follow RFC3986 with respect to the document's base URI).

```suggestion
  - API paths are appended to the URL in the Server Object, which itself might be relative and resolved against the retrieval URI of the document; note that API URLs do NOT use the document's base URI (`$self` or its RFC3986-defined alternate sources).  Examples are now provided in the Server Object section.
```

draft-release-notes.md
handrews: Why hyphenation?  AFAICT we don't hyphenate it anywhere in the OAS.  I'd go with:

```suggestion
- Use `itemSchema` in a Media Type Object to describe each item.
```

to avoid any confusion over what an "entry" is.

draft-release-notes.md
```diff
@@ -0,0 +1,354 @@
...
+OpenAPI specification v3.2 brings consistent, modular, and extensible media type and parameter support, while also expanding the set of media types supported in response to both emerging and legacy use cases. In addition, the ambiguity regarding how to present examples in a variety of complex scenarios is reduced.
+
+There is also a new Media Types Registry, to provide further resources for working with different media types.
+
+#### Support for sequential media types
+
+- Support for sequential media types such as `text/event-stream` for server-sent events (SSE) and `multipart/mixed`, `application/jsonl`, `application/json-seq` and others for sequential data.
+- Responses can be a repeating data structure, and are treated as if they are an array of schema objects.
+- Use `itemSchema` in a mediatype entry to describe each item.
+- Related: a new media types registry is published to give more context for each of the media types.
```
handrews:
```suggestion
- Related: a new media types registry is published to give more context for each of the media types and allow registration of future media type support outside of the OAS publication schedule.
```

draft-release-notes.md
```diff
@@ -0,0 +1,354 @@
...
+- Also a "Complete vs Streaming Content" section for guidance on streaming binary payloads.
```
handrews: This actually applies more to non-binary payloads, but we can just say "streaming payloads" and it is accurate (`multipart/byteranges` is binary).

```suggestion
- Also a "Complete vs Streaming Content" section for guidance on streaming payloads.
```

draft-release-notes.md
```diff
@@ -0,0 +1,354 @@
...
+
+#### Parameter and header changes
+
+- Additional parameter location `querystring`, to allow parsing the entire query string as a single field similar to the way a request body is handled, using the `content` field. Useful for complex or unconventional query data. Mutually exclusive with the `query` field.
```
handrews:
```suggestion
- Additional parameter location `querystring`, to allow parsing the entire query string as a single field similar to the way a request body is handled, using the `content` field. Useful for complex or unconventional query data. Mutually exclusive with the `query` location.
```

draft-release-notes.md
```diff
@@ -0,0 +1,354 @@
...
+- Parameters can therefore be `in` the `querystring` as an alternative to the existing `header`, `cookie`, `query` and `path` values.
+- `allowReserved` field is now permitted on headers and on parameters with any value of `in`, and applies where the combination of `in` and `style` automatically percent-encode the value.
```
handrews: It's easier to say this (which is matched in the text) than explain the exact conditions which are now pretty complicated because cookies are Like That.

```suggestion
- `allowReserved` now applies anywhere the combination of `in` and `style` automatically percent-encode the value.
```

draft-release-notes.md
handrews: I'm not entirely thrilled with my suggested wording, but `form` is only a close fit under certain circumstances.  I will not be upset if this suggestion is ignored (in which case do take Karen's version for the grammar fix!)

```suggestion
  The `form` style can sometimes work, but see Appendix D for caveats.
```

draft-release-notes.md
```diff
@@ -0,0 +1,354 @@
...
+- Remove incorrect mention of Reference Object in the header `schema` field. The JSON Schema `$ref` would be the correct thing to use in this context.
+- The `examples` (and older `example`) field is now supported with `content`.
+
+#### Specific `style` option for cookies
+
+- Additional `style` option `cookie` for content in a cookie, which uses semicolon as a separator and does not encode data values.
```
handrews, line 170:
```suggestion
- Additional `style` option `cookie` for content in a cookie, which conforms to RFC6265 by using semicolon as a separator and does not encoding data values.
```

draft-release-notes.md
```diff
@@ -0,0 +1,354 @@
...
+  The `form` style is a close fit, but you may wish to switch to `cookie` for more correct description.
+
+#### Multipart media types
+
+Multipart media types are much better supported in OpenAPI 3.2.
+
+- New `itemSchema` field, for the schema that describes each of the items in a sequential media type.
```
handrews, line 178:
```suggestion
- `multipart/mixed`, on which most other `multipart` formats are based, is now supported
- New `itemSchema` field, for the schema that describes each of the items in a sequential media type.
```

draft-release-notes.md
```diff
@@ -0,0 +1,354 @@
...
+  - Use `prefixEncoding` with an array of positional encoding information where the position of each item in the content is known.
+  - The `itemEncoding` is a single encoding entry that is used for information about multiple items.
+- The specification also contains examples of sequential JSON and Server-Sent events that show these fields in use.
+- Nested content types are supported to at least one level of nesting.
+
+#### Improvements for APIs using XML as a content format
+
+- New `nodeType` field allows mapping schemas to common XML node types: `element`, `attribute`, `text`, `cdata`, or `none`.
+- `attribute: true` is now deprecated in favor of `nodeType: attribute`.
+- `wrapped: true` deprecated in favor of `nodeType: element` (as `nodeType` defaults to `none` for arrays to preserve compatibility).
```
handrews:
```suggestion
- New `nodeType` field allows mapping schemas to common XML node types: `element`, `attribute`, `text`, `cdata`, or `none`, defaulting to `element` for most data types, or `none` for arrays.
- `attribute: true` is now deprecated in favor of `nodeType: attribute`.
- `wrapped: true` for arrays deprecated in favor of `nodeType: element` (set explicitly to override the default `nodeType: none` for arrays)
```

draft-release-notes.md
handrews: I agree with @karenetheridge , although if this wording is seen as too detailed, "suitable for evaluating with a `schema`" or "suitable for evaluating with a Schema Object" is the most important part to me.

draft-release-notes.md
```diff
@@ -0,0 +1,354 @@
...
+
+- The Example Object (used in `examples` fields) gets two new fields: `dataValue` and `serializedValue`.
+- `dataValue` describes the example in structured format.
+- `serializedValue` shows how the example will be formatted when it is sent/received by the API.
+- The existing `value` field can still be used, which means that you can safely upgrade to 3.2 and then revisit these fields and update them to use either `dataValue` or `serializedValue` as appropriate.
+  Use the new fields for examples you add after upgrading to 3.2.
+- The existing `externalValue` field can still be used to give a reference to an example, but this is now clearly documented as being a serialized value.
+- There are lots of examples of the examples (ha!) and clear documentation of how the fields can be combined.
+- Summary of what to do: use `examples` over `example`, and use `dataValue` to show a clear, structured example of the data.
+  If your use case could benefit from an example of how the data will look when it's transmitted, use `serializedValue` as well.
```
handrews: I trimmed the "but this is now documented" partially for flow, but partially because it was always clear (Dan's investigation didn't find any examples of incorrect implementations).

```suggestion
- For UTF-8 compatible serializations, `serializedValue` shows how the example will be formatted when it is sent/received by the API.
- The existing `externalValue` field supports non-UTF-8-compatible serializations, and can still be used to give a reference to an example.
- The existing `value` field can still be used, which means that you can safely upgrade to 3.2 and then revisit these fields and update them to use either `dataValue` or `serializedValue` as appropriate.
  Use the new fields for examples you add after upgrading to 3.2.
- There are lots of examples of the examples (ha!) and clear documentation of how the fields can be combined.
- Summary of what to do: use `examples` over `example`, and use `dataValue` to show a clear, structured example of the schema-ready data.
  If your use case could benefit from an example of how the data will look when it's transmitted, use `serializedValue` as well.
```

draft-release-notes.md
```diff
@@ -0,0 +1,354 @@
...
+
+#### Better polymorphic support
+
+Discriminator is a great way to match the correct schema to a payload. This release gives more robustness by adding support for a default type so that unknown objects can be handled safely.
+
+- The discriminator `propertyName` can now be an optional field.
+- New field `defaultMapping` to indicate which schema to use if the `propertyName` is not set, or if the value is unrecognized.
+- No change from previous versions: use `discriminator` to hint which entry in `anyOf` or `oneOf` is expected.
+- No change from previous versions: use `mapping` to link the discriminator property value to the Schema name if they aren't an exact match.
+- Implementations now SHOULD (rather than MAY) support templates/generics using `$dynamicRef`.
```
handrews: This probably belongs somewhere else, like the miscellaneous section at the end- it's not related to discriminator, and isn't _really_ a polymorphism thing either.  Or if it stays here, it should go at the top or bottom and not among the Discriminator-specific things.

draft-release-notes.md
```diff
@@ -0,0 +1,354 @@
...
+#### In-place updates to existing specifications and standards that we reference
+
+- Update to <https://www.ietf.org/archive/id/draft-bhutton-json-schema-01.html> of JSON Schema Specification.
+- Update to <https://www.ietf.org/archive/id/draft-bhutton-json-schema-validation-01.html> of JSON Schema Validation Specification.
+- Use [RFC8529](https://tools.ietf.org/html/rfc8259) for JSON.
+- Use [RFC9110](https://tools.ietf.org/html/rfc9110) for HTTP.
+
+#### Editorial changes
+
+- Extensive additions around media types, encoding, sequential media types, SSE examples, working with binary data.
```
handrews, line 334:
```suggestion
- Re-organization of the top-level sections to group more content with the most relevant Objects and bring the start of the Objects and Fields (formerly Schema) section close to the top of the OAS.
- Substantial streamlining and simplification of document parsing, relative URI-reference resolution, and implicit connection resolution, including examples of base URI determination
- Extensive additions around media types, encoding, sequential media types, SSE examples, working with binary data.
```

draft-release-notes.md
```diff
@@ -0,0 +1,354 @@
...
+- Non-Schema examples no longer "override" Schema examples; tools are free to use the most appropriate example for any given use case.
+- A new key `mediaTypes` is supported under `components` to support re-use of Media Type Objects.
+
+#### In-place updates to existing specifications and standards that we reference
+
+- Update to <https://www.ietf.org/archive/id/draft-bhutton-json-schema-01.html> of JSON Schema Specification.
+- Update to <https://www.ietf.org/archive/id/draft-bhutton-json-schema-validation-01.html> of JSON Schema Validation Specification.
+- Use [RFC8529](https://tools.ietf.org/html/rfc8259) for JSON.
+- Use [RFC9110](https://tools.ietf.org/html/rfc9110) for HTTP.
+
```
handrews, line 329:
```suggestion
- Use [RFC9512](https://www.rfc-editor.org/rfc/rfc9512.html) for YAML, particularly for defining JSON-compatible YAML.
- Use [RFC9110](https://tools.ietf.org/html/rfc9110) for HTTP.
```

draft-release-notes.md
```diff
@@ -0,0 +1,354 @@
...
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  type: string
+```
+
+#### Minor updates that are worth a mention
+
```
handrews, line 318: ***If*** PR #4927 is merged (probably needs rewriting as this is my perspective):

```suggestion

- The `openapi` field now only requires the major and minor version (e.g. `openapi: "3.2"`) to emphasize that patch within a minor release line clarify and replace earlier patch releases, and are not distinct versions with their own behavior.
```

summary-release-notes.md
handrews: @karenetheridge after your Slack comments about an hour or so ago, I re-read this suggestion again, and my comment stands.

Also, why do you object to just dropping this line from the change log?  What benefit is there here?

draft-release-notes.md
lornajane: Note to self: adapt this for grammar (does not encoding -> does not encode)

handrews: @lornajane if you want a fixed suggestion (but feel free to do your own):

```suggestion
- Additional `style` option `cookie` for content in a cookie, which conforms to RFC6265 by using a semicolon followed by a space as a separator and by not encoding data values.
```

draft-release-notes.md
karenetheridge: We also don't say "mediatype" either.

summary-release-notes.md
karenetheridge: I don't object. I made these comments before you suggested dropping them.

draft-release-notes.md
karenetheridge: Right, "base URI" is vague as it doesn't explain _which_ base, and there are multiple: the document's `$self`, the retrieval URI, the server url (which still has to go through a resolution step even though it feels like it's concatenated, and that's a debate that we need to have later, not here), and the URI of the runtime request (for operations that are relevant for a live request), so that term is best avoided when possible to be more specific.

summary-release-notes.md
handrews: @karenetheridge apologies, it's gotten a little hard to follow it all!

draft-release-notes.md
handrews: @karenetheridge agreed

draft-release-notes.md
lornajane: Good catch. I'll reword I think, but I had to go back and double check how that part works (useful as this is a headline feature for the talk version of this material too!)

draft-release-notes.md
handrews: @karenetheridge the Server Object's `url` field does not use a "base URI" in the RFC3986 sense at all.  It uses the retrieval URI of the document, which _can_, **coincidentally**, be the base URI, but is not always (there is only one base URI for the resource, there are just multiple possible sources from which it might come, some of which might be base URIs _for other sources in the list_ but not for the document as a whole).

lornajane:
Thank for all the brilliant reviews!! Split a bunch of the content out to become some new docs: https://github.com/OAI/learn.openapis.org/pull/137

draft-release-notes.md
karenetheridge: That doesn't make sense at all. Let's pick this up when you're back. I think we are not using this terminology in the same way.

draft-release-notes.md
karenetheridge: I would change "with their own behavior" to "with intentionally distinct behaviour", to allow for unintentionally distinct behaviour (which is the reason why referencing a point version can be significant).

karenetheridge: (the change from Simplified English to Traditional English is of course intentional) ;)

draft-release-notes.md
```diff
@@ -0,0 +1,357 @@
...
+#### Tags: nested, multipurpose, and not just for documentation
+
+Tags get an upgrade, with some new fields to make them more useful and reflect some of the features that need extensions to achieve today:
+
+- `summary` field to allow short descriptions, used when displaying lists of tags. If you use the `x-displayName` extension, you can now use `summary` instead.
+- `parent` field to point to the tag that this tag is nested under. If you use `x-tagGroups`, adjust to use a tags hierarchy.
+- `kind` field to allow a tag to be classified into a category such as navigation, or audience.
+   The `kind` field is free-form text, however there are some expected/conventional values such as `nav` (in line with the most common current usage as grouping for documentation output).
+- A [registry](https://spec.openapis.org/registry/tag-kind/index.html) to establish conventions for values used in `kind`.
```
karenetheridge, line 21: Pulled this out of the bullets as the registry isn't a new field.

```suggestion

And we've added a [registry](https://spec.openapis.org/registry/tag-kind/index.html) to establish conventions for values used in `kind`.
```

draft-release-notes.md
```diff
@@ -0,0 +1,357 @@
...
+    description: Limited-time seasonal offerings
+    kind: badge
+    externalDocs:
+      description: Seasonal menu planning guide
+      url: https://docs.cakestore.com/seasonal-planning
+```
+
+#### Support additional HTTP methods
+
+- Support the new `query` method alongside the existing `get`/`post`/`put`/`delete`/`options`/`head`/`patch`/`trace`.
```
karenetheridge, line 55: Added "field" to distinguish the field name from the actual method name which is QUERY.

```suggestion
- Support the new `query` method field alongside the existing `get`/`post`/`put`/`delete`/`options`/`head`/`patch`/`trace`.
```

draft-release-notes.md
```diff
@@ -0,0 +1,357 @@
...
+  QUERY is a [draft standard](https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-11.html) new HTTP method nearing final publication, which is designed to allow complex filtering query data to be sent in the body of a request when it doesn't fit in the query string.
+- Under an `additionalOperations` entry in a Path, use any other methods not listed as keys using the correct capitalization, e.g. you can add `LINK` under this field, but do NOT add `HEAD`; use the existing Operation Object `head` field.
```
karenetheridge, line 57:
```suggestion
- Under an `additionalOperations` entry in a Path Item Object, use any other methods not listed as keys using the correct capitalization, e.g. you can add `LINK` under this field, but do NOT add `HEAD`; use the existing Operation Object `head` field.
```

draft-release-notes.md
```diff
@@ -0,0 +1,357 @@
...
+
+The following example describes both QUERY and LINK methods for the `cakes/` endpoint:
```
karenetheridge, line 59:
```suggestion
The following example describes both QUERY and LINK methods for the `/cakes` endpoint:
```

summary-release-notes.md
```diff
@@ -0,0 +1,132 @@
...
+    - `summary` field to allow short descriptions, used when displaying lists of tags.
+    - `parent` field to point to the tag that this tag is nested under.
+    - `kind` field to allow a tag to be classified into a category such as navigation, or audience.
+       The `kind` field is free-form text, however there are some expected/conventional values such as `nav` (in line with the most common current usage as grouping for documentation output).
+    - A [registry](https://spec.openapis.org/registry/tag-kind/index.html) to establish conventions for values used in `kind`.
+
+- Support for additional HTTP methods
+
+    - Support the new `query` method alongside the existing `get`/`post`/`put`/`delete`/`options`/`head`/`patch`/`trace`.
+    - Under an `additionalOperations` entry in a Path, use any other methods not listed as keys using the correct capitalization, e.g. `LINK`. Do NOT add `HEAD` under this, use the existing sibling `head`.
```
karenetheridge, line 20: as in the main document:

```suggestion
    - Under an `additionalOperations` entry in a Path Item Object, use any other methods not listed as keys using the correct capitalization, e.g. you can add `LINK` under this field, but do NOT add `HEAD`; use the existing Operation Object `head` field.
```

summary-release-notes.md
```diff
@@ -0,0 +1,132 @@
...
+
+- Document identity and URL resolution
+
+    - Additional top-level field `$self` is added to allow users to define the base URI of the document, used to resolve relative references.
+    - More explanation and examples regarding URL resolution.
```
karenetheridge, line 25:
```suggestion
    - More explanation and examples regarding URI resolution.
```

summary-release-notes.md
ralfhandl: "_and_ use the existing" instead of a
semicolon?

lornajane:
Thanks for all your input, the content became the release notes and also some documentation updates, so the pull requests is no longer needed

PR #4610, opened 2025-05-18
v3.2: Support summary alongside every description

### LINKED ISSUE #1591: Suggestion for new 'summary' field on 'response' object
I sometimes need to write a lengthy explanation for a specific `response` (possibly with lists and other formatted text). AFAIK `description` is the only field for this, but a lengthy description here doesn't look too good in most documentation renderers I've tried (e.g. Swagger UI, ReDoc). In those tools, `response.description` seems to be intended to be kept short, almost as a "heading" for the response.

I've posted an example with a screenshot on https://github.com/Rebilly/ReDoc/issues/500.

It would be great to have a `summary` field on the `response` object, so that renderers can use this as a summary for the response, allowing `description` to be used for a lengthier explanation where needed.

handrews:
_**NOTE:** For this PR, if everyone reacts with "yeah let's do this!" then we should put it in 3.2.  If it's controversial, I'm fine with bumping it for further consideration.  I spent about 10 minutes on this PR, which was faster than asking anyone if I should do it._

Fixes #1591.

This adds a `summary` field to every Object that has a `description` field but did not already have `summary`.  As @LasneF noted in the linked issue, it makes sense for us to be consistent with these fields.

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -469,7 +469,8 @@ An object representing a Server.
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
 | <a name="server-url"></a>url | `string` | **REQUIRED**. A URL to the target host. This URL supports Server Variables and MAY be relative, to indicate that the host location is relative to the location where the document containing the Server Object is being served. Query and fragment MUST NOT be part of this URL. Variable substitutions will be made when a variable is named in `{`braces`}`. |
-| <a name="server-description"></a>description | `string` | An optional string describing the host designated by the URL. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation. |
+| <a name="server-summary"></a>summary | `string` | A short summary of the purpose of the server. |
+| <a name="server-description"></a>description | `string` | A longer description of the host designated by the URL. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation. |
 | <a name="server-name"></a>name | `string` | An optional unique string to refer to the host designated by the URL. |
 | <a name="server-variables"></a>variables | Map[`string`, [Server Variable Object](#server-variable-object)] | A map between a variable name and its value. The value is used for substitution in the server's URL template. |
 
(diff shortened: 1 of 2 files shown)


miqui (approved):
@handrews >> if everyone reacts with "yeah let's do this!" then we should put it in 3.2

+1 for this and IMHO should not be controversial itself.  I believe this makes the overall spec more consistent.

ralfhandl (approved):

lornajane:
I don't think this makes sense in every context, honestly, people don't use one field for words, so I don't think the users need two (parameters are a particular bald patch). However we already made this change for tags, and I think for security schemes, servers and responses it does make sense partly because those are often displayed in a condensed, pick-one format, which is where the summary fields are most needed.

handrews:
@lornajane my sense from the issue is that

1. It definitely makes sense for more places than we had it in 3.1
2. It is the lowest-effort sort of field to support
3. It's easier to just be consistent than to debate every possible location in detail (example:  You did not include 
4. Consistency in the data structure allows consistency in UIs and other handling

That said, whatever the TSC decides on this is fine with me.  I'm not interested in debating this Object by Object so if you and others want to I'll leave you to it.  I think this comment covers everything I have to say on this one.

lornajane:
My personal opinion is that if we're not prepared to consider this object-by-object, we should not merge this as-is, but instead revisit when we're ready to look at it all in detail.

miqui:
> honestly, people don't use one field for words,

@lornajane Not sure what you mean by this?  On more than one occasion, I have had tech writers (involved in API gov initiatives yay!!) share that summary and description are both essential components in technical documentation, (and more so when the API is to be eventually mature to be a product on its own). 

I don't in anyway presume to be a tech writer, but at least to me their point made a lot of sense.

I would rather use a summary attribute than to leverage the new name attribute in the server object, for example.

lornajane:
@miqui Sorry, let me try to make my point a bit more clearly. As a specific example, most parameter fields have very minimal descriptions, such as "The ID of the user". There usually isn't a need to write a great deal for each individual field in a request or response - although with description fields of course people can write as much or as little as needed so they work for organisations that write more, and those who write less. I'd say the same for links objects, server variables, external docs too. I also do see quite a bit of confusion on the difference between summary and description - many tools only use one or other where both are described the the specification.

By adding summary _everywhere_, we are making a change that I expect to be ignored by vendors and users alike, which devalues the value of the specification.

When proposing new fields it's common to first use the `x-` extension to see if it gets much adoption. We added `summary` for tags already, partly because there's a very widely used `x-tag-name` extension that the majority of tools support and the need was clear. A very unscientific search shows that `x-summary` does seem to get used, but only for `info` and for `responses` as the original ticket requested.

As it is written, I don't think this change should be accepted (although I'd happily support adding the summary field for responses which can require detailed explanation and are displayed in list form in many contexts). I'll close the PR but we can re-open it if someone feels strongly that this current form is the correct one.

PR #4611, opened 2025-05-18
v3.1: Clarify how to describe no content

### LINKED ISSUE #3536: Do we need to clarify what a response missing content means
I've been writing up guides for contract testing with OpenAPI in a bunch of different languages, and [the Ruby on Rails one](https://docs.bump.sh/guides/openapi/design-first-rails/) using [https://github.com/mkon/openapi_contracts/](mkon/openapi_contracts) gave this great example, letting me know an empty response has been defined, but a response has been provided by the API implementation.

```
1) widgets POST /widgets responds with 201 when valid
    Failure/Error: expect(response).to match_openapi_doc(OPENAPI_DOC)
      * Expected empty response body
    # ./spec/requests/widgets_spec.rb:17:in `block (3 levels) in <top (required)>'
```

Here's some OpenAPI for visual people:

```yaml
post:
  summary: Create Widget
  operationId: create-widget
  requestBody:
    description: Widget to create
    required: true
    content:
      application/json:
        schema:
          $ref: "../components/schema/widget.yaml"
  responses:
    '201':
      description: Created
```

Does this mean "There is definitely no content" or "I haven't bothered defining it but dont worry about it"?

There are ways to define a "dont worry about it" thats a bit more specific, like: 

```
post:
  summary: Create Widget
  operationId: create-widget
  requestBody:
    description: Widget to create
    required: true
    content:
      application/json:
        schema:
          $ref: "../components/schema/widget.yaml"
  responses:
    '201':
      description: Created
      content:
        application/json:
          schema:
            type: object
```

Or you can pop an example in instead of worrying about defining a schema but still providing something tangible for docs/mocks to think about.

Scrabbling round the spec I coudn't find anything in 3.0 or 3.1 but the [SmartBear Guide on Describing Responses](https://swagger.io/docs/specification/describing-responses/#empty) does explicitly say: 

> Some responses, such as 204 No Content, have no body. To indicate the response body is empty, do not specify a content for the response: 

This leads me to think all responses missing a `content` should have no body, because otherwise there's going to be some weird logic going "If 204 then missing means definitely no body but otherwise it just means I dunno!" 

Would that be a breaking change to 3.1.x if its clarifying intent? Or does this have to go to Moonwalk?

handrews:
Fixes #3536.

We decided that the lack of `content` doesn't guarantee a lack of response content, but that noting a `Content-Length` of `0` does.  @karenetheridge pointed out that we can set the `Content-Length` header to clarify this.

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -2079,16 +2079,27 @@ headers:
       type: integer
 ```
 
-Response with no return value:
+Response with no return value (note that the `Content-Length` header is not marked as required, as some implementations might not set it at all if there is no content):
 
 ```json
 {
   "description": "object created"
+  "headers": {
+    "Content-Length": {
+      "schema": {
+        "const": 0
+      }
+    }
+  }
 }
 ```
 
 ```yaml
 description: object created
+headers:
+  Content-Length:
+    schema:
+      const: 0
 ```
 
 #### Callback Object


handrews:
I'm a little uncertain on this because the HTTP spec is very squirrelly around `Content-Length`, but it seems to be the best way to set expectations given the current options.

src/oas.md
```diff
@@ -2079,7 +2079,7 @@ headers:
       type: integer
 ```
 
-Response with no return value:
+A response with no return value can be specified by omitting the `content` field, although this technically does not enforce a lack of content, it merely indicates that none is documented:
```
karenetheridge, line 2082: "no return value" is not very clear (some might say a response's "return value" is its HTTP response code) -- we should say "with no message body", or perhaps "zero length response body".

handrews:
@karenetheridge I'm starting to question this entire approach.

I know that we generally say that OADs are not comprehensive, and something that is not documented is not necessarily excluded (paging @mikekistler, @earth2marsh ).

But by this logic, having a Response Object that only documents `application/json` would also allow, say, `application/xml`.  Or `text/html`.  Which is a feature rather than a bug, as you might document some 400 responses with known JSON payloads, but an intermediary might send one back with a `text/html` response.  This is not _wrong_ exactly, but was also not anticipated (or anticipatable) by the API designer.

So I think that the original advice to just omit `content` should send just a clear of a signal that no content is expected as specifying `application/json` content only signals that XML, HTML, etc. are not expected.

And if that is the case, then I feel like getting into the use of `Content-Length` is getting too deep in the weeds.  Particularly given that RFC9110 makes it clear that you can't rely on it (many no-content responses won't include `Content-Length`, but chunked responses don't include it either).  People who really want to get detailed with `Content-Length` are likely people who can figure this out already.  And it's a bit misleaing/confusing to everyone else. Maybe?  I think?

I'm moving this to draft to indicate that I don't think it's ready for merge.

karenetheridge:
Sure, leaving out `content` entirely sends the signal that the body is not important -- for documenting APIs for consumers, that's totally fine.

But what about server implementers, if they want to verify that there is definitely no message body?  That's a valid usecase (at $lastjob, I built exactly such a thing into the test framework for a server application), and it would be nice for there to be a way to signal "if there is a message body, then this is invalid".

But yes, this doesn't have to go into the spec. I think we have an issue/discussion somewhere discussing a set of docs that would act as a cookbook? I  might be thinking of JSON Schema.

handrews:
@karenetheridge the use case is definitely valid!  It's just that (as you note) this doesn't  _have_ to go in the spec to be usable, and the fact that `Content-Length` is not a sure-fire way to verify this has me thinking that it shouldn't.  An article about interesting or non-obvious content modeling cases on the learn site would probably be a good thing, though.

To really enforce it, we'd need to add some sort of additional field, which would mean 3.2/3.3.

It sounds to me like you're OK with closing this, though?  It is a 100% valid use case, I just don't think this is clear and reliable enough to go into the spec.

handrews:
Closing in favor of OAI/learn.openapis.org#132.

PR #4613, opened 2025-05-18
v3.1.2: Allow `xml` field/keyword in any schema as a MAY

handrews:
Suggested by @ralfhandl in https://github.com/OAI/OpenAPI-Specification/pull/4576#discussion_r2090451308 on the grounds that some vendors already do this anyway.

This allows implementations to support the `xml` Schema Object field / JSON Schema extension keyword in any Schema Object, with a recommendation that the behavior be forward-compatible with OAS 3.2.

This does not break 3.1 compatibility, but it is a bit unusual as a patch release change as it does allow for additional functionality, and unlike the additional MAY guidance on `$dynamicRef` in 3.1.1, it is not just adding guidance to an area where we were silent.

On the other hand, we can look at it as a "we're not adding a new feature, we're just telling you how you can be compatible with future changes" thing.  idk how that fits with our compatibility story.

I'm fine with including this in 3.1.2, and I'm fine rejecting it and just letting people figure out that they can, of course, "backport" 3.2 functionality as "extensions".  But I figured it was worth a quick write-up based on @ralfhandl 's comment.

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -2749,7 +2749,7 @@ JSON Schema implementations MAY choose to treat keywords defined by the OpenAPI 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
 | <a name="schema-discriminator"></a>discriminator | [Discriminator Object](#discriminator-object) | Adds support for polymorphism. The discriminator is used to determine which of a set of schemas a payload is expected to satisfy. See [Composition and Inheritance](#composition-and-inheritance-polymorphism) for more details. |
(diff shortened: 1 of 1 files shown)


lornajane:
I'm really not sure we should allow these fields in new places in a patch release. I like the idea, are we doing it in 3.2 anyway?

handrews:
@lornajane We are doing it in 3.2.  Whatever you and @ralfhandl can agree on is fine with me, I posted this because he advocated for it.

ralfhandl:
Let's do this in 3.2 and not in 3.1.2

handrews:
Seems like we can close this- thanks for sorting it out, folks!

PR #4621, opened 2025-05-20
docs: adds missing tag definition causing document to be invalid

baywet:
reason https://github.com/microsoft/kiota/actions/runs/15145628198/job/42580662697?pr=6481
related https://github.com/microsoft/kiota/pull/6481
this description is missing the tag definition, which technically makes it invalid.
This is both an effort to "clean up the documents" and unblock our ci with upgrade of oai.net
If you don't want that change, that's fine, I'll make a local copy for our CI


baywet:
nevermind, this is actually valid

PR #4627, opened 2025-05-27
v3.1.2: Style examples cannot include `?`

handrews:
_[**EDIT:** Due to the complex and confusing history, it would be far better to go with #4647 which aligns this table with new example fields that have well-defined behavior.]_

We say that the example table is "as would be shown with the `example` or `examples` keywords", but then we include the `?` which is not correct for a single parameter with `in: query`.


handrews:
I'm going to put this in draft b/c I want to make a more coherent proposal after discussions in #4622.  We may or may not need this separate PR after that.

handrews:
See PR #4647 for a much more thorough treatment that describes what is going on much better.  Even if we do this PR instead, I need to update it a bit based on what I learned doing the other one.

handrews:
I'm closing this in favor of #4647 given that PRs positive reception in this morning's call

PR #4630, opened 2025-05-28
v3.2: Add "summary" field to Response Object

### LINKED ISSUE #1591: Suggestion for new 'summary' field on 'response' object

handrews:
Fixes #1591 (the initial request, not the expanded idea for more `summary` fields).

This is extracted from #4610 where @lornajane noted that we were not considering actual usage with the broader proposal, but that a `summary` field on the Response Object would be reasonable.  So this is just for the Response Object.

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -1831,6 +1831,7 @@ Describes a single response from an API operation, including design-time, static
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
+| <a name="response-summary"></a>summary | `string` | A short summary of the meaning of the response. |
 | <a name="response-description"></a>description | `string` | A description of the response. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation. |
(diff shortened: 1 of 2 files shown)


handrews:
Since we had approval of the general concept with the earlier PR, I'm going to go ahead and merge even though this has only been up for a day- no one previously objected for the Response Object, just to the other Objects.

Issue #4595, opened 2025-05-16
Prepare 3.2.0 and 3.1.2 releases


lornajane:
Creating this tracking issue for the tasks that need doing around the release.

- [ ] Release notes for 3.2.0 and 3.1.2
- [ ] Announcement blog post for releases
- [ ] Check that the 3.2 schema is in sync with the spec changes

These are the pull requests we won't release without (other PRs may still merge via the normal processes, these are must-haves)
 
 - https://github.com/OAI/OpenAPI-Specification/pull/4556
 - https://github.com/OAI/OpenAPI-Specification/pull/4589

mikekistler:
The releases have been published so I think we can close this one out.

PR #4647, opened 2025-06-02
v3.??? Data vs Serialized Examples in Example Objects

handrews:
This is most of what we would do for the [first and most essential](https://github.com/OAI/OpenAPI-Specification/discussions/4622#discussioncomment-13311493) of three Example Object related proposals I went over in the last TDC call.  I'm posting it because:

1.  I cannot remember the last time I learned this much writing a PR.  Using these new example fields helped me find errors in the examples and think through the spec _and_ user implications more clearly.  I think this shows the tremendous value these changes would bring compared to the confusing an inconsistently-supported status quo.
2. This change fits the 3.2 theme of better data modeling and serialization.  Without these changes, the other changes we are adding are harder to understand and use correctly.  With these changes, _most_ complex modeling and serializaton scenarios can now be shown clearly (the others require the other two parts of this overall proposal, but we could take just this part and not those and still add tremendous value).

Please look through this and particularly focus on how this allowed me to change our in-spec examples to make many things more clear, _and_ show how OAD authors can similarly make their intent clear.

Note that the changes to the XML examples don't (I think) assume the `nodeType` change in PR #4592, but I might have lost track of that if you see something strange there.  The point is that this helps tremendously in showing expected XML behavior.

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -2325,7 +2325,7 @@ content:
     examples:
       Red:
         externalDataValue: ./examples/2-by-2-red-pixels.png
-        serializedDataValue: ./examples/2-by-2-red-pixels.png
+        externalSerializedValue: ./examples/2-by-2-red-pixels.png
 ```
 
 ###### Boolean Query Parameter Examples
--- a/src/oas.md
+++ b/src/oas.md
@@ -1757,7 +1757,7 @@ requestBody:
               "zip": "99999+1234"
             }
           }
-          serializedValue: ./examples/urlencoded
+          externalSerializedValue: ./examples/urlencoded
 ```
 
 Where `./examples/urlencoded` would be:
--- a/src/oas.md
+++ b/src/oas.md
@@ -994,7 +994,7 @@ Serializing with `schema` is NOT RECOMMENDED for `in: "cookie"` parameters, `in:
 | <a name="parameter-explode"></a>explode | `boolean` | When this is true, parameter values of type `array` or `object` generate separate parameters for each value of the array or key-value pair of the map. For other types of parameters this field has no effect. When [`style`](#parameter-style) is `"form"`, the default value is `true`. For all other styles, the default value is `false`. Note that despite `false` being the default for `deepObject`, the combination of `false` with `deepObject` is undefined. |
 | <a name="parameter-allow-reserved"></a>allowReserved | `boolean` | When this is true, parameter values are serialized using reserved expansion, as defined by [RFC6570](https://datatracker.ietf.org/doc/html/rfc6570#section-3.2.3), which allows [RFC3986's reserved character set](https://datatracker.ietf.org/doc/html/rfc3986#section-2.2), as well as percent-encoded triples, to pass through unchanged, while still percent-encoding all other disallowed characters (including `%` outside of percent-encoded triples). Applications are still responsible for percent-encoding reserved characters that are not allowed by the rules of the `in` destination or media type, or are [not allowed in the path by this specification](#path-templating); see Appendices [C](#appendix-c-using-rfc6570-based-serialization) and [E](#appendix-e-percent-encoding-and-form-media-types) for details. The default value is `false`. |
 | <a name="parameter-schema"></a>schema | [Schema Object](#schema-object) | The schema defining the type used for the parameter. |
-| <a name="parameter-example"></a>example | Any | **Deprecated**. Example of the parameter's potential value; see [Working With Examples](#working-with-examples). |
+| <a name="parameter-example"></a>example | Any | Example of the parameter's potential value; see [Working With Examples](#working-with-examples). |
 | <a name="parameter-examples"></a>examples | Map[ `string`, [Example Object](#example-object) \| [Reference Object](#reference-object)] | Examples of the parameter's potential value; see [Working With Examples](#working-with-examples). |
 
 See also [Appendix C: Using RFC6570-Based Serialization](#appendix-c-using-rfc6570-based-serialization) for additional guidance.
@@ -1276,7 +1276,7 @@ See [Working With Examples](#working-with-examples) for further guidance regardi
 | ---- | :----: | ---- |
 | <a name="media-type-schema"></a>schema | [Schema Object](#schema-object) | A schema describing the complete content of the request, response, parameter, or header. |
 | <a name="media-type-item-schema"></a>itemSchema | [Schema Object](#schema-object) | A schema describing each item within a [sequential media type](#sequential-media-types). |
-| <a name="media-type-example"></a>example | Any | **Deprecated**. Example of the media type; see [Working With Examples](#working-with-examples). |
+| <a name="media-type-example"></a>example | Any | Example of the media type; see [Working With Examples](#working-with-examples). |
 | <a name="media-type-examples"></a>examples | Map[ `string`, [Example Object](#example-object) \| [Reference Object](#reference-object)] | Examples of the media type; see [Working With Examples](#working-with-examples). |
 | <a name="media-type-encoding"></a>encoding | Map[`string`, [Encoding Object](#encoding-object)] | A map between a property name and its encoding information, as defined under [Encoding Usage and Restrictions](#encoding-usage-and-restrictions).  The `encoding` field SHALL only apply when the media type is `multipart` or `application/x-www-form-urlencoded`. If no Encoding Object is provided for a property, the behavior is determined by the default values documented for the Encoding Object. |
 
@@ -1920,7 +1920,10 @@ requestBody:
       examples:
         multipart:
           description: |
-            The input example data for the 
+            The input example data for the serialized value
+            is taken from the Schema Object `examples` fields
+            for the "id" and "addresses" field, and is a solid
+            red 2x2 pixel PNG for the "profileImage" field.
           externalSerializedValue: ./examples/multipart
 ```
 
@@ -2216,8 +2219,8 @@ As further explained under [Working With Examples](#working-with-examples), this
 | <a name="example-external-data-value"></a>externalDataValue | `string` | A URI that identifies the data example in a separate document, allowing for values not easily expressed in JSON or YAML.  This is usually only needed when working with binary data. The value MUST be valid according to the relevant Schema Object. If this field is present, then `dataValue`, `value`, and `externalValue` MUST be absent. See also the rules for resolving [Relative URI References](#relative-references-in-api-description-uris). |
 | <a name="example-serialized-value"></a>serializedValue | `string` | An example of the serialized form of the value, including encoding and escaping as described under [Validity of Data and Serialized Examples](#validity-of-data-and-serialized-examples).  If `dataValue` or `externalDataValue` are present, then this field SHOULD contain the serialization of the given data.  Otherwise, it SHOULD be the valid serialization of a data value that itself MUST be valid as described for `dataValue`.  This field SHOULD NOT be used if the serialization format is JSON, as the data form is easier to work with. If this field is present, `externalSerializedValue`, `value`, and `externalValue` MUST be absent. |
 | <a name="example-eternal-serialized-value"></a>externalSerializedValue | `string` | A URI that identifies the serialized example in a separate document, allowing for values not easily or readably expressed in JSON or YAML strings.  If `dataValue` or `externalDataValue` are present, then this field SHOULD identify a serialization of the given data.  Otherwise, the value SHOULD be a valid serialization as described for `serializedValue`.  If this field is present, `serializedValue`, `value`, and `externalValue` MUST be absent. See also the rules for resolving [Relative References](#relative-references-in-api-description-uris). |
-| <a name="example-value"></a>value | Any | **Deprecated**: It is implementation-defined whether this field is used for data or serialized forms. Embedded literal example. The `value` field and `externalValue` field are mutually exclusive. To represent examples of media types that cannot naturally represented in JSON or YAML, use a string value to contain the example, escaping where necessary. |
-| <a name="example-external-value"></a>externalValue | `string` | **Deprecated**: It is implementation-defined whether this field is used for data or serialized forms. A URI that identifies the literal example. This provides the capability to reference examples that cannot easily be included in JSON or YAML documents. The `value` field and `externalValue` field are mutually exclusive. See also the rules for resolving [Relative URI References](#relative-references-in-api-description-uris). |
+| <a name="example-value"></a>value | Any | Embedded literal example. The `value` field and `externalValue` field are mutually exclusive. To represent examples of media types that cannot naturally represented in JSON or YAML, use a string value to contain the example, escaping where necessary. |
+| <a name="example-external-value"></a>externalValue | `string` | A URI that identifies the literal example. This provides the capability to reference examples that cannot easily be included in JSON or YAML documents. The `value` field and `externalValue` field are mutually exclusive. See the rules for resolving [Relative References](#relative-references-in-api-description-uris). |
 
 This object MAY be extended with [Specification Extensions](#specification-extensions).
 
@@ -2225,11 +2228,15 @@ This object MAY be extended with [Specification Extensions](#specification-exten
 
 Due to the history of example fields in both this specification and JSON Schema, there are a number of different ways to specify examples.
 
-Outside of [Schema Objects](#schema-object), it is RECOMMENDED to use Example Objects with the `examples` (plural) fields available in the [Parameter](#parameter-object), [Header](#header-object), [Media Type](#media-type-object), and [Encoding](#encoding-object) Objects.
+In the case of the non-Schema Object `example` (singular) fields, and of the `value` and`externalValue` fields of the Example Object, the example value SHOULD be compatible with the schema of its associated value.
+Because of variations in tooling support for `value` and `externalValue` and ambiguity regarding what can be "easily" represented in JSON or YAML, the `dataValue` `externalDataValue`, `serializedValue`, and `externalSerializedValue` fields have been introduced.
+ For compatibility requirements of these new fields, see [Validity of Data and Serialized Examples](#validity-of-data-and-serialized-examples).
+
+Tooling implementations MAY choose to validate compatibility automatically, and reject the example value(s) if incompatible.
 
 ###### Example Objects vs Other Fields Allowing Examples
 
-However, there are several other ways to provide examples: The deprecated `example` (singular) field that is mutually exclusive with `examples` in all Objects that support both fields, and two keywords (the deprecated singular `example` and the current plural `examples`, which takes an array of examples) in the [Schema Object](#schema-object).
+However, there are several other ways to provide examples: The `example` (singular) field that is mutually exclusive with `examples` in all Objects that support both fields, and two keywords (the deprecated singular `example` and the current plural `examples`, which takes an array of examples) in the [Schema Object](#schema-object).
 When a schema is present alongside of the non-Schema Object `example` and `examples`, this results in four possible ways to specify an example.
 Each of these fields has slightly different considerations.
 
@@ -2237,15 +2244,8 @@ The Schema Object's fields are used to show example values without regard to how
 The `examples` array is part of JSON Schema and is the preferred way to include examples in the Schema Object, while `example` is retained purely for compatibility with older versions of the OpenAPI Specification.
 
 The fields outside of the Schema Object can be used to show how data in either or both of serialized (`serializedValue`, `externalSerializedValue`) or unserialized (`dataValue`, `externalDataValue`) forms.
-However, the `example` (singular) field is deprecated as it shares the same difficulties as the likewise deprecated `value` and `externalValue` fields of the Example Object.
 
 Because examples using these non-Schema Object fields can show both the data and its serialization more clearly, they SHALL _override_ any `example` in the corresponding Schema Object.
-
-###### Deprecated Example Object Fields
-
-Historically, the Example Object's `value` and `externalValue` field and the non-Schema Object singular `example` fields were intended to show examples of the serialized form, while allowing JSON or YAML examples to be included inline rather than as serialized strings.
-Due to some ambiguity in this guidance, these fields have been implemented in several different ways.
-To preserve existing tooling behavior, these fields have been deemed to have implementation-defined behavior, and are deprecated in favor of the new unambiguous fields that allow OAD authors to choose which form(s) to show.
 
 ###### Validity of Data and Serialized Examples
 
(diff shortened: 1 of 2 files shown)


src/oas.md
```diff
@@ -1125,6 +1140,10 @@ schema:
   additionalProperties:
     type: integer
 style: form
+examples:
+  freeForm:
+    dataValue: {"yeah": "I'm", "free": "forming"}
```
mikekistler, line 1145: This doesn't look right to me. Doesn't `additionalProperties: type: integer` mean that all the properties in this object must have integer values?

handrews: @mikekistler 

🤦

handrews:
@mikekistler fixed that example- it had been done with integers but those aren't interesting serialization cases!  So I switched to strings.

I also added some text about it being easier/less ambiguous to parse the serialized example and validate the data in many cases than to try to go the other way around.

src/oas.md
```diff
@@ -1157,13 +1176,17 @@ content:
...
-      "flag": null
-    }
+    examples:
+      minimized:
+        summary: JSON should be serialized with minimal whitespace
+        dataValue: {
+          "numbers": [1, 2],
+          "flag": null
+        }
+        serializedValue: '{"numbers":[1,2],"flag":null}'
```
mikekistler, line 1186: The description of `serializedValue` below says it is "An example of the serialized form of the value, including all relevant encoding and escaping." So why isn't the value shown here the same as what is shown on line 1192?

handrews: @mikekistler I'll update the text, as it should really be something like "all relevant encoding and escaping for the Object containing the `examples` field"

This example is in a Media Type Object, so it is only serialized as far as the Media Type Object specifies.  To show `%7B%22numbers%22%3A%5B1%2C2%5D%2C%22flag%22%3Anull%7D` the example would need to be in the Parameter Object, which is what applies that level of encoding/escaping.  See PR #4647 for that.

This is important for two reasons:

1. It allows implementations to handle each Object the same regardless of its context.  A Media Type Object is a Media Type Object, and the only context needed is the parent key which says which media type is being described.
2. It allows showing intermediate serialization concerns.  In this example, we are showing the the application recommends removing unnecessary whitespace.  That's a bit contrived, but in `multipart` scenarios, we can end up with complex rules at both the single-part and overall levels.

handrews:
@mikekistler I have clarified the serialization rules in the latest added commit.

src/oas.md
```diff
@@ -1972,14 +2075,26 @@ headers:
...
+        dataValue: 2
+        serializedValue: 'X-Rate-Limit-Remaining: 2'
   X-Rate-Limit-Reset:
     description: The number of seconds left in the current period
     schema:
       type: integer
+    examples:
+      oneMinute:
+        dataValue: 60
+        serializedValue: 'X-Rate-Limit-Reset: 60'
```
mikekistler, line 2100: I love that you have added all these examples!

src/oas.md
```diff
@@ -2087,141 +2202,151 @@ transactionCallback:
...
 | <a name="example-description"></a>description | `string` | Long description for the example. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation. |
-| <a name="example-value"></a>value | Any | Embedded literal example. The `value` field and `externalValue` field are mutually exclusive. To represent examples of media types that cannot naturally represented in JSON or YAML, use a string value to contain the example, escaping where necessary. |
-| <a name="example-external-value"></a>externalValue | `string` | A URI that identifies the literal example. This provides the capability to reference examples that cannot easily be included in JSON or YAML documents. The `value` field and `externalValue` field are mutually exclusive. See the rules for resolving [Relative References](#relative-references-in-api-description-uris). |
+| <a name="example-data-value"></a>dataValue | Any | An example of the data structure that MUST be valid according to the relevant [Schema Object](#schema-object).  If this field is present, `externalDataValue`, `value`, and `externalValue` MUST be absent. |
```
mikekistler, line 2218: Since JSON Schemas validate "instances", should we use "instance" instead of "data structure"?

src/oas.md
```diff
@@ -2087,141 +2202,151 @@ transactionCallback:
...
+| <a name="example-eternal-serialized-value"></a>externalSerializedValue | `string` | A URI that identifies the serialized example in a separate document, allowing for values not easily or readably expressed in JSON or YAML strings.  If `dataValue` or `externalDataValue` are present, then this field SHOULD identify a serialization of the given data.  Otherwise, the value SHOULD be a valid serialization as described for `serializedValue`.  If this field is present, `serializedValue`, `value`, and `externalValue` MUST be absent. See also the rules for resolving [Relative References](#relative-references-in-api-description-uris). |
+| <a name="example-value"></a>value | Any | **Deprecated**: It is implementation-defined whether this field is used for data or serialized forms. Embedded literal example. The `value` field and `externalValue` field are mutually exclusive. To represent examples of media types that cannot naturally represented in JSON or YAML, use a string value to contain the example, escaping where necessary. |
```
mikekistler: I'm not sure it is necessary to mark this field as "Deprecated". I think we could add that in practice this field contains a value that is valid according to the relevant Schema Object.

src/oas.md
```diff
@@ -2087,141 +2202,151 @@ transactionCallback:
...
+| <a name="example-external-value"></a>externalValue | `string` | **Deprecated**: It is implementation-defined whether this field is used for data or serialized forms. A URI that identifies the literal example. This provides the capability to reference examples that cannot easily be included in JSON or YAML documents. The `value` field and `externalValue` field are mutually exclusive. See also the rules for resolving [Relative URI References](#relative-references-in-api-description-uris). |
 
 This object MAY be extended with [Specification Extensions](#specification-extensions).
 
-In all cases, the example value SHOULD be compatible with the schema of its associated value.
-Tooling implementations MAY choose to validate compatibility automatically, and reject the example value(s) if incompatible.
-
```
mikekistler: I think we should leave this language in place if we decide that `example` does not need to be deprecated.

src/oas.md
```diff
@@ -2087,141 +2202,151 @@ transactionCallback:
...
-The exact serialization and encoding is determined by various fields in the Parameter Object, Header Object, or in the Media Type Object's [Encoding Object](#encoding-object).
-Because examples using these fields represent the final serialized form of the data, they SHALL _override_ any `example` in the corresponding Schema Object.
+The fields outside of the Schema Object can be used to show how data in either or both of serialized (`serializedValue`, `externalSerializedValue`) or unserialized (`dataValue`, `externalDataValue`) forms.
+However, the `example` (singular) field is deprecated as it shares the same difficulties as the likewise deprecated `value` and `externalValue` fields of the Example Object.
 
-The singular `example` field in the Parameter, Header, or Media Type Object is concise and convenient for simple examples, but does not offer any other advantages over using Example Objects under `examples`.
+Because examples using these non-Schema Object fields can show both the data and its serialization more clearly, they SHALL _override_ any `example` in the corresponding Schema Object.
```
mikekistler, line 2248: I don't see why _overriding_ is necessary here. `examples` in a Schema object can have multiple values, and `examples` outside Schema objects can have multiple values.  Why not simply make this additive? There is no "override" -- these all can exist "in harmony".

src/oas.md
```diff
@@ -2087,141 +2202,151 @@ transactionCallback:
...
 
-Some examples cannot be represented directly in JSON or YAML.
-For all three ways of providing examples, these can be shown as string values with any escaping necessary to make the string valid in the JSON or YAML format of documents that comprise the OpenAPI Description.
-With the Example Object, such values can alternatively be handled through the `externalValue` field.
+###### Deprecated Example Object Fields
 
-##### Example Object Examples
+Historically, the Example Object's `value` and `externalValue` field and the non-Schema Object singular `example` fields were intended to show examples of the serialized form, while allowing JSON or YAML examples to be included inline rather than as serialized strings.
```
mikekistler: I can't speak to the "original intent" for these fields -- I wasn't involved at the time. But I wonder if this was really consciously decided as the intent, and then somehow omitted from the spec, or if instead it was not really understood that there is a difference between the data value and serialized value and that is why the distinction was not included in the spec.

src/oas.md
```diff
@@ -2087,141 +2202,151 @@ transactionCallback:
...
+        dataValue:
+          author: A. Writer
+          title: An Older Book
+          rating: 4.5
+        serializedValue: |
+          {
+            "author": "A. Writer",
+            "title": "An Older Book",
+            "rating": 4.5
+          }
```
mikekistler, line 2314: Should this serialized value be a string, which condensed white-space (as was done above)?

src/oas.md
```diff
@@ -2434,7 +2559,7 @@ The `example` and `examples` fields are mutually exclusive, and if either is pre
 | <a name="header-style"></a>style | `string` | Describes how the header value will be serialized. The default (and only legal value for headers) is `"simple"`. |
 | <a name="header-explode"></a>explode | `boolean` | When this is true, header values of type `array` or `object` generate a single header whose value is a comma-separated list of the array items or key-value pairs of the map, see [Style Examples](#style-examples). For other data types this field has no effect. The default value is `false`. |
 | <a name="header-schema"></a>schema | [Schema Object](#schema-object) \| [Reference Object](#reference-object) | The schema defining the type used for the header. |
-| <a name="header-example"></a>example | Any | Example of the header's potential value; see [Working With Examples](#working-with-examples). |
+| <a name="header-example"></a>example | Any | **Deprecated**. Example of the header's potential value; see [Working With Examples](#working-with-examples). |
```
mikekistler: Again, I'm not convinced that we need to deprecate "example".

src/oas.md
```diff
@@ -3207,67 +3336,120 @@ The `namespace` field is intended to match the syntax of [XML namespaces](https:
...
+  schema:
+    type: object
+    xml:
+      name: document
+    properties:
+      animals:
+        type: string
+  examples:
+    pets:
+      dataValue: "dog, cat, hamster"
```
mikekistler: This does not look right to me. I think `dataValue` should have "animals" in it. Otherwise, how could I have an schema with two properties, "animals" and "owners", and show that correctly in the `dataValue`?

src/oas.md
```diff
@@ -3207,67 +3336,120 @@ The `namespace` field is intended to match the syntax of [XML namespaces](https:
...
+    xml:
+      name: document
+    properties:
+      animals:
+        type: array
+        items:
+          type: string
+  examples:
+    pets:
+      dataValue: [dog, cat, hamster]
```
mikekistler: Ditto my comment above.

src/oas.md
handrews: This was language carried over from before, so I just didn't change it.  I'm open to changing it but can we make that a separate chunk of work?

src/oas.md
handrews: @mikekistler it is a string, that's what the `|` at the end of the `serializedValue: |` does (YAML block literal).

As for minimizing, I did that for the parameter because it has to be shoved into the query string.  This is presumed to be for a body, so it might be sent without that minimizing.  But TBH I was just trying out all sorts of different things, I'm not super-attached to this.

src/oas.md
handrews: @mikekistler I'm totally fine with separating any discussion of deprecating `example` from the Example Object stuff.  Maybe I'll go ahead and do that before the Thursday call.

src/oas.md
handrews: @mikekistler I think you're right.  Trying to work these out made me think of so many things!  But yes I got this one wrong I'll update.

src/oas.md
handrews: I couldn't decide which language was better.  Easy to tweak at any point- I'll think about it as I do the other updates. _[**EDIT:** I decided "instance" was too JSON Schema jargon-y, plus we go outside of the JSON Schema instance model with things like raw binary data.]_

src/oas.md
handrews: Honestly it was one of the best things about writing the PR and is what convinced me it's worth making a strong case for it.

src/oas.md
handrews: @mikekistler I'll see about separating all of the deprecating and rewording for the old fields into another PR.  They're really separate things and I don't want to get either one hung up on the other.  To me, `example`, `value`, and `externalValue` are kind of all the same thing, but I can see handling them separately.

src/oas.md
handrews: @mikekistler I understand that tooling vendors have implemented it differently, but these quotes from 3.0.3 seem pretty clear to me, and we've debated this before and always agreed that the encoded/serialized form was the intent.  AFAIK, this effort by you right now is the first time anyone deeply involved has taken the opposite position in terms of what the wording means:

From the Parameter Object:

> When `example` or `examples` are provided in conjunction with the schema object, the example MUST follow the prescribed serialization strategy for the parameter.

From the Media Type Object `example` field:

> The example object SHOULD be in the correct format as specified by the media type.

From the Media Type Object `examples` field:

> Examples of the media type. Each example object SHOULD match the media type and specified schema if present. 

From the Example Object `value` field:

> To represent examples of media types that cannot naturally represented in JSON or YAML, use a string value to contain the example, escaping where necessary.

From the Example Object `externalValue` field:

> This provides the capability to reference examples that cannot easily be included in JSON or YAML documents.

From the Example Object after the fixed fields table:

> In all cases, the example value is expected to be compatible with the type schema of its associated value. Tooling implementations MAY choose to validate compatibility automatically, and reject the example value(s) if incompatible.

I honestly don't see how all of this above, taken together, can be interpreted in any other way.

handrews:
@mikekistler I have updated quite a few things, and left other things the same, with the upshot being that this PR is _only_ about adding new fields to the Example Object and showing their use:

* I have un-deprecated all of the fields, and done my best to retain the prior language around them- some language moved from just after the bit about `x-` extensions to right at the top of the "Working With Examples" section because otherwise it was just too awkward
* This means keeping the "override" language- we can talk about changing that separately as it is completely independent of whether we add new fields
* This also means _not_ providing new guidance on how to interpret the old fields, as that needs to be done whether we add new fields or not
* I decided against using "instance" as that is too jargon-y, and more a JSON Schema thing, and arguably we work with data beyond the JSON Schema instance model anyway (e.g. raw binary).  If we decide to add the fields we can debate the language, but let's decide whether the fields are worth adding first, please.
* I did fix the XML examples and various other small things.
* I'm probably forgetting something, I'll check again in the morning

lornajane:
There's a lot of support for this change and especially for the examples that are added. We need to discuss naming before merging, but we think this can go into 3.2.

src/oas.md
```diff
@@ -1699,26 +1742,33 @@ requestBody:
...
+          dataValue: {
+            "id": "f81d4fae-7dec-11d0-a765-00a0c91e6bf6",
+            "address": {
+              "streetAddress": "123 Example Dr.",
+              "city": "Somewhere",
+              "state": "CA",
+              "zip": "99999+1234"
+            }
+          }
+          serializedValue: ./examples/urlencoded
```
hudlow:
```suggestion
          externalSerializedValue: ./examples/urlencoded
```

handrews:
@OAI/tsc and all, I have filed several issues to split out various related conversations and keep this PR moving to conclusion:

* #4658
    * @mikekistler @lornajane I have added this as a release-blocking issue to #4595 so that we can merge this PR with the current names while we keep debating the permanent names.  Let's not keep this held open once we agree on the behvior
* #4659 
* #4660 
* #4661 
* #4662 
* #4664

src/oas.md
```diff
@@ -1087,6 +1087,10 @@ schema:
     type: integer
     format: int64
 style: simple
+examples:
+  number:
+    dataValue: 12345678
+    serializedValue: "token: 12345678"
```
karenetheridge, line 1093: Can we add a second example, demonstrating the use of an array coming from deserializing the simple style?

```suggestion
examples:
  one number:
    dataValue: 12345678
    serializedValue: "token: 12345678"
  two numbers:
    dataValue:
      - 12345678
      - 99999999
    serializedValue: "token: 12345678,99999999"
```

src/oas.md
```diff
@@ -1424,16 +1449,30 @@ components:
...
+          "timestamp": "1985-04-12T23:20:51.37Z",
+          "level": 1,
+          "message": "Bye!"
+        }
+      ]
       # JSON Text Sequences require an unprintable character
       # that cannot be escaped in a YAML string, and therefore
       # must be placed in an external document shown below
-      externalValue: examples/log.json-seq
+      externalSerializedValue: examples/log.json-seq
```
karenetheridge, line 1467: It would be really cool if we had the html renderer detect this keyword and add a hyperlink to a real file named "./examples/log.json-seq" in the github-pages repo.

src/oas.md
```diff
@@ -2087,141 +2205,148 @@ transactionCallback:
 
 #### Example Object
 
-An object grouping an internal or external example value with basic `summary` and `description` metadata.
-This object is typically used in fields named `examples` (plural), and is a [referenceable](#reference-object) alternative to older `example` (singular) fields that do not support referencing or metadata.
+Example values along with `summary` and `description` metadata, showing either a valid data structure for use with a schema, a valid serialized form of the data, or both.
 
-Examples allow demonstration of the usage of properties, parameters and objects within OpenAPI.
+As further explained under [Working With Examples](#working-with-examples), this Object is used in fields named `examples` (plural),and is a [referenceable](#reference-object) alternative to older `example` (singular) fields that do not support referencing, metadata, or data structure vs serialized forms.
```
```suggestion
As further explained under [Working With Examples](#working-with-examples), this Object is used in fields named `examples` (plural), and is a [referenceable](#reference-object) alternative to older `example` (singular) fields that do not support referencing, metadata, or data structure vs serialized forms.
```

src/oas.md
karenetheridge: If the schema author is using the new fields, which are unambiguous, why would they also combine them with the old fields, which are ambiguous?  Tooling will know how to treat the new fields but not know the intent for the old ones.

My first preference would be to outright disallow mixing old and new together, but second would be as it is stated here - the old is overridden (i.e. content ignored) in the presence of the new.

src/oas.md
handrews: @karenetheridge as noted in the past TDC call, I'd like to handle newly added examples separately, otherwise we'll never finish.

handrews: @karenetheridge upon further thought, I'm going to split this PR up to separate the new fields from all of the extra example updates.  I'll take this request into consideration then- I think that might be better than either allowing endless new example requests here or trying to keep them out of this process entirely.

src/oas.md
```diff
@@ -2087,141 +2205,148 @@ transactionCallback:
...
 
 ##### Fixed Fields
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
 | <a name="example-summary"></a>summary | `string` | Short description for the example. |
 | <a name="example-description"></a>description | `string` | Long description for the example. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation. |
+| <a name="example-data-value"></a>dataValue | Any | An example of the data structure that MUST be valid according to the relevant [Schema Object](#schema-object).  If this field is present, `externalDataValue`, `value`, and `externalValue` MUST be absent. |
+| <a name="example-external-data-value"></a>externalDataValue | `string` | A URI that identifies the data example in a separate document, allowing for values not easily expressed in JSON or YAML.  This is usually only needed when working with binary data. The value MUST be valid according to the relevant Schema Object. If this field is present, then `dataValue`, `value`, and `externalValue` MUST be absent. See also the rules for resolving [Relative URI References](#relative-references-in-api-description-uris). |
```
hudlow, line 2219: Aside from some clarifications that could be postponed to another PR, my big concern is that I can't make heads or tails out of what a _non-serialized_ external value looks like. In my mind, external values are _inherently_ serialized: if you can provide `externalDataValue: ./image.png` and the image is treated opaquely then it implies that when you provide `externalDataValue: ./data.json` then the JSON document will be treated opaquely, and this would semantically match `dataValue: "{ ... }"` _not_ `dataValue: { ... }`.

If you _wanted_ the latter then the phrase "allowing for values not easily expressed in JSON or YAML" doesn't seem right, plus if all you want is to move a JSON data structure to an external URI, you can just use `$ref` at the usage point for the example value? Yes, that means the whole Example Object is externalized but... that seems fine?

It is my belief that we should merely clarify that `externalValue` is treated opaquely (i.e., as a serialized value) and skip adding `externalDataValue` and `externalSerializedValue`.

hudlow (changes requested):

src/oas.md
```diff
@@ -2087,141 +2205,148 @@ transactionCallback:
...
+This example shows both `externalDataValue` and `externalSerializedValue` to emphasize that no encoding is taking place, but it is also valid to show only one or the other.
+
+```yaml
+content:
+  image/png:
+    schema: {}
+    examples:
+      Red:
+        externalDataValue: ./examples/2-by-2-red-pixels.png
+        serializedDataValue: ./examples/2-by-2-red-pixels.png
```
hudlow:
```suggestion
        externalSerializedValue: ./examples/2-by-2-red-pixels.png
```

src/oas.md
```diff
@@ -2087,141 +2205,148 @@ transactionCallback:
...
+          rating: 4.5
+        serializedValue: |
+          {
+            "author": "A. Writer",
+            "title": "An Older Book",
+            "rating": 4.5
+          }
+```
+
+###### Binary Examples
```
hudlow, line 2317: I think it might also be nice to show a [`data`-URL-scheme](https://datatracker.ietf.org/doc/html/rfc2397) example. My assumption is that what URL schemes an interpreter of an OpenAPI description supports are completely out of scope, but there's no reason one couldn't support the data scheme?

Something like: 
```yaml
externalSerializedValue: "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAIAAAACCAIAAAD91JpzAAAABGdBTUEAALGPC_xhBQAAADhlWElmTU0AKgAAAAgAAYdpAAQAAAABAAAAGgAAAAAAAqACAAQAAAABAAAAAqADAAQAAAABAAAAAgAAAADO0J6QAAAAEElEQVQIHWP8zwACTGCSAQANHQEDqtPptQAAAABJRU5ErkJggg%3D%3D"`

src/oas.md
karenetheridge: `externalDataValue` acts just like the `$ref` keyword -- the serialization level of the file is the same as the referencing document, but it's in a separate file solely because of its size (presumably), or to perhaps allow it to be modified at a different cadence than the referencing document itself (although that makes less sense here when the example needs to align with the schema).

I would imagine it would be the least-used of the four new keywords, but I see no reason to omit it.

hudlow: @karenetheridge so you disagree with the text as written in this table and with the example on line 2328? If it works like this, why not just allow `dataValue` to be a reference?

src/oas.md
handrews: @hudlow can you please file an issue requesting that the OAS support this for external example fields (regardless of _which_ external fields we end up having)?  I think it is an interesting idea worth discussing but orthogonal to this PR.

src/oas.md
handrews: Because then you can't have an example that uses a property named `$ref`, this has been discussed extensively in past issues, search for an issue allowing for combining examples.

handrews:
I am closing this in favor of the following PRs, because this is too big and the conversation has a lot of complex threads that have already been resolved.

* #4671 (adds fields- @hudlow I made a section about encodings)
* #4672 (Example Object example updates- these messed up the diff if included in 4671)
* #4673 (Parameter/Header style serialization rules- I was unsure of these on the last call so I thought through the rules more and propose these)
* Numerous forthcoming PRs updating examples for Parameter, Header, Request Body, Response, Media Type, Encoding, and XML Objects

src/oas.md
hudlow: Done: https://github.com/OAI/OpenAPI-Specification/discussions/4674

Issue #2415, opened 2020-12-03
Consider keeping semver for *future* 3.x releases


pimterry:
I think I understand the v3.1 vs v4.0 decision, and while I don't love it, I'm taking that as a given.

However, I don't see that that necessitates the current versioning text in the 3.1.0 RC spec. Is there any flexibility to provide more guarantees there? I'm talking about this paragraph specifically:

> Occasionally, non-backwards compatible changes may be made in minor versions of the OAS where impact is believed to be low relative to the benefit provided.

As a developer building on top of OpenAPI, this means today I cannot write code that accepts any OAS spec inputs with versions >3.1. All code & libraries written today _must_ reject those as incompatible, because they're defined as being potentially incompatible in all possible ways.

I suspect that's suboptimal, because it seems very likely that a hypothetical v3.2, 3.3, 3.4... and other future specs will be exciting feature releases, backwards compatible with 3.1, rather than a series of new breaking changes.

It would be unfortunate if OpenAPI's versioning system unnecessarily made adoption of these future specifications more difficult by forcing today's tools and libraries to reject these compatible specifications out of hand.

I'm proposing that this paragraph could instead say something along the lines of:

> This specification breaks compatibility with previous 3.0.x specifications, in some specific cases with minimal impact and high benefit. This is intended to be a one-off occurrence. Future 3.x.x specifications MUST remain backwards compatible with this specification (v3.1.0) and tools MAY assume that future 3.x.x specifications are fully backwards compatible.

I.e. don't drop semver entirely, and instead say "just this one time, we're bending the rules".

This has a few benefits:

* Code that consumes OpenAPI specs today can make useful assumptions about more future spec versions.
* OpenAPI releases will be able to easily communicate both size and compatibility of future specifications just through the version number. Right now that's not possible (is 3.2 a shiny new feature set, or an incompatible change? Is 3.1.1 a typo fix, or a big-but-compatible change?).
* It provides clearer documentation of the current decision and intent within the spec itself.
* It provides future pressure to either a) stay backwards compatible or b) more clearly communicate breaking changes with a version bump. Staying backwards compatible where possible is clearly valuable. While I'm aware that a v4 bump was undesirable, if there will be more breaking changes in future, doing the bump and matching standard versioning expectations will eventually be preferable.

[...]

PR #4656, opened 2025-06-05
v3.2: makes the versioning and deprecation non-breaking for minor versions

baywet:
motivation: having 3.1 being a minor version on paper but in reality bringing a lot of breaking changes caused confusion and friction for consumers, tools implementers etc...
This is partly due because nowadays, a lot of the industry is following, and expects others to follow semver.
This pull request updates the verbiage of the versioning and deprecation policies so it forbids making breaking changes in minor/patch versions, so we can conform to semver, after 3.1.


lornajane (changes requested):
I am not in favour of this change.

Additionally, I find it particularly unhelpful for it to be proposed at the _end_ of a development cycle which has been worked on under different constraints. Hard no from me.

See also https://github.com/OAI/OpenAPI-Specification/issues/2415

baywet:
> I am not in favour of this change.

@lornajane besides the reasons articulated in the linked issue, do you have additional reasons why you're against such a change?

> Additionally, I find it particularly unhelpful for it to be proposed at the end of a development cycle which has been worked on under different constraints. Hard no from me.

What do you encompass as the development cycle here? 3.2? 3.X?
- In the case of 3.X, well, I wasn't here at the time :)
- In the case of 3.2, I've submitted this PR against that branch to start with, but if this ends up being accepted, I suggest we backport it to 3.1 as well.

handrews:
That's two solid "no" votes from the TSC, so I'm going to close this.

Issue #4658, opened 2025-06-05
v3.2: Decide on names of new Example Object fields


handrews:
PR #4647 adds four new fields to the Example Object.  There seems to be consensus that we should add the fields once that PR is out of draft, but there is less consensus on the names.  

In this issue, "JSON/YAML" refers to JSON (as specified by the RFC) and the subset of YAML that we allow for OAD authoring.  This subset is intended to be essentially equivalent to JSON, as all OAD keys and values MUST be possible to write in JSON.

The names in the PR are:

* `dataValue`:  An inline example value of the data structure that is validated by a Schema Object, possibly trivially if the data is raw binary as that is outside of JSON Schema's standard data model
* `externalDataValue`: Like `dataValue` but for values not possible to inline in JSON/YAML, such as a raw binary image or video
* `serializedValue`: An inline example of the fully serialized value, which MUST be a string
* `externalSerializedValue`: Like `serializedValue` but for values not possible to inline as a JSON/YAML string.

@mikekistler has questioned "data" and brought up that JSON Schema uses "instance."  My feeling is that "instance" is too jargon-y, and the data is not necessarily a proper JSON Schema instance when binary data is involved, but I could be convinced to go with it if I'm in the minority here.

handrews:
BTW I do not care what the names are.  Just lmk when there's consensus and I'm happy to write a PR to update the text with the final names.  I have just separated this out (and [made it a 3.2 release blocker](https://github.com/OAI/OpenAPI-Specification/issues/4595)) so that PRs #4647 and #4648 do not get held up on naming debates.

hudlow:
I can't usefully contribute to a naming discussion _separate from a semantics discussion_ unless the semantics are clear to me (and finalized), which they are not. I have commented with my objections, here: https://github.com/OAI/OpenAPI-Specification/pull/4671#discussion_r2141172449.

mikekistler:
Regarding "DataValue", I would prefer a more descriptive name since "data" is very generic. My first thought was "JsonValue", since we want this to match the JSON Schema.

ralfhandl:
I'm fine with either `dataValue` or `jsonValue`.

karenetheridge:
jsonValue could suggest that the value is json-encoded, which it isn't.

handrews:
@hudlow and I have been discussing and are leaning towards keeping `externalValue` and not adding any further `external*` fields.  I think `serializedValue` is pretty solid now.

[...]

PR #4671, opened 2025-06-09
v3.2: Add data vs serialized Example Object fields (Revised)

handrews:
_**NOTE 1:** Please review this for the functionality only, **not** the field names which are being tracked in release-blocker issue #4658._

_**NOTE 2:** This PR is the main part of PR #4647, extracted and reworked to minimize the diff while addressing various points that have come up since I posted that PR, which will now be closed. The updated examples will be posted in different PRs (I tried to include the updated Example Object examples in this PR, but the diff gets really messy for reasons that are not clear to me- `git` can't figure out the "Example Object Examples" heading as a stable line for some reason)._

This adds four fields to the Example Object.

`dataValue` and `externalDataValue` apply to the data that would be passed to schema validation.

`serializedValue` (which MUST be a string) and `externalSerializedValue` apply to the serialized form.

External values are used when the appropriate value cannot be included in JSON or YAML,


handrews:
Force-pushed to fix check errors since no one had looked at this yet anyway.

src/oas.md
```diff
@@ -2088,23 +2088,28 @@ transactionCallback:
...
 
 ##### Fixed Fields
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
 | <a name="example-summary"></a>summary | `string` | Short description for the example. |
 | <a name="example-description"></a>description | `string` | Long description for the example. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation. |
+| <a name="example-data-value"></a>dataValue | Any | An example of the data structure that MUST be valid according to the relevant [Schema Object](#schema-object).  If this field is present, `externalDataValue`, `value`, and `externalValue` MUST be absent. |
+| <a name="example-external-data-value"></a>externalDataValue | `string` | A URI that identifies the data example in a separate document, allowing for values not easily expressed in JSON or YAML.  This is usually only needed when working with binary data. The value MUST be valid according to the relevant Schema Object. If this field is present, then `dataValue`, `value`, and `externalValue` MUST be absent. See also the rules for resolving [Relative URI References](#relative-references-in-api-description-uris). |
```
hudlow: Following up on this thread: https://github.com/OAI/OpenAPI-Specification/pull/4647#discussion_r2135909371

@karenetheridge said (and @handrews endorsed):

> `externalDataValue` acts just like the `$ref` keyword -- **the serialization level of the file is the same as the referencing document**, but it's in a separate file **solely because of its size (presumably), or to perhaps allow it to be modified at a different cadence than the referencing document itself** (although that makes less sense here when the example needs to align with the schema).
>
> I would imagine it would be the least-used of the four new keywords, but I see no reason to omit it.

But this is not what the proposal _says_. In contrast, it says:

> [An `externalDataValue` is a] URI that identifies the data example in a separate document, allowing for values **not easily expressed in JSON or YAML**.  This is usually only needed when working with binary data.

and 

> When the validation-ready data consists of **a value outside of the JSON data model, such as a raw binary image**, the `externalDataValue` field can be used. While `externalDataValue` **can be used for entirely binary data**, there is no format suitable for mixing JSON Schema data model-compatible data with binary data as might happen in a `multipart` media type. In such cases, it is only possible to show the serialized form.

I have 3 objections here:

1. ~First, I am getting mixed messages about the purpose/behavior of these fields, and this presents a barrier to weighing in at all on things like their names. I need clarity on how these are proposed work, and I do not have it. Alternatively, if the name discussion is pulled back into this PR, we can discuss the subtle implications of a proposed name and their impact on how the fields _should_ work around the edges.~ (Edit: now resolved.)
2. ~Responding directly to the text proposed in this PR, I know roughly what is being proposed in terms of supporting "a value outside of the JSON data model" but I strongly oppose either of the "data value" fields supporting such a thing. It is such an incredibly subtle, incredibly sharp edge, and I believe it undermines the whole purpose of splitting these fields out. I believe a "data value" should be required, with flashing red lights, to be _data in the JSON data model which conforms to the applicable JSON schema_. If someone can explain to me why referencing something like binary data in a "data value" field is useful in a way that referencing it in a "serialized value" field is not, please do.~ (Edit: now resolved.)
3. I believe Henry's proposal here is insightful and correct in addressing the problems that (1) there was lack of clarity around whether `<ExampleObject>.value` was meant to contain a logical value or a literal one, that (2) both use cases are importantly useful, and (3) our best option was to split them up and deprecate the older, ambiguous `value`. 

   I also strongly believe that the work to rationalize taking the same course with `externalValue` **has not been done** and that to justify this additional transformation we must (a) determine that there is confusion and inconsistency with respect to `externalValue` that corresponds to the confusion and inconsistency with respect to `value` and (b) that there is demand for supporting two distinct use cases that cannot be met with the combination of (i) using `$ref` in the `examples` array for Media Type, Parameter, and Header Objects to reference an externalized Example Object containing a logical value and (ii) using `externalValue` (or `serializedExternalValue`) to reference an externalized, serialized example.

   I think it's vital that we be conservative in adding new fields to OpenAPI, and I simply haven't seen the rationalization for this beyond an _assumption_ that if it's useful for inline values it's useful for external values or that... the symmetry is elegant? I don't think that's good enough, but I am ready and waiting to be enlightened.

src/oas.md
```diff
@@ -2113,18 +2118,64 @@ In all three Objects, this is done through the `examples` (plural) field.
...
+The `value` and `externalValue` fields were intended to hold serialized values, with `value` allowing inline JSON/YAML structures in place of a string if the serialization format is JSON or otherwise compatible with JSON/YAML.
+However, many implementations treat them as data values, so these fields are ambiguous and not interoperable in practice.
+
+###### Character Encodings and Binary Data
+
+Note that `serializedValue`, which MUST be a string, is by necessity a sequence of Unicode code points, which may need to be re-encoded based on the character set or other constraints of the target location.
+This conversion is purely one of encoding.  All escaping necessary to reduce the example to the set of characters valid for the target location (e.g. URI percent-encoding) MUST be included in the `serializedValue`.
+If the target location's encoding is ambiguous, `externalSerializedValue` can be used to demonstrate the exact serialization, including character set encoding, that is intended.
+
+While converting from one character encoding to another is well-defined, binary data does not always correspond to valid Unicode strings.
```
hudlow: > While converting from one character encoding to another is well-defined [...]

This surely isn't true for _every_ [character set](https://www.iana.org/assignments/character-sets/character-sets.xhtml)?

> [...] binary data does not always correspond to valid Unicode strings.

I don't think this is strong enough.

Perhaps we could say something like:

```suggestion
`serializedValue` can be used for any textual value with a character set encoding which has an unambiguous mapping to Unicode code points. Since Unicode strings have no inherent binary representation, `serializedValue` cannot be used for binary data.
```

src/oas.md
handrews: If there is consensus to forbid non-JSON data in `dataValue` and `externalDataValue` that is fine with me.  @OAI/tsc please decide.  That does not remove the need for `externalDataValue`, as @karenetheridge has noted elsewhere- it's just useful to manage large and/or independently updated examples a separate documents sometimes.

Regarding the external fields, you have not convinced me that there is any actual problem here.

src/oas.md
handrews: @hudlow the problems of converting between character sets are a matter for character set specifciations, and are firmly outside of our scope.

src/oas.md
karenetheridge: The wording "...allowing for values not easily expressed in JSON or YAML. This is usually only needed when working with binary data" does appear wrong to me, because if it's not serialized data, then it's the "inflated" data, which must match the JSON document model.

Although, just like JSON Schema's (or OpenAPI's) `$ref` keyword, the actual file format doesn't need to match the referencing document's format, it still needs to be something that an OAD parser can interpret.  I could see the hypothetical-v4-ADA possibly inlining that content in the document, to be equivalent to a `dataValue` keyword.

Therefore, this content wouldn't ever be binary, and this bit of the spec should be fixed.  If it's binary the data can be referenced with a `externalSerializedValue` instead, representing the literal data before it goes through the media-type encoding step.

handrews:
I am not able to fix the schema test coverage issues due to issue #4693.

src/oas.md
handrews: @karenetheridge @hudlow I will deem two people "consensus", update the PR, and we'll see how it's received in reviews :-)  (@hudlow I also already merged your other wording suggestion).

I'm still not sold on other problems with the `external*` fields, but as this line of discussion (hopefully) shows, I'm not impossible ot convince.

handrews:
@hudlow @karenetheridge I have updated this and PR #4672 to not use the `data` fields for binary.  @hudlow I dropped both your updated comment on unicode and binary and some of my remaining sentences as they seemed redundant.  I _think_ what remains covers what is needed.

src/oas.md
```diff
@@ -2113,18 +2118,65 @@ In all three Objects, this is done through the `examples` (plural) field.
...
+If either of `dataValue` or `externalDataValue` are also present, the serialized value MUST be a serialization of the data value, and SHOULD be a valid according to the serialization format.
+
+When using `serializedValue`, the value MUST be a string that is suitably escaped for inclusion in JSON or YAML in addition to any escaping that is part of the target format.
+See [Character Encodings and Binary Data](#character-encodings-and-binary-data) for details.
+
+The `externalSerializedValue` field supports any format for the external value.
+Note that this serialization may or may not exactly match what is transmitted over the wire, as different versions of HTTP use different text or binary encodings, and HTTP content may be subject to compression or other transformations not captured in the OpenAPI Description.
+
+The `value` and `externalValue` fields were intended to hold serialized values, with `value` allowing inline JSON/YAML structures in place of a string if the serialization format is JSON or otherwise compatible with JSON/YAML.
+However, many implementations treat them as data values, so these fields are ambiguous and not interoperable in practice.
```
hudlow, line 2165: I haven't found any evidence of implementations that treat `externalValue` as a data value, and the text of 3.0 and 3.1 doesn't seem to imply that an `externalValue` could be anything but a serialized value.

src/oas.md
```diff
@@ -2113,18 +2118,65 @@ In all three Objects, this is done through the `examples` (plural) field.
...
-The exact serialization and encoding is determined by various fields in the Parameter Object, Header Object, or in the Media Type Object's [Encoding Object](#encoding-object).
+The singular `example` field in non-Schema Objects is a shorthand for a single Example Object using only the `value` field, and all guidance for the `value` field applies to it.
+
+As originally conceived, the Schema Object's keywords are intended to show data as it would be validated by the containing Schema Object, while the fields in the other Objects were intended to show the serialized media type, parameter, or header.
+However, the non-Schema Object fields allowed serialized forms that could be "easily" represented as JSON or JSON-compatible YAML to be shown as inline data structures rather than serialized to strings, which means that in the most common case (`appication/json` documents), this sort of "serialized" example would be represented in the same way as the validation-ready Schema Object examples.
```
hudlow:
```suggestion
However, the non-Schema Object fields allowed serialized forms that could be "easily" represented as JSON or JSON-compatible YAML to be shown as inline data structures rather than serialized to strings, which means that in the most common case (`application/json` documents), this sort of "serialized" example would be represented in the same way as the validation-ready Schema Object examples.
```

src/oas.md
```diff
@@ -2113,18 +2118,65 @@ In all three Objects, this is done through the `examples` (plural) field.
...
+Authors who wish to clearly show serialized and unserialized forms of the same data together are RECOMMENDED to use the new fields in the Example Object to do so.
+
+Due to the lack of any format for mixing JSON Schema data model-compatible data with binary data (as might happen in a `multipart` media type), such examples can only be given using the `externalSerializedValue` field.
+
 Because examples using these fields represent the final serialized form of the data, they SHALL _override_ any `example` in the corresponding Schema Object.
 
-The singular `example` field in the Parameter, Header, or Media Type Object is concise and convenient for simple examples, but does not offer any other advantages over using Example Objects under `examples`.
+###### Validating Examples
+
+Tooling implementations MAY choose to validate compatibility automatically, and reject the example value(s) if incompatible.
```
hudlow, line 2144:
```suggestion
Tooling implementations MAY choose to automatically validate compatibility, and reject the example value(s) if incompatible.
```

hudlow: 🤷 I think the split infinitive reads better, but it's subjective.

src/oas.md
```diff
@@ -2113,18 +2118,65 @@ In all three Objects, this is done through the `examples` (plural) field.
...
+However, many implementations treat them as data values, so these fields are ambiguous and not interoperable in practice.
+
+###### Character Encodings and Binary Data
+
+For `externalDataValue`, it is always the parsed data that is significant, not its character encoding.
+
+While JSON Schema allows for applying schemas to data outside of its data model, this allowance is intended to support consistent use of Schema Objects, including fields such as `readOnly` that do not impact JSON Schema's validation outcome.
+No standard format suitable for passing such non-JSON data to JSON Schema has been defined, and therefore data that is outside of the JSON data model cannot be represented with `dataValue` or `externalDataValue`, and MUST use `serializedValue` or `externalSerializedValue`.
+
+Note that `serializedValue`, which MUST be a string, is by necessity a sequence of Unicode code points, which may need to be re-encoded based on the character set or other constraints of the target location.
```
hudlow:
```suggestion
Note that `serializedValue`, which MUST be a string, is by necessity a sequence of Unicode code points, which will need to be encoded based on the character set or other constraints of the target location.
```

Nit: I don't think a code point vector is encoded at all.

src/oas.md
```diff
@@ -2113,18 +2118,65 @@ In all three Objects, this is done through the `examples` (plural) field.
...
+This conversion is purely one of encoding.  All escaping necessary to reduce the example to the set of characters valid for the target location (e.g. URI percent-encoding) MUST be included in the `serializedValue`.
+If the target location's encoding is ambiguous, `externalSerializedValue` can be used to demonstrate the exact serialization, including character set encoding, that is intended.
 
-Some examples cannot be represented directly in JSON or YAML.
-For all three ways of providing examples, these can be shown as string values with any escaping necessary to make the string valid in the JSON or YAML format of documents that comprise the OpenAPI Description.
-With the Example Object, such values can alternatively be handled through the `externalValue` field.
+Per [[!RFC8259]] [Section 8.2](https://www.rfc-editor.org/rfc/rfc8259.html#section-8.2), using escape sequences that cannot encode Unicode characters to represent binary data is not portable and may cause runtime errors.
+Therefore, data formats such those including binary data that are not always representable as Unicode code points SHOULD use `externalSerializedValue`.
```
hudlow: I still don't think this is right, but RFC 8259 itself is a little muddled on this aspect as well.

The point I'm trying to make is that JSON strings cannot _ever_ represent binary data because parsed JSON only gives the consumer access to a vector[^vector] of code points, which is a vector of integer values, but is _not_ a sequence of _bits_, since it is agnostic to encoding.

To drive this point home, in a UTF-8 JSON document, the strings `"❤️"` and `"\u2764\uFE0F"` are indistinguishable to a consumer of the parsed document, but the former is literally encoded as `E29DA4EFB88F`, not `2764FE0F`. Yes, the escape sequences use UTF-16 code units to represent escaped characters but these code units are instructing a parser to look up a Unicode character, they're not telling a parser to literally insert the represented bit sequence into the string. That wouldn't work!

So I don't think it's right to say that "using escape sequences that cannot encode Unicode characters to represent binary data is not portable." Whether or not you use escape sequences, whether or not those escape sequences are valid UTF-16 encodings of Unicode characters, _a JSON string does not ever represent a bit sequence_. It represents an integer (code point) vector which you COULD encode binary data on top of (e.g., in base64), but which itself has any number of possible binary serializations, none of which are implied to a consumer of parsed JSON.

[^vector]: I'm using "vector" instead of "sequence" in this context to try to emphasize that a Unicode string is a data structure whose in-memory representation may or may not be contiguous.

karenetheridge: > JSON strings cannot ever represent binary data because parsed JSON only gives the consumer access to a vector of code points  
> a JSON string does not ever represent a bit sequence  

That's the exact point I was trying to make as well, with respect to `dataValue`! :)

src/oas.md
hudlow: Thanks Henry!

> I'm still not sold on other problems with the `external*` fields, but as this line of discussion (hopefully) shows, I'm not impossible ot convince.

I think the point of disagreement might be where the burden of proof is. I'm making the case that the burden of proof is to demonstrate that someone needs all three `external*` fields, and I suspect your position is that the burden of proof is to demonstrate that the `external*` fields should be treated differently from the non-`external*` fields.

> That does not remove the need for externalDataValue, as @karenetheridge has noted elsewhere- it's just useful to manage large and/or independently updated examples a separate documents sometimes.

But you have to need to externalize the data value for size or lifecycle reasons _and_ need to avoid just externalizing its `ExampleObject` envelope, and needing both those things seems kind of implausible to me.

handrews: @hudlow people use `externalValue` now, and as far as I know they use it for JSON values as well as non-JSON values, so for feature parity we have both.  In the case where the serialization is JSON, of course the fields are redundant.  In the case of other serializations, they are not.

And yes, since we're replacing fields (even if the inconsistency was more problematic with `value` than `externalValue`, I do not think it is necessary to re-litigate this, as I consider `externalValue` sufficient precedent.  But if you build a consensus around your position, I will go with whatever @OAI/tsc decides (either formally or by just commenting in favor without others being opposed).

src/oas.md
handrews: @hudlow I still don't see how anyone reads any version of 3.x and sees `value` and `externalValue` as anything other than serialized, even if weirdly represented in `value`, so I feel like if one is too ambiguous, so is the other, regardless of the actual implementation out there.

On the other hand, if it is really true that `externalValue` is always handled correctly as serialized (`multipart/form-data` would be an ideal test case), and if you build a consensus that `externalDataValue` is not needed, then I suppose we could just keep `externalValue`, say it is serialized, and only add `dataValue` and `serializedValue`.

As with the last comment, build a consensus and I'll go with it.

src/oas.md
handrews: This was just copied verbatim from prior 3.1.1 language so I'd rather leave it.

src/oas.md
handrews: @hudlow (and @karenetheridge )

>  still don't think this is right, but RFC 8259 itself is a little muddled on this aspect as well.

I based this language on RFC8259's wording, which is why I've stuck with it.  But I do see your point here, let me go back and look at the wording you'd added before and sort that back out again.  Me removing that wording was supposed to trim redundancy, not change the outcome.

src/oas.md
handrews: (And to anyone reading who is wondering why I'm saying "build a consensus" instead of just accepting the argument, it's because I don't feel able to make a call one way or the other, still prefer my position, and the outcome shouldn't come down to "Henry wins by default because he posted the PR")

src/oas.md
handrews: @hudlow OK I pulled your prior wording back out, dropped the sentence about non-portability and runtime parse errors, and did some minor tweaks to smooth that into a paragraph.  Please see if you and @karenetheridge think it is correct now.  My apologies for not getting the point previously- as you mentioned the RFC is a bit confusing to read in that section.

ralfhandl:
@handrews This PR extends `schema.yaml` and now also needs to add positive test cases for the new features. Please add.

src/oas.md
karenetheridge: > But you have to need to externalize the data value for size or lifecycle reasons and need to avoid just externalizing its ExampleObject envelope,

True, you could just `$ref` to the entire example object in another file... but then you lose visibility of the `summary` and `description` fields, which might be useful to keep in the main document (I'm thinking of people who are using the raw documents, rather than viewing the OAD through a GUI tool of some kind).

My inclination here is that since it _may_ be useful, and is harmless to include, it is better to err on the side of keeping it.

hudlow: > @hudlow people use externalValue now, and as far as I know they use it for JSON values as well as non-JSON values, so for feature parity we have both.

I tried to test this hypothesis; surprisingly, in my unscientific test, even the _first_ assertion was a little shaky.

My methodology was to clone https://github.com/APIs-guru/openapi-directory and grep for `externalValue`. Out of the ~700 API descriptions (not counting multiple versions for one API), only **one** uses `externalValue` _at all_, and they appear to use it only as a serialized value (i.e., matching media type).

<details>
<summary>methodology</summary>

```sh
cpy.re/peertube/5.1.0/openapi.yaml-            application/json:
cpy.re/peertube/5.1.0/openapi.yaml-              examples:
cpy.re/peertube/5.1.0/openapi.yaml-                nightly:
cpy.re/peertube/5.1.0/openapi.yaml:                  externalValue: https://peertube2.cpy.re/api/v1/config
--
cpy.re/peertube/5.1.0/openapi.yaml-            application/json:
cpy.re/peertube/5.1.0/openapi.yaml-              examples:
cpy.re/peertube/5.1.0/openapi.yaml-                nightly:
cpy.re/peertube/5.1.0/openapi.yaml:                  externalValue: https://peertube2.cpy.re/api/v1/config/about
--
cpy.re/peertube/5.1.0/openapi.yaml-            application/json:
cpy.re/peertube/5.1.0/openapi.yaml-              examples:
cpy.re/peertube/5.1.0/openapi.yaml-                nightly:
cpy.re/peertube/5.1.0/openapi.yaml:                  externalValue: https://peertube2.cpy.re/api/v1/video-playlists/privacies
--
cpy.re/peertube/5.1.0/openapi.yaml-            application/json:
cpy.re/peertube/5.1.0/openapi.yaml-              examples:
cpy.re/peertube/5.1.0/openapi.yaml-                nightly:
cpy.re/peertube/5.1.0/openapi.yaml:                  externalValue: https://peertube2.cpy.re/api/v1/videos/categories
--
cpy.re/peertube/5.1.0/openapi.yaml-            application/json:
cpy.re/peertube/5.1.0/openapi.yaml-              examples:
cpy.re/peertube/5.1.0/openapi.yaml-                nightly:
cpy.re/peertube/5.1.0/openapi.yaml:                  externalValue: https://peertube2.cpy.re/api/v1/videos/languages
--
cpy.re/peertube/5.1.0/openapi.yaml-            application/json:
cpy.re/peertube/5.1.0/openapi.yaml-              examples:
cpy.re/peertube/5.1.0/openapi.yaml-                nightly:
cpy.re/peertube/5.1.0/openapi.yaml:                  externalValue: https://peertube2.cpy.re/api/v1/videos/licences
--
cpy.re/peertube/5.1.0/openapi.yaml-            application/json:
cpy.re/peertube/5.1.0/openapi.yaml-              examples:
cpy.re/peertube/5.1.0/openapi.yaml-                nightly:
cpy.re/peertube/5.1.0/openapi.yaml:                  externalValue: https://peertube2.cpy.re/api/v1/videos/privacies
--
cpy.re/peertube/5.1.0/openapi.yaml-            application/atom+xml:
cpy.re/peertube/5.1.0/openapi.yaml-              examples:
cpy.re/peertube/5.1.0/openapi.yaml-                nightly:
cpy.re/peertube/5.1.0/openapi.yaml:                  externalValue: https://peertube2.cpy.re/feeds/video-comments.atom?filter=local
--
cpy.re/peertube/5.1.0/openapi.yaml-            application/json:
cpy.re/peertube/5.1.0/openapi.yaml-              examples:
cpy.re/peertube/5.1.0/openapi.yaml-                nightly:
cpy.re/peertube/5.1.0/openapi.yaml:                  externalValue: https://peertube2.cpy.re/feeds/video-comments.json?filter=local
--
cpy.re/peertube/5.1.0/openapi.yaml-            application/rss+xml:
cpy.re/peertube/5.1.0/openapi.yaml-              examples:
cpy.re/peertube/5.1.0/openapi.yaml-                nightly:
cpy.re/peertube/5.1.0/openapi.yaml:                  externalValue: https://peertube2.cpy.re/feeds/video-comments.rss?filter=local
--
cpy.re/peertube/5.1.0/openapi.yaml-            application/xml:
cpy.re/peertube/5.1.0/openapi.yaml-              examples:
cpy.re/peertube/5.1.0/openapi.yaml-                nightly:
cpy.re/peertube/5.1.0/openapi.yaml:                  externalValue: https://peertube2.cpy.re/feeds/video-comments.xml?filter=local
--
cpy.re/peertube/5.1.0/openapi.yaml-            text/xml:
cpy.re/peertube/5.1.0/openapi.yaml-              examples:
cpy.re/peertube/5.1.0/openapi.yaml-                nightly:
cpy.re/peertube/5.1.0/openapi.yaml:                  externalValue: https://peertube2.cpy.re/feeds/video-comments.xml?filter=local
--
cpy.re/peertube/5.1.0/openapi.yaml-            application/atom+xml:
cpy.re/peertube/5.1.0/openapi.yaml-              examples:
cpy.re/peertube/5.1.0/openapi.yaml-                nightly:
cpy.re/peertube/5.1.0/openapi.yaml:                  externalValue: https://peertube2.cpy.re/feeds/videos.atom?filter=local
--
cpy.re/peertube/5.1.0/openapi.yaml-            application/json:
cpy.re/peertube/5.1.0/openapi.yaml-              examples:
cpy.re/peertube/5.1.0/openapi.yaml-                nightly:
cpy.re/peertube/5.1.0/openapi.yaml:                  externalValue: https://peertube2.cpy.re/feeds/videos.json?filter=local
--
cpy.re/peertube/5.1.0/openapi.yaml-            application/rss+xml:
cpy.re/peertube/5.1.0/openapi.yaml-              examples:
cpy.re/peertube/5.1.0/openapi.yaml-                nightly:
cpy.re/peertube/5.1.0/openapi.yaml:                  externalValue: https://peertube2.cpy.re/feeds/videos.rss?filter=local
--
cpy.re/peertube/5.1.0/openapi.yaml-            application/xml:
cpy.re/peertube/5.1.0/openapi.yaml-              examples:
cpy.re/peertube/5.1.0/openapi.yaml-                nightly:
cpy.re/peertube/5.1.0/openapi.yaml:                  externalValue: https://peertube2.cpy.re/feeds/videos.xml?filter=local
--
cpy.re/peertube/5.1.0/openapi.yaml-            text/xml:
cpy.re/peertube/5.1.0/openapi.yaml-              examples:
cpy.re/peertube/5.1.0/openapi.yaml-                nightly:
cpy.re/peertube/5.1.0/openapi.yaml:                  externalValue: https://peertube2.cpy.re/feeds/videos.xml?filter=local

APIs (main)$ curl "https://peertube2.cpy.re/feeds/video-comments.xml?filter=local" --head -s | grep "content-type"
content-type: application/xml; charset=utf-8
```

</details>

> My inclination here is that since it may be useful, and is harmless to include, it is better to err on the side of keeping it.

I don't see it this way at all; it constrains future directions for the 3.x spec, and it burdens implementors who wish to be comprehensive. And _we can always add it later, if there is demand_.

src/oas.md
hudlow: Responded here: https://github.com/OAI/OpenAPI-Specification/pull/4671#discussion_r2143510760

handrews:
@ralfhandl new commits added with tests.  Otherwise the force-push is just a rebase with not conflicts (same commits).

src/oas.md
```diff
@@ -2088,23 +2088,28 @@ transactionCallback:
 #### Example Object
 
 An object grouping an internal or external example value with basic `summary` and `description` metadata.
+The examples can show either data suitable for schema validation, or serialized data as required by tehe containing [Media Type Object](#media-type-object), [Parameter Object](#parameter-object), or [Header Object](#header-object).
```
```suggestion
The examples can show either data suitable for schema validation, or serialized data as required by the containing [Media Type Object](#media-type-object), [Parameter Object](#parameter-object), or [Header Object](#header-object).
```

src/oas.md
```diff
@@ -2088,23 +2088,28 @@ transactionCallback:
...
+| <a name="example-eternal-serialized-value"></a>externalSerializedValue | `string` | A URI that identifies the serialized example in a separate document, allowing for values not easily or readably expressed in JSON or YAML strings.  If `dataValue` or `externalDataValue` are present, then this field SHOULD identify a serialization of the given data.  Otherwise, the value SHOULD be a valid serialization as described for `serializedValue`.  If this field is present, `serializedValue`, `value`, and `externalValue` MUST be absent. See also the rules for resolving [Relative References](#relative-references-in-api-description-uris). |
```
```suggestion
| <a name="example-external-serialized-value"></a>externalSerializedValue | `string` | A URI that identifies the serialized example in a separate document, allowing for values not easily or readably expressed in JSON or YAML strings.  If `dataValue` or `externalDataValue` are present, then this field SHOULD identify a serialization of the given data.  Otherwise, the value SHOULD be a valid serialization as described for `serializedValue`.  If this field is present, `serializedValue`, `value`, and `externalValue` MUST be absent. See also the rules for resolving [Relative References](#relative-references-in-api-description-uris). |
```

src/oas.md
```diff
@@ -2113,18 +2118,65 @@ In all three Objects, this is done through the `examples` (plural) field.
...
+However, the non-Schema Object fields allowed serialized forms that could be "easily" represented as JSON or JSON-compatible YAML to be shown as inline data structures rather than serialized to strings, which means that in the most common case (`application/json` documents), this sort of "serialized" example would be represented in the same way as the validation-ready Schema Object examples.
+Furthermore, some parameters and headers are simple enough that there is no difference between the serialized and validation-ready forms.
+
+Possibly due to these ambiguities, many implementations appear to have implemented the fields that existed prior to version 3.2 of this specification as requiring validation-ready data structures.
+For this reason, OpenAPI Description authors who want portable behavior with examples that are not for `application/json` media types or trivial strings without any formatting or escaping are RECOMMENDED to use the new fields (`dataValue` or `externalDataValue` for validation-ready data, and `serializedValue` or `externalSerializedValue` for serialized forms)
```
```suggestion
For this reason, OpenAPI Description authors who want portable behavior with examples that are not for `application/json` media types or trivial strings without any formatting or escaping are RECOMMENDED to use the new fields (`dataValue` or `externalDataValue` for validation-ready data, and `serializedValue` or `externalSerializedValue` for serialized forms).
```

src/oas.md
```diff
@@ -2113,18 +2118,65 @@ In all three Objects, this is done through the `examples` (plural) field.
...
+
+In addition, it can be challenging to correlate the validation-ready Schema Object example with serialized Example Object examples when all are part of shared Objects reached through (possibly multiple) references.
+Authors who wish to clearly show serialized and unserialized forms of the same data together are RECOMMENDED to use the new fields in the Example Object to do so.
+
+Due to the lack of any format for mixing JSON Schema data model-compatible data with binary data (as might happen in a `multipart` media type), such examples can only be given using the `externalSerializedValue` field.
```
ralfhandl, line 2138:
```suggestion
Due to the lack of any format for mixing JSON Schema data-model compatible data with binary data (as might happen in a `multipart` media type), such examples can only be given using the `externalSerializedValue` field.
```
I learned that if three consecutive words belong together, the first two can be connected with a dash (as opposed to German, where the three words would just be strung together into one word).

Also this seems to mean "compatible with the data model of JSON Schema", and not "model-compatible with data for JSON Schema" which doesn't make any sense to me.

src/oas.md
```diff
@@ -2088,23 +2088,28 @@ transactionCallback:
...
+| <a name="example-eternal-serialized-value"></a>externalSerializedValue | `string` | A URI that identifies the serialized example in a separate document, allowing for values not easily or readably expressed in JSON or YAML strings.  If `dataValue` or `externalDataValue` are present, then this field SHOULD identify a serialization of the given data.  Otherwise, the value SHOULD be a valid serialization as described for `serializedValue`.  If this field is present, `serializedValue`, `value`, and `externalValue` MUST be absent. See also the rules for resolving [Relative References](#relative-references-in-api-description-uris). |
 | <a name="example-value"></a>value | Any | Embedded literal example. The `value` field and `externalValue` field are mutually exclusive. To represent examples of media types that cannot naturally represented in JSON or YAML, use a string value to contain the example, escaping where necessary. |
```
```suggestion
| <a name="example-value"></a>value | Any | Embedded literal example. The `value` field and `externalValue` field are mutually exclusive. To represent examples of media types that cannot be naturally represented in JSON or YAML, use a string value to contain the example, escaping where necessary. |
```

ralfhandl (commented):
+1, only minor nits

src/oas.md
```diff
@@ -2113,18 +2118,65 @@ In all three Objects, this is done through the `examples` (plural) field.
...
+This conversion is purely one of encoding.  All escaping necessary to reduce the example to the set of characters valid for the target location (e.g. URI percent-encoding) MUST be included in the `serializedValue`.
+If the target location's encoding is ambiguous, `externalSerializedValue` can be used to demonstrate the exact serialization, including character set encoding, that is intended.
 
-Some examples cannot be represented directly in JSON or YAML.
-For all three ways of providing examples, these can be shown as string values with any escaping necessary to make the string valid in the JSON or YAML format of documents that comprise the OpenAPI Description.
-With the Example Object, such values can alternatively be handled through the `externalValue` field.
+The `serializedValue` field can be used for any textual value with a character set encoding which has an unambiguous mapping to Unicode code points. Since Unicode strings have no inherent binary representation, `serializedValue` cannot be used for binary data.
+Therefore, data formats such those including binary data that are not always representable as Unicode code points SHOULD use `externalSerializedValue`.
```
hudlow: There was a minor editing error in this sentence, so I'm suggesting something that I don't think changes the intended meaning.

```suggestion
Instead, binary data and other formats that are not always representable as Unicode code points SHOULD use `externalSerializedValue`.
```

src/oas.md
handrews: @ralfhandl I think it reads better as "cannot naturally be" so I'm going to do that, but thanks for catching!

src/oas.md
handrews: This is to be read as (JSON Schema data model)-compatible, where "JSON Schema data model" is logically one term, so it is correct as written.  The phrase "model-compatible with data for JSON Schema" does not make sense to me, and I'm pretty sure that the way I wrote it is compatible with other places where we mention similar things, because I'm pretty sure I wrote those places as well.

src/oas.md

> My inclination here is that since it may be useful, and is harmless to include, it is better to err on the side of keeping it.

Strong agree.  In fact I just needed it yesterday as I was showing two different media types for the same data, where the example data value was sizeable so putting it in a file and using `externalDataValue` in both Example Objects was much preferred over inlining a non-trivial example twice and having to maintain it in both places.

I'm quite convinced of its usefulness, and see no real way in which this causes problems down the line.

handrews:
Force-pushed a rebase to pick up the latest test runner changes.  No conflicts or additions- these are literally the same commits as before.

handrews:
@karenetheridge @hudlow @mikekistler in addition to the [previously-mentioned use case](https://github.com/OAI/OpenAPI-Specification/pull/4671#discussion_r2145488723) for `externalDataValue` (which arose without me looking for such a thing), I also ended up using it in [the XML example updates](https://github.com/OAI/OpenAPI-Specification/pull/4711/files#diff-a662b9afa95a60c2a67cc48014a8d5e00fa6ab2bd5ba2a54531674a5dd398c73R3700) because again I needed to show the same input data serialized in two different ways in two different contexts.  While this second use case is more contrived, it's not unrealistic, and that's two that have come up without me particularly trying to make it happen.

At this point, I am quite firm in thinking that we need both `externalDataValue` and `externalSerializedValue` (even if `externalValue` is well-defined as behaving like `externalSerializedValue`, the naming asymmetry is problematic).

_[**EDIT:** PR #4648 shows even more potential use cases, where you are using the same `dataValue` at two different levels with different serializations]_

src/oas.md
```diff
@@ -2088,23 +2088,28 @@ transactionCallback:
...
+| <a name="example-value"></a>value | Any | Embedded literal example. The `value` field and `externalValue` field are mutually exclusive. To represent examples of media types that cannot naturally be represented in JSON or YAML, use a string value to contain the example, escaping where necessary. |
 | <a name="example-external-value"></a>externalValue | `string` | A URI that identifies the literal example. This provides the capability to reference examples that cannot easily be included in JSON or YAML documents. The `value` field and `externalValue` field are mutually exclusive. See the rules for resolving [Relative References](#relative-references-in-api-description-uris). |
```
hudlow, line 2106: Below is my proposal based on ongoing conversations.



```suggestion
| <a name="example-data-value"></a>dataValue | Any | An example of the data structure that MUST be valid according to the relevant [Schema Object](#schema-object). This field MUST not be used with `dataValueReference` or `value`. |
| <a name="example-external-data-value"></a>dataValueReference | [Reference Object](#reference-object) | A reference to an example of the data structure that MUST be valid according to the relevant [Schema Object](#schema-object). This field MUST not be used with `dataValue` or `value`. |
| <a name="example-serialized-value"></a>serializedValue | `string` | An example of the Unicode-serialized form of a value that MUST be valid according to the relevant [Schema Object](#schema-object) and includes encoding and escaping as described under [Validating Examples](#validating-examples). If `dataValue` or `dataValueReference` are present, then this field SHOULD contain a serialization of the given data value. This field MUST not be used with `value` or `externalValue`. |
| <a name="example-value"></a>value | Any | An example value valid as described for either `dataValue` or `serializedValue`. This field MUST not be used with `dataValue`, `dataValueReference`, `serializedValue`, or `externalValue`. |
| <a name="example-external-value"></a>externalValue | `string` | A URI that identifies the binary-serialized form of the value. This provides the capability to reference examples that cannot easily be included in JSON or YAML documents, as well as to demonstrate binary encoding for any value. If `dataValue` or `dataValueReference` are present, then this field SHOULD identify a serialization of the given data. This field MUST not be used with `value`. See the rules for resolving [Relative References](#relative-references-in-api-description-uris). |
```

One part of this analysis involved mapping out possible interpretations of the existing values:

image

Some general thoughts:
- Unicode and binary serialization are importantly different.
- `$ref` behavior is unique for its implicit deserialization and parsing of the target document.

On the ambiguity of `externalValue`, while it is conceivable that someone has implemented a `$ref`-style abstraction over `externalValue`:
- I don't see anything in previous versions of the OpenAPI spec that would suggest this is supported or encourage it. The language itself doesn't imply the same dichotomy as the language for `value` and the examples don't show a JSON (or YAML) document being referenced for a non-JSON media type.
- I can't find any real-world examples of OpenAPI documents that seem to assume a `$ref`-style abstraction.

> @karenetheridge @hudlow @mikekistler in addition to the https://github.com/OAI/OpenAPI-Specification/pull/4671#discussion_r2145488723 for externalDataValue (which arose without me looking for such a thing), I also ended up using it in [the XML example updates](https://github.com/OAI/OpenAPI-Specification/pull/4711/files#diff-a662b9afa95a60c2a67cc48014a8d5e00fa6ab2bd5ba2a54531674a5dd398c73R3700) because again I needed to show the same input data serialized in two different ways in two different contexts. While this second use case is more contrived, it's not unrealistic, and that's two that have come up without me particularly trying to make it happen.

Okay, I'm sold, but I think we need to leverage a reference object to clarify that (1) this can be done in-document and that (2) if it's done out-of-document the implicit deserialization/parsing process applies. Hence the proposal for `dataValueReference` above.

> even if `externalValue` is well-defined as behaving like `externalSerializedValue`, the naming asymmetry is problematic

I actually think the naming asymmetry is a feature and not a bug because:
1. `externalValue` lacks `$ref`-style behavior
2. `externalValue` is a binary serialization unlike `serializedValue` which is a Unicode serialization

handrews: @hudlow it sounds like you have a thoroughly thought-through proposal here- I'm not weighing in on the merits because I have only had time to skim it, but it definitely looks worth a proper standalone write-up/PR.  I've done two rounds of PRs on this subject and don't have the bandwidth for a third, but if you open your own PR I can close this one in favor of it.  Feel free to lift whatever you want from this one, or start over completely, whatever works best for you.

You might also want to take over #4672, and possibly #4648 depending on how that dovetails with your proposal, but the others (particularly #4673) I'll keep and update as needed with the outcome.  There are some more examples to update that I hadn't gotten around to posting yet- I can finish those once you've sorted out the new fields, or you can do them if you prefer.  But I need to focus on finishing up `multipart` and other media type things such as the media type registry.

handrews:
I've marked this as draft as I am expecting to close it in favor of a new PR I've [invited @hudlow to submit](https://github.com/OAI/OpenAPI-Specification/pull/4671#discussion_r2146360222) with his alternative proposal above.

src/oas.md
hudlow: @handrews busy week ahead, but I will get something together as soon as I can. I'll aim to have it done by Thursday, but not positive how realistic that is.

handrews: Thanks @hudlow — I have more than enough to occupy the Thursday call with `multipart` and media registry things this week if you want to target next week :-)

handrews: @hudlow having had a chance to consider this in more detail, it makes sense even if it is not how I would do it.

Personally, I find your distinction between "unicode serialization" and "binary serialization" to be far more confusing than useful- YAML/JSON is unicode, and that's all you need to know.  "Unicode serialization" sounds like something involving escape sequences of the sort that we actually _don't_ want.  But if you can sell the TSC on it, I won't block it.  I continue to dislike it and find it unnecessarily confusing, though, and think it makes too big of a deal out of things that other specs handle just fine.

Regardless of all that, `dataReference` needs to be `dataRef` and take a URI-reference (not a Reference Object) for consistency with `operationRef` and possibly something else I'm forgetting.  You would also need a `components` entry for it.

hudlow: @handrews thanks for the additional feedback — will work on refining the language.

handrews:
Closing in favor of #4799.

PR #4593, opened 2025-05-16
v3.2: Allow allowReserved everywhere

handrews:
Fixes:

* #1710 (allow `;` in path parameters outside of `matrix` expansion)

The restriction on `allowReserved` make some useful configurations impossible, and do not actually prevent pathological scenarios like path parameter values containing `/`, from occurring.  Such pathological scenarios are already possible by using `content` instead of `style`/`explode`/`allowReserved`.

Lifting the restriction also makes the handling of this field more consistent, as it is no longer necessary to keep track of the destination.

- [ ] schema changes are included in this pull request
- [ ] schema changes are needed for this pull request but not done yet
- [X] no schema changes are needed for this pull request

PR #4680, opened 2025-06-10
v3.2: Allow `allowReserved` in the Header Object (left out of PR #4593 by mistake)

handrews:
We added it in the Parameter and Encoding Objects in PR #4593 but forgot the Header Object (again).  Appendix D, which deals with header and cookie serialization, was already updated in the earlier PR.  We just forgot the Header Object.

See PR #4673 for an example where this is used, which is what made me realize we forgot it!


lornajane (approved):
Good catch

PR #4700, opened 2025-06-13
Use the full schema (schema-base.yaml) for testing and coverage

handrews:
We were not actually using the correct schema for our testing, which meant that we were not covering the Schema Object at all.  This fixes that, at least for the `npm run tests` with coverage.  The same could also be done in `tests/schema/schema.test.mjs` but I can't figure out what that script is for- is it redundant?  Can we delete it or consolidate it with the coverage one?

Note that we will need to modify the `jsonSchemaDialect` test per dev branch, which is why I split it out into its own file instead of leaving it in `mega.yaml`.

We do not have full coverage anyway on `dev`, but I will work on it for `v3.1-dev` and `v3.2-dev`.

### DIFF
--- a/scripts/schema-test-coverage.mjs
+++ b/scripts/schema-test-coverage.mjs
@@ -46,7 +46,10 @@ class TestCoveragePlugin {
     this.allLocations = [];
 
     for (const schemaLocation in context.ast) {
-      if (schemaLocation === "metaData") {
+      if (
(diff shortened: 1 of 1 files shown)


handrews:
Arg- apparently this needs to start on main even though we don't have tests there?  I'll close this and re-submit

PR #4701, opened 2025-06-13
Use full schema (schema-base.yaml) for coverage

handrews:
But don't count standard JSON Schema keywords.

We were not actually using the correct schema for our testing, which meant that we were not covering the Schema Object at all.

### DIFF
--- a/scripts/schema-test-coverage.mjs
+++ b/scripts/schema-test-coverage.mjs
@@ -200,5 +200,5 @@ console.log(
   (allocations.length - notCovered.length),
   "of",
   allLocations.length,
-  "(" + Math.floor((visitedLocations.size / allLocations.length) * 100) + "%)",
+  "(" + Math.floor(((allocations.length - notCovered.length) / allLocations.length) * 100) + "%)",
 );
--- a/scripts/schema-test-coverage.mjs
+++ b/scripts/schema-test-coverage.mjs
@@ -48,7 +48,7 @@ class TestCoveragePlugin {
     for (const schemaLocation in context.ast) {
       if (
         schemaLocation === "metaData" ||
-        // Do not reqiure coverage of standard JSON Schema
+        // Do not require coverage of standard JSON Schema
         schemaLocation.includes("json-schema.org") ||
         // Do not require coverage of default $dynamicAnchor
         // schemas, as they are not expected to be reached
--- a/scripts/schema-test-coverage.mjs
+++ b/scripts/schema-test-coverage.mjs
@@ -52,7 +52,6 @@ class TestCoveragePlugin {
         schemaLocation.includes("json-schema.org") ||
         // Do not require coverage of default $dynamicAnchor
         // schemas, as they are not expected to be reached
-        // schemaLocation.includes("/schema/WORK-IN-PROGRESS#/$defs/schema/")
         schemaLocation.endsWith("/schema/WORK-IN-PROGRESS#/$defs/schema")
       ) {
         continue;
--- a/scripts/schema-test-coverage.mjs
+++ b/scripts/schema-test-coverage.mjs
@@ -193,16 +193,13 @@ if (notCovered.length > 0) {
   const firstNotCovered = notCovered.slice(0, maxNotCovered);
   if (notCovered.length > maxNotCovered) firstNotCovered.push("...");
   console.log(firstNotCovered);
+  process.exitCode = 1;
 }
 
 console.log(
   "Covered:",
-  visitedLocations.size,
+  (allocations.length - notCovered.length),
   "of",
   allLocations.length,
   "(" + Math.floor((visitedLocations.size / allLocations.length) * 100) + "%)",
 );
-
-if (visitedLocations.size != allLocations.length) {
-  process.exitCode = 1;
-}
--- a/scripts/schema-test-coverage.mjs
+++ b/scripts/schema-test-coverage.mjs
@@ -1,10 +1,11 @@
+import { readFileSync } from "node:fs";
 import { readdir, readFile } from "node:fs/promises";
 import YAML from "yaml";
 import { join } from "node:path";
 import { argv } from "node:process";
-import { validate } from "@hyperjump/json-schema/draft-2020-12";
+import { registerSchema, validate } from "@hyperjump/json-schema/draft-2020-12";
 import "@hyperjump/json-schema/draft-04";
-import { BASIC } from "@hyperjump/json-schema/experimental";
+import { BASIC, addKeyword, defineVocabulary } from "@hyperjump/json-schema/experimental";
 
 /**
  * @import { EvaluationPlugin } from "@hyperjump/json-schema/experimental"
@@ -45,7 +46,10 @@ class TestCoveragePlugin {
     this.allLocations = [];
 
     for (const schemaLocation in context.ast) {
-      if (schemaLocation === "metaData") {
+      if (
+        schemaLocation === "metaData" ||
+        schemaLocation.includes("json-schema.org")
+      ) {
         continue;
       }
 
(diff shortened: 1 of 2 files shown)


scripts/schema-test-coverage.mjs
```diff
@@ -110,6 +119,68 @@ const runTests = async (schemaUri, testDirectory) => {
...
+  interpret: (xml, instance, context) => {
+    return true;
+  },
+  annotation: (xml) => {
+    return xml;
+  },
+});
+
+defineVocabulary(
+  "https://spec.openapis.org/oas/3.1/vocab/base",
```
ralfhandl, line 166: This apparently needs to be parameterized with the OpenAPI version of the branch it is running on, because the script fails if run as-is on the `v3.2-dev` branch.

No spontanous idea on how to reliably do that, other than parsing `dialect.yaml`.

scripts/schema-test-coverage.mjs
```diff
@@ -45,7 +46,15 @@ class TestCoveragePlugin {
...
 
     for (const schemaLocation in context.ast) {
-      if (schemaLocation === "metaData") {
+      if (
+        schemaLocation === "metaData" ||
+        // Do not reqiure coverage of standard JSON Schema
+        schemaLocation.includes("json-schema.org") ||
+        // Do not require coverage of default $dynamicAnchor
+        // schemas, as they are not expected to be reached
+        // schemaLocation.includes("/schema/WORK-IN-PROGRESS#/$defs/schema/")
```
ralfhandl: Was this an idea that did not work, or is it a TODO for further investigation?

scripts/schema-test-coverage.mjs
```diff
@@ -110,6 +119,68 @@ const runTests = async (schemaUri, testDirectory) => {
...
+  const schemaYaml = readFileSync(filePath, "utf8");
+  return YAML.parse(schemaYaml, { prettyErrors: true });
+};
+registerSchema(parseYamlFromFile("./src/schemas/validation/meta.yaml"));
+registerSchema(parseYamlFromFile("./src/schemas/validation/dialect.yaml"));
+registerSchema(parseYamlFromFile("./src/schemas/validation/schema.yaml"));
+
 ///////////////////////////////////////////////////////////////////////////////
 
 const { allLocations, visitedLocations } = await runTests(argv[2], argv[3]);
```
ralfhandl, line 185: `visitedLocations` is off, merging this PR into `v3.1-dev` and running it produces

    Covered: 701 of 656 (106%)

Which can be fixed by replacing `visitedLocations.size` with `(allLocations.length - notCovered.length)` and moving `process.exitCode = 1;` up into the `if (notCovered.length > 0) {...}` block.

scripts/schema-test-coverage.mjs
```diff
@@ -45,7 +46,15 @@ class TestCoveragePlugin {
     this.allLocations = [];
 
     for (const schemaLocation in context.ast) {
-      if (schemaLocation === "metaData") {
+      if (
+        schemaLocation === "metaData" ||
+        // Do not reqiure coverage of standard JSON Schema
```
ralfhandl:
```suggestion
        // Do not require coverage of standard JSON Schema
```
My editor complains, I wouldn't have noticed 😎

ralfhandl:
@handrews Thanks for taking this up, I was wondering how to test our Schema Object extensions without having to cover all standard JSON Schema keywords.

scripts/schema-test-coverage.mjs
```diff
@@ -110,6 +119,68 @@ const runTests = async (schemaUri, testDirectory) => {
...
+
+addKeyword({
+  id: "https://spec.openapis.org/oas/schema/vocab/keyword/xml",
+  interpret: (xml, instance, context) => {
+    return true;
+  },
+  annotation: (xml) => {
+    return xml;
+  },
+});
```
ralfhandl, line 163: These seem to be unnecessary, commenting them out does not change the test result.

scripts/schema-test-coverage.mjs
handrews: @ralfhandl I was just going to modify the line on the 3.2 port

scripts/schema-test-coverage.mjs
handrews: Last line wasn't supposed to be commented.

scripts/schema-test-coverage.mjs
handrews: @ralfhandl I don't understand what you mean here because `visitedLocations.size` is not used until after that `if` block.  Can you just make a suggestion so I can merge it in?

handrews: Oh wait maybe I see it now, lemme try, but if this isn't right please make a mergeable suggestion

scripts/schema-test-coverage.mjs
handrews: Oh actually yeah last line was an older version with an error and I forgot to remove it when I figured out the trailing `/` was wrong.

scripts/schema-test-coverage.mjs
handrews: @ralfhandl they make the vocabulary be handled correctly if we want to further utilize those keywords in tests.  I prefer to set up the vocabulary correctly rather than stub it out just because we're not using it right now.

handrews:
@ralfhandl I have fixed the blocking things, I think.

I realize the `3.1` hardcoding is not ideal, and we can debate the `addKeyword` thing.  Please let's get this merged so literally everything else can be unblocked.  We can work on parametrizing or re-organizing but I'll just modify the PR on `v3.2-dev` to hardcode `3.2` instead so we can get the other PRs mergeable.  We can keep improving this afterwards.

Among other things, it's extremely awkward that this script has to start on `main` but the non-coverage script that runs the fail tests has to start on `dev`.  But let's get this going first.

scripts/schema-test-coverage.mjs
ralfhandl: Which would end up in main (and then dev and v3.1-dev) after the release unless someone fiddles with the v3.2.0-rel branch.

Automatically detecting the release seems preferable.

handrews: @ralfhandl Which is fine but let's fix it separately.  I do not have the mental bandwidth today for that, and want to get the XML PR merged so I can unblock further work that depends on it.

scripts/schema-test-coverage.mjs
ralfhandl: Would checks of the keyword's correct use go into the `interpret` stubs?

scripts/schema-test-coverage.mjs
```diff
@@ -122,16 +192,13 @@ if (notCovered.length > 0) {
...
+  process.exitCode = 1;
 }
 
 console.log(
   "Covered:",
-  visitedLocations.size,
+  (allocations.length - notCovered.length),
   "of",
   allLocations.length,
   "(" + Math.floor((visitedLocations.size / allLocations.length) * 100) + "%)",
```
ralfhandl:
```suggestion
  "(" + Math.floor(((allocations.length - notCovered.length) / allLocations.length) * 100) + "%)",
```

ralfhandl (changes requested):
Need to eliminate all uses of `visitedLocations.size`

scripts/schema-test-coverage.mjs
handrews: @ralfhandl they're not stubs.  Annotations are always valid, returning `true` is correct.

handrews: At least I think.  That was as far as I got in the library's documentation.  It's at least closer to correct than it was, and we can work on it more later.

handrews:
@ralfhandl merged your latest suggestion, thanks.

ralfhandl (approved):

scripts/schema-test-coverage.mjs
jdesrosiers: > These seem to be unnecessary, commenting them out does not change the test result.

It works because the vocabulary is set as optional. If there is no keyword implementation, it's treated like an unknown keyword, which is allowed, but not quite the same thing.

These keyword implementations are already provided, so you shouldn't need to define them here. Instead of importing `@hyperjump/json-schema/draft-2020-12`, use `@hyperjump/json-schema/openapi-3-1`. Then the keywords and vocabulary definitions will be loaded and you can skip this. However, you still need to register the local schemas because you're testing the local schemas, not the ones included with the `@hyperjump/json-schema`.

scripts/schema-test-coverage.mjs
jdesrosiers: > This apparently needs to be parameterized with the OpenAPI version of the branch it is running on

I didn't realize you were versioning the vocabulary now. So, ignore what I said about not needing `defineVocabulary`, but you still shouldn't need to implement the keywords as long as you use the correct import.

PR #4702, opened 2025-06-13
Remove apparenlty redundant test script.

handrews:
_[**NOTE:** The schema coverage will fail because the script is not smart enough to figure out that this branch derives from `dev`.]_

This script confused me greatly when trying to fix the coverage, as it _looks_ like the test script that runs.  But it is not, the tests run out of the scripts/ directory.

Let's remove the confusion (or if this can't be removed, make the situation more clear).


ralfhandl:
> This script confused me greatly when trying to fix the coverage, as it _looks_ like the test script that runs. But it is not, the tests run out of the scripts/ directory.

This script is run by `vitest` and is the "parent" of the test coverage script.

handrews:
@ralfhandl I have changed this PR to make the same changes as to the other script.  It's not idea with the duplication, but please let's get this unblocked first and optimize/refactor later.

handrews:
BTW the commit to fix the `jsonSchemaDialect` stuff was cherry-picked from `v3.1-dev`, hopefully git will recognize it is already on that branch, and I can resolve conflicts on the port to `v3.2-dev`

ralfhandl (approved):

tests/schema/schema.test.mjs
```diff
@@ -26,7 +26,64 @@ const parseYamlFromFile = (filePath) => {
...
+
+defineVocabulary(
+  "https://spec.openapis.org/oas/3.1/vocab/base",
+  {
+    "discriminator": "https://spec.openapis.org/oas/schema/vocab/keyword/discriminator",
+    "example": "https://spec.openapis.org/oas/schema/vocab/keyword/example",
+    "externalDocs": "https://spec.openapis.org/oas/schema/vocab/keyword/externalDocs",
+    "xml": "https://spec.openapis.org/oas/schema/vocab/keyword/xml",
+  },
+);
```
jdesrosiers, line 81: You shouldn't need to define the vocabulary or keywords. They're already loaded when you import from `@hyperjump/json-schema/openapi-3-1`.

handrews: @jdesrosiers thanks for commenting- the problem is that we're using a schema with a different `$id` from the published 3.1 schema, as we're testing the in-development branches.  Also I was mostly trying to get something else unblocked so I didn't dig as deeply as I might have otherwise.

jdesrosiers:
> Not sure if/how we can fold the "fail" test cases into the coverage script.

There shouldn't be any reason why you can't run the fail tests as well. You're throwing an error when a test fails, but for the purpose of test coverage, we shouldn't care whether the test passes or fails. You could just remove that check and run both sets of tests.

tests/schema/schema.test.mjs
jdesrosiers: > the problem is that we're using a schema with a different `$id` from the published 3.1 schema

Right. Registering the meta-schemas immediately after this block addresses that concern. It's just defining the vocabulary (`defineVocabulary`) and the keywords (`addKeyword`) that's unnecessary.

ralfhandl: > It's just defining the vocabulary (`defineVocabulary`) and the keywords (`addKeyword`) that's unnecessary.

Apparently we still need `defineVocabulary`: just registering the meta-schemas results in 

    Error: Unrecognized vocabulary: https://spec.openapis.org/oas/3.2/vocab/base.
    You can define this vocabulary with the 'defineVocabulary' function.

when validating 3.2 OpenAPI descriptions.

I expected that registering `meta.yaml` - whose title and description claim that it defines the OAS Base Vocabulary - would actually define the vocabulary 😞

jdesrosiers: Oh, yeah. I never added support for 3.2. I'll work on it.

PR #4725, opened 2025-06-14
Use matching jsonSchemaDialect

ralfhandl:
Since we are testing with a placeholder, we need to match the placeholder.  This will unfortunately need to be different on each new release line branch, so let's separate this test case into its own file.


ralfhandl:
@handrews Maybe this helps resolving the conflict: one commit cherry-picked unchanged from `dev`, then a commit to make it 3.2-compatible.

PR #4726, opened 2025-06-15
v3.2: (port of #4614) Clarify that Request Body Objects need a body

handrews:
Port of PR #4614, which AFAICT is the only 3.1 PR that still needed to be ported to 3.2.

We require `content` but failed to require it to be non-empty, even though a request body without a body does not make any sense.


mikekistler (approved):
Looks good. 👍

PR #4728, opened 2025-06-16
v3.2: Allow Media Type Object re-use

handrews:
_[**NOTE:** See [this later comment](https://github.com/OAI/OpenAPI-Specification/pull/4728#issuecomment-3091349130) for a why this PR has been changed to only support `$ref`-ing Media Type Objects, and **not** Encoding Objects.]_

Add the Media Type Object ~~and Encoding Object~~ to the Components Object, and allow a Reference Object anywhere they are allowed.  To ensure that re-usable Objects can be documented clearly, add a `description` field ~~to both Objects~~. _[Note:  If adding `description` is more controversial than just supporting re-use, I could be persuaded to drop it.]_

One important thing to do as a release approaches is to try to use the new features more.  As I have done this, I've realized that our increased usage and functionality of Media Type and Encoding Objects (which now handle streaming JSON, and are expected to handle a wider variety of multipart media types and be more useful with XML) make re-use of those objects a much more compelling prospect.  While your average `application/json` Media Type Object is trivial, once you start using more complex media types, the duplication becomes substantial.

### Explanation and Examples

One use case might come from @jeremyfiel who, IIRC, works with an API  where `multipart/mixed` usage that combines JSON metadata with binary data is common.  Depending on how complex and common that format is, the media type object might be worth re-using in many places.  We're expanding the support for `multipart` a lot, and in general `multipart` Media Type Objects are quite complex, with many Encoding Objects (which may have substantial commonality themselves if similar parts are used in different `multipart` configurations) plus Example Objects.

But to bring things back to a concrete example, here is a pattern that was at one time fairly common- allowing querying either through the URL query string, or through putting the query contents in the body of a POST in case the URL got too long (although here I am showing QUERY instead of POST).

Assume both of the examples to follow start with these components (I pulled them out to make the length comparisons more obvious):

```yaml
openapi: 3.2.0 
info:
  title: API
components:
  schemas:
    PetFields:
      enum: 
      - name
      - petType
      - color
      - gender
      - breed
    Pet:  
      propertyNames:
        $ref: '#/components/schemas/PetFields'
      additionalProperties:
        type: string
      required:
      - name
      - petType
    ListForQuery:
      type: array 
      items:
        type: string      
    PetQueryLists:
      additionalProperties:
        $ref: '#/components/schemas/ListForQuery'
      propertyNames:
        $ref: '#/components/schemas/PetFields'
  examples:
    CatSearch:
      summary: Search for Persian or Ragamuffin cats
      dataValue:
        petType:
        - Cat
        breed:
        - Persian
        - Ragamuffin
      serializedValue: petType=Cat&breed=Persian,Ragamuffin
  responses:
    Pets:
      content:
        application/json:
          schema:
            type: array
              items:
                $ref: '#/components/schemas/Pet'
```

Here it is with re-use, at 50 additional lines:

```yaml
...
components:
   ...
  parameters:
    PetQuery:
      name: pets
      in: querystring
      content:
        application/x-www-form-urlencoded:
          $ref: '#/components/mediaTypes/PetQuery'
      examples:
        CatSearch:
          $ref: '#/components/examples/CatSearch'
  mediaTypes:
    PetQuery:
      description: A query format for pets.
      schema:
        $ref: '#/components/schemas/PetQueryLists'
      encoding:
        name: 
          style: form
          explode: false
        petType:
          style: form
          explode: false
        color:
          style: form
          explode: false
        gender:
          style: form
          explode: false
        breed:
          style: form
          explode: false
      examples:
        CatSearch:
          $ref: '#/components/examples/CatSearch'
paths:
  /pets:
    get:
      parameters:
      - $ref: '#/components/parameters/PetQuery'
      responses:
        "200":
          $ref: '#/components/responses/Pets'
    query:
      requestBody:
        content:
          application/x-www-form-urlencoded:
            $ref: '#/components/mediaTypes/PetQuery'
      responses:
        "200":
          $ref: '#/components/responses/Pets'
```

Without re-use, it takes 66 additional lines.  ~~However, there is another wrinkle if we wanted to put in some descriptive text, which we'll look at afterwards~~ _[without Encoding Object re-use included, the descriptive text issue is no longer relevant]_:

```yaml
...
components:
   ...
  parameters:
    PetQuery: 
      name: pets
      in: querystring
      content:
        application/x-www-form-urlencoded:
          schema:
            $ref: '#/components/schemas/PetQueryLists'
          encoding:
            name:
              style: form
              explode: false
            petType:
              style: form
              explode: false
            color:
              style: form
              explode: false
            gender:
              style: form
              explode: false
            breed:
              style: form
              explode: false
          examples:
            CatSearch:
              $ref: '#/components/examples/CatSearch'
      examples:
        CatSearch:
          $ref: '#/components/examples/CatSearch'
paths:
  /pets:
    get:
      parameters:
      - $ref: '#/components/parameters/PetQuery'
      responses:
        "200":
          $ref: '#/components/responses/Pets'
    query:
      requestBody:
        content:
          application/x-www-form-urlencoded:
            schema:
              $ref: '#/components/schemas/PetQueryLists'
            encoding:
              name:
                style: form
                explode: false
              petType:
                style: form
                explode: false
              color:
                style: form
                explode: false
              gender:
                style: form
                explode: false
              breed:
                style: form
                explode: false
            examples:
              CatSearch:
                $ref: '#/components/examples/CatSearch'
      responses:
        "200":
          $ref: '#/components/responses/Pets'
```

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -633,6 +633,7 @@ All objects defined within the Components Object will have no effect on the API 
 | <a name="components-links"></a> links | Map[`string`, [Link Object](#link-object) \| [Reference Object](#reference-object)] | An object to hold reusable [Link Objects](#link-object). |
 | <a name="components-callbacks"></a> callbacks | Map[`string`, [Callback Object](#callback-object) \| [Reference Object](#reference-object)] | An object to hold reusable [Callback Objects](#callback-object). |
 | <a name="components-path-items"></a> pathItems | Map[`string`, [Path Item Object](#path-item-object)] | An object to hold reusable [Path Item Objects](#path-item-object). |
+| <a name="components-media-types"></a> mediaTypes | Map[`string`, [Media Type Objects](#media-type-object) \| [Reference Object](#reference-object)] | An object to hold reusable [Media Type Objects](#media-type-object). |
 
 This object MAY be extended with [Specification Extensions](#specification-extensions).
 
@@ -1014,7 +1015,7 @@ For use with `in: "querystring"` and `application/x-www-form-urlencoded`, see [E
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
-| <a name="parameter-content"></a>content | Map[`string`, [Media Type Object](#media-type-object)] | A map containing the representations for the parameter. The key is the media type and the value describes it. The map MUST only contain one entry. |
+| <a name="parameter-content"></a>content | Map[`string`, [Media Type Object](#media-type-object) \| [Reference Object](#reference-object)] | A map containing the representations for the parameter. The key is the media type and the value describes it. The map MUST only contain one entry. |
 
 ##### Style Values
 
@@ -1204,7 +1205,7 @@ Describes a single request body.
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
 | <a name="request-body-description"></a>description | `string` | A brief description of the request body. This could contain examples of use. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation. |
-| <a name="request-body-content"></a>content | Map[`string`, [Media Type Object](#media-type-object)] | **REQUIRED**. The content of the request body. The key is a media type or [media type range](https://www.rfc-editor.org/rfc/rfc9110.html#appendix-A) and the value describes it. The map SHOULD have at least one entry; if it does not, the behavior is implementation-defined. For requests that match multiple keys, only the most specific key is applicable. e.g. `"text/plain"` overrides `"text/*"` |
+| <a name="request-body-content"></a>content | Map[`string`, [Media Type Object](#media-type-object) \| [Reference Object](#reference-object)] | **REQUIRED**. The content of the request body. The key is a media type or [media type range](https://www.rfc-editor.org/rfc/rfc9110.html#appendix-A) and the value describes it. The map SHOULD have at least one entry; if it does not, the behavior is implementation-defined. For requests that match multiple keys, only the most specific key is applicable. e.g. `"text/plain"` overrides `"text/*"` |
 | <a name="request-body-required"></a>required | `boolean` | Determines if the request body is required in the request. Defaults to `false`. |
 
 This object MAY be extended with [Specification Extensions](#specification-extensions).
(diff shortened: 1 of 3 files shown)


handrews:
Accidentally pushed an intermediate commit, had to force-push to fix.  This should be correct now- also fixed the title as this is proposed for 3.2 based on trying to use new 3.2 features in non-trivial ways.

src/oas.md
```diff
@@ -1249,11 +1251,12 @@ See [Working With Examples](#working-with-examples) for further guidance regardi
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
+| <a name="media-type-description"></a>description | `string` | A brief description of the media type. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation. |
```
ralfhandl: Do we also want to add `summary` which we seem to have added everywhere else?

Same for the Encoding Object.

tests/schema/pass/media-type-examples.yaml
ralfhandl, line 1: We now also need fail test cases with non-object entries in the new components maps, and non-string descriptions in these objects.

src/oas.md
handrews: @ralfhandl according to @lornajane , [no we do not](https://github.com/OAI/OpenAPI-Specification/pull/4610).  Or at least, "no we should not do that automatically but should debate each one," and I (eventually) agreed with her [reasoning](https://github.com/OAI/OpenAPI-Specification/pull/4610#issuecomment-2906917215), which is grounded in her experience with documentation tools.

tests/schema/pass/media-type-examples.yaml
handrews: @ralfhandl We do not have those sorts of tests at the time of this PR, and I'm going to draw a line on moving the testing goal posts while a PR is in flight.  I'm happy to add them later.

ralfhandl (approved):

handrews:
Hmm... I have run into a problem with `$ref`-ing Encoding Objects, at least if we add support for examples to them.  And it feels like we should given that we want people to use Encoding Objects for `style`/`explode`/etc. in the query string.

Examples of individual form fields include the name (e.g. `foo=bar`, not just `bar`), but the name is not part of the Encoding Object.  That's fine if the Encoding Object MUST be inline, as that keeps it under the parent `encoding` field which does have the name.  But if it's `$ref`-able, it would be used for fields with different names, and any `example`/`examples` field that we added to the Encoding Object would only be right for one of the names. 😵‍💫 

The Header Object would have the same problem, except header examples do _not_ include the header name (see PR #4800 for why).  So it's fine to `$ref` Header Objects.

But at the moment, it feels like we can either `$ref` Encoding Objects, _or_ we can add example support to them, and I lean towards example support.

Of course it is possible to just do an example of the whole query string or `multipart` document at the Media Type Object level, but I feel like limiting examples to that might be problematic.  I need to think on it more.

lornajane:
I'd be in favour of adding the extra components field for just media types if that's simpler? It seems like it would have the widest use case too (at least the way I write OpenAPI!)

handrews:
@lornajane given the potential confusion with Encoding Objects I'm leaning that way.  I will update this PR.  There are also always Overlays if people really want to re-use Encoding Objects and can manage the name-vs-example thing in their own way.

handrews:
With apologies for the rebase, it was easier to `git reset --soft` and re-add the Media Type parts than to manually remove the Encoding Object parts.  This should now _only_ address Media Type Object re-use.

src/oas.md
```diff
@@ -633,6 +633,7 @@ All objects defined within the Components Object will have no effect on the API
 | <a name="components-links"></a> links | Map[`string`, [Link Object](#link-object) \| [Reference Object](#reference-object)] | An object to hold reusable [Link Objects](#link-object). |
 | <a name="components-callbacks"></a> callbacks | Map[`string`, [Callback Object](#callback-object) \| [Reference Object](#reference-object)] | An object to hold reusable [Callback Objects](#callback-object). |
 | <a name="components-path-items"></a> pathItems | Map[`string`, [Path Item Object](#path-item-object)] | An object to hold reusable [Path Item Objects](#path-item-object). |
+| <a name="components-media-types"></a> mediaTypes | Map[`string`, [Media Type Objects](#media-type-object) \| [Reference Object](#reference-object)] | An object to hold reusable [Media Type Objects](#media-type-object). |
```
```suggestion
| <a name="components-media-types"></a> mediaTypes | Map[`string`, [Media Type Object](#media-type-object) \| [Reference Object](#reference-object)] | An object to hold reusable [Media Type Objects](#media-type-object). |
```
To be in sync with the other table lines

ralfhandl (changes requested):
+1, minor nit

ralfhandl (approved):

Issue #667, opened 2016-04-22
Make it easier to define link headers


ePaul:
[RFC 5988, section 5](https://tools.ietf.org/html/rfc5988#section-5) is defining the `Link`HTTP header.
It is mainly used on responses to GET requests (but can happen on any kind of response), but can also be used sensible as request headers for PUT, and possibly for the (less common) LINK and UNLINK methods.

It can look like this:

``` http
Link: <http://example.com/TheBook/chapter2>; rel="previous";
      title="previous chapter"
```

or this (having a list of two links):

``` http
Link: </TheBook/chapter2>;
      rel="previous"; title*=UTF-8'de'letztes%20Kapitel,
      </TheBook/chapter4>;
      rel="next"; title*=UTF-8'de'n%c3%a4chstes%20Kapitel
```

(Examples are from the RFC.) As usual with HTTP headers, you could have written this last example also in two headers, like this:

``` http
    Link: </TheBook/chapter2>;
          rel="previous"; title*=UTF-8'de'letztes%20Kapitel
    Link: </TheBook/chapter4>;
          rel="next"; title*=UTF-8'de'n%c3%a4chstes%20Kapitel
```

I would like to be able to define that a link with a certain relation type (the value of the `rel` attribute) will/might be available in the response (or must/can be sent in a request). Currently (with OpenAPI 2.0), the best I came up with was this:

``` yaml
      responses:
        200:
          description: Retrieval was successful.
          headers:
            Link:
              type: string
              description: |
                Links to related resources, in the format defined by
                [RFC 5988](https://tools.ietf.org/html/rfc5988#section-5).
                This will include a link with relation type `next` to the
                next page, if there is a next page.
```

I guess I could make this an array and use `collectionFormat: csv`, though I'm not sure how either of these would handle the two-headers version.

For OpenAPI.next (i.e. 3.0, if possible) I would like some better way of doing this, including to say links of which relation are expected to be there (but not forbidding other ones, of course).

I might be that the "parameters can have schemas" (in #654) with serialization strategies (in #665) might help here, if it gets extended to response headers, too, and we get a special serialization strategy for those kinds of headers.

Any other ideas?

handrews:
@ePaul While it's certainly not ideal, PR #4673 shows one way to describe a `Link` header requiring a specific set of `rel` values.

I hope to tackle this more directly in v3.3.

handrews:
This is fixed in 3.2 which is now out!

PR #4740, opened 2025-06-18
v3.2: Add support for application/linkset[+json]

handrews:
Fixes:
*  #667

These media types solve our long-standing problems with modeling HTTP Link headers.  We do not need to define anything beyond noting how to use them in a Media Type Ojbect.  [RFC9264](https://www.rfc-editor.org/rfc/rfc9264.html) already defines the mapping between the JSON data model, the standalone HTTP-like format, and the actual HTTP `Link` header syntax.  _This fits well with our 3.2 theme of data modeling improvements._

When someone hands us a standardized solution for our problem, we should improve the standards ecosystem by adopting it :-)  Note that this is _much_ more powerful and flexible than the hack I used as an example [elsewhere](https://github.com/OAI/OpenAPI-Specification/pull/4673/files#diff-a662b9afa95a60c2a67cc48014a8d5e00fa6ab2bd5ba2a54531674a5dd398c73R2531), which required a mind-boggling combination of `anyOf`, `contains`, and `pattern` and can't do much beyond checking for the link relations that way.  Its also fragile due to the need to manage quoting and whitespace concerns, and the possibility of something elsewhere in a header matching a link relation type string.  RFC9264 has none of these problems.

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -4354,7 +4354,7 @@ This will expand to the result:
 
 [RFC6570](https://www.rfc-editor.org/rfc/rfc6570)'s percent-encoding behavior is not always appropriate for `in: "header"` and `in: "cookie"` parameters.
 In many cases, it is more appropriate to use `content` with a media type such as `text/plain` and require the application to assemble the correct string.
-Other media types, such as `application/linkset` (see [Modeling Link Headers](#modeling-link-headers) are directly suitable for use as `content` for specific headers.
+Other media types, such as `application/linkset` (see [Modeling Link Headers](#modeling-link-headers)), are directly suitable for use as `content` for specific headers.
 
 In some cases, setting `allowReserved: true` will be sufficient to avoid incorrect encoding, however many characters are still percent-encoded with this field enabled, so care must be taken to ensure no unexpected percent-encoding will take place.
 
--- a/src/oas.md
+++ b/src/oas.md
@@ -2451,6 +2451,67 @@ Using `content` with a `text/plain` media type is RECOMMENDED for headers where 
 | ---- | :----: | ---- |
 | <a name="header-content"></a>content | Map[`string`, [Media Type Object](#media-type-object)] | A map containing the representations for the header. The key is the media type and the value describes it. The map MUST only contain one entry. |
 
+##### Modeling Link Headers
+
+[[!RFC9264]] defines the `application/linkset` and `application/linkset+json` media types.
+The former is exactly the format of HTTP link header values except allowing additional whitespace for readability, while the latter is an equivalent JSON representation of such headers.
+
+To use either of these media types, the `schema` in the [Media Type Object](#media-type-object) MUST describe the links as they would be structured in the `application/linkset+json` format.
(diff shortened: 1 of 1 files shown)


dret:
I am always glad to see standards being useful and being adopted. And of course even more so when I played a role in creating these standards. But, @handrews, is there anything I need to actually review here other than being happy to see this happening?

src/oas.md
```diff
@@ -2451,6 +2451,67 @@ Using `content` with a `text/plain` media type is RECOMMENDED for headers where
 | ---- | :----: | ---- |
 | <a name="header-content"></a>content | Map[`string`, [Media Type Object](#media-type-object)] | A map containing the representations for the header. The key is the media type and the value describes it. The map MUST only contain one entry. |
 
+##### Modeling Link Headers
+
+[[!RFC9264]] defines the `application/linkset` and `application/linkset+json` media types.
```
dret, line 2456: I am assuming this pulls in the reference info automatically? If not, it probably needs to be added somewhere in the references section.

handrews: @dret yes, the `[[!]]` adds it to the normative references.  Which I think is correct as we do that with the main `multipart` RFCs?  Although I see I added the RFCs for `application/json-seq` and `+json-seq` to informative using `[[?]]`... so maybe I don't have the best grasp on which goes where.

handrews:
@dret 

> But, @handrews, is there anything I need to actually review here other than being happy to see this happening?

I'm mostly just checking to make sure I didn't egregiously misunderstand or misuse the RFC :-)

src/oas.md
```diff
@@ -4293,6 +4354,7 @@ This will expand to the result:
 
 [RFC6570](https://www.rfc-editor.org/rfc/rfc6570)'s percent-encoding behavior is not always appropriate for `in: "header"` and `in: "cookie"` parameters.
 In many cases, it is more appropriate to use `content` with a media type such as `text/plain` and require the application to assemble the correct string.
+Other media types, such as `application/linkset` (see [Modeling Link Headers](#modeling-link-headers) are directly suitable for use as `content` for specific headers.
```
lornajane: Small formatting fix

```suggestion
Other media types, such as `application/linkset` (see [Modeling Link Headers](#modeling-link-headers)), are directly suitable for use as `content` for specific headers.
```

src/oas.md
lornajane: So do we need to add that target to another file somewhere? Or is just the number enough for the build to know what to do?

handrews: The build knows what to do.  There is a database of things like RFCs that it consults.  When we have less-standard-standards (lol) _then_ we do need to add an entry in part of the build script.  But not for this.

darrelmiller:
I like the concept here and support adding this feature.

ralfhandl (approved):
Nice!

ralfhandl:
> I like the concept here and support adding this feature.

@darrelmiller Do you like it enough to approve this PR? 😎

lornajane (approved):

PR #4743, opened 2025-06-21
v3.2: Guidance on searching and evaluating schemas

handrews:
_**NOTE 1:** This is intended to clarify requirements that **already exist** but have never been well-defined, both by making certain things required and stating clearly that other things are not.  It is particularly relevant in light of the Encoding Object changes, although the vaguely-defined behavior predates the new features._

Some OAS features casually state that they depend on the type of data being examined, or implicitly carry ambiguity about how to determine how to parse the data.

This section attempts to provide some guidance and limits, requiring only that implementations follow the unambiguous, statically deterministic keywords `$ref` and `allOf`.

It also provides for just validating the data (when possible) and using the actual in-memory type when a schema is too complex to analyze statically.

One use of this is breaking apart schemas to use them with mixed binary and JSON-compatible data, and a new section has been added to address that.

Finally, a typo in a related section was fixed.

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -1641,7 +1641,7 @@ These fields MAY be used either with or without the RFC6570-style serialization 
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
-| <a name="encoding-content-type"></a>contentType | `string` | The `Content-Type` for encoding a specific property. The value is a comma-separated list, each element of which is either a specific media type (e.g. `image/png`) or a wildcard media type (e.g. `image/*`). Default value depends on the type (determined by a [schema search](#searching-schemas)) as shown in the table below. |
+| <a name="encoding-content-type"></a>contentType | `string` | The `Content-Type` for encoding a specific property. The value is a comma-separated list, each element of which is either a specific media type (e.g. `image/png`) or a wildcard media type (e.g. `image/*`). The default value depends on the type (determined by a [schema search](#searching-schemas)) as shown in the table below. |
 | <a name="encoding-headers"></a>headers | Map[`string`, [Header Object](#header-object) \| [Reference Object](#reference-object)] | A map allowing additional information to be provided as headers. `Content-Type` is described separately and SHALL be ignored in this section. This field SHALL be ignored if the media type is not a `multipart`. |
 
 This object MAY be extended with [Specification Extensions](#specification-extensions).


src/oas.md
```diff
@@ -288,6 +288,40 @@ The formats defined by the OAS are:
...
+
+If the data is in a form that can be validated by the relevant Schema Object and is determined to be valid, implementations MUST support detecting characteristics such as JSON type or property or item structure from the data, whether it can be gleaned from the schema(s) or not.
+If `format` or the `content*` keywords are involved in further characterizing the data, these can be obtained as [annotation results](#extended-validation-with-annotations).
+
+##### Locating Schemas and Keywords
+
+When the data is in a non-JSON format, particularly one such as XML or various form media types where data is stored as strings without type information, it can be necessary to find this information through the relevant Schema Object to determine how to parse the format into a structure that can be validated by the schema.
+As schema organization can become very complex, implementations are not expected to handle every possible schema layout.
+However, given a known starting point schema (usually the value of the nearest `schema` field), implementations MUST search the following for the relevant keywords (e.g. `type`, `format`, `contentMediaType`, etc.):
```
karenetheridge: Is this suggesting that if `contentMediaType` is present at the top level of the schema that the entire data instance should be decoded with that media type before passing to validation? That would conflict with the expectation by the schema itself that the data instance is a string and that `contentSchema` should be used for validation of the decoded instance -- if the instance is already decoded to an object, `contentSchema` will do nothing.  I think we should recommend that an explicit media type should be specified instead, which is available for all places that a schema is used (parameters as well as message bodies).

Additionally, we could specify that schema keywords that require a specific type (e.g. `uniqueItems`, `properties`) can be used to infer a specific data type for the instance. I also wouldn't mind if we _didn't_ do this, and instead required the user to be explicit with a `type` keyword, so an implementation doesn't have to enumerate all of the type-specific keywords present in JSON Schema (it's version specific, e.g. an OAD that happened to use the a schema dialect under draft2019-09 would need to include `additionalItems` -- and asking the OAD parser to also have to parse the `$schema` keyword is going a little far).

src/oas.md
```diff
@@ -288,6 +288,40 @@ The formats defined by the OAS are:
 
 As noted under [Data Type](#data-types), both `type: number` and `type: integer` are considered to be numbers in the data model.
 
+#### Determining Type and Structure
+
+Several features of the OpenAPI Specification depend on detecting data characteristics such as type, format, media type, and object property or array item structure.
+
+If the data is in a form that can be validated by the relevant Schema Object and is determined to be valid, implementations MUST support detecting characteristics such as JSON type or property or item structure from the data, whether it can be gleaned from the schema(s) or not.
```
karenetheridge: I'm not sure what this means. What other detection methods are available other than from the schema, that aren't already present as a media type, encoding, or charset via one of the headers?  Is there a simple example that can be provided?

src/oas.md
```diff
@@ -288,6 +288,40 @@ The formats defined by the OAS are:
...
+If `format` or the `content*` keywords are involved in further characterizing the data, these can be obtained as [annotation results](#extended-validation-with-annotations).
```
karenetheridge: As a side note, future versions of JSON Schema will be validating `format` by default, and perhaps the content keywords too, so it might be worth noting somewhere that this behaviour is dialect-specific, and implementations that contain out-of-band configuration options to enable these validation features need not have to make use of annotations.

src/oas.md
karenetheridge: Also, is this suggesting that data deeper within the instance (for example items in an array) need type inference as well, or is this intended only as a mechanism for determining the type of the overall data instance?

If the latter, there is no need to perform schema inspection for an XML document, as surely the media type would be indicating that.  The only application I've found in my implementation for needing type inference is for the parameter style and explode features, which require an array or object to operate on and do not use media types.

> Is this suggesting that if contentMediaType is present at the top level of the schema that the entire data instance should be decoded with that media type before passing to validation?

No.  I'll have to go back over this in detail to try to figure out how it seems to be implying that and improve it.  I'm trying to cover several scenarios here and I probably muddled them together a bit.

> Additionally, we could specify that schema keywords that require a specific type

I spent too much effort ripping a rule somewhat like that out of 3.x (`type: array` required `items` in 3.0) to be willing to do such a thing now (and I appreciate that you would rather not do that either).  More seriously, the goal here is to strike a balance of allowing some authoring flexibility in arranging schemas without requiring implementors to do too much.  Inferring types from other keywords, which can quickly get contradictory when dealing with dynamically typed structures, is definitely over the line into "too much" for me.

> and instead required the user to be explicit with a `type` keyword

I don't want to outright require one in any specific place either.  I'd like to make it clear where things _definitely_ MUST work, and make it clear that there are a lot of things beyond that that won't be expected to work.  But there might be unexpected configurations that will work fine in practice, or that specific tools support (perhaps already implemented when deciding how to handle the ambiguity in the first place).

> Also, is this suggesting that data deeper within the instance (for example items in an array) need type inference as well, or is this intended only as a mechanism for determining the type of the overall data instance?

It's for whatever requirements in the spec need this kind of things.  Correlating Encoding Objects with Schema Objects is probably the most obvious and complicated.  And then once you've done that, figuring out the default `contentType`.  Other use cases are figuring out the destination type of ambiguous strings (e.g. query parameters).

> If the latter, there is no need to perform schema inspection for an XML document, as surely the media type would be indicating that.

No this is the _"where data is stored as strings without type information"_ use case, where it might be unclear how to parse `<foo attr="true">42</foo>`.  Are those the boolean `true` and the number `42` or the strings `"true"` and `"42"` or some combination?  That's distinct from the `multipart` part use case or other embedded media type use case where you might need to figure out a media type that does not have a Media Type Object with parent key- see PR #4744 for this specific use case.

src/oas.md

If `{"foo": 42}` is valid then you can determine that it is an object and that the value of `"foo"` is a number and mathematical integer, whether you could find a `type` keyword or not.

src/oas.md
handrews: I'll leave updating the OAS for any future JSON Schema developments to someone else.

handrews:
@karenetheridge while I have your attention, do you think this is fine where it is or should it go under the Schema Object somewhere?  I really could not decide.

src/oas.md
handrews: @karenetheridge to make this part more clear:

> But there might be unexpected configurations that will work fine in practice

For example, if you set the Encoding Object's `contentType` explicitly, there's no need to locate any schema information to determine its default.  But also I don't want complex rules around "if you don't set `contentType` explicitly, your schema MUST look like...."  Rather, if it is unclear whether tooling can detect the default behavior, authors should heed the limitations in the spec and decide to set `contentType` explicitly themselves.

Similarly, when parsing query parameters and trying to figure out the type, maybe tools just try parsing them in that type order I give, on the grounds that the precedence is probably right.  Basically treating it as an "any" type and guessing.  As long as they validate it afterwards, that should be pretty reliable (I say "should" because somewhere there's some schema that could work out either of two ways and was intended to go the other way, but I can't figure out what that would look like off the top of my head).

handrews:
I'm putting this in draft because based on @karenetheridge's feedback I'm going to rework it fairly substantially, but it's still of use when understanding how it fits with the other related PRs.

The effect of the rewrite should be the same, but I think the wording and organization will be significantly different.  It's clear that the different use cases here need to be separated out and clarified.  I think this ended up being a bit oddly abstract because of how I tried to split things up into PRs that don't conflict.

handrews:
I have added a commit that almost totally rewrites this- you probably just want to review the whole thing and not look at the per-commit diff as it will be a mess.  The new version:

* Puts most things under the Schema Object
* Organizes use cases by the point in the process they occur relative to schema evaluation
* Links from elsewhere in the spec so that we do not need to include quite as much in the main part of the text

I do not think that has changed anything substantial, but it's essentially a new PR now.

src/oas.md
karenetheridge: Ok I'm not understanding then what you mean by "is determined to be valid". For what sort of usecase are you thinking here, and how would we know it is valid?  e.g. is this for client generation, where the input is the valid data and it then needs to be serialized into an HTTP message?

src/oas.md
karenetheridge: > I don't want to outright require one in any specific place either. I'd like to make it clear where things definitely MUST work, and make it clear that there are a lot of things beyond that that won't be expected to work.

One scenario here where a `type` keyword is required is for a header parameter -- an incoming value of "blue,black,brown" could be legitimately parsed as a string or an array, and we'd need to look at the schema to know which is expected. If we see `"type":"array"` at the top level of the schema then we can confidently parse into the array `["blue", "black", "brown"]` and at least have a chance of validating against the schema as a whole (whereas leaving it as a string will definitely be invalid).

In my implementation I don't do any of the guessing that you describe (try parsing into one type, see if it's valid, then parse it a different way and try again). The schema gets one crack at trying to validate the data, and if it failed to specify what type was expected, too bad!

It's nice that this is getting into the spec in clear language.  When I implemented parameter parsing I was following https://swagger.io/docs/specification/v3_0/serialization/ and tried to reverse-engineer what they were assuming, since these docs are all written from the client-side perspective (constructing an HTTP message from raw data) as opposed to the parsing and deserializing perspective.

src/oas.md
handrews: @karenetheridge have you read the new version of the PR?  I'd rather see if that makes things clear than continue here on outdated text.

src/oas.md
handrews: > The schema gets one crack at trying to validate the data, and if it failed to specify what type was expected, too bad!

There are too many scenarios where it's reasonable not to have a `type` right there in the inline schema.  If nothing else, when there is a `$ref`.  I wrestled a bit with how much to require, but following `$ref` and `allOf` seems straightforward enough.  But I am open to other requirements.  I have just seen plenty of implementations (including one I worked on - not publicly available sadly) that search those and roll up examples or flatten `allOf`s to make the documentation rendering better.  It seemed like a reasonable amount of searching.

I also considered allowing implementations to put some sort of step limit on it, like not following more than X number of `$ref`s, but it's not like that gets any harder the more you do, and figuring out how to count "steps" with both `$ref` and `allOf` hurt my brain :-)

-----

> One scenario here where a type keyword is required is for a header parameter -- an incoming value of "blue,black,brown" could be legitimately parsed as a string or an array, and we'd need to look at the schema to know which is expected.

This is a really good point as `"type": ["array", "string"]` would be a problem for that parameter.  I think the rules I wrote would end up with it treated as an array, as string is always the type of last resort.  But do we want that?  Unlike the use cases I was thinking of, this truly could go either way, so we're picking a winner here based on rules I made up as I wrote them.

The one multi-value `type` use case that I think we MUST (heh) require support for is adding `"null"` to another type (e.g. `"type": ["number", "null"]`).  If we think the multi-type rules are too complex or too likely to have unexpected behavior, we could just special-case it for one non-`"null"` type plus `"null"`.   I'm really uncertain as to what is best here and would love to hear more opinions.

handrews:
@karenetheridge I'm going to mark various threads as resolved since the text is now so different that they are confusing- please do not take that to mean I'm dismissing open questions, please just re-start whatever is needed with comments on the new text, or as new top-level comments.  Apologies for the inconvenience.

src/oas.md
karenetheridge: > This is a really good point as "type": ["array", "string"] would be a problem for that parameter. I think the rules I wrote would end up with it treated as an array, as string is always the type of last resort.

That's what I do -- array is selected first if it's possible, then object, then string is the final fallback as the "original" format of the data.

BTW after you created this PR I expanded my type checking to look for `allOf`s, which it does recursively. It was already following $refs for this (bombing out after a configurable max stack depth value is reached).  I pondered adding type inference from seeing array- or object-specific keywords but decided to shelve that for now. But `allOf` is a quite reasonable addition.

src/oas.md
```diff
@@ -2599,6 +2601,10 @@ Note that JSON Schema Draft 2020-12 does not require an `x-` prefix for extensio
 The [`format` keyword (when using default format-annotation vocabulary)](https://www.ietf.org/archive/id/draft-bhutton-json-schema-validation-01.html#section-7.2.1) and the [`contentMediaType`, `contentEncoding`, and `contentSchema` keywords](https://www.ietf.org/archive/id/draft-bhutton-json-schema-validation-01.html#section-8.2) define constraints on the data, but are treated as annotations instead of being validated directly.
 Extended validation is one way that these constraints MAY be enforced.
 
+In addition to extended validation, annotations are the most effective way to determine whether these keywords impact the type and structure of the fully parsed data.
+For example, formats such as `int64` can be applied to JSON strings, as JSON numers have limitations that make large integers non-portable.
```
```suggestion
For example, formats such as `int64` can be applied to JSON strings, as JSON numbers have limitations that make large integers non-portable.
```

src/oas.md
```diff
@@ -2599,6 +2601,10 @@ Note that JSON Schema Draft 2020-12 does not require an `x-` prefix for extensio
...
+If annotation collection is not available, implementations MUST perform a [schema search](#searching-schemas) for these keywords, and MUST document the limitatioons this imposes.
```
```suggestion
If annotation collection is not available, implementations MUST perform a [schema search](#searching-schemas) for these keywords, and MUST document the limitations this imposes.
```

src/oas.md
```diff
@@ -2611,6 +2617,106 @@ Even when read-only fields are not required, stripping them is burdensome for cl
...
+For example, if `foo` had the schema `{"type": "string", "format": "int64")`, the data structure used for validation would still be the same, but the application will need to convert the string `"42"` to the 64-bit integer `42`.
+Similarly, the `content*` keywords can indicate further structure within a string.
+
+Implementations MUST either use [annotation collection](#extended-validation-with-annotations) to gather this information, or perform a [schema search](#searching-schemas), and MUST document which approach it implements.
+
+Note that parsing string contents based on `contentMediaType` carries the same security risks as parsing HTTP message bodies based on `Content-Type`; see [Handling External Resources](#handling-external-resources) for further information.
+
+###### Schema Evaluation and Binary Data
+
+As noted under [Working with Binary Data](#working-with-binary-data), Schema Objects for binary documents do not use any standard JSON Schema assertions, as the only ones that could apply (`const` and `enum`) would require embedding raw binary into JSON which is not possible.
```
karenetheridge: well not quite.. there's also `minLength` and `maxLength`. But I think I'd be inclined to check the `Content-Length` header for this instead.

src/oas.md
```diff
@@ -2611,6 +2617,106 @@ Even when read-only fields are not required, stripping them is burdensome for cl
...
+
+However, `multipart` media types can mix binary and text-based data, leaving implementations with two options for schema evaluations:
+
+1. Use a placeholder value, on the assumption that no assertions will apply to the binary data and no conditional schema keywords will cause the schema to treat the placeholder value differently (e.g. a part that could be either plain text or binary might behave unexpectedly if a string is used as a binary placeholder, as it would likely be treated as plain text and subject to different subschemas and keywords).
```
karenetheridge, line 2687: `"format": "binary"` would be an ideal signal for this.

src/oas.md
```diff
@@ -2611,6 +2617,106 @@ Even when read-only fields are not required, stripping them is burdensome for cl
...
+Implementations MUST document which strategy or strategies they use, as well as any known limitations.
+
+##### Searching Schemas
+
+Several OAS features require searching Schema Objects for keywords indicating the data type and/or structure.
+Even if the requirement is given in terms of schema keywords, if the data is in a form [suitable for schema evaluation](#preparing-data-for-schema-evaluation) and the necessary information (including type) can be determined by inspecting the data (and possibly also annotations such as `format`), implementations MUST support doing so as this is effective regardless of how schemas are structured.
+
+If this is not possible, the schemas MUST be searched to see if the information can be determined without performing evaluation.
+As schema organization can become very complex, implementations are not expected to handle every possible schema layout.
+However, given a known starting point schema (usually the value of the nearest `schema` field), implementations MUST search the following for the relevant keywords (e.g. `type`, `format`, `contentMediaType`, `properties`, `prefixItems`, `items`, etc.):
```
karenetheridge: If we say MUST, we can't give an incomplete example list of keywords - we need to spell out exactly which ones are supported.

Instead we could say "..implementations SHOULD search the following for relevant keywords such as ..., and MUST document which keywords are supported."

src/oas.md
```diff
@@ -2611,6 +2617,106 @@ Even when read-only fields are not required, stripping them is burdensome for cl
...
+
+In some cases, such as correlating [Encoding Objects](#encoding-object) with Schema Objects using fields in a [Media Type Object](#media-type-object), it is be necessary to first find a keyword such as `properties`, and then treat its subschema(s) as starting point schemas for further searches.
+
+Implementations MAY analyze subschemas of other keywords such as `oneOf` or `dependentSchemas`, or examine possible `$dynamicRef` targets, and MUST document the extent and nature of any such additional support.
+
+###### Handling Multiple Types
+
+When a `type` keyword with multiple values (e.g. `type: ["number", "null"]`) is found, implementations MUST attempt to use the types as follows, ignoring any types not present in the `type` list:
+
+1. Determine if the data can be parsed as whichever of `null`, `number`, `object`, or `array` are present in the `type` list, treating `integer` as `number` for this step.
```
karenetheridge: We can also parse booleans from the strings "0", "1", "false", "true". I have often seen query parameters using 0 or 1 as a flag value.

src/oas.md
handrews: @karenetheridge so do you think we should keep the type precedence list for multi-valued `type`?  Sounds like you were doing something close to this?

Regarding the header parameter, I did some digging and there seem to be three cases:

1.  Headers that _cannot_ be list-valued, in which case `blue,black,brown` need not be quoted and is just a string
2. Headers that _can_ be list-valued in the  normal way, where the values can appear on a single line, separated by commas, in which case `blue,black,brown` is an array of strings but `"blue,black,brown"` is a string (because of the quotes... `"blue","black","brown"` would also be an array, with unnecessary but allowed quotes)
3. `Set-Cookie`, which has to be treated differently from everything else per its own RFC and per RFC9110 (because it can be list-valued, but it also uses commas in unquoted values so you can't use commas as a delimiter on a single line, you have to put it on multiple lines)

src/oas.md
handrews: @karenetheridge oh good point, we even say something about `maxLength` in binary streaming... I'll rework this a bit.

src/oas.md
handrews: That's not a format in 3.1+. the appropriate signal would be `"contentMediaType": "application/octet-stream"` or similar.  But because of how the Encoding Object works, it's all more complicated.

src/oas.md
handrews: The MUST for that is at the point of each use.  So in some cases you MUST look for `type` but in otherse you MUST look for `properties`, etc.... it's not always the same list.

src/oas.md
handrews: That's basically why string is the last.  I'm not sure what to do here and honestly it's a strong argument for saying that the only thing people need to support is `type: ["X", "null"]` and let everything else be implementation-defined.  Because if you have an instance of `0` or the empty string and `type: "boolean"` it's easy enough to parse it as `false`, but if you have `type: ["boolean", "number"]` then how do you  know whether `0` is the boolean `false` or the number `0`?  I am increasingly inclined to _not_ try to solve this level of type determination. 🤔

src/oas.md
karenetheridge: There are many other binary media types, e.g. `image/jpeg`, where it's not useful to apply a schema to the content and we might want to use a placeholder.

I'm now wondering if we need a signal for "do not bother trying to deserialize this content for the purpose of validation". At present my implementation always applies the appropriate media type deserialization to the content and applies the schema to it (and throws an error if it doesn't know how to process that media type, or if the deserializer encountered an error), but bypassing that might be useful.

src/oas.md
karenetheridge: Ok I see what you mean.

handrews: @karenetheridge I added some language to clarify this in the most recent commit [EDIT: 2nd-most-recent commit now].

src/oas.md
handrews: @karenetheridge I meant "some binary media type in `contentMediaType`" more than specifically `application/octet-stream`, but arguably tacking "or similar" on as I did was not sufficiently clear!

I'm open to better ideas here.  Perhaps it is best to remove the placeholder idea as it is pretty squishy and can easily cause problems, and just document the schema search (and break apart) process, which leverages things implementations have to do anyway.  We could include a brief note about the possibilities and dangers of using a placeholder instead?  idk.

> I'm now wondering if we need a signal for "do not bother trying to deserialize this content for the purpose of validation". At present my implementation always applies the appropriate media type deserialization to the content and applies the schema to it (and throws an error if it doesn't know how to process that media type, or if the deserializer encountered an error), but bypassing that might be useful.

I'm not entirely sure what to do with this.  I sort-of follow you, but I would not be surprised if you have the only JSON Schema implementation that can handle this at all, and I want to write the requirements for the typical case.

handrews:
@karenetheridge I trimmed back the multi-valued `type` requirements as from our discussion I just see too many ways it can go wrong.  Now it's just "if you have [X, "null"] treat it like X" and everything else is optional guidance.  How does that sit with you?

src/oas.md
handrews: @karenetheridge fixed in the most recent commit.

handrews:
@karenetheridge I'm marking various threads resolved as I think subsequent commits addressed them, and it's a lot of at least somewhat outdated discussion for folks to have to read through before tomorrow's call.  Please feel free to re-raise anything that is still not addressed.

src/oas.md
```diff
@@ -1639,7 +1641,7 @@ These fields MAY be used either with or without the RFC6570-style serialization
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
-| <a name="encoding-content-type"></a>contentType | `string` | The `Content-Type` for encoding a specific property. The value is a comma-separated list, each element of which is either a specific media type (e.g. `image/png`) or a wildcard media type (e.g. `image/*`). Default value depends on the property type as shown in the table below. |
+| <a name="encoding-content-type"></a>contentType | `string` | The `Content-Type` for encoding a specific property. The value is a comma-separated list, each element of which is either a specific media type (e.g. `image/png`) or a wildcard media type (e.g. `image/*`). Default value depends on the type (determined by a [schema search](#searching-schemas)) as shown in the table below. |
```
```suggestion
| <a name="encoding-content-type"></a>contentType | `string` | The `Content-Type` for encoding a specific property. The value is a comma-separated list, each element of which is either a specific media type (e.g. `image/png`) or a wildcard media type (e.g. `image/*`). The default value depends on the type (determined by a [schema search](#searching-schemas)) as shown in the table below. |
```

karenetheridge (commented):
one small edit (not a change introduced by you, but still an improvement I think).

src/oas.md
```diff
@@ -2599,6 +2601,10 @@ Note that JSON Schema Draft 2020-12 does not require an `x-` prefix for extensio
 The [`format` keyword (when using default format-annotation vocabulary)](https://www.ietf.org/archive/id/draft-bhutton-json-schema-validation-01.html#section-7.2.1) and the [`contentMediaType`, `contentEncoding`, and `contentSchema` keywords](https://www.ietf.org/archive/id/draft-bhutton-json-schema-validation-01.html#section-8.2) define constraints on the data, but are treated as annotations instead of being validated directly.
 Extended validation is one way that these constraints MAY be enforced.
 
+In addition to extended validation, annotations are the most effective way to determine whether these keywords impact the type and structure of the fully parsed data.
+For example, formats such as `int64` can be applied to JSON strings, as JSON numbers have limitations that make large integers non-portable.
+If annotation collection is not available, implementations MUST perform a [schema search](#searching-schemas) for these keywords, and MUST document the limitations this imposes.
```
lornajane, line 2606:
```suggestion
If annotation collection is not available, implementations MUST perform a [schema search](#searching-schemas) for these keywords, and SHOULD document the limitations this imposes.
```

src/oas.md
```diff
@@ -2611,6 +2617,108 @@ Even when read-only fields are not required, stripping them is burdensome for cl
...
+
+In order to prepare the correct data structure for evaluation in such cases, implementations MUST perform a [schema search](#searching-schemas) for the `type` keyword.
+
+###### Applying Further Type Information
+
+The `format` keyword provides more fine-grained type information, and can even change the underlying data type for the purposes of the application.
+For example, if `foo` had the schema `{"type": "string", "format": "int64")`, the data structure used for validation would still be the same, but the application will need to convert the string `"42"` to the 64-bit integer `42`.
+Similarly, the `content*` keywords can indicate further structure within a string.
+
+Implementations MUST either use [annotation collection](#extended-validation-with-annotations) to gather this information, or perform a [schema search](#searching-schemas), and MUST document which approach it implements.
```
lornajane, line 2674:
```suggestion
Implementations MUST either use [annotation collection](#extended-validation-with-annotations) to gather this information, or perform a [schema search](#searching-schemas), and SHOULD document which approach it implements.
```

src/oas.md
```diff
@@ -2611,6 +2617,108 @@ Even when read-only fields are not required, stripping them is burdensome for cl
...
+When the entire instance is binary, this is straightforward as few keywords are relevant.
+
+However, `multipart` media types can mix binary and text-based data, leaving implementations with two options for schema evaluations:
+
+1. Use a placeholder value, on the assumption that no assertions will apply to the binary data and no conditional schema keywords will cause the schema to treat the placeholder value differently (e.g. a part that could be either plain text or binary might behave unexpectedly if a string is used as a binary placeholder, as it would likely be treated as plain text and subject to different subschemas and keywords).
+2. Perform [schema searches](#searching-schemas) to find the appropriate keywords (`properties`, `prefixItems`, etc.) in order to break up the subschemas and apply them separately to binary and JSON-compatible data.
+
+Implementations MUST document which strategy or strategies they use, as well as any known limitations.
+
+##### Searching Schemas
```
lornajane, line 2692: Question about moving this section a little further up the document, who has thoughts?

src/oas.md
```diff
@@ -2611,6 +2617,108 @@ Even when read-only fields are not required, stripping them is burdensome for cl
...
+
+OAS Implementations that do not have access to a binary-instance-supporting JSON Schema implementation MUST examine schemas and apply them in accordance with [Working with Binary Data](#working-with-binary-data).
+When the entire instance is binary, this is straightforward as few keywords are relevant.
+
+However, `multipart` media types can mix binary and text-based data, leaving implementations with two options for schema evaluations:
+
+1. Use a placeholder value, on the assumption that no assertions will apply to the binary data and no conditional schema keywords will cause the schema to treat the placeholder value differently (e.g. a part that could be either plain text or binary might behave unexpectedly if a string is used as a binary placeholder, as it would likely be treated as plain text and subject to different subschemas and keywords).
+2. Perform [schema searches](#searching-schemas) to find the appropriate keywords (`properties`, `prefixItems`, etc.) in order to break up the subschemas and apply them separately to binary and JSON-compatible data.
+
+Implementations MUST document which strategy or strategies they use, as well as any known limitations.
```
lornajane, line 2690:
```suggestion
Implementations SHOULD document which strategy or strategies they use, as well as any known limitations.
```

lornajane (commented):
Minor suggestions from the TSC call

src/oas.md
karenetheridge: (removed, commented in wrong section)

src/oas.md
karenetheridge: As discussed in the meeting, if implementations don't do this, what would they do instead?  If there isn't anything they can do, then I think the MUST would stand.

handrews: I really did not expect this PR to get hung up on a debate about how much to require implementations to document their behavior.  Which I _thought_ would be thoroughly non-controversial.  Why would we _not_ want them to do so?

So... I have no idea.  I want everyone else to resolve their differences around documentation requirements so it doesn't hang up this PR, that's my opinion on the matter.

handrews:
~~@lornajane @karenetheridge @duncanbeevers Can y'all sort out what we should be doing on documentation requirements and why?  I have no idea why MUST requirements around documenting behavior are controversial, but all I really care about is that this does not hang up this PR.  It sounds like @karenetheridge is disagreeing on one?  I just want a broadly applicable rule that tells me what to do here.~~ _See next comment_

handrews:
It seems like we really do not have any other requirements about documenting things in the spec, which I really thought we did.  Since this turns out to be controversial, I'm removing those statements entirely.

handrews:
In addition to removing the documentation requirements, I'm also reorganizing this a fair amount- stay tuned for an update, but it might be a few days.

handrews:
After continued consideration of last week's TDC call feedback from @whitlockjc and others, I'm reworking this to take a substantially different approach.

PR #4744, opened 2025-06-21
v3.2: Provide guidance for multiple contentType values

handrews:
We have always allowed multiple `contentType` values in the Encoding Object, but do not provide any guidance on how to determine which to use when performing the encoding.  This adds such guidance.

This has become more relevant with `multipart/mixed` support, as the only way to define a set of parts where the order of the media types is not predictable (perhaps because the number of each media type is variable) is to use `itemEncoding` and put all of the media types in `contentType`.  

***A PR with an example will be linked shortly.***

(also fix a copy-paste error that left a `+` at the beginning of a line)

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -1768,10 +1768,19 @@ See [Encoding Usage and Restrictions](#encoding-usage-and-restrictions) for guid
 
 Note that there are significant restrictions on what headers can be used with `multipart` media types in general ([RFC2046](https://www.rfc-editor.org/rfc/rfc2046.html#section-5.1)) and `multi-part/form-data` in particular ([RFC7578](https://www.rfc-editor.org/rfc/rfc7578.html#section-4.8)).
 
+###### Handling Multiple `contentType` Values
+
+When multiple values are provided for `contentType`, parsing remains straightforward as the part's actual `Content-Type` is included in the document.
+
+For encoding and serialization, implementations MUST provide a mechanism for applications to indicate which media type is intended.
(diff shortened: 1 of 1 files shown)


src/oas.md
```diff
@@ -1768,10 +1768,19 @@ See [Encoding Usage and Restrictions](#encoding-usage-and-restrictions) for guid
 
 Note that there are significant restrictions on what headers can be used with `multipart` media types in general ([RFC2046](https://www.rfc-editor.org/rfc/rfc2046.html#section-5.1)) and `multi-part/form-data` in particular ([RFC7578](https://www.rfc-editor.org/rfc/rfc7578.html#section-4.8)).
 
+###### Handling Multiple `contentType` Values
+
+When multiple values are provided for `contentType`, parsing remains straightforward as the part's actual `Content-Type` is included in the document.
+
+For encoding and serialization, implementations MUST provide a mechanism for applications to indicate which media type is intended.
+Implementations MAY choose to offer media type sniffing ([[!SNIFF]]) as an alternative, but this MUST NOT be the default behavior due to the security risks inherent in the process.
```
```suggestion
Implementations MAY choose to offer media type sniffing ([[SNIFF]]) as an alternative, but this MUST NOT be the default behavior due to the security risks inherent in the process.
```

The leading exclamation mark for a normative reference was valid in older respec versions, now default is normative and a question mark indicates non-normative.

ralfhandl (commented):
+1, minor nit

src/oas.md
handrews: @ralfhandl I'll merge this but we'll need to go do a pass because there are `[[!...]]]` constructs all over right now.

lornajane (approved):

ralfhandl (approved):

PR #4745, opened 2025-06-21
v3.2: (Split and smaller!) Support ordered multipart including streaming

handrews:
_**NOTE: This is just the new fields and their description, schema, and schema tests, with several things removed: 1.) the elaborate rules for determining type, which are now in PR #4743, 2.) the examples, which are in PR #4746 so they can be discussed in-depth and updated based on the pending Example Object changes.  The contents of this PR are otherwise identical to #4589._

-----

This adds support for all `multipart` media types that do not have named parts, including support for streaming such media types. Note that `multipart/mixed` defines the basic processing rules for all `multipart` types, and implementations that encounter unrecognized `multipart` subtypes are required to process them as `multipart/mixed`.  Therefore support for `multipart/mixed` addresses all other subtypes to some degree.

This builds on the recent support for sequential media types:

* `multipart/mixed` and similar meet the definition for a sequential media type, requiring it to be modeled as an array.  This does use an expansive definition of "repeating the same structure", where the structure is literally any content with a media type.
* As a sequential media type, it also supports `itemSchema`
* Adding a parallel `itemEncoding` is the obvious solution to `multipart/mixed` streams requiring an Encoding Object
* We have regularly received requests to support truly mixed `multipart/mixed` payloads, and previously claimed such support from 3.0.0 onwards, without actually supporting it. Adding `prefixEncoding` along with `itemEncoding` supports this use case with a clear parallel to `prefixItems`, which is the schema construct needed to support this case.
* There is no need for a `prefixSchema` field because the streaming use case requires a repetition of the same schema for each item. Therefore all mixed use cases can use `schema` and `prefixItems`

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -103,14 +103,18 @@ Some examples of sequential media types (including some that are not IANA-regist
   application/json-seq
   application/geo+json-seq
   text/event-stream
+  multipart/mixed
 ```
 
 In the first three above, the repeating structure is any [JSON value](https://tools.ietf.org/html/rfc8259#section-3).
-The fourth repeats `application/geo+json`-structured values, while the last repeats a custom text format related to Server-Sent Events.
+The fourth repeats `application/geo+json`-structured values, while `text/event-stream` repeats a custom text format related to Server-Sent Events.
+The final media type listed above, `multipart/mixed`, provides an ordered list of documents of any media type, and is sometimes streamed.
+Note that while `multipart` formats technically allow a preamble and an epilogue, the RFC directs that they are to be ignored, making them effectively comments, and this specification does not model them.
 
 Implementations MUST support mapping sequential media types into the JSON Schema data model by treating them as if the values were in an array in the same order.
 
 See [Complete vs Streaming Content](#complete-vs-streaming-content) for more information on handling sequential media types in a streaming context, including special considerations for `text/event-stream` content.
+For `multipart` types, see also [Encoding By Position](#encoding-by-position).
 
 #### Media Type Registry
 
@@ -1260,7 +1264,9 @@ See [Working With Examples](#working-with-examples) for further guidance regardi
 | <a name="media-type-item-schema"></a>itemSchema | [Schema Object](#schema-object) | A schema describing each item within a [sequential media type](#sequential-media-types). |
 | <a name="media-type-example"></a>example | Any | Example of the media type; see [Working With Examples](#working-with-examples). |
 | <a name="media-type-examples"></a>examples | Map[ `string`, [Example Object](#example-object) \| [Reference Object](#reference-object)] | Examples of the media type; see [Working With Examples](#working-with-examples). |
-| <a name="media-type-encoding"></a>encoding | Map[`string`, [Encoding Object](#encoding-object)] | A map between a property name and its encoding information, as defined under [Encoding Usage and Restrictions](#encoding-usage-and-restrictions).  The `encoding` field SHALL only apply when the media type is `multipart` or `application/x-www-form-urlencoded`. If no Encoding Object is provided for a property, the behavior is determined by the default values documented for the Encoding Object. |
+| <a name="media-type-encoding"></a>encoding | Map[`string`, [Encoding Object](#encoding-object)] | A map between a property name and its encoding information, as defined under [Encoding By Name](#encoding-by-name).  The `encoding` field SHALL only apply when the media type is `multipart` or `application/x-www-form-urlencoded`. If no Encoding Object is provided for a property, the behavior is determined by the default values documented for the Encoding Object. This field MUST NOT be present if `prefixEncoding` or `itemEncoding` are present. |
+| <a name="media-type-prefix-encoding"></a>prefixEncoding | [[Encoding Object](#encoding-object)] | An array of positional encoding information, as defined under [Encoding By Position](#encoding-by-position).  The `prefixEncoding` field SHALL only apply when the media type is `multipart`. If no Encoding Object is provided for a property, the behavior is determined by the default values documented for the Encoding Object. This field MUST NOT be present if `encoding` is present. |
+| <a name="media-type-item-encoding"></a>itemEncoding | [Encoding Object](#encoding-object) | A single Encoding Object that provides encoding information for multiple array items, as defined under [Encoding By Position](#encoding-by-position). The `itemEncoding` field SHALL only apply when the media type is `multipart`. If no Encoding Object is provided for a property, the behavior is determined by the default values documented for the Encoding Object. This field MUST NOT be present if `encoding` is present. |
 
 This object MAY be extended with [Specification Extensions](#specification-extensions).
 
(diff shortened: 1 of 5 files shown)


philsturgeon (approved):

miqui (dismissed):

handrews:
The most recent commit slightly relaxes the wording around using Encoding Objects and makes it clear that if you have more Encoding Objects than properties or items, this is not an error (which I think is not a change, but it depended on how you read a certain MUST).

I also relaxed the requirement that the property MUST exist "in the `schema`" (presumably under `properties`) as this is supplanted by the schema search guidance in PR #4743, which notes that you can just validate the data and then work with its type/structure directly whenever that is possible.

src/oas.md
```diff
@@ -1324,15 +1334,30 @@ See [Encoding the `x-www-form-urlencoded` Media Type](#encoding-the-x-www-form-u
...
+As with `prefixItems`, it is _not_ an error if the instance array is shorter than the `prefixEncoding` array; the additional Encoding Objects SHALL be ignored.
+
+The `itemEncoding` field can also be used with `itemSchema` to support streaming `multipart` content.
+
+###### Additional Encoding Approaches
+
+The `prefixEncoding` field can be used with any `multipart` content to require a fixed part order.
+This includes `multipart/form-data`, for which the Encoding Object's `headers` field MUST be used to provide the `Content-Disposition` and part name, as no property names exist to provide the names automatically.
+
+Prior versions of this specification advised using the `name` [`Content-Disposition` parameter](https://www.iana.org/assignments/cont-disp/cont-disp.xhtml#cont-disp-2) of the `form-data` [`Content-Disposition` value](https://www.iana.org/assignments/cont-disp/cont-disp.xhtml#cont-disp-1) with `multipart` media types other than `multipart/form-data` in order to work around the limitations of the `encoding` field.
```
ralfhandl: Line 1334 uses the formulation 

> the [`name` parameter](https://www.rfc-editor.org/rfc/rfc7578#section-4.2) of the `Content-Disposition: form-data` header of each part

which I find easier to parse. Could you use similar text here, and also reference RFC7578?

src/oas.md
```diff
@@ -1309,13 +1316,16 @@ properties:
...
-Array properties MUST be handled by applying the given Encoding Object to one part per array item, each with the same `name`, as is recommended by [[?RFC7578]] [Section 4.3](https://www.rfc-editor.org/rfc/rfc7578.html#section-4.3) for supplying multiple values per form field.
-For all other value types for both top-level non-array properties and for values, including array values, within a top-level array, the Encoding Object MUST be applied to the entire value.
+###### Encoding By Name
 
 The behavior of the `encoding` field is designed to support web forms, and is therefore only defined for media types structured as name-value pairs that allow repeat values, most notably `application/x-www-form-urlencoded` and `multipart/form-data`.
+
+To use the `encoding` field, each key under the field MUST exist as a property; `encoding` entries with no corresponding property SHALL be ignored.
+Array properties MUST be handled by applying the given Encoding Object to produce one encoded value per array item, each with the same `name`, as is recommended by [[?RFC7578]] [Section 4.3](https://www.rfc-editor.org/rfc/rfc7578.html#section-4.3) for supplying multiple values per form field.
```
ralfhandl: Why an informative reference to RFC7578 here?

We reference it as normative elsewhere, so the question mark here should not make any difference in the respec output.

ralfhandl (commented):
+1, minor nits

src/oas.md
handrews: Huh.  No clue.  Honestly we should go through and assess the normative vs informative-ness of all of our reference holistically before releasing.  Some are obvious to me and some are not and I doubt this is the only place I've been inconsistent.

handrews:
@ralfhandl updated - both concerns should be addressed now!

src/oas.md
```diff
@@ -1309,30 +1316,48 @@ properties:
 
 ##### Encoding Usage and Restrictions
 
-The `encoding` field defines how to map each [Encoding Object](#encoding-object) to a specific value in the data.
+The three encoding fields define how to map each [Encoding Object](#encoding object) to a specific value in the data.
```
lornajane:
```suggestion
These encoding fields define how to map each [Encoding Object](#encoding object) to a specific value in the data.
```

src/oas.md
ralfhandl: Unresolved because I don't see a change to the text in question.

ralfhandl: The force-push effect again 🙄

ralfhandl (approved):

handrews:
With apologies to @ralfhandl, I had to rebase this to fix conflicts after #4747 was merged.  The conflict was just in the test file, and really I just removed the conflict markers as it put everything in the right place anyway.  So nothing has actually changed.

ralfhandl:
> nothing has actually changed

I trust you.

Just out of curiosity: how could I verify this?

handrews:
@ralfhandl well, it's the same commits in terms of what is in each commit, so me changing them would have been a lot more work.  Honestly it hadn't occurred to me to need to check.

handrews:
@ralfhandl mostly, I put a lot of effort into making sure all of these PRs didn't conflict.  The nested multipart was the only one that did, and I thought it would get done last so I'd easily be able to rebase it before the final review and just manage that.  I didn't anticipate having to do it the other way around, and also when I first posted these they didn't need those tests, so things changed in-flight and messed up my no-conflicts, no-rebase plan :-/

ralfhandl:
No worries, I’ll re-approve once someone else has approved.

tests/schema/fail/media-type-enc-item-exclusion.yaml
```diff
@@ -0,0 +1,10 @@
+openapi: 3.2.0
+info:
+  title: API
+  version: 1.0.0
+components:
+  requestBodies:
+    content:
+      multipart/mixed:
+        encoding: {}
+        itemEncoding: {}
```
ralfhandl:
```suggestion
    encoding-with-itemEncoding-not-allowed:
      content:
        multipart/mixed:
          encoding: {}
          itemEncoding: {}
```
Name of reusable request body is missing, making this fail for the wrong reason.

tests/schema/fail/media-type-enc-prefix-exclusion.yaml
```diff
@@ -0,0 +1,10 @@
+openapi: 3.2.0
+info:
+  title: API
+  version: 1.0.0
+components:
+  requestBodies:
+    content:
+      multipart/mixed:
+        encoding: {}
+        prefixEncoding: {}
```
ralfhandl:
```suggestion
    encoding-with-prefixEncoding-not-allowed:
      content:
        multipart/mixed:
          encoding: {}
          prefixEncoding: []
```
Name of reusable request body is missing, making this fail for the wrong reason.

ralfhandl (approved):

miqui (approved):

PR #4746, opened 2025-06-21
v3.2: Ordered `multipart` examples

handrews:
These are the examples that go with PRs #4745 and #4744.

This includes the recently (this past week) updated `multipart/related` example that was provoking a lot of discussion.  Otherwise, the examples are the same as in the old PR #4589.

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -2017,6 +2017,7 @@ The nested parts are XML, plain text, and a PNG image.
 ```yaml
 multipart/mixed:
   schema:
+    type: array
     prefixItems:
     - type: array
     - type: array


src/oas.md
```diff
@@ -1869,6 +1869,121 @@ requestBody:
...
+      {}
+  prefixEncoding:
+  - # Encoding Object defaults are correct for JSON
+    {}
+  - contentType: image/*
+```
+
+###### Example: Ordered Multipart With Required Header
+
+As described in [[?RFC2557]], a set of resources making up a web pages can be sent in a `multipart/related` payload, preserving links among themselves by defining a `Content-Location` header for each page.
```
miqui: @handrews "pages" should be "page"

miqui (commented):
@handrews looks good other than the minor typo

src/oas.md
ralfhandl: Or "making up a set of web pages" because "preserving links among themselves" suggests multiple pages.

src/oas.md
```diff
@@ -1869,6 +1869,121 @@ requestBody:
...
+            - type: string
+            items:
+              anyOf:
+              - type: string
+              - $comment: To allow binary, this must always pass
+          prefixEncoding:
+          - contentType: text/html
+            headers:
+              Content-ID:
+                $ref: '#/components/headers/RFC2557ContentId'
```
ralfhandl: Interesting way of forbidding this header

src/oas.md
```diff
@@ -1869,6 +1869,121 @@ requestBody:
...
+              Content-Location:
+                $ref: '#/components/headers/RFC2557ContentLocation'
+          itemEncoding:
+            contentType: text/html, text/css, text/javascript, image/*
```
ralfhandl: I would not have expected whitespace here, and my hypothetical implementation would brutally split at the commas and preserve the leading spaces.

handrews: @ralfhandl that is surprising as this is effecrtively an `Accept` header without `q`-values, and whitespace is allowed there: https://www.rfc-editor.org/rfc/rfc9110.html#section-12.5.1

The phrase "comma-separated list" is often used in specifications with whitespace allowed.  In fact, in places where whitespace is _not_ allowed such as around the `=` in an [RFC9112 field parameter](https://www.rfc-editor.org/rfc/rfc9112.html#section-2.2), specifications go out of their way to emphasize that.

src/oas.md
handrews: Um... thanks?  :-D

Did you want me to change something here?  I could inline the reference as I thnk I made it a `$ref` on autopilot.  Using a boolean JSON Schema of `false` is a common idiom for forbidding things, and was one of the reasons we added boolean schemas back in draft whatever that was (draft-06, I think).  It's much more clear than `{"not": {}}`, or (for a property) `{"not": {"required": ["dontusethis"]}`. And there's no way to do the not-required thing for a header like this.

handrews: Oh wait it's a ref b/c I use it twice.  I think I called it RFC2557ContentId because RFC2557 is what advises against it, but I could make that RFC2557NoContentId which might be more clear.

src/oas.md
handrews: @miqui @ralfhandl In this case it should be "page", I think, as that is the _primary_ intent of RFC2557:  A single `text/html` document and its subsideary style sheets, images, etc.  But RFC2557 allows use of the general technique for other linked sets of documetns, so it could also be used for a set of pages.  I'll clarify this in the text somehow.

handrews:
@ralfhandl @miqui I have updated the above and will mark them resolved to reduce noise when we look at this in the call, but here is a summary:

* I removed the whitespace in `contentType` and filed issue #4788 to track if we should say something about it.  The no-whitespace version is _definitely_ valid so we'll go with that here
* The "page" vs "pages" was a bit of a mess as I was allowing additional `text/html` resources, so I changed that and reworded things a bit to make it clear that _this_ example is for a single page, even if you could also do a linked set of pages
* I changed the ref name to RFC2557NoContentId which makes it easier for me to parse- it took me a minute to realize what I'd been doing there with the other name.  The referenced schema is the same- I don't actually know any other way to forbid a header except with `not: {}` which is 😵‍💫

src/oas.md
```diff
@@ -1869,6 +1869,121 @@ requestBody:
...
+  itemSchema:
+    $comment: A single range of bytes from a video
+  itemEncoding:
+    contentType: video/mp4
+    headers:
+      Content-Range:
+        required: true
+        content:
+          text/plain:
+            schema:
```
lornajane, line 2006: @handrews a note to review the wording here, and consider if we need both this example and the one above

lornajane:
We discussed the media range in the multipart/related example - the type param is actually required so we agreed to keep it in the example this time.

src/oas.md
ralfhandl: Let‘s revisit this once
* #4788 

is decided.

ralfhandl (approved):

ralfhandl (dismissed):

handrews:
With (more) apologies to @ralfhandl I rebased to clear conflicts with #4747, in this case it was just that both added new examples in the same place at the end of the section, so now this PR adds them after the old examples but before the nested multipart example.  The commits are otherwise identical to before.

ralfhandl (dismissed):

lornajane (approved):

PR #4747, opened 2025-06-21
v3.2: Support nested multipart with nested Encoding Objects

handrews:
In PR #4589, both of the most in-depth reviewers (@jeremyfiel and @ralfhandl) requested support for nested `multipart/mixed` with real-world use cases.

This adds the Media Type Object's three encoding fields to the Encoding Object to support nested multipart documents.  It only requires one level of nesting, but allows implementations to support more, which seemed like a reasonable compromise as a fully-recursive system is quite complex.

This PR relies on PR #4743 to address the complexities of correlating Encoding Objects and Schema Objects, and determining the instance type in order to properly default `contentType`.

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -1681,7 +1681,6 @@ The absence of all three of those fields is the equivalent of using `content`, b
 
 Nested formats requiring encoding, most notably nested `multipart/mixed`, can be supported with this Object's `encoding`, `prefixEncoding`, and / or `itemEncoding` fields.
 Implementations MUST support one level of nesting, and MAY support additional levels.
-If supporting additional levels, any limits on nesting levels MUST be documented.
 
 ##### Encoding the `x-www-form-urlencoded` Media Type
 
--- a/src/oas.md
+++ b/src/oas.md
@@ -1641,6 +1641,9 @@ These fields MAY be used either with or without the RFC6570-style serialization 
 | ---- | :----: | ---- |
 | <a name="encoding-content-type"></a>contentType | `string` | The `Content-Type` for encoding a specific property. The value is a comma-separated list, each element of which is either a specific media type (e.g. `image/png`) or a wildcard media type (e.g. `image/*`). Default value depends on the property type as shown in the table below. |
 | <a name="encoding-headers"></a>headers | Map[`string`, [Header Object](#header-object) \| [Reference Object](#reference-object)] | A map allowing additional information to be provided as headers. `Content-Type` is described separately and SHALL be ignored in this section. This field SHALL be ignored if the media type is not a `multipart`. |
+| <a name="encoding-encoding"></a>encoding | Map[`string`, [Encoding Object](#encoding-object)] | Applies nested Encoding Objects in the same manner as the [Media Type Object](#media-type-object)'s `encoding` field. |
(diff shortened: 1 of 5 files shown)


karenetheridge:
Oh come on, we can do full recursion ;)

src/oas.md
```diff
@@ -1869,6 +1878,30 @@ requestBody:
...
+
+This defines a two-part `multipart/mixed` where the first part is JSON and the second part is a nested `multipart/mixed` document.
+The nested parts are JSON, plain text, and a PNG image.
+
+```yaml
+multipart/mixed:
+  schema:
+    type: array
+    prefixItems:
+    - type: array
```
ralfhandl, line 1890: Shouldn't the first prefix item be`type: object` and the second have `type: array` and `prefixItems`?

If not, comments in or more prose around the example would be helpful.

ralfhandl (changes requested):

src/oas.md
handrews: The first one is correct, as I wanted to highlight that in ordered multipart, `type: array` defaults to `application/json` as opposed to the convoluted (but RFC-based) logic for arrays for `form-data`).  I have changed "is JSON" to "is a JSON array" in the text to clarify.

The second was missing `type`, and also the first nested entry is XML and not JSON, which I have corrected in the text.

ralfhandl (dismissed):

handrews:
@ralfhandl with apologies, you'll have to hit the review button again- I stripped out the line requiring documentation of nesting level limitations as documentation requirements proved controversial on another PR.  Which I still find baffling but at this point I just want the changes in and don't want to argue about the documentation stuff.

miqui (approved):

ralfhandl (approved):

src/schemas/validation/schema.yaml
```diff
@@ -561,9 +561,24 @@ $defs:
...
+      itemEncoding:
+        $ref: '#/$defs/encoding'
     allOf:
       - $ref: '#/$defs/specification-extensions'
       - $ref: '#/$defs/styles-for-form'
+      - dependentSchemas:
+          encoding:
+            properties:
+              prefixEncoding: false
+              itemEncoding: false
```
ralfhandl, line 581: @handrews These lines seem to have no effect: when I comment them out, the two `encoding-enc-*-exclusion.yaml` fail cases still fail, for whatever reason.

tests/schema/fail/encoding-enc-item-exclusion.yaml
```diff
@@ -0,0 +1,12 @@
+openapi: 3.2.0
+info:
+  title: API
+  version: 1.0.0
+components:
+  requestBodies:
+    content:
```
ralfhandl, line 7: The name for this request body is missing, making this case fail.

Fixed in #4780.

tests/schema/fail/encoding-enc-prefix-exclusion.yaml
```diff
@@ -0,0 +1,12 @@
+openapi: 3.2.0
+info:
+  title: API
+  version: 1.0.0
+components:
+  requestBodies:
+    content:
```
ralfhandl, line 7: The name for this request body is missing, making this case fail.

ralfhandl: Fixed in #4780.

Issue #1237, opened 2017-06-29
Multiple 'Set-Cookie' headers in one response


JohnnyNiu:
Is it possible to have multiple 'Set-Cookie' headers in one response?   As is known there are two ways to set cookies header in the response:
        - Having separated headers
	- Folding into 1 header and using comma separated

The later way however is deprecated in (RFC6265)[http://www.rfc-editor.org/rfc/rfc6265.txt] and not supported by some latest browsers. 
```  
   Origin servers SHOULD NOT fold multiple Set-Cookie header fields into
   a single header field.  The usual mechanism for folding HTTP headers
   fields (i.e., as defined in [RFC2616]) might change the semantics of
   the Set-Cookie header field because the %x2C (",") character is used
   by Set-Cookie in a way that conflicts with such folding.
```

So that below can be valid:
```
responses:
   200:
   description: "Response with content"
   headers:
       Set-Cookie:
           type: String
           description: "eg. key1=value1"
       Set-Cookie:
           type: String
           description: "eg. key2=value2"
```

ePaul:
You can of course have multiple headers with the same name in your HTTP response, but it looks like OpenAPI 2.0 has no way of documenting that. The way shown by you doesn't work (a JSON or YAML object can have each key just once, and if you repeat one, most parsers will retain just one of them).

I think it looks like the same with the current RC of OpenAPI 3.0 – while the handling of plural parameters changed (via `style` instead of `collectionFormat`), I didn't find anything for "multiple headers of the same name". (I guess that could be handled by adding a new style value here.)

RobDolinMS:
Per @darrelmiller - We have a way to capture this in the request but not in the response.  

@JohnnyNiu Is that sufficient for your use case?

honzajavorek:
The `Set-Cookie` header is typically a response header.

jwalton:
@RobDolinMS Also, you can't have multiple `cookie` headers in the request.  As per [RFC 6265 S5.4](https://tools.ietf.org/html/rfc6265#section-5.4):

> When the user agent generates an HTTP request, the user agent MUST NOT attach more than one Cookie header field.

mikejav:
## I got rid of this problem with this one simple trick:
just surround the next same header by quotation marks and add null char at the beginning.
```yaml
headers:
  Set-Cookie:
    description: Session cookie
    schema:
      type: string
      example: SESSIONID=abcde12345; Path=/
  "\0Set-Cookie":
[...]

PR #4748, opened 2025-06-24
v3.2: Provide guidance for the `Set-Cookie` response header

handrews:
Partially addresses issue #1237 

The Set-Cookie response header breaks the normal rules for headers with multiple values and requires special handling.

There are two options here:

1. Rework how we handle headers to accommodate `Set-Cookie` in a consistent manner
2. Treat `Set-Cookie` as a special case and define how it relates to normal behavior

Since [RFC9110 §5.3](https://www.rfc-editor.org/rfc/rfc9110.html#section-5.3) advises clients to treat `Set-Cookie` as a special case, I went with special-casing it.   This assumes the approach to header serialization that is described in PR #4648 and discussed in one of its comment threads (specifically, that it does not include the header name which is required for it to be consistent with how `style: "simple"` and `explode` are treated elsewhere and are defined in RFC6570).

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -2559,10 +2559,9 @@ components:
       content:
         text/plain:
           schema:
+            # Due to lack of support for multiline regular expressions
+            # in the `pattern` keyword, not much validation can be done.
             type: string
-            allOf:
-            - pattern: "^lang=[^;];.*Expires="
-            - pattern: "^foo=[^;];.*Expires="
       examples:
         WithExpires:
           # This demonstrates that the text is required to be provided
@@ -2582,7 +2581,7 @@ components:
         - foo
         additionalProperties:
           type: string
-          pattern: "^[^[:space:]]$"
+          pattern: "^[^[:space:]]*$"
       style: simple
       explode: true
       examples:


src/oas.md
```diff
@@ -2451,6 +2451,72 @@ Using `content` with a `text/plain` media type is RECOMMENDED for headers where
...
+##### Representing the `Set-Cookie` Header
+
+As noted in [[!RFC9110]] [Section 5.3](https://www.rfc-editor.org/rfc/rfc9110.html#section-5.3) the `Set-Cookie` response header violates the requirements for representing multiple values as a comma-separated list, as `style: "simple"` produces.
+
+```http
+Set-Cookie: lang=en-US; Expires=Wed, 09 Jun 2021 10:18:14 GMT
+Set-Cookie: foo=bar; Expires=Wed, 09 Jun 2021 10:18:14 GMT
+```
+
+If these values were to be place on a single line using `style: "simple"`, the result would be `lang=en-US; Expires=Wed, 09 Jun 2021 10:18:14 GMT,foo=bar; Expires=Wed, 09 Jun 2021 10:18:14 GMT`, which when split would then produce four values: `lang=en-US; Expires=Wed`, `09 Jun 2021 10:18:14 GMT`, `foo=bar; Expires=Wed`, and `09 Jun 2021 10:18:14 GMT`.
```
philsturgeon:
```suggestion
If these values were to be placed on a single line using `style: "simple"`, the result would be `lang=en-US; Expires=Wed, 09 Jun 2021 10:18:14 GMT,foo=bar; Expires=Wed, 09 Jun 2021 10:18:14 GMT`, which when split would then produce four values: `lang=en-US; Expires=Wed`, `09 Jun 2021 10:18:14 GMT`, `foo=bar; Expires=Wed`, and `09 Jun 2021 10:18:14 GMT`.
```

philsturgeon (approved):

handrews:
With apologies to @philsturgeon for clearing his approval (its' set to automatically do that if anything changes) I realized we don't need`allowReserved: true` in that example (an earlier version I tried out did, and it was left over from that).  And while I was at it I added a paragraph emphasizing that this only really matters if you need an intermedia multi-value single-string representation.  If you just have an array of values and `style: simple` to map them to HTTP, they end results are the same as for other headers that support multiple values.  `Set-Cookie` is only strange in that you cannot put multiple values on a single line.

handrews:
OK apparently I was more out-of-it yesterday with this cold than I thought.  _Now_ I have added the commit that does things I mentioned in the last comment! 🤦

src/oas.md
```diff
@@ -2451,6 +2451,75 @@ Using `content` with a `text/plain` media type is RECOMMENDED for headers where
...
+
+```http
+Set-Cookie: lang=en-US; Expires=Wed, 09 Jun 2021 10:18:14 GMT
+Set-Cookie: foo=bar; Expires=Wed, 09 Jun 2021 10:18:14 GMT
+```
+
+If these values were to be placed on a single line using `style: "simple"`, the result would be `lang=en-US; Expires=Wed, 09 Jun 2021 10:18:14 GMT,foo=bar; Expires=Wed, 09 Jun 2021 10:18:14 GMT`, which when split would then produce four values: `lang=en-US; Expires=Wed`, `09 Jun 2021 10:18:14 GMT`, `foo=bar; Expires=Wed`, and `09 Jun 2021 10:18:14 GMT`.
+While the two dates (`09...`) are not valid `Set-Cookie` values on their own, [[?RFC6265]] does not provide any guarantee that all such embedded uses of commas will produce detectable errors when split in this way.
+
+RFC9110 therefore advises recipients to 'handle "Set-Cookie" as a special case while processing fields,' so the OAS similarly special-cases its handling of `Set-Cookie`.
```
lornajane: Special case is not a verb!

```suggestion
RFC9110 therefore advises recipients to 'handle "Set-Cookie" as a special case while processing fields,' so the OAS similarly applies a special case to its handling of `Set-Cookie`.
```

src/oas.md
```diff
@@ -2451,6 +2451,75 @@ Using `content` with a `text/plain` media type is RECOMMENDED for headers where
...
+
+When an OAS implementation is mapping directly between the multi-`Set-Cookie:` header line format and an array representation, without any intermediate single string holding the multiple values, no special handling is needed as the behavior is the same as for headers that can be either on a single line with comma-separated values or on multiple lines.
```
lornajane: This wasn't instantly clear to me. Multi-header? Multi-line? I _think_ we want to explain that we're mapping between the format of multiple set-cookie headers and an array representation but I may have confused myself by now.

src/oas.md
handrews: I'll verb whatever I want to verb!  ;-P

(ok yeah this is better)

src/oas.md
handrews: @lornajane yeah this is a bit weird and I could use some help streamlining it.  It might take a bit before I can come back and explain because I want to update PR #4673 to be more clear, which I think will help explain this.  Or at least help me sort out my thoughts.

handrews:
@lornajane I have rewritten this to (hopefully) focus more on examples and less on explanations.  This should also go well with PR #4800 which defines `style` example behavior in terms of whether the name is included in the example or not, and explains why it is for forms and is not for headers.

Meh... now there are conflicts, though, I will rebase and force-push, but it will be the same commits.  It's probably best to just look at it fresh anyway as the rewrite commit on its own is noisier than just looking at it as one thing.

lornajane:
I commented this somewhere else too: I don't think we have the serializedValue PR merged yet and I see this one uses the feature (and it's a good use of it!) so let's get that one in before we approve this one.

src/oas.md
```diff
@@ -2524,6 +2524,92 @@ components:
             $ref: '#/components/mediaTypes/CollectionLinks'
 ```
 
+##### Representing the `Set-Cookie` Header
+
+The `Set-Cookie` header is noted in [[!RFC9110]] [Section 5.3](https://www.rfc-editor.org/rfc/rfc9110.html#section-5.3) as an exception to the normal rules of headers with multiple values.
+
+For most headers using the general syntax defined in RFC9110, the multiple-line and comma-separaed single-line forms are interchangeable, meaning that this:
```
```suggestion
For most headers using the general syntax defined in RFC9110, the multiple-line and comma-separated single-line forms are interchangeable, meaning that this:
```

src/oas.md
```diff
@@ -2524,6 +2524,92 @@ components:
...
+    SetCookieWithExpires:
+      # Spaces within the Expires values prevent the use of `schema` and
+      # `style` as they would be percent-encoded, even with `allowReserved`.
+      content:
+        text/plain:
+          schema:
+            type: string
+            allOf:
+            - pattern: "^lang=[^;];.*Expires="
+            - pattern: "^foo=[^;];.*Expires="
```
ralfhandl: How can a string start with both l and f?

I couldn't find text in JSON Schema or OAS that suggests or requires multiline ECMAScript pattern matching for `type: string`.

src/oas.md
```diff
@@ -2524,6 +2524,92 @@ components:
...
+            foo=bar; Expires=Wed, 09 Jun 2021 10:18:14 GMT
+    SetCookieWithNoSpaces:
+      schema:
+        type: object
+        required:
+        - lang
+        - foo
+        additionalProperties:
+          type: string
+          pattern: "^[^[:space:]]$"
```
ralfhandl: Exactly one character that is not whitespace?

Or is a `*` missing at the second-to-last position in the pattern?

ralfhandl (dismissed):
+1 with one nit and two questions

src/oas.md
handrews: blarg.  This was supposed to be a bit more complex with `anyOf`, I'll fix it.

handrews: Hmm... good point about the multiline.  I had adapted this from a `Link` header example, but multiple values are on one line there, while the whole point of this PR is that that is not true for `Set-Cookie`.

I guess there's really not much that can be done here, then.

src/oas.md
handrews: `*` was missing, fixed now.

Issue #2951, opened 2022-06-22
YAML JSON Schema is not enough for roundtrip-safe


ioggstream:
## I suggest

- OAS 3.1 can reference   https://www.ietf.org/archive/id/draft-ietf-httpapi-yaml-mediatypes-01.html#name-yaml-and-json once it's released

## Because

- The "YAML JSON Schema" has some issues with inf/nan that do not map to JSON datatypes. 
- DRY

cc: @darrelmiller https://github.com/ietf-wg-httpapi/mediatypes/issues/8#issuecomment-1073116351

LasneF:
@handrews; @darrelmiller 
  should this be updated to leverage https://www.rfc-editor.org/rfc/rfc9512.html as it is now published

handrews:
@LasneF I think it should definitely be updated in 3.2.  I hesitate to update it in the patch releases, in part because a.) I'm uncertain of the exact impact, and b.) I doubt anyone actually depends on this guidance for round-trip safety.  I don't even know how to enforce it- if I want round-trip safety, I'll test it.

I am open to other opinions on this, although if we want it in 3.0.4 we need to decide that ASAP.

LasneF:
i would rather use the RFC as even for the 3.04  
* our patch is pushing should be leveraging the latest precision of RFC if there is no breaking change but just precision
* it s better to mention a stable RFC than a draft 
* in our case the content of the paragraph is similar 

we need to introduce a sentence about it as currenlty there is no such precision , and it should be an advise only 

my opinion would be here to have a consistent update for all the version 

that said if we push it only to 3.2 it  would be ok as well

handrews:
@LasneF 

> our patch is pushing should be leveraging the latest precision of RFC if there is no breaking change but just precision

We already decided not to update RFCs in patch releases (search issues / PRs for 9110).

> it s better to mention a stable RFC than a draft

Yes, and we have made those updates, but the YAML 1.2 spec is not a draft.  It just was not a formal media type registration for IANA purposes.

> my opinion would be here to have a consistent update for all the version

Yeah I'd prefer that but if we do that we need to update all RFCs and revisit the decision made about RFC9110 replacing RFC7231 (which we decided to deal with in 3.2).  Feel free to lobby the TSC for it or pick up this work, but I don't have the time/energy for it.  You could bring it up in the Style Guide / release checklist issue as it could arguably be part of the checklist if we want to update.

ioggstream:
[...]

PR #4758, opened 2025-07-04
v3.2: Clarify JSON-compatible YAML and UTF-8 use

handrews:
I'm stuck at home bored and getting over a cold, so have a PR as I distract myself from missing holiday plans!

This fixes:
* #2951 (the YAML "JSON schema ruleset" turns out to [allow values that can't be expressed in JSON](https://github.com/yaml/yaml-spec/issues/29) which is a problem for us)
* #4643 (more about how the YAML spec is hard to comprehend than anything else, but it demonstrates problems with our existing text)

Everything in this area is a RECOMMENDED (== SHOULD) so we have some leeway.  I have tightened what is recommended but acknowledged previous guidance and simply said that depending on anything allowed by the old guidance but not by the new is NOT RECOMMENDED.  I suspect most tools never even noticed, TBH.

Referencing the new RFC also ties YAML in to JSON's requirement that only UTF-8 be treated as interoperable, which is relevant to the ongoing debates on serialized examples (see PR #4671).  Being more clear about our own encoding requirements simplifies that problem.


src/oas.md
```diff
@@ -169,12 +169,22 @@ The [schema](#schema) exposes two types of fields: _fixed fields_, which have a
...
-* Keys used in YAML maps MUST be limited to a scalar string, as defined by the [YAML Failsafe schema ruleset](https://yaml.org/spec/1.2/spec.html#id2802346).
+#### JSON and YAML Compatibility
 
-**Note:** While APIs may be described by OpenAPI Descriptions in either YAML or JSON format, the API request and response bodies and other content are not required to be JSON or YAML.
+In order to preserve the ability to round-trip between YAML and JSON formats, YAML version [1.2](https://yaml.org/spec/1.2/spec.html) is RECOMMENDED along with the additional constraints listed in [[!RFC9512]] [Section 3.4](https://www.rfc-editor.org/rfc/rfc9512.html#name-yaml-and-json).
+
+Previous versions of this specification expressed this requirement in terms of [YAML's JSON schema ruleset](https://yaml.org/spec/1.2/spec.html#id2803231), which defines a subset of the YAML syntax and is unrelated to [[JSON-Schema-2020-12|JSON Schema]].
+Despite its name, this ruleset supports features that are not part of JSON, such as distinguishing between integers and floats (see [Data Types](#data-types)) or supporting `!!float .nan` as a value.
```
mikekistler: I don't know that this is necessary. This content seems fine as part of the description in the PR, in case some one wants to understand the change later, but I don't think we need it in the spec. Just my 2 cents.

mikekistler (dismissed):
Looks good. 👍

src/oas.md
handrews: I'd be OK with taking it out.

handrews: @mikekistler the only thing that's really happening here is that the first sentence acknowledges why there may be a compatibility weirdness where something previously within the "RECOMMENDED" is now "NOT RECOMMENDED", but maybe we don't need to make as much of a point about that?  The second sentence here is really just me over-explaining.

hudlow: FWIW, I think better to put it in the spec than force someone to track it down in the PR. (I think we need a better compromise in general, like footnotes.)

handrews: @hudlow let me see if I can just word it better, so it feels less like a tangent.

handrews:
After further discussions with @hudlow the character set stuff may not be needed regarding the Example Object, so it's worth considering whether that part needs to be here at all, or whether what's in the JSON and YAML RFCs is sufficient (it probably is).

handrews:
@hudlow @mikekistler I have removed the bit about character encodings (it is truly not needed, after getting a night of sleep- I've been under the weather, please forgive the confusion) and have slimmed down the explanation as much as I can while still conveying that there's a bit of a danger zone that was _technically_ previously recommended, but was never really intended to work.

ralfhandl (approved):

mkistler (approved):
Looks good! 👍

handrews:
@mkistler @mikekistler GitHub won't let me merge unless you approve as @mikekistler

lornajane (commented):
Indeed, much clearer

lornajane (approved):

src/oas.md
lornajane: This was re-worded and all comments must be reviewed before merge so I'll do that.

Issue #4759, opened 2025-07-05
Registry: Support `date-time-local` and `time-local`


ya7010:
**Which registry do you want to contribute to**
- [ ] [Alternative Schema Type Registry](https://spec.openapis.org/registry/alternative-schema)
- [ ] [Draft Features Registry](https://spec.openapis.org/registry/draft-feature)
- [ ] [Specification Extension Registry](https://spec.openapis.org/registry/extension)
- [x] [Format Registry](https://spec.openapis.org/registry/format)
- [ ] [Extension Namespace Registry](https://spec.openapis.org/registry/namespace)
- [ ] [Tag Kind Registry](https://spec.openapis.org/registry/tag-kind)

**Describe your contribution**
To clarify the use of JSON Schema in TOML, add a date/time format without a time zone.

**Describe alternatives you've considered**

**Additional context**
https://github.com/OAI/OpenAPI-Specification/discussions/4533

handrews:
@ya7010 If you want to move this forward, you can open a PR against the format registry on the `gh-pages` branch.  I'm going to close this issue as a duplicate of #4533 (although thanks for filling out our new registry checkboxes correctly!), but really all that is needed is for someone to make a PR and answer questions while it is reviewed.  In the column for the format where it wants a specification, just link to the TOML specification as you did in the other issue.

handrews:
Oh wait this is an issue and that's a discussion... meh, we need to sort out how to handle issues vs discussions, for now I'll re-open as opening an issue for this is not wrong.

ya7010:
Merged 🎉

PR #4760, opened 2025-07-06
feat: add date-time-local and time-local.

ya7010:
Fix: https://github.com/OAI/OpenAPI-Specification/issues/4759

### DIFF
--- a/registries/_format/date-time-local.md
+++ b/registries/_format/date-time-local.md
@@ -5,7 +5,7 @@ description: RFC3339 date-time without the timezone component
 base_type: string
 layout: default
 source_label: RFC 3339
-source: https://datatracker.ietf.org/doc/html/rfc3339#section-5.6
+source: https://www.rfc-editor.org/rfc/rfc3339#section-5.6
 ---
 
 {% capture summary %}
--- a/registries/_format/date-time-local.md
+++ b/registries/_format/date-time-local.md
@@ -4,8 +4,8 @@ issue: 4759
 description: RFC3339 date-time without the timezone component
 base_type: string
 layout: default
-source_label: TOML
-source: https://toml.io/en/v1.0.0#local-date-time
+source_label: RFC 3339
+source: https://datatracker.ietf.org/doc/html/rfc3339#section-5.6
 ---
 
 {% capture summary %}
--- a/registries/_format/date-time-local.md
+++ b/registries/_format/date-time-local.md
@@ -1,7 +1,7 @@
 ---
 owner: ya7010
 issue: 4759
-description: date and time without timezone
+description: RFC3339 date-time without the timezone component
 base_type: string
 layout: default
 source_label: TOML
@@ -9,7 +9,7 @@ source: https://toml.io/en/v1.0.0#local-date-time
 ---
 
 {% capture summary %}
-The `{{page.slug}}` format represents a date and time without timezone for [TOML Local Date-Time](https://toml.io/en/v1.0.0#local-date-time) compatibility.
+The `{{page.slug}}` format represents a RFC3339 date-time without timezone.
 {% endcapture %}
 
 {% include format-entry.md summary=summary %}
--- a/registries/_format/date-time-local.md
+++ b/registries/_format/date-time-local.md
@@ -1,0 +1,15 @@
+---
+owner: ya7010
+issue: 4759
+description: date and time without timezone
+base_type: string
+layout: default
+source_label: TOML
+source: https://toml.io/en/v1.0.0#local-date-time
+---
+
+{% capture summary %}
+The `{{page.slug}}` format represents a date and time without timezone for [TOML Local Date-Time](https://toml.io/en/v1.0.0#local-date-time) compatibility.
+{% endcapture %}
+
+{% include format-entry.md summary=summary %}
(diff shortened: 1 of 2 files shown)


handrews:
@baywet @mikekistler thoughts?

mikekistler:
No objection from me.

registries/_format/time-local.md
```diff
@@ -0,0 +1,15 @@
...
+layout: default
+source_label: TOML
+source: https://toml.io/en/v1.0.0#local-time
+---
+
+{% capture summary %}
+The `{{page.slug}}` format represents a time without timezone for [TOML Local Time](https://toml.io/en/v1.0.0#local-time) compatibility.
+{% endcapture %}
+
+{% include format-entry.md summary=summary %}
```
baywet, line 15: Why are we not also including date local?

registries/_format/time-local.md
```diff
@@ -0,0 +1,15 @@
+---
+owner: ya7010
+issue: 4759
+description: time without timezone
```
baywet: Maybe mention it's based on RFC 3339?

registries/_format/time-local.md
```diff
@@ -0,0 +1,15 @@
+---
```
baywet, line 1: For the name, I've seen programming languages call this a time only. I think it'd be interesting to see whether multiple languages use the same name (time only or otherwise) and check whether it's worth using a different name for clarity.

registries/_format/time-local.md
ya7010: Time zone is required by [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339#section-5.6). 
This PR proposes a format without time zone, which is not defined in the RFC but defined in TOML.

registries/_format/time-local.md
ya7010: Since Time Zone is tied to time, `date-local` was not suggested.

registries/_format/time-local.md
ya7010: Your point is correct.

On the other hand, the already existing `time` and `date-time` format follow RFC3339, and thus enforce the time zone.

So, I think the proposed format with `*-local` is the right direction.

registries/_format/time-local.md
baywet: sorry, what I meant was maybe we should update the text to say something like "RFC3339 time without the timezone component" to clarify the base for it is the RFC

ya7010: I misunderstood.

That makes sense.

ya7010: fixed: https://github.com/OAI/OpenAPI-Specification/pull/4760/commits/33017465f9064774bde79ec92b6be61a54d07ddc

registries/_format/time-local.md
handrews: I agree that given that `time` is already a format, `time-local` seems the most clear.  Unless you had a specific alternative in mind @baywet ?

baywet: time-only was the alternative I was thinking about here. But I'm probably coming with [too much of a dotnet angle](https://learn.microsoft.com/en-us/dotnet/standard/datetime/how-to-use-dateonly-timeonly)

java calls it [local time](https://stackoverflow.com/questions/8384252/best-way-to-store-time-only-time-and-not-date-in-database)
go uses the same struct for everything

I guess I'm good with time-local, resolving.

baywet: (well, I can't resolve, please go ahead and resolve this comment)

registries/_format/time-local.md
```diff
@@ -0,0 +1,15 @@
...
+owner: ya7010
+issue: 4759
+description: RFC3339 time without the timezone component
+base_type: string
+layout: default
+source_label: TOML
+source: https://toml.io/en/v1.0.0#local-time
```
ralfhandl: This is defined in [RFC3339, section 5.6](https://datatracker.ietf.org/doc/html/rfc3339#section-5.6) as `partial-time` and not specific to TOML. Please adjust source reference and remove `source_label`.

registries/_format/date-time-local.md
```diff
@@ -0,0 +1,15 @@
+---
+owner: ya7010
+issue: 4759
+description: RFC3339 date-time without the timezone component
+base_type: string
+layout: default
+source_label: TOML
+source: https://toml.io/en/v1.0.0#local-date-time
```
ralfhandl: This is not TOML-specific, so I wouldn't reference it.

If you want to be exact about the allowed pattern, you can define it in RFC3339 terms as `full-date "T" partial-time`.

registries/_format/time-local.md
ya7010: fix: 1f4191427576b299dbd61068c7733bbca02f67af

registries/_format/date-time-local.md
ya7010: fix: 1f4191427576b299dbd61068c7733bbca02f67af

Would it be better to propose the format name `partial-date-time` / `partial-time`?

Personally, I feel that `partial-*` looks like a temp name for defining specifications.

registries/_format/time-local.md
ya7010: fix: source link 41bb4955fad5d5709b5effd3e71452a2587977cc

ralfhandl (approved):

mikekistler (approved):
Looks good. 👍

PR #4762, opened 2025-07-09
Switch to @hyperjump/json-schema-coverage for schema test coverage

jdesrosiers:
Since our last update to the JSON Schema test coverage script, I've been working on a more complete solution and it's finally ready to share! I was able to generate istanbul coverage documents with the information collected from the @hyperjump/json-schema EvaluationPlugin we introduced in the last update. That allows us to generate familiar looking reports. I also figured out how to integrate it with Vitest, so it's super easy to use. I even included support for YAML out-of-the-box. See, [@hyperjump/json-schema-coverage](https://github.com/hyperjump-io/json-schema-coverage) for more details and examples.

**WARNING**: Because this does a more thorough job than the current solution, it will no longer report 100% coverage. More tests will need to be added. The biggest difference is that I added branch coverage support. Instead of just checking that the keyword was covered, it now also tracks whether both the true and false result of a keyword are covered.

**WARNING**: In the current version, we added a hack to ignore `#/$defs/schema` in `schema.yaml` because it isn't possible to cover that location if we're testing against `schema-base.yaml`. This new package doesn't have a solution for ignoring things. We could add a test that uses `schema.yaml` to cover that location, or look into some way to ignore coverage in the new solution.

**NOTE**: I think the `dev` branch is the right place to make this PR, but please let me know if that's wrong. The PR template says `main`, but the coverage tests don't exist on that branch.

---


jdesrosiers:
Here's an example of what we'll get for the 3.1 branch.

```
 % Coverage report from @hyperjump/json-schema-coverage/vitest/coverage-provider                     
------------------|---------|----------|---------|---------|-------------------                      
File              | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s                       
------------------|---------|----------|---------|---------|-------------------                      
All files         |   99.42 |    56.34 |   92.58 |   99.42 |                                         
 dialect.yaml     |     100 |      100 |     100 |     100 |                                         
 meta.yaml        |   98.14 |    53.12 |    82.6 |   98.14 | 18                                      
 schema-base.yaml |     100 |       50 |     100 |     100 | 13                                      
 schema.yaml      |   99.49 |    56.57 |   93.05 |   99.49 | 630,726-729,945                         
------------------|---------|----------|---------|---------|-------------------                      
```

image

tests/schema/schema.test.mjs
```diff
@@ -36,22 +18,20 @@ defineVocabulary(oasBaseVocab, {
...
   describe("Pass", () => {
     readdirSync(`${fixtures}/pass`, { withFileTypes: true })
       .filter((entry) => entry.isFile() && /\.yaml$/.test(entry.name))
       .forEach((entry) => {
-        test(entry.name, () => {
+        test(entry.name, async () => {
           const instance = parseYamlFromFile(`${fixtures}/pass/${entry.name}`);
-          const output = validateOpenApi(instance, BASIC);
-          expect(output).to.deep.equal({ valid: true });
+          await expect(instance).to.matchJsonSchema("https://spec.openapis.org/oas/3.1/schema-base/WORK-IN-PROGRESS");
```
ralfhandl: No, we expect the instance to match the schema in ./src/schemas/validation/schema-base.yaml, whose `$id` depends on the branch this test is running in, for example (for v3.2)

https://github.com/OAI/OpenAPI-Specification/blob/18c753aa74885268b06a0aafcb08cee3226df0f1/src/schemas/validation/schema-base.yaml#L1

tests/schema/schema.test.mjs
```diff
@@ -60,10 +40,9 @@ describe("v3.1", () => {
     readdirSync(`${fixtures}/fail`, { withFileTypes: true })
       .filter((entry) => entry.isFile() && /\.yaml$/.test(entry.name))
       .forEach((entry) => {
-        test(entry.name, () => {
+        test(entry.name, async () => {
           const instance = parseYamlFromFile(`${fixtures}/fail/${entry.name}`);
-          const output = validateOpenApi(instance, BASIC);
-          expect(output.valid).to.equal(false);
+          await expect(instance).to.not.matchJsonSchema("https://spec.openapis.org/oas/3.1/schema-base/WORK-IN-PROGRESS");
```
ralfhandl: Will these negative test cases count into the coverage? They should not.

vitest.config.mjs
```diff
@@ -2,7 +2,12 @@ import { defineConfig } from 'vitest/config'
 
 export default defineConfig({
   test: {
+    coverage: {
```
ralfhandl, line 8: Setting [coverage thresholds](https://vitest.dev/config/#coverage-thresholds) to 100 seems to have no effect.

We want the tests to fail if schema coverage falls below 100% to avoid adding schema features without test cases.

ralfhandl (commented):
@jdesrosiers Really nice!

Unfortunately we'd need some additions to use it, see comments.

tests/schema/schema.test.mjs
jdesrosiers: Not a problem. The matcher can take the path to the file instead of a URI.

vitest.config.mjs
jdesrosiers: Threshold enforcement through vitest isn't supported. Not sure it's possible, but I'll look into it.

But you can do it with `nyc` (the istanbul CLI) from the coverage report that's generated. For example,

```bash
nyc check-coverage -t coverage --extension .yaml --lines 100 --statements 100 --functions 100 --branches 100
```
Result for v3.1-dev:
```
ERROR: Coverage for lines (99.42%) does not meet global threshold (100%)
ERROR: Coverage for functions (92.58%) does not meet global threshold (100%)
ERROR: Coverage for branches (56.34%) does not meet global threshold (100%)
ERROR: Coverage for statements (99.42%) does not meet global threshold (100%)
```

tests/schema/schema.test.mjs
jdesrosiers: > Will these negative test cases count into the coverage?

Yes

> They should not.

Why not? The branch coverage feature ensures that your tests cover both the failing and passing cases of keywords. For example, branch coverage makes sure that for `type: object` you have a test that asserts that the instance passes when it's an object and another test that asserts that the instance fails when the document is something else, like `42`. You can only do that with a negative assertion.

If the test fails (either the positive tests or the negative tests) it will still record the coverage, but Vitest won't generate reports. Reports will only be generated if all the tests pass.

jdesrosiers:
It looks like defining a vocabulary isn't quite working, so this isn't working on the v3.2-dev branch yet.

ralfhandl:
It does work with the "old" solution, both for validation pass/fail and for coverage.

jdesrosiers:
I fixed the problem with registering the vocabulary, so this now works for all branches including v3.2.

I also rebased to fix the conflict.

As far as I know, the only thing missing at this point is adding additional tests to cover the parts the new solution catches that the old one didn't. I'll leave that part to someone else.

ralfhandl:
Unfortunately this does not run on Windows machines, it fails with

~~~
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Unhandled Error ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
RetrievalError: Unable to load resource 'c:\git\OpenAPI-Specification\src\schemas\validation\dialect.yaml'.
 ❯ get node_modules/@hyperjump/browser/lib/browser/browser.js:29:13
 ❯ getSchema node_modules/@hyperjump/json-schema/lib/schema.js:33:24
 ❯ FileCoverageMapService.addFromFile node_modules/@hyperjump/json-schema-coverage/src/coverage-map-service.js:43:26
 ❯ FileCoverageMapService.addFromFile node_modules/@hyperjump/json-schema-coverage/src/vitest/file-coverage-map-service.js:57:35
 ❯ Object.setup node_modules/@hyperjump/json-schema-coverage/src/vitest/build-coverage-maps.js:55:29
 ❯ TestProject._initializeGlobalSetup node_modules/vitest/dist/chunks/cli-api.BkDphVBG.js:7074:21
 ❯ Vitest.initializeGlobalSetup node_modules/vitest/dist/chunks/cli-api.BkDphVBG.js:9724:35
 ❯ node_modules/vitest/dist/chunks/cli-api.BkDphVBG.js:9636:5
 ❯ Vitest.runFiles node_modules/vitest/dist/chunks/cli-api.BkDphVBG.js:9666:10  
 ❯ Vitest.start node_modules/vitest/dist/chunks/cli-api.BkDphVBG.js:9554:18     

Caused by: UnsupportedUriSchemeError: The 'c:' URI scheme is not supported. Use the 'addUriSchemePlugin' function to add support for 'c:' URIs.
 ❯ retrieve node_modules/@hyperjump/browser/lib/uri-schemes/uri-schemes.js:19:11
 ❯ get node_modules/@hyperjump/browser/lib/browser/browser.js:23:30
 ❯ getSchema node_modules/@hyperjump/json-schema/lib/schema.js:33:24
 ❯ FileCoverageMapService.addFromFile node_modules/@hyperjump/json-schema-coverage/src/coverage-map-service.js:43:26
 ❯ FileCoverageMapService.addFromFile node_modules/@hyperjump/json-schema-coverage/src/vitest/file-coverage-map-service.js:57:35
 ❯ Object.setup node_modules/@hyperjump/json-schema-coverage/src/vitest/build-coverage-maps.js:55:29
 ❯ TestProject._initializeGlobalSetup node_modules/vitest/dist/chunks/cli-api.BkDphVBG.js:7074:21
 ❯ Vitest.initializeGlobalSetup node_modules/vitest/dist/chunks/cli-api.BkDphVBG.js:9724:35
 ❯ node_modules/vitest/dist/chunks/cli-api.BkDphVBG.js:9636:5
 ❯ Vitest.runFiles node_modules/vitest/dist/chunks/cli-api.BkDphVBG.js:9666:10  

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
Serialized Error: { scheme: 'c' }
~~~

The previous implementation worked fine on Windows and Ubuntu. Maybe something is missing when resolving relative paths into file URLs, it should actually try to load resource

    file:///C:/git/OpenAPI-Specification/src/schemas/validation/dialect.yaml

which it would find.

package.json
```diff
@@ -15,7 +15,7 @@
   "scripts": {
     "build": "bash ./scripts/md2html/build.sh",
     "build-src": "npm run validate-markdown && bash ./scripts/md2html/build.sh src && bash ./scripts/schema-publish.sh src",
-    "test": "c8 --100 vitest --watch=false && bash scripts/schema-test-coverage.sh",
+    "test": "c8 --100 vitest run --coverage && nyc check-coverage",
```
ralfhandl:
```suggestion
    "test": "c8 --100 vitest run --coverage && ([ \"$BASE\" = \"dev\" ] || nyc check-coverage )",
```
We need to skip `nyc` for pull requests targeting the `dev` branch. Once this suggestion is accepted, the `schema-test / test` check should succeed.

jdesrosiers:
> Unfortunately this does not run on Windows machines

I always forget about Windows! I made some fixes and made sure the tests pass with Github actions on windows. I don't have a windows machine to test locally, so let me know if there are any more issues.

vitest.config.mjs
jdesrosiers: I got threshold enforcement through vitest working, so I removed nyc.

ralfhandl:
> I don't have a windows machine to test locally, so let me know if there are any more issues.

Works like a charm, and you even got rid of the dependency to `nyc` 😄

vitest.config.mjs
```diff
@@ -1,8 +1,17 @@
 import { defineConfig } from 'vitest/config'
+import { jsonSchemaCoveragePlugin } from "@hyperjump/json-schema-coverage/vitest"
 
 export default defineConfig({
+  plugins: [jsonSchemaCoveragePlugin()],
   test: {
+    globalSetup: ["tests/schema/oas-schema.mjs"],
+    coverage: {
+      include: ["src/schemas/validation/**/*.yaml"],
+      thresholds: process.env.BASE !== "dev" ? {
```
ralfhandl, line 10: This is elegant!

ralfhandl (approved):

PR #4776, opened 2025-07-12
v3.2: Drop the idea of example overriding

handrews:
@mikekistler questioned why we have this language in one of the other example PRs, and after spending some time with it I agree.  This is a very simple change that will make the other example updates more clear.  Also, the new example fields (which @hudlow and I are pretty close to agreeing on at this point, I think) further clarify how to use different examples.

-----
Commit message:

Examples at different levels serve different purposes, and it is not clear what is meant by "overriding."  Removing this will allow tools to determine the most appropriate example(s) to show in a given context.


mikekistler (approved):
Looks good! 👍

ralfhandl (approved):

PR #4780, opened 2025-07-14
3.2 improve schema coverage

ralfhandl:
This PR currently contains a schema change that seems to have no negative effects and increases coverage of `schema.yaml`. I've opened
* https://github.com/hyperjump-io/json-schema-coverage/issues/1

to verify whether this is necessary, and if so, whether it is a good idea.

This PR also includes the changes in
* #4779 

which can't be merged due to the insufficient schema test coverage.


tests/schema/fail/invalid-components-object.yaml
```diff
@@ -0,0 +1,16 @@
...
+  schemas: true
+  responses: true
+  parameters: true
+  examples: true
+  requestBodies: true
+  headers: true
+  securitySchemes: true
+  links: true
+  callbacks: true
+  pathItems: true
```
handrews: Does the JSON Schema implementation actually check every single one of these?  Many if not most stop evaluating on the first error.  This is why most of the time negative test cases test only one condition at a time while postive test cases can test many conditions.

Also, why is `paths: true`?

handrews (commented):
I'm concerned that the actual validation coverage is not as thorough as the files make it appear- I have commented on one such example.  This all depends on how the JSON Schema implementation handles things, and whether we expect this suite to be usable with other implementations.  If we want it to work with any conforming JSON Schema implementation, we need to only test one failure condition per case.

tests/schema/fail/invalid-components-object.yaml
ralfhandl: > whether we expect this suite to be usable with other implementations

I don't expect that. These tests are for us to verify whether the schemas we publish describe what we intend. The schemas did not fully describe what former TSCs intended before we added test coverage and were wrong in "exotic" areas.

> Does the JSON Schema implementation actually check every single one of these?

Yes, it does. This single fail test case covers a lot of branches in a negative way.

> Many if not most stop evaluating on the first error.

These implementations would then not count the remaining negative cases, and test coverage would noticeably drop. If the tool we use would switch its behaviour, we would immediately know.

> why is `paths: true`?

Negative test for `type: object`. `paths: 42` would also work, and `paths: []`, with the latter being a more probable error than the other variants.

I could go over the fail test cases and generally use an object to fail `type: array` and an array to fail `type: object`.

handrews: @ralfhandl 

>> why is paths: true?
>
> Negative test for type: object. paths: 42 would also work, and paths: [], with the latter being a more probable error than the other variants.
> 
> I could go over the fail test cases and generally use an object to fail type: array and an array to fail type: object.

No need, this was more "why is this in the components test file" but we have other files that contain test conditions beyond what is mentioned in the file name so that's fine.

ralfhandl:
@handrews all test cases are now documented

tests/schema/fail/invalid-components-object.yaml
karenetheridge: > Yes, it does. This single fail test case covers a lot of branches in a negative way.

How is that done? If you are just checking for validation failure of the entire document, you don't know if only one of those invalid properties generated an error, or some of them, or all of them.  Is the test suite counting failures? I don't see any way of indicating that per test.

ralfhandl: Our test suite runs the schema validation tool on a set of „pass“ and „fail“ instances, one by one.

The schema validation tool evaluates the schema on a given instance, reports *all failed checks* as defined for the [BASIC output structure](https://json-schema.org/draft/2020-12/json-schema-core#section-12.4.2), collects all touched schema branches, and then decides whether the instance passes or fails validation.

The test suite then checks whether the instance passes or fails as expected.

If all test cases work as expected, the test suite succeeds.

Finally the overall coverage of all schema branches across all evaluated instances is checked against the threshold of 100%.

karenetheridge: Side note -- if you resolve the conversation before I've had a chance to read it, it makes it really hard to see that you responded.  I only saw that you requested a review for me but couldn't figure out why.

karenetheridge: "collects all touched schema branches" is that just the top level of the schema?  so if there is a deeper `properties` - for example testing something deep within a `parameters` object - that won't be seen?

It would be useful to publish this algorithm so that the test suite could be used by other implementations as well for validating their correctness.

For example, the JSON Schema Test Suite (https://github.com/json-schema-org/JSON-Schema-Test-Suite) includes documentation describing the structure of the test data and how to use it with any implementation, and I pull that into my JSON Schema implementation using git submodules which are run when extended testing is enabled.

karenetheridge:
Given the number of commits in this PR, including multiple merge commits, could this be squashed at merge time? There should be an option for it right in the merge button. TIA!

tests/schema/fail/invalid-components-object.yaml
ralfhandl: > if you resolve the conversation before I've had a chance to read it, it makes it really hard to see that you responded

You followed up to a resolved conversation between Henry and me without unresolving it.

I also had difficulties finding your comment, and assumed you intentionally left it resolved.

ralfhandl: > It would be useful to publish this algorithm so that the test suite could be used by other implementations as well for validating their correctness.

We are only using the publicly available tool https://github.com/hyperjump-io/json-schema-coverage.

ralfhandl: > "collects all touched schema branches" is that just the top level of the schema?

No, the tool seems to consider all nesting levels, I had to add a lot of test cases to cover all branches.

ralfhandl: @karenetheridge Did I answer all your questions?

karenetheridge: > You followed up to a resolved conversation between Henry and me without unresolving it.

Ok, yes, sorry!  I could see Henry's comment above the fold so I didn't think it was in a resolved conversation.

> We are only using the publicly available tool https://github.com/hyperjump-io/json-schema-coverage.

Ok, I'll pursue that with them.

ralfhandl:
> Given the number of commits in this PR, including multiple merge commits, could this be squashed at merge time?

I've squashed my last commits after the merge-commit from v3.2-dev.

I'm a bit hesitant to squash more because that may cause problems with merging the overdue PR
* #4787

which contains six commits that are also in this PR.

ralfhandl:
> There should be an option for [squash merging] right in the merge button

After a discussion in the TDC on preserving the commit history of PRs we disabled squash merging and rebase merging for this repo.

PR #4791, opened 2025-07-17
v3.2: Clarify template variable uniqueness

handrews:
_NOTE: This takes the opposite view as the thread in https://github.com/OAI/OpenAPI-Specification/issues/4763#issuecomment-3074650101 based on continued slack conversations where we dug deeper into the spec wording.  Please see the note after the list as well:_

This clarifies that variables in both path and server URL templates MUST be unique.  This is justified as compatible with our minor release policy on the following grounds:

* Due to Parameter Object `name` + `in` constraints and the Server Object's `variables` field being a map, each variable can only map to one paramater or server variable (although the "one" paramater might be defined separately yet uniquely for each Operation).
* The Path Templating section uses the phrase "Each template expression in the path MUST correspond to ***a*** path parameter" (emphasis added).
* The Parameter Object uses similar language in the other direction: "If in is "path", the name field MUST correspond to ***a*** template expression".
* The word "a" is interpreted to mean an unknown but unique thing in both directions.
* The Server Object's `url` template has always been assumed to function in an analogous way to path templating.

_Note: I think this is reasonably solid, and no one has actually presented a use case or known use of duplicate fields, but it does rely on interpreting language in a very specific way.  I am very much interested in whether people see this as a clarification or change, as my intent is for it to be a clarification._

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -62,6 +62,8 @@ sub-delims          = "!" / "$" / "&" / "'" / "(" / ")"
 
 Here, `pchar`, `unreserved`, `pct-encoded` and `sub-delims` definitions are taken from [RFC 3986](https://tools.ietf.org/html/rfc3986). The `path-template` is directly derived from [RFC 3986, section 3.3](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3).
 
+Each template expression MUST NOT appear more than once in a single path template.
+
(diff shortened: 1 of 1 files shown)


ralfhandl (approved):

mikekistler (approved):
Looks good. 👍

PR #4793, opened 2025-07-18
v3.2: Provide parsing and serialization guidance

handrews:
_**NOTE**: This PR is a replacement for #4743.  @whitlockjc and @baywet it is largely inspired by your questions of how far we need to go to explain how to build on schemas and other parts of the OAS, and how that fits with tools that do not work with runtime data (e.g. code gen).  This is now substantially less ambitious and I think much better for it._

This creates a "Working with Data" section that incorporates the existing "Data Types" section (with some section level adjustments) along with new guidance on mapping different kinds of data between serialized, data, and application forms.  This terminology matches the terminology currently being considered for examples.  Note that "Working with Binary Data" is just moved, but the diff got confused and sort-of makes it look like the heading got deleted.

The application form is largely out of scope for the OAS, and is mainly included to clarify this scope while acknowledging that the OAS may influence such things.

Most of the new material is on parsing and serializing, briefly addressing JSON as the common case before going into detail on non-JSON data, with examples.  This is where the requirements for schema and/or instance inspection/searching are listed.

The only additional change is no longer mentioning the property schema in the Encoding Object, in part because with the new `multipart/mixed` support Encoding Objects can be used with arrays as well as objects.

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -294,7 +294,7 @@ As noted under [Data Type](#data-types), both `type: number` and `type: integer`
 
 API data has several forms:
 
-1. The serialized form, which is either a document of a particular media type, part of an HTTP header value, or part of a URI.
+1. The serialized form, which is either a document of a particular media type, an HTTP header value, or part of a URI.
 2. The data form, intended for use with a [Schema Object](#schema-object).
 3. The application form, which incorporates any additional information conveyed by JSON Schema keywords such as `format` and `contentType`, and possibly additional information such as class hierarchies that are beyond the scope of this specification, although they MAY be based on specification elements such as the [Discriminator Object](#discriminator-object) or guidance regarding [Data Modeling Techniques](#data-modeling-techniques).
 


src/oas.md
```diff
@@ -288,7 +290,115 @@ The formats defined by the OAS are:
...
+
+API data has three forms:
+
+1. The serialized form, which is either a document of a particular media type, part of an HTTP header value, or part of a URI.
+2. The data form, intended for use with a [Schema Object](#schema-object).
+3. The application form, which incorporates any additional information conveyed by JSON Schema keywords such as `format` and `contentType`, and possibly additional information such as class hierarchies that are beyond the scope of this specification, although they MAY be based on specification elements such as the [Discriminator Object](#discriminator-object) or guidance regarding [Data Modeling Techniques](#data-modeling-techniques).
+
+##### JSON Data
+
+JSON-serialized data is nearly equivalent to the data form because the [JSON Schema data model](https://www.ietf.org/archive/id/draft-bhutton-json-schema-01.html#section-4.2.1) is nearly equivalent to the JSON representation.
```
ralfhandl, line 303: "Nearly equivalent"? What are the differences?

Just curios.

src/oas.md
```diff
@@ -288,7 +290,115 @@ The formats defined by the OAS are:
...
+    type: string
+    format: date-time
+```
+
+Some applications might leave the string as a string regardless of programming language, while others might notice the `format` and use it as a `datetime.datetime` instance in Python, or a `java.time.ZonedDateTime` in Java.
+This specification only requires that the data is valid according to the schema, and that [annotations](#extended-validation-with-annotations) such as `format` are available in accordance with the JSON Schema specification.
+
+##### Non-JSON Data
+
+Non-JSON serializetions can be substantially different from their corresponding data form, and might require several steps to parse.
```
ralfhandl: S
```suggestion
Non-JSON serializations can be substantially different from their corresponding data form, and might require several steps to parse.
```

src/oas.md
```diff
@@ -288,7 +290,115 @@ The formats defined by the OAS are:
...
+API data has three forms:
+
+1. The serialized form, which is either a document of a particular media type, part of an HTTP header value, or part of a URI.
+2. The data form, intended for use with a [Schema Object](#schema-object).
+3. The application form, which incorporates any additional information conveyed by JSON Schema keywords such as `format` and `contentType`, and possibly additional information such as class hierarchies that are beyond the scope of this specification, although they MAY be based on specification elements such as the [Discriminator Object](#discriminator-object) or guidance regarding [Data Modeling Techniques](#data-modeling-techniques).
+
+##### JSON Data
+
+JSON-serialized data is nearly equivalent to the data form because the [JSON Schema data model](https://www.ietf.org/archive/id/draft-bhutton-json-schema-01.html#section-4.2.1) is nearly equivalent to the JSON representation.
+The serialized UTF-8 JSON string `{"when": "1985-04-12T23%3A20%3A50.52"}` represents an object with one data field, named `when`, with a string value, `1985-04-12T23%3A20%3A50.52`.
```
ralfhandl: I would expect colons instead of %3A in JSON.

src/oas.md
```diff
@@ -288,7 +290,115 @@ The formats defined by the OAS are:
...
+
+Some applications might leave the string as a string regardless of programming language, while others might notice the `format` and use it as a `datetime.datetime` instance in Python, or a `java.time.ZonedDateTime` in Java.
+This specification only requires that the data is valid according to the schema, and that [annotations](#extended-validation-with-annotations) such as `format` are available in accordance with the JSON Schema specification.
+
+##### Non-JSON Data
+
+Non-JSON serializetions can be substantially different from their corresponding data form, and might require several steps to parse.
+
+To continue our "when" example, if we serialized the object as `application/x-www-form-urlencoded`, it would appear as the ASCII string `when=1985-04-12T23%3A20%3A50.52`.
+This example is still straightforward to use as it is all string data, and the only differences from JSON are the URI percent-encoding and the delimiter syntax (`=` instead of JSON punctuation and quoting).
```
ralfhandl, line 324: The JSON example currently shows percentage-encoding, which it probably should not based on this text.

src/oas.md
```diff
@@ -288,7 +290,115 @@ The formats defined by the OAS are:
...
+
+However, many non-JSON text-based formats can be complex, requiring examination of the appropriate schema(s) in order to correctly parse the text into a schema-ready data structure.
+Serializing data into such formats requires either examing the schema-validated data or performing the same schema inspections.
```
```suggestion
Serializing data into such formats requires either examining the schema-validated data or performing the same schema inspections.
```

src/oas.md
```diff
@@ -288,7 +290,115 @@ The formats defined by the OAS are:
...
+```uri
+code=1234&count=42&extra=%3Cinfo%3Eabc%3C/info%3E
+```
+
+We must first search the schema for `properties` or other property-defining keywords, and then use each property schema as a starting point for a search for that property's `type` keyword, as follows (the exact order is implementation-defined):
+
+* `#/components/requestBodies/Form/content/application~1x-www-form-urlencoded/schema` (initial starting point schema, only `$ref`)
+* `#/components/schemas/FormData` (follow `$ref`, found `properties`)
+* `#/components/schemas/FormData/properties/code` (starting point schema for `code` property)
+* `#/components/schemas/FormData/properties/code/allOf/0` (follow `allOf`, but no `type`)
```
ralfhandl: Currently the first `allOf` alternative has a `type`.

src/oas.md
```diff
@@ -288,7 +290,115 @@ The formats defined by the OAS are:
...
+            extra:
+              contentType: application/xml
+  schemas:
+    FormData:
+      type: object
+      properties:
+        code:
+          allOf:
+          - type: string
+            pattern: "1"
```
ralfhandl: Just
```suggestion
          - pattern: "1"
```
to make the fourth bullet point after the example correct?

src/oas.md
```diff
@@ -288,7 +290,115 @@ The formats defined by the OAS are:
...
+
+We must first search the schema for `properties` or other property-defining keywords, and then use each property schema as a starting point for a search for that property's `type` keyword, as follows (the exact order is implementation-defined):
+
+* `#/components/requestBodies/Form/content/application~1x-www-form-urlencoded/schema` (initial starting point schema, only `$ref`)
+* `#/components/schemas/FormData` (follow `$ref`, found `properties`)
+* `#/components/schemas/FormData/properties/code` (starting point schema for `code` property)
+* `#/components/schemas/FormData/properties/code/allOf/0` (follow `allOf`, but no `type`)
+* `#/components/schemas/FormData/properties/code/allOf/1` (follow `allOf`, found `type: string`)
+* `#/components/schemas/FormData/properties/count` (starting point schema for `count` property, found `type: integer`)
+* `#/components/schemas/FormData/properties/extra` (starting point schema for `count` property, found `type: object`)
```
```suggestion
* `#/components/schemas/FormData/properties/extra` (starting point schema for `extra` property, found `type: object`)
```

ralfhandl (changes requested):
+1, some nits and questions

src/oas.md
handrews: @ralfhandl If you go to the given link (https://www.ietf.org/archive/id/draft-bhutton-json-schema-01.html#section-4.2.1) it will tell you.  (I'm not trying to be snarky, I'm trying to see if it's clear enough as i really, really, really do not want to repeat this obscure but real distinction in our spec).

src/oas.md
handrews: Ah- good catch, but I had meant to use this to show that you might find `type` more than once but it's OK if it all matches.  I will update the bullet point instead.

src/oas.md
handrews: Oops, copy-pasted the wrong thing, will fix- thanks!

handrews:
@ralfhandl I have made the other updates, and in the process expanded the example and guidance somewhat.  I also restored the MAY allowing further inspection, and put some boundaries on it.  Yesterday I had not been able to figure out a good way to bound that but today it seems more tractable.  @baywet this might interest you in particular as it does place some restrictions on inferences that can be made about type.

ralfhandl (dismissed):

whitlockjc (dismissed):
👍

src/oas.md
```diff
@@ -288,7 +290,141 @@ The formats defined by the OAS are:
 
 As noted under [Data Type](#data-types), both `type: number` and `type: integer` are considered to be numbers in the data model.
 
-#### Working with Binary Data
+#### Parsing and Serializing
+
+API data has three forms:
```
```suggestion
API data has several forms:
```

src/oas.md
```diff
@@ -288,7 +290,141 @@ The formats defined by the OAS are:
...
+
+1. The serialized form, which is either a document of a particular media type, part of an HTTP header value, or part of a URI.
```
```suggestion
1. The serialized form, which is either a document of a particular media type, an HTTP header value, or part of a URI.
```

lornajane (approved):
Repeat approval since we just dismissed the others' on minor changes. Thanks!

PR #4799, opened 2025-07-18
v3.2 Add data vs serialized example fields (3rd try)

handrews:
_This replaces PR #4671, and is the result of numerous discussions with @hudlow.  Note that PR #4672 is still the update to the Example Objects examples as it required only a one-line update to conform with this PR._

This adds two fields to the Example Object and partially deprecates the existing `value` field.

`dataValue` applies to the data that would be passed to schema validation.

`serializedValue` (which MUST be a string), like `externalValue`, applies to the serialized form.

Guidance is provided that `value` (and the shorthand singluar `example`) are safe for JSON serialization targets and for strings serialized to targets that do not apply further escaping, but are otherwise deprecated due to ambiguous behavior.


miqui (approved):

src/oas.md
```diff
@@ -2101,42 +2101,84 @@ transactionCallback:
...
+| <a name="example-external-value"></a>externalValue | `string` | A URI that identifies the serialized example in a separate document, allowing for values not easily or readably expressed as a Unicode string.  If `dataValue` is present, then this field SHOULD identify a serialization of the given data.  Otherwise, the value SHOULD the valid serialization of a data value that itself MUST be valid as described for `dataValue`. If this field is present, `serializedValue`, and `value` MUST be absent. See also the rules for resolving [Relative References](#relative-references-in-api-description-uris). |
```
```suggestion
| <a name="example-external-value"></a>externalValue | `string` | A URI that identifies the serialized example in a separate document, allowing for values not easily or readably expressed as a Unicode string.  If `dataValue` is present, then this field SHOULD identify a serialization of the given data.  Otherwise, the value SHOULD be the valid serialization of a data value that itself MUST be valid as described for `dataValue`. If this field is present, `serializedValue`, and `value` MUST be absent. See also the rules for resolving [Relative References](#relative-references-in-api-description-uris). |
```

src/oas.md
```diff
@@ -2101,42 +2101,84 @@ transactionCallback:
...
+| <a name="example-value"></a>value | Any | Embedded literal example. The `value` field and `externalValue` field are mutually exclusive. To represent examples of media types that cannot naturally be represented in JSON or YAML, use a string value to contain the example, escaping where necessary.<br /><br>**Deprecated for non-JSON serialization targets:** Use `dataValue` and/or `serializedValue`, which both have unambiguous syntax and semantics, instead. |
```
ralfhandl: Using `<br /><br>` here is ... interesting. Other places use one form twice, and unfortunately both forms are used.

src/oas.md
```diff
@@ -2101,42 +2101,84 @@ transactionCallback:
...
+
+The `serializedValue` and `externalValue` fields both MUST show the serialized form of the data.
+For Media Type Objects, this is a document of the appropriate media type, with any Encoding Object effects applied.
+For Parameter and Header Objects using `schema` and `style` rather than a Media Type Object, see [Style Examples](#style-examples) for what constitutes a serialized value.
+
+###### Criteria for `serializedExample`
+
+A serialization can be represented as a valid Unicode string in `serializedValue` if any of the following are true of the serialization:
+
+* It is for a media type that supports a `charset` parameter that indicates a Unicode encoding such as UTF-8, or any valid subset of such an encoding, such as US-ASCII.
```
ralfhandl: How would I express a UTF-32 example within a JSON OAD that uses UTF-8?

Maybe it is exactly UTF-8 and not "such as UTF-8"?

ralfhandl (changes requested):

src/oas.md
handrews: @ralfhandl since UTF-32 is a Unicode character set, that example can be represented as Unicode and therefore stored in any UTF-(whatever) document.  This is explaining that the `serializedValue` field works with Unicode code points, not with character sets (lack of clarity on this is what @hudlow was objecting over).

So "such as UTF-8" is correct, as that is just the most common "UTF", but I can say "such as UTF-8 or any other Unicode encoding"?

Your other option is to use `externalValue` and point to a UTF-32-encoded document, which you would want to do if there is no way to express the UTF-32-ness in the OAD.

src/oas.md
handrews: @ralfhandl I'll update it to `<br><br>` which seems the most used, and we can go make all of them consistent after this is merged.  I do not care which we choose, but agree that mixing styles was strange!  And unintended.

handrews:
@ralfhandl both of the issues so far should now be fixed.

handrews:
@ralfhandl sorry, one more commit as I realized I could improve the language in the other bullet points, and provide an explantation of the expected conversions involved.

src/oas.md
```diff
@@ -2101,42 +2101,86 @@ transactionCallback:
...
+
+* It is for a media type that supports a `charset` parameter that indicates any Unicode encoding (UTF-8, UTF-16, etc.), or any valid subset of such an encoding, such as US-ASCII.
+* It is for a format (such as URIs or HTTP fields) or character-based media type that requires or defaults to a Unicode encoding, or any valid subset of such an encoding, such as US-ASCII, and this is not overridden by `charset`.
+* It is for a compound format where all parts meet at least one of the above criteria, e.g. a `multipart/mixed` media type with parts that are `application/json` (a media type that defaults to UTF-8) and `application/xml; charset=utf-8` (a media type with an explicit `charset` parameter).
+
+In all of these cases, the conversion from the character set of the OAD (presumed to be UTF-8 as the only interoperable character set for JSON, an therefore also for JSON-compatible YAML as noted in [[RFC9512]] [Section 3.4](https://www.rfc-editor.org/rfc/rfc9512.html#section-3.4)) first to Unicode code points and then to the actual serialization character set is well-defined.
```
```suggestion
In all of these cases, the conversion from the character set of the OAD (presumed to be UTF-8 as the only interoperable character set for JSON, and therefore also for JSON-compatible YAML as noted in [[RFC9512]] [Section 3.4](https://www.rfc-editor.org/rfc/rfc9512.html#section-3.4)) first to Unicode code points and then to the actual serialization character set is well-defined.
```

ralfhandl (changes requested):
Almost there 😎

ralfhandl (approved):

lornajane (approved):
I know it was a lot of round trips but this is a good outcome. Thank you!

PR #4800, opened 2025-07-18
v3.2: Explain Param and Header example serialization

handrews:
_Note: This replaces PR #4673, except for updates to examples outside of the style table which are in PR #4801.  This takes a somewhat different approach to explaining things compared to the older PR._

The rules for this have not been clear, and are not always intuitive. This states and explains them directly and ensures that the Style Examples table matches the rules.

Unlike past efforts, this provides a rule system regarding what is and is not included, based on a combination of what is produced by RFC6570 (or the nearest RFC6570 equivalent), modified by removing leading delimiters that are not correct for our usage due to differences from the assuptions made by RFC6570.

This also shows some uses of the new Example Object fields, including some that would be redundant but are included to clarify the different options; the redundancy is noted in the text.

- [ ] schema changes are included in this pull request
- [ ] schema changes are needed for this pull request but not done yet
- [X] no schema changes are needed for this pull request

PR #4801, opened 2025-07-18
v3.2: Parameter and Header example updates

handrews:
_Note: This replaces the other half of PR #4673 (the first half being replaced by PR #4800), and updates parameter and header examples for PRs #4800, #4799., and #4802._

This updates for both the new example fields and for examples with `content`

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -2637,8 +2637,12 @@ ETag:
     text/plain:
       schema:
         type: string
+        # Note that quotation markes are part of the
+        # ETag value, unlike many other headers that
+        # use a quoted string purely for managing
+        # reserved characters.
         pattern: ^"
-  example: xyzzx
+  example: '"xyzzy"'
 ```
 
 #### Tag Object


miqui (approved):

miqui (approved):

src/oas.md
```diff
@@ -1102,24 +1106,39 @@ description: username to fetch
...
+    dataValue: diṅnāga
+    serializedValue: di%E1%B9%85n%C4%81ga
+examples:
+  Al-Khwarizmi:
+    dataValue: "الخوارزميّ"
+    serializedValue: "%D8%A7%D9%84%D8%AE%D9%88%D8%A7%D8%B1%D8%B2%D9%85%D9%8A%D9%91"
 ```
 
-An optional query parameter of a string value, allowing multiple values by repeating the query parameter:
+An optional query parameter of a integer value, allowing multiple values by repeating the query parameter
```
```suggestion
An optional query parameter of a string value, allowing multiple values by repeating the query parameter
```

src/oas.md
```diff
@@ -1102,24 +1106,39 @@ description: username to fetch
...
 style: form
 explode: true
+examples:
+  ObjectList:
+    dataValue: ["one thing", "another thing"]
+    serializedValue: "thing=one%20thing&thing=another%20thing"
 ```
 
-A free-form query parameter, allowing undefined parameters of a specific type:
+A free-form query parameter, allowing arbitrary parameters of a `type: "integer"`:
```
```suggestion
A free-form query parameter, allowing arbitrary parameters of `type: "integer"`:
```

src/oas.md
```diff
@@ -1183,11 +1273,14 @@ content:
     schema:
       type: string
     example: $.a.b[1:1]
+examples:
```
ralfhandl, line 1276: `examples` and `example` can currently only be used with `schema`, not with `content`, see https://github.com/OAI/OpenAPI-Specification/blob/v3.2-dev/src/oas.md#fixed-fields-for-use-with-content.

This affects multiple examples :sad:

handrews:
@ralfhandl the email notification contained this:

> In [src/oas.md](https://github.com/OAI/OpenAPI-Specification/pull/4801#discussion_r2217350898):
> 
> > @@ -1183,11 +1273,14 @@ content:
>      schema:
>        type: string
>      example: $.a.b[1:1]
> +examples:
> examples and example can currently only be used with schema, not with content, see https://github.com/OAI/OpenAPI-Specification/blob/v3.2-dev/src/oas.md#fixed-fields-for-use-with-content.
> 
> This affects multiple examples :sad:

but I don't see it now... was this just because I had split this from PR #4802? (and accidentally posted them in the wrong order...)

src/oas.md
handrews: Ugh... every time I redo this part of a PR (which I've done like 4 times now as these examples have been impacted by multiple other PRs that have multiple revisions) I think I have gotten the type mis-aligned one way or another.  Thanks for catching!

ralfhandl:
Yes, I got confused for a minute and lost sight of the target picture.

src/oas.md
```diff
@@ -2545,6 +2638,7 @@ ETag:
       schema:
         type: string
         pattern: ^"
+  example: xyzzx
```
ralfhandl: I don’t understand why the example does not match the pattern required by the schema.

handrews: LOL neither do I 🤦

handrews: @ralfhandl OK so it turns out the `ETag` header makes strange use of quotes, and I'm not really sure how to represent it.  Here are examples from RFC9110:

```http
ETag: "xyzzy"
ETag: W/"xyzzy"
ETag: ""
```

I was trying to ensure the first or third would be valid, but the second would not be.  Which I guess means that unlike most headers where you strip the quotes (the `quoted-string` ABNF production), the quotes are significant in the `ETag`?  I think?

So I think maybe this should be `pattern: ^"` and `example: "\"xyzzy\""`?  Or maybe I should find a different example?  I just want an example of a real header....

handrews: @ralfhandl I updated it with that approach and added a comment about the unusual quote use.... I am open to dropping this as too confusing but it's actually a semantically significant documentation use case.

handrews: (I have no idea if anyone would _use_ this in an actual OAD, but it does tell you something useful).

ralfhandl (approved):

miqui (approved):

PR #4802, opened 2025-07-18
Allow Parameter/Header examples w/content

handrews:
_Note:  This replaces PR #4648, except for the examples which are in PR #4801._

Parameter and Header serialization is complex, particularly when using the `content` field to use a Media Type Object.

In such scenarios, the serialization occurs in two steps: The first step is to serialize the data to the media type, which can be captured by the `examples` field of the Media Type Object.

The second is the encoding/escaping of the media type document for use in a URI, HTTP header, or other location with its own rules.

Sometimes the part needing illustration with an example is at one level, sometimes at another, and sometimes it is helpful to show both.

For simplicity, the "data" examples are always treated as the overall input data, so they would be the same at both levels. This is also because it is not always possible to show each step, particularly when there are binary serializations. This allows showing either step (or both steps) with both data and serialization, depending on what makes sense for the use case.

### DIFF
--- a/src/schemas/validation/schema.yaml
+++ b/src/schemas/validation/schema.yaml
@@ -366,6 +366,8 @@ $defs:
       - required:
         - content
     allOf:
+      - $ref: '#/$defs/examples'
+      - $ref: '#/$defs/specification-extensions'
       - if:
           properties:
             in:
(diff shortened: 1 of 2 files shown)


miqui (dismissed):

handrews:
@ralfhandl good catch!  There should be an award for catching things that aren't actually part of the diff, that's advance reviewing skills there!

src/schemas/validation/schema.yaml
```diff
@@ -733,6 +733,9 @@ $defs:
             type: boolean
         $ref: '#/$defs/examples'
     $ref: '#/$defs/specification-extensions'
```
ralfhandl: This line seems to be implied by the second allOf entry.

ralfhandl (changes requested):

src/schemas/validation/schema.yaml
handrews: @ralfhandl thanks, fixed!

ralfhandl (approved):

miqui (approved):

PR #4812, opened 2025-07-23
v3.2: Move a paragraph, unchanged, to make more sense in multipart encoding

handrews:
When I made this a separate section in the nested multipart PR, I did not realize that the flow no longer worked.  Also, that was put up front when we only supported `form-data`, but now that we support other `multipart` subtypes, transfer encoding is relevant for those subtypes.  So just warn about `form-data` at the end of the section.


ralfhandl (approved):

mikekistler (approved):
Looks good. 👍

PR #4818, opened 2025-08-02
v3.2: Bug fix: deepObject explode default behavior.

handrews:
The `explode` field defaults to `false` with `style: deepObject`, but the behavior of that combination is undefined in 3.0 and 3.1.

For 3.2, define it to behave the same as
`style: deepObject, explode: true`, which maintains compatibility (by defining previously undefined behavior rather than changing the defined default value) and produces intuitive results.

Note that there is no need to update the Header Object because it does not allow `style: deepObject`.


src/oas.md
```diff
@@ -1236,7 +1236,7 @@ The following table shows serialized examples, as would be shown with the `seria
 | spaceDelimited | true | _n/a_ | _n/a_ | _n/a_ | _n/a_ |
 | pipeDelimited | false | _n/a_ | _n/a_ | <span style="white-space: nowrap;">color=blue%7Cblack%7Cbrown</span> | <span style="white-space: nowrap;">color=R%7C100%7CG%7C200%7CB%7C150</span> |
 | pipeDelimited | true | _n/a_ | _n/a_ | _n/a_ | _n/a_ |
-| deepObject | false | _n/a_ | _n/a_ | _n/a_ | _n/a_ |
+| deepObject | false | _n/a_ | _n/a_ | _n/a_ | <span style="white-space: nowrap;">color%5BR%5D=100&color%5BG%5D=200&color%5BB%5D=150</span> |
```
ralfhandl: Can we express that with one table line where the "explode" column says "false or true"? Then readers don't have to mentally diff the values in the two table lines.

handrews: done

whitlockjc (approved):
👍

PR #4819, opened 2025-08-02
v3.1.2: Fix guidance on headers and RFC6570 percent-encoding

handrews:
_**Note:** Since this is just correcting incorrect guidance rather than changing the actual behavior, it is suitable for a patch release.  I will port it to 3.2 once merged._

After much debate and research, we agreed that percent-encoding was never meant to be applied to headers.

Exactly how to handle RFC6570 with cookie parameters remains TBD. For now, this preserves (but streamlines) the existing guidance for cookies.

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -1164,7 +1164,10 @@ For simpler scenarios, a [`schema`](#parameter-schema) and [`style`](#parameter-
 When `example` or `examples` are provided in conjunction with the `schema` field, the example SHOULD match the specified schema and follow the prescribed serialization strategy for the parameter.
 The `example` and `examples` fields are mutually exclusive, and if either is present it SHALL _override_ any `example` in the schema.
 
-Serializing with `schema` is NOT RECOMMENDED for `in: "cookie"` parameters, `in: "header"` parameters that use HTTP header parameters (name=value pairs following a `;`) in their values, or `in: "header"` parameters where values might have non-URL-safe characters; see [Appendix D](#appendix-d-serializing-headers-and-cookies) for details.
+When serializing `in: "header"` parameters with `schema`, URI percent-encoding MUST NOT be applied; if using an RFC6570 implementation that automatically applies it, it MUST be removed before use.
+Implementations MUST NOT attempt to automatically quote header values, as the quoting rules vary too widely among different headers; users are expected to provide pre-quoted data as needed.
+
+Serializing with `schema` is NOT RECOMMENDED for `in: "cookie"` parameters; see [Appendix D](#appendix-d-serializing-headers-and-cookies) for details.
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
@@ -2563,7 +2566,8 @@ This object MAY be extended with [Specification Extensions](#specification-exten
 For simpler scenarios, a [`schema`](#header-schema) and [`style`](#header-style) can describe the structure and syntax of the header.
 When `example` or `examples` are provided in conjunction with the `schema` field, the example MUST follow the prescribed serialization strategy for the header.
 
-Serializing with `schema` is NOT RECOMMENDED for headers with parameters (name=value pairs following a `;`) in their values, or where values might have non-URL-safe characters; see [Appendix D](#appendix-d-serializing-headers-and-cookies) for details.
+When serializing `in: "header"` parameters with `schema`, URI percent-encoding MUST NOT be applied; if using an RFC6570 implementation that automatically applies it, it MUST be removed before use.
+Implementations MUST NOT attempt to automatically quote header values, as the quoting rules vary too widely among different headers; users are expected to provide pre-quoted data as needed.
 
 When `example` or `examples` are provided in conjunction with the `schema` field, the example SHOULD match the specified schema and follow the prescribed serialization strategy for the header.
 The `example` and `examples` fields are mutually exclusive, and if either is present it SHALL _override_ any `example` in the schema.
@@ -2581,7 +2585,6 @@ See also [Appendix C: Using RFC6570-Based Serialization](#appendix-c-using-rfc65
 ###### Fixed Fields for use with `content`
 
 For more complex scenarios, the [`content`](#header-content) field can define the media type and schema of the header, as well as give examples of its use.
-Using `content` with a `text/plain` media type is RECOMMENDED for headers where the `schema` strategy is not appropriate.
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
@@ -2612,13 +2615,9 @@ Requiring that a strong `ETag` header (with a value starting with `"` rather tha
 ```json
 "ETag": {
   "required": true,
-  "content": {
-    "text/plain": {
-      "schema": {
-        "type": "string",
-        "pattern": "^\""
-      }
-    }
+  "schema": {
+    "type": "string",
+    "pattern": "^\""
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


handrews:
@ralfhandl the "no spaces inside code span element" is causing a linting failure even though the space is there intentionally.  Is it possible to disable the rule for this particular use?  Or would it need to be disabled for the whole document?

handrews:
@ralfhandl I can't even get it to turn off locally.  I reworded the sentence- I dislike this option but I care more about getting this merged than blocking on linter pickiness.

earth2marsh:
LGTM, thanks!

handrews:
> Do you want to mention Structured Headers in the appendix?

I'd rather not, as it was previously mentioned only to given an example of conflicting escaping guidance, and there is no longer a conflict.  Structured Headers are still quite rare, and bringing that in is more likely to add confusion than clarity, I think.  For any header, you now need to look up the escaping rules yourself.

src/oas.md
```diff
@@ -1164,7 +1164,10 @@ For simpler scenarios, a [`schema`](#parameter-schema) and [`style`](#parameter-
 When `example` or `examples` are provided in conjunction with the `schema` field, the example SHOULD match the specified schema and follow the prescribed serialization strategy for the parameter.
 The `example` and `examples` fields are mutually exclusive, and if either is present it SHALL _override_ any `example` in the schema.
 
-Serializing with `schema` is NOT RECOMMENDED for `in: "cookie"` parameters, `in: "header"` parameters that use HTTP header parameters (name=value pairs following a `;`) in their values, or `in: "header"` parameters where values might have non-URL-safe characters; see [Appendix D](#appendix-d-serializing-headers-and-cookies) for details.
+When serializing `in: "header"` parameters with `schema`, URI percent-encoding MUST NOT be applied; if using an RFC6570 implementation that automatically applies it, it MUST be removed before use.
+Implementations MUST NOT attempt to automatically quote header values, as the quoting rules vary too widely among different headers; see [Appendix D](#appendix-d-serializing-headers-and-cookies) for guidance on quoting and escaping.
```
ralfhandl: This requirement here raises the question "What should implementations do instead?". 

Which is answered in Appendix D, so why not simply point to Appendix D without repeating parts of it here?

src/oas.md
```diff
@@ -2563,7 +2566,8 @@ This object MAY be extended with [Specification Extensions](#specification-exten
 For simpler scenarios, a [`schema`](#header-schema) and [`style`](#header-style) can describe the structure and syntax of the header.
 When `example` or `examples` are provided in conjunction with the `schema` field, the example MUST follow the prescribed serialization strategy for the header.
 
-Serializing with `schema` is NOT RECOMMENDED for headers with parameters (name=value pairs following a `;`) in their values, or where values might have non-URL-safe characters; see [Appendix D](#appendix-d-serializing-headers-and-cookies) for details.
+When serializing `in: "header"` parameters with `schema`, URI percent-encoding MUST NOT be applied; if using an RFC6570 implementation that automatically applies it, it MUST be removed before use.
+Implementations MUST NOT attempt to automatically quote header values, as the quoting rules vary too widely among different headers; see [Appendix D](#appendix-d-serializing-headers-and-cookies) for guidance on quoting and escaping.
```
ralfhandl: Dito: can we simply point to the appendix without repeating parts of it here?

src/oas.md
```diff
@@ -4522,25 +4519,25 @@ This will expand to the result:
...
+While the `quoted-string` ABNF rule given in [[RFC7230]] [Section 3.2.6](https://httpwg.org/specs/rfc7230.html#field.components) is the most common escaping solution, it is not sufficiently universal to apply automatically.
+For example, a strong `ETag` looks like `"foo"` (with quotes, regardless of the contents), and a weak `ETag` looks like `W/"foo"` (note that only part of the value is quoted); the contents of the quotes for this header are also not escaped in the way `quoted-string` contents are.
+
+For this reason, any data being passed to a header by way of a [Parameter](#parameter-object) or [Header](#header-object) Object needs to be quoted and escaped prior to passing it to the OAS implementation, and the parsed header values are expected to contain the quotes and escapes.
+
+_**Note:** OAS v3.0.4 and v3.1.1 applied the advice in this section to avoid RFC6570-style serialization to both headers and cookies.
+However, further research has indicated that percent-encoding was never intended to apply to headers, so this section has been corrected to apply only to cookies._
```
ralfhandl, line 4531: But we link here from the a paragraph on headers in the Parameter Object section.

Why, if it does not apply?

handrews:
@ralfhandl I have tried to ensure that all of the formal requirements are in non-appendix sections (this may not be entirely true, I should go through and check it once we finish the new PRs), so I do not want to remove those sections in favor of a link to Appendix D.  I have clarified them.  I have also broken up Appendix D so that the comment about what no longer applies to headers is more clear.  Note that for 3.2, PR #4822 would remove this part as well.

ralfhandl:
> Is it possible to disable the rule for this particular use?

Sorry, I’m not familiar with that linter.

src/oas.md
```diff
@@ -2563,7 +2566,8 @@ This object MAY be extended with [Specification Extensions](#specification-exten
 For simpler scenarios, a [`schema`](#header-schema) and [`style`](#header-style) can describe the structure and syntax of the header.
 When `example` or `examples` are provided in conjunction with the `schema` field, the example MUST follow the prescribed serialization strategy for the header.
 
-Serializing with `schema` is NOT RECOMMENDED for headers with parameters (name=value pairs following a `;`) in their values, or where values might have non-URL-safe characters; see [Appendix D](#appendix-d-serializing-headers-and-cookies) for details.
+When serializing headers with `schema`, URI percent-encoding MUST NOT be applied; if using an RFC6570 implementation that automatically applies it, it MUST be removed before use.
```
earth2marsh, line 2569: Should RFCs always get a link? The spec has some inconsistencies (and what looks like some occassional formatting issues, but that's a separate issue!), but if we chose to link all of them, this would need it. :)

src/oas.md
```diff
@@ -4522,25 +4519,27 @@ This will expand to the result:
...
+HTTP headers have inconsistent rules regarding what characters are allowed, and how some or all disallowed characters can be escaped and included.
+While the `quoted-string` ABNF rule given in [[RFC7230]] [Section 3.2.6](https://httpwg.org/specs/rfc7230.html#field.components) is the most common escaping solution, it is not sufficiently universal to apply automatically.
+For example, a strong `ETag` looks like `"foo"` (with quotes, regardless of the contents), and a weak `ETag` looks like `W/"foo"` (note that only part of the value is quoted); the contents of the quotes for this header are also not escaped in the way `quoted-string` contents are.
+
+For this reason, any data being passed to a header by way of a [Parameter](#parameter-object) or [Header](#header-object) Object needs to be quoted and escaped prior to passing it to the OAS implementation, and the parsed header values are expected to contain the quotes and escapes.
+
+### Percent-Encoding and Cookies
+
+_**Note:** OAS v3.0.4 and v3.1.1 applied the advice in this section to avoid RFC6570-style serialization to both headers and cookies.
```
earth2marsh, line 4530: (nit) it looks like most of the time notes aren't italicized, but this is one of maybe 3 that is? Minor inconsistency, but seeing it here made me wonder and look. #carryOn!

src/oas.md
handrews: It's linked elsewhere.  We should do a [style guide](https://github.com/OAI/OpenAPI-Specification/blob/main/style-guide.md) pass at the end for this sort of thing, and for the italicized note thing.

PR #4822, opened 2025-08-04
v3.2: First pass at a `style` solution for cookies

handrews:
`style: form` (the default and only allowed value for cookies) has never really made sense, and a casual survey of implementations shows that it has been interpreted in ways that are rather surprising (e.g. stuffing an entire query string, percent-encoded _including the delimiters_, as the cookie value), where it is supported at all.

Rather than try to redefine how `style: form` behaves in the context of `in: cookie`, we can just define `style: cookie` that does what makes sense.  Regrettably, compatibility rules prevent changing the default.

This is a draft because if this is accepted, there's a bit more tweaking to do, particularly with Appendix D.  Also, I'd probably add an example.

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -1151,8 +1151,8 @@ Serializing with `schema` is NOT RECOMMENDED for `in: "cookie"` parameters, `in:
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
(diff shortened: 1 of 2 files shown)


ralfhandl:
This is probably the best we can do at the moment, so +1 from me.

Can we have an example for `in: cookie, style: cookie` with `dataValue` and `serializedValue`, and ideally a "pass" test case with the example?

handrews:
@ralfhandl 

> Can we have an example for in: cookie, style: cookie with dataValue and serializedValue, and ideally a "pass" test case with the example?

Yup.  Given two generally-positive reactions it's worth rounding this PR out as you suggest and taking it out of draft.  Will do shortly.

miqui (approved):

handrews:
Replaced by PR #4902

PR #4825, opened 2025-08-10
v3.1.2 Update examples and appendicies for percent-encoding

### LINKED ISSUE #4813: v3.1: application/x-www-form-urlencoded potential inconsistency between examples and referenced RFC

handrews:
Fixes #4813.

After adding a new section on percent-encoding guidance, this updates the examples and other supplemental text to match it. _This is a bit challenging to review because the most likely error is that I missed something that should be updated._


ralfhandl (dismissed):

lornajane:
Looks good but we need to update the branch (probably because I just merged something else!)

handrews:
@lornajane OK rebased and should be fine now.

ralfhandl (approved):

lornajane (approved):
Thank you!

PR #4826, opened 2025-08-10
v3.1.2: Explain Param and Header example serialization (backports PR 4800)

handrews:
Backport of #4800, with the parts referencing the new Example Object fields and guidance on `Set-Cookie` removed.  Note that we do have one use of the Example Object's `value` field for an `application/x-www-form-urlencoded` Media Type Object, and this change to the style examples table makes the table line up with the example.  Which I think is a good thing.  Since we're doing a bunch of other percent-encoding-related tidying, we might as well include this and make 3.1.2 and 3.2.0 consistent.

Commit message follows:

-----

The rules for this have not been clear, and are not always intuitive. This states and explains them directly and ensures that the Style Examples table matches the rules.

Unlike past efforts, this provides a rule system regarding what is and is not included, based on a combination of what is produced by RFC6570 (or the nearest RFC6570 equivalent), modified by removing leading delimiters that are not correct for our usage due to differences from the assuptions made by RFC6570.

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -1230,9 +1230,27 @@ In some cases, such as inserting `/` into path parameter values, doing so is [ex
 
 See [Appendix E](#appendix-e-percent-encoding-and-form-media-types) for a thorough discussion of percent-encoding options, compatibility, and OAS-defined delimiters that are not allowed by RFC3986, and [Appendix C](#appendix-c-using-rfc6570-based-serialization) for guidance on using RFC6570 implementations.
 
+##### Serialization and Examples
+
+The rules in this section apply to both the Parameter and [Header](#header-object) Objects, both of which use the same mechanisms.
+
+When showing serialized examples using the `example` field or [Example Objects](#example-object), in most cases the value to show is just the value, with all relevant percent-encoding or other encoding/escaping applied, and also including any delimiters produced by the `style` and `explode` configuration.
+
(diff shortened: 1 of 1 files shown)


miqui (approved):

src/oas.md
```diff
@@ -1230,23 +1230,40 @@ In some cases, such as inserting `/` into path parameter values, doing so is [ex
...
+
+For headers, the header name MUST NOT be included as part of the serialization, as it is never part of the RFC6570-derived result.
+However, names produced by `style: "simple", explode: "true"` are included as they appear within the header value, not as separate headers.
+
+The following section illustrates these rules.
+
 ##### Style Examples
 
-Assume a parameter named `color` has one of the following values:
+Assume a parameter named `color` has one of the following values, where the value to the right of the `->` is what would be shown in the `dataValue` field of an Example Object:
```
ralfhandl: We did not add `dataValue` in 3.1.2

ralfhandl (changes requested):
+1 in general, please remove `dataValue`

src/oas.md
handrews: Argh!  I specifically remember noting that to take out, but I guess I forgot?  Blah.  Fixed now.

lornajane: This is a GREAT catch @ralfhandl :) Thanks for fixing, @handrews

PR #4840, opened 2025-08-12
Reusable request bodies need a name

ralfhandl:
Two of the new "fail" test cases fail for the wrong reason: the name of the reuse component was missing.

With this change they fail for the right reason.


ralfhandl:
Merging with one approval as it only changes test cases

PR #4853, opened 2025-08-12
3.2 schema tests: correct test suite name

ralfhandl:
The old suite name "v3.1" was confusing when checking workflow logs.


ralfhandl:
Merged with one approval because it only changes display text in the test suite

PR #4857, opened 2025-08-13
3.2 schema tests: 100% subschema (function) coverage

ralfhandl:
This covers all occurrences of `unevaluatedProperties: true` and brings subschema (function) coverage up to 100%.

This would allow setting the coverage threshold to 100 and then automatically spotting if someone added a new fixed field to an Object without adding a "pass" test case for that fixed field.

See also discussion
* https://github.com/OAI/OpenAPI-Specification/discussions/4841 


miqui (approved):

ralfhandl:
Replaced with better approach in 
* #5095

PR #4876, opened 2025-08-16
3.2 typos and minor editorial changes

ralfhandl:

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -240,9 +240,9 @@ Unless specified otherwise, all fields that are URIs MAY be relative references 
 
 ##### Establishing the Base URI
 
-Relative URI references are resolved using the appropriate base URI, which MUST be determined in accordance with [[RFC3986]] [Section 5.1.1 – 5.1.4](https://tools.ietf.org/html/rfc3986#section-5.1.1) and, for Schema objects, [JSON Schema draft 2020-12 Section 8.2](https://www.ietf.org/archive/id/draft-bhutton-json-schema-01.html#section-8.2), as illustrated by the examles in [Appendix G: Examples of Base URI Determination and Reference Resolution](#appendix-g-examples-of-base-uri-determination-and-reference-resolution).
+Relative URI references are resolved using the appropriate base URI, which MUST be determined in accordance with [[RFC3986]] [Section 5.1.1 – 5.1.4](https://tools.ietf.org/html/rfc3986#section-5.1.1) and, for Schema objects, [JSON Schema draft 2020-12 Section 8.2](https://www.ietf.org/archive/id/draft-bhutton-json-schema-01.html#section-8.2), as illustrated by the examples in [Appendix G: Examples of Base URI Determination and Reference Resolution](#appendix-g-examples-of-base-uri-determination-and-reference-resolution).
 
-If `$self` is a relative URI-reference, it is resolved agains the next possible base URI source ([[RFC3986]] [Section 5.1.2 – 5.1.4](https://tools.ietf.org/html/rfc3986#section-5.1.2)) before being used for the resolution of other relative URI-references.
+If `$self` is a relative URI-reference, it is resolved against the next possible base URI source ([[RFC3986]] [Section 5.1.2 – 5.1.4](https://tools.ietf.org/html/rfc3986#section-5.1.2)) before being used for the resolution of other relative URI-references.
 
 The most common base URI source that is used in the event of a missing or relative `$self` (in the [OpenAPI Object](#openapi-object)) and (for [Schema Object](#schema-object)) `$id` is the retrieval URI.
 Implementations MAY support document retrieval, although see the [Security Considerations](#security-considerations) sections for additional guidance.
@@ -263,8 +263,8 @@ API endpoints are by definition accessed as locations, and are described by this
 
 Unless specified otherwise, all fields that are URLs MAY be relative references as defined by [RFC3986](https://tools.ietf.org/html/rfc3986#section-4.2).
 
-Because the API Is a distinct entity from the OpenAPI Document, RFC3986's base URI rules for the OpenAPI Document do not apply.
-Unless specified otherwise, relative references are resolved using the URLs defined in the [Server Object](#server-object) as a Base URL. Note that these themselves MAY be relative to the referring document.
+Because the API is a distinct entity from the OpenAPI Document, RFC3986's base URI rules for the OpenAPI Document do not apply.
+Unless specified otherwise, relative references are resolved using the URLs defined in the [Server Object](#server-object) as a base URL. Note that these themselves MAY be relative to the referring document.
 
 ##### Examples of API Base URL Determination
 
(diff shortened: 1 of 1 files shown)


mikekistler (approved):
Looks good! 👍

src/oas.md
```diff
@@ -883,7 +883,7 @@ The path is appended to the URL from the [Server Object](#server-object) in orde
...
+| <a name="paths-path"></a>/{path} | [Path Item Object](#path-item-object) | A relative path to an individual endpoint. The field name MUST begin with a forward slash (`/`). The path is **appended** (no relative URL resolution) to the resolved and template-variable-substituted URL from the [Server Object](#server-object)'s `url` field in order to construct the full URL. [Path templating](#path-templating) is allowed. When matching URLs, concrete (non-templated) paths would be matched before their templated counterparts. Templated paths with the same hierarchy but different templated names MUST NOT exist as they are identical. In case of ambiguous matching, it's up to the tooling to decide which one to use. |
```
handrews: I don't agree with this change.  "template variable" is not hyphenated, and hyphenating it, as a whole, with something else, does not create a hyphen.  Although as always, if you can cite a rule from a reliable source, I'll drop my objection.  Or if others agree with you.  I don't feel _that_ strongly about it.

src/oas.md
```diff
@@ -1578,6 +1578,8 @@ Unlike `schema`, which is applied to the complete content (treated as an array a
 Both `schema` and `itemSchema` MAY be used in the same Media Type Object.
 However, doing so is unlikely to have significant advantages over using the `items` keyword within the `schema` field.
 
+<a name="considerations-event-stream"></a><!-- works and satisfies both markdownlint and ReSpec -->
+
```
handrews: Can you explain this a bit?  I don't recall having problems linking to this and I have definitely linked to it including from the registry which worked locally just fine.

src/oas.md
```diff
@@ -1821,7 +1823,7 @@ Our `application/json-seq` example has to be an external document because of the
 
 ###### Server-Sent Event Streams
 
-For this example, assume that the generic event schema provided in the "Special Considerations for `text/event-stream` Content" section is available at `#/components/schemas/Event`:
+For this example, assume that the generic event schema provided in the [Special Considerations for `text/event-stream` Content](#considerations-event-stream) section is available at `#/components/schemas/Event`:
```
handrews, line 1826: Just in case it was a consideration, this wasn't un-linked because of a problem, it was un-linked because of the complexity of getting all of the PRs through while avoiding circular dependencies.

src/oas.md
```diff
@@ -2068,8 +2070,8 @@ requestBody:
           name:
             type: string
           icon:
-            # The default content type with "contentEncoding" present
-            # is application/octet-stream, so we need to set the correct
+            # The default content type with `contentEncoding` present
+            # is "application/octet-stream", so we need to set the correct
```
handrews: I agree with "contentEncoding" to `contentEncoding, but I question the quotes on `application/octet-stream`, as media types are _always_ written in backticks elsewhere.  So if we're keeping backticks for field names, why not also for media types?  I find it confusing to do one but not the other.

src/oas.md
```diff
@@ -2126,20 +2128,20 @@ requestBody:
...
-          # is `application/octet-stream`
+          # is "application/octet-stream"
           profileImage: {}
 
           # for arrays, the `encoding` field applies the Encoding Object
           # to each item individually and determines the default content type
           # based on the type in the `items` subschema, which in this example
           # is an object, so the default content type for each item is
-          # `application/json`
+          # "application/json"
```
handrews: Same with these and other changes from backticks to quotes on media types.  We should either keep the regular backticks formatting in the comments for everything that is formatted that way, or for nothing.

src/oas.md
```diff
@@ -4507,47 +4506,47 @@ This example does not define properties for `"related"` as it is showing how
...
-      "count": 42,
-      "description: "Thing"
-      "related": {}
-    }
-   externalValue: ./examples/productNoNulls.xml
+application/xml:
+  schema:
+    xml:
+      name: product
+    type: object
```
handrews, line 4513: 🤦 🤦 🤦

src/oas.md
```diff
@@ -5395,7 +5395,7 @@ components:
 In this example, the retrieval URIs are irrelevant because both documents define `$self`.
 
 The relative `$ref` in the first document is resolved against `$self` to produce `https://example.com/api/shared/foo#/components/requestBodies/Foo`.
-The portion of that URI before the '#' matches the `$self` of the second document, so the reference target is resolved to `#/components/requestBodies/Foo` in that second document.
+The portion of that URI before the "#" matches the `$self` of the second document, so the reference target is resolved to `#/components/requestBodies/Foo` in that second document.
```
handrews: This was meant to be backticks, as that is how I usually format characters from URIs.  Although I'm not sure we have a rule for that?  I partially do it because it makes them visually much more distinctive and it can otherwise be easier to miss a single character while reading.

handrews (commented):
@ralfhandl I really, _really_ appreciate your thorough read-through here.  I have some minor quibbles and questions, but nothing major and nothing I feel so strongly about as to really dig in if other folks are fine with it.

ralfhandl:
@handrews Let's discuss your questions, please comment on the changes you don't fully agree with.

src/oas.md
lornajane:
```suggestion
| <a name="paths-path"></a>/{path} | [Path Item Object](#path-item-object) | A relative path to an individual endpoint. The field name MUST begin with a forward slash (`/`). The URL from the [Server Object](#server-object)'s `url` field, resolved and with template variables substituted, has the path **appended** (no relative URL resolution) to it in order to construct the full URL. [Path templating](#path-templating) is allowed. When matching URLs, concrete (non-templated) paths would be matched before their templated counterparts. Templated paths with the same hierarchy but different templated names MUST NOT exist as they are identical. In case of ambiguous matching, it's up to the tooling to decide which one to use. |
```

Does rewording it help?

ralfhandl: @handrews Please resolve if you agree.

src/oas.md
ralfhandl: Yes, I recall.

src/oas.md
ralfhandl: Problem is that respec and markdownlint generate anchors for headings slightly differently:
* respec replaces `/` with `-`, resulting in
  * `#special-considerations-for-text-event-stream-content`
* markdownlint removes `/`, resulting in
  * `#special-considerations-for-textevent-stream-content`

If we use the respec anchor, it will work in the spec HTML and markdownlint will complain.

If we use the markdownlink anchor, markdownlint is satisfied and the link will _not_ work in the spec HTML.

Using a custom anchor avoids that problem.

src/oas.md
ralfhandl: That depends whether the media type mentioned in the comment is a field name or a field value.

In 3.1.1 we used backticks, will revert that.

src/oas.md
ralfhandl: Will change

src/oas.md
handrews: UGH.

lornajane: Thanks for explaining and fixing it, @ralfhandl

handrews: @ralfhandl I tested this and it breaks my expectations because it does not change how the little `§` anchor link that I use to find the right section-specific URI still uses the auto-generated one.  There is no indication to the user that there is a special link thing happening here.  Can we get markdownlint to ignore this section?  I really don't like the idea of either breaking user expectations or changing our section titles because markdownlint can't handle something that is, in fact, 100% valid.

handrews:
@ralfhandl 

> @handrews Let's discuss your questions, please comment on the changes you don't fully agree with.

I think most things are now fine.  I replied to the markdownlint section header problem elsewhere

> That depends whether the media type mentioned in the comment is a field name or a field value.

There's also just using a media type as a media type regardless of its position.  But in all three cases, it should be in backticks, just possibly _also_ in quotes.  I think.  Like this:

* Field name: `text/event-stream`
* Field value: `"text/event-stream"`
* Media type, regardless of position (or position ambiguous): `text/event-stream`

miqui (approved):

src/oas.md
ralfhandl: @handrews Both fragments work and navigate to the same place in the HTML:
- the auto-generated fragment `#special-considerations-for-text-event-stream-content` that you can pick up from the `§` link and from the table of contents in the HTML, and
- the custom fragment `#considerations-event-stream` that is used in the spec text.

It doesn't matter which one you use.

ralfhandl:
@mikekistler Could you please re-approve?

PR #4877, opened 2025-08-16
3.1 Runtime Expression ABNF: use exact same name as in RFC7159

ralfhandl:
We already did this in v3.2.


mikekistler (approved):
Looks good. 👍

handrews (approved):
Oh I see what happened.  I deliberately did _not_ backport this b/c I misread it as coming from RFC5234 which uses `CHAR`, but RFC7159 (cited here) does use `char` (as does RFC8259 which is cited in 3.2).

PR #4881, opened 2025-08-17
main CONTRIBUTING.md: update Publishing section

ralfhandl:
We now have a version-independent file `tests/schema/oas-schema.mjs` that needs to be preserved when preparing a release branch because we don't want it deleted from `main`. The `adjust-release-branch.sh` script already does this, now adjusting the prose text.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -213,7 +213,7 @@ The steps for creating a `vX.Y.Z-rel` branch are:
    - move `src/oas.md` to `versions/X.Y.Z.md`
    - copy `EDITORS.md` to `versions/X.Y.Z-editors.md`
    - delete `src/schemas`
-   - delete `tests/schema`
+   - delete version-specific files and folders from `tests/schema`
    - bash script `scripts/adjust-release-branch.sh` performs these steps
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -213,7 +213,7 @@ The steps for creating a `vX.Y.Z-rel` branch are:
    - move `src/oas.md` to `versions/X.Y.Z.md`
    - copy `EDITORS.md` to `versions/X.Y.Z-editors.md`
    - delete `src/schemas`
-   - delete `tests/schema`
+   - delete version-specific files and folders from `tests/schema`
```
lornajane: Can we name the list of files and folders for this step like we do for the others?

ralfhandl: Sure. I intentionally kept it vague because the concrete list is in the reference bash script.

lornajane: I think it would be clearer to say as part of the main text for this step that there is a bash script to run, and that the script does the things listed. Because if I were following this checklist I would have done all the steps before reaching the final entry and discovering I should have run a script. Does that make sense?

miqui (approved):

PR #4882, opened 2025-08-17
3.2 code block language: convention is to use lowercase

ralfhandl:
Some tools are picky and get confused by mixed/upper case.

Apparently not (at the moment) the tools we use, but better be consistent. The vast majority of the almost 200 code blocks already uses lowercase.


handrews:
Does this need to go into 3.1.2 as well?  I've been trying to put things that go into both in 3.1.2 first and then merge forward (as we did last time to start things in 3.0.4 and merge forward to 3.1.1).

miqui (approved):

PR #4899, opened 2025-08-21
v3.2: Enc style defaults

handrews:
While working on cookie-related style changes, I noticed that we are unclear on how `style` defaulting works for the Encoding Object.  Specifically, in 3.1.1 we clarified that all of `style`, `explode`, and `allowReserved` apply when _any_ of those three are explicitly present.  Otherwise, `contentType` applies (even if it is absent- its defaults apply).

So the default behavior for `style` takes place if either of the other two fields are there, but not if _none_ of them are there (whether `contentType` is there or not).
This requires some rather complex schema changes and tests, which have not been done yet.

On the one hand, this is a clarification and should probably be backported to 3.1.2.  On the other hand, it has schema changes, but more in the "we got the schema wrong" sense than in the "we're changing what was expected."  I'm not entirely sure what to do with this for 3.1.2.


ralfhandl (approved):

lornajane (approved):
Helpful clarification, thanks.

karenetheridge:
I've got a schema change for this, but I don't see how it can be tested, since "default" is an annotation-only keyword and its presence or absence cannot cause a validation failure.

handrews:
@karenetheridge I would not expect a test is necessary- does the coverage go down in a way that makes things fail when you add the schema change?  If not, at least for right now I think that's sufficient.  Testing `default` is really impossible (as you know but others might not) because it's an annotation that only applies when the instance is not there to be annotated 🤦

Issue #4598, opened 2025-05-16
Check that the 3.2 schema is in sync with the spec changes


lornajane:

lornajane:
We think it is in sync

karenetheridge:
We can add something for https://github.com/OAI/OpenAPI-Specification/pull/4776.

(reopening as we haven't finalized the list of changes for 3.2 yet)

handrews:
@karenetheridge what does #4776 require?  It does not change anything that was in the schema AFAIK?  (I agree with re-opening it, though, there will be schema changes with the `style: "cookie"` and `allowReserved` PRs).

karenetheridge:
"The `example` and `examples` fields are mutually exclusive."

My understanding is that specification restrictions are reflected in the schema whenever it is possible to express that restriction.  e.g. we include the restriction that "schema" and "content" are mutually exclusive in the parameter and header objects, and also "operationRef" and "operationId" are mutually exclusive in the link object.

So for #4776, we would do:

```yaml
oneOf:
  - required: [example]
  - required: [examples]
```

handrews:
Other PRs needing schema changes (draft PR that may or may not be at all correct:

* #4904
* #4899 
* #4900 
* #4902

handrews:
@karenetheridge 

> "The `example` and `examples` fields are mutually exclusive."

I thought that was already checked in the schemas- most such conditions are.  This is not new in 3.2, it just got moved because those fields (together) are no longer conditional on other fields in the Parameter and Header Objects.

If they're not, I agree that they should be.

karenetheridge:
After editing out the unchecked boxes from all issues in the 3.2.0 milestone, this search now works:

https://github.com/OAI/OpenAPI-Specification/issues?q=%22schema%20changes%20are%20needed%20for%20this%20pull%20request%20but%20not%20done%20yet%22%20milestone%3Av3.2.0&page=1

PR #4900, opened 2025-08-21
v3.2: Fix allowReserved

handrews:
We want it to apply only when percent-encoding applies:

* `in: path`
* `in: query`
* `in: cookie` [no `style`- deafult is `form`]
* `in: cookie, style: form`

But not:

* `in: header`
* `in: cookie, style: cookie`

The schema changes needed for this are a bit complex to handle `in: "cookie"` with both no `style`, and with an expicit `style: "form"`.


lornajane (approved):
Good clarification, thanks

lornajane:
Can we get a separate issue for the followup schema change so we don't lose it? I'm not sure I understand enough to create the issue.

karenetheridge:
I can provide the schema changes to this PR.

ralfhandl (approved):

handrews:
@lornajane I added https://github.com/OAI/OpenAPI-Specification/issues/4598#issuecomment-3211268653 to the schema-tracking issue that I know @karenetheridge is already following.

karenetheridge:
"This field only applies to.." -- is that intending to say that other combinations of properties are ILLEGAL? or simply that if they're there, they are ignored?

If the latter, then there are no schema changes to be made here.

handrews:
@karenetheridge the "only applies" language has been there since 3.0.0, we're only broadening it's applicability a bit.  So... whatever it meant before?  I think we have interpreted that as "ignore", but I'm not 100% sure.  Ignoring is it reasonable-ish, particularly with the new wording, as there are no reserved characters to allow for the field combinations to which it does not apply.

(TL;DR:  don't know)

karenetheridge:
I feel like we can't add exclusions to the schema here, at least not until 4.0 when breaking changes are allowed.

PR #4902, opened 2025-08-21
Add style: cookie, explain pct-encoding

handrews:
Replaces PR #4822.  _Please pay close attention while reviewing, I wrote this through various bouts of illness and then split it from other smaller PRs, and I'm not entirely sure I got it all._

This adds a `style: cookie` to provide a non-percent-encoding option for managing cookie parameters.

It also explains the conditions under which automatic percent-encoding with `style: form` might work, and also links to a draft update of the cookie RFC that explains that things that look like percent-encoding should _not_ be decoded for storage, which means that `style: cookie` is providing the most compliant behavior for that updated guidance.


src/oas.md
```diff
@@ -1148,15 +1148,13 @@ For simpler scenarios, a [`schema`](#parameter-schema) and [`style`](#parameter-
 
 These fields MUST NOT be used with `in: "querystring"`.
 
-When serializing `in: "header"` parameters with `schema`, URI percent-encoding MUST NOT be applied; if using an RFC6570 implementation that automatically applies it, it MUST be removed before use.
-Implementations MUST pass header values through unchanged rather than attempting to automatically quote header values, as the quoting rules vary too widely among different headers; see [Appendix D](#appendix-d-serializing-headers-and-cookies) for guidance on quoting and escaping.
-
-Serializing with `schema` is NOT RECOMMENDED for `in: "cookie"` parameters; see [Appendix D](#appendix-d-serializing-headers-and-cookies) for details.
+When serializing `in: "header"` or `in: "cookie", style: "cookie"` parameters with `schema`, URI percent-encoding MUST NOT be applied, and when parsing any apparent percent-encoding MUST NOT be decoded; if using an RFC6570 implementation that automatically performs these steps, the steps MUST be reversed before use.
```
lornajane:
```suggestion
Care is needed for parameters with `schema` that have `in: "header"` or `in: "cookie", style: "cookie"`:
* When serializing these values, URI percent-encoding MUST NOT be applied.
* When parsing these parameters, any apparent percent-encoding MUST NOT be decoded.
* If using an RFC6570 implementation that automatically performs encoding or decoding steps, the steps MUST be undone before use.
```

We found this confusing, so I've tried to make it clearer. If it is all still valid, hopefully it's an improvement.

src/oas.md
```diff
@@ -1211,7 +1209,11 @@ Reserved characters MUST NOT be percent-encoded when being used for reserved pur
 The result of inserting non-percent-encoded delimiters into data using manual percent-encoding, including via RFC6570's reserved expansion rules, is undefined and will likely prevent implementations from parsing the results back into the correct data structures.
 In some cases, such as inserting `/` into path parameter values, doing so is [explicitly forbidden](#path-templating) by this specification.
 
-See [Appendix E](#appendix-e-percent-encoding-and-form-media-types) for a thorough discussion of percent-encoding options, compatibility, and OAS-defined delimiters that are not allowed by RFC3986, and [Appendix C](#appendix-c-using-rfc6570-based-serialization) for guidance on using RFC6570 implementations.
+See also:
+
+* [Appendix C](#appendix-c-using-rfc6570-based-serialization) for guidance on using or simulating RFC6570 implementations.
+* [Appendix D](#appendix-d-serializing-headers-and-cookies) for guidance on percent-encoding and cookies, as well as other escaping approaches for headers and cookies
```
baywet:
```suggestion
* [Appendix D](#appendix-d-serializing-headers-and-cookies) for guidance on percent-encoding and cookies, as well as other escaping approaches for headers and cookies.
```

src/oas.md
```diff
@@ -1211,7 +1209,11 @@ Reserved characters MUST NOT be percent-encoded when being used for reserved pur
...
+* [Appendix E](#appendix-e-percent-encoding-and-form-media-types) for a thorough discussion of percent-encoding options, compatibility, and handling OAS-defined delimiters that are not allowed by RFC3986
```
baywet:
```suggestion
* [Appendix E](#appendix-e-percent-encoding-and-form-media-types) for a thorough discussion of percent-encoding options, compatibility, and handling OAS-defined delimiters that are not allowed by RFC3986.
```

src/oas.md
```diff
@@ -1211,7 +1209,11 @@ Reserved characters MUST NOT be percent-encoded when being used for reserved pur
 The result of inserting non-percent-encoded delimiters into data using manual percent-encoding, including via RFC6570's reserved expansion rules, is undefined and will likely prevent implementations from parsing the results back into the correct data structures.
 In some cases, such as inserting `/` into path parameter values, doing so is [explicitly forbidden](#path-templating) by this specification.
 
-See [Appendix E](#appendix-e-percent-encoding-and-form-media-types) for a thorough discussion of percent-encoding options, compatibility, and OAS-defined delimiters that are not allowed by RFC3986, and [Appendix C](#appendix-c-using-rfc6570-based-serialization) for guidance on using RFC6570 implementations.
+See also:
+
+* [Appendix C](#appendix-c-using-rfc6570-based-serialization) for guidance on using or simulating RFC6570 implementations.
```
lornajane: Why are we simulating it here? Not just using?

src/oas.md
```diff
@@ -1297,6 +1301,54 @@ examples:
     serializedValue: "12345678,90099"
 ```
 
+
+A cookie parameter with an exploded object (the default for `style: "cookie"`):
+
+```yaml
```
lornajane, line 1312: Could we have one more level of surrounding tag maybe?

src/oas.md
```diff
@@ -1297,6 +1301,54 @@ examples:
...
+    description: |
+        Note that the comma (,) has been pre-percent-encoded
+        to "%2C" in the data, as it is forbidden in
+        cookie values.  However, the exclamation point (!)
+        is legal in cookies, so it can be left unencoded.
+    dataValue: {
+      "greeting": "Hello%2C world!",
+      "code": 42
+    }
+    serializedValue: "greeting=Hello%2C world!; code: 42"
```
lornajane: Why is this example `code: 42` at the end and not `code=42`?

src/oas.md
```diff
@@ -1297,6 +1301,54 @@ examples:
...
+```
+
+A cookie parameter relying on the percent-encodingn behavior of the default `style: "form"`:
```
lornajane:
```suggestion
A cookie parameter relying on the percent-encoding behavior of the default `style: "form"`:
```

src/oas.md
```diff
@@ -1297,6 +1301,54 @@ examples:
...
+  Object:
+    description: |
+        Note that the comma (,) has been pre-percent-encoded
+        to "%2C" in the data, as it is forbidden in
+        cookie values.  However, the exclamation point (!)
+        is legal in cookies, so it can be left unencoded.
+    dataValue: {
+      "greeting": "Hello%2C world!",
+      "code": 42
+    }
```
karenetheridge, line 1334: Mixing json into yaml is confusing...

```suggestion
    dataValue:
      greeting: Hello%2C world!
      code: 42
```

src/oas.md
```diff
@@ -1297,6 +1301,54 @@ examples:
...
+    serializedValue: "greeting=Hello%2C world!; code: 42"
+```
+
+A cookie parameter relying on the percent-encodingn behavior of the default `style: "form"`:
+
+```yaml
+name: greeting
+in: cookie
+schema:
```
baywet, line 1343: would it be worth updating this example to match the schema of the previous one? I think the point was to compare encodings and having a different "structure" gets in the way of that IMHO

src/oas.md
```diff
@@ -5127,22 +5179,14 @@ For this reason, any data being passed to a header by way of a [Parameter](#para
...
+The behavior of `style: "cookie"` assumes this usage, and _does not_ apply or remove percent-encoding.
 
+If automatic percent-encoding is desired, `style: "form"` with a primitive value or with the non-default `explode` value of `false` provides this behavior.
+However, note that the default value of `explode: true` for `style: "form"` with non-primitive values uses the wrong delimiter for cookies (`&` instead of `;` followed by a single space) to set multiple cookie values.
 Using `style: "form"` with `in: "cookie"` via an RFC6570 implementation requires stripping the `?` prefix, as when producing `application/x-www-form-urlencoded` message bodies.
-
-For multiple values, `style: "form"` is always incorrect, even if no characters are subject to percent-encoding, as name=value pairs in cookies are delimited by a semicolon followed by a space character rather than `&`.
+To allow the full use of `style: "form"` with `in: "cookie"`, the `allowReserved` field is now supported for cookies.
```
lornajane:
```suggestion
To allow the full use of `style: "form"` with `in: "cookie"`, use the `allowReserved` field.
```

(it doesn't need to have a time element, it just is)

src/oas.md
handrews: It's extremely complicated- read appendix C.  TL;DR: Sometimes they don't do waht we need because we munged RFC6570 with other things.

src/oas.md
handrews: Same as https://github.com/OAI/OpenAPI-Specification/pull/4903#discussion_r2291805318 - let's get these merged and then you're free to change them all at once if you'd like.

src/oas.md
handrews: Yeah I'll add that.

src/oas.md
handrews: Error on my part, thanks for catching!

src/oas.md
handrews: I'd forgotten about the overlap here as I added the percent-encoding part to the previous example at the last minute.  I agree, I'll make them fit together better.

handrews (commented):
I'll fix the remaining things soon- I think I have either accepted the suggestions or otherwise replied to all comments for now.

src/oas.md
handrews: OK, I said yes, but this matches how all of the other examples are done and I don't have it in me to go change them all.  If someone wants to do it as a follow-on that would be great, I'm going to leave it like the others here.

src/oas.md
handrews: @baywet this is more complicated than it seems, I'll follow up with you separately but can we leave it for this PR?  We still have to restructure the top-level sections so there's time for a follow-on for this sort of thing.

handrews:
I have updated most of the things.  Due to the state of my health and general exhaustion, I'd like to plead that we focus only on accuracy and not what examples or formatting might be nice to have.

@lornajane – I changed "simulating" to "simulating/extending" which might make it a bit more clear why it's not just useing?  idk.  For the extra level on the examples, I'd happily approve a follow-on PR doing that to all of the examples in this section (although see next point below).

@baywet – I'll follow up with you on slack, but the example request is a bit more complicated.  I'll happy approve a follow-on PR from you, and answer any questions you might have while making it.

@karenetheridge – I will gladly approve a PR removing the JSON-like formatting from `dataValue` throughout the entire spec.  It's a convention that makes sense to me, but... I'm a werid data point and if no one else finds it compelling and you find it confusing, let's toss it.  But for the whole spec at once.

src/oas.md
```diff
@@ -5127,22 +5184,14 @@ For this reason, any data being passed to a header by way of a [Parameter](#para
...
-
-[RFC6265](https://www.rfc-editor.org/rfc/rfc6265) recommends (but does not strictly required) base64 encoding (`contentEncoding: "base64"`) if "arbitrary data" will be stored in a cookie.
-Note that the standard base64-encoding alphabet includes non-URL-safe characters that are percent-encoded by RFC6570 expansion; serializing values through both encodings is NOT RECOMMENDED.
-While `contentEncoding` also supports the `base64url` encoding, which is URL-safe, the header and cookie RFCs do not mention this encoding.
+While percent-encoding seems more common as an escaping mechanism than the base64 encoding (`contentEncoding`: "base64") recommended by [[RFC6265]], [section 5.6 of draft-ietf-httpbis-rfc6265bis-20](https://www.ietf.org/archive/id/draft-ietf-httpbis-rfc6265bis-20.html#section-5.6), the proposed update to that RFC, notes that cookies sent in the `Set-Cookie` response header that appear to be percent-encoded MUST NOT be decoded when stored by the client, which would mean that they are already encoded when retrieved from that storage for use in the `Cookie` request header.
```
```suggestion
While percent-encoding seems more common as an escaping mechanism than the base64 encoding (`contentEncoding`: "base64") recommended by [[RFC6265]], [section 5.6 of draft-ietf-httpbis-rfc6265bis-20](https://www.ietf.org/archive/id/draft-ietf-httpbis-rfc6265bis-20.html#section-5.6), the proposed update to that RFC notes that cookies sent in the `Set-Cookie` response header that appear to be percent-encoded MUST NOT be decoded when stored by the client, which would mean that they are already encoded when retrieved from that storage for use in the `Cookie` request header.
```

src/oas.md
baywet: what I was trying to express there is that when looking at those two examples in vacuum, I think the point is to demonstrate the differences in style, and how it impacts the serialized representation.
If that is the case (maybe I misunderstood something), I believe this goal would be better achieved by making sure the schema is aligned between the two examples (object with properties), this way the difference in the serialized representation would stand out.
If outlining the difference in data representation is not the goal of those two examples, and/or they are independent or each other, my comment can be disregarded.

handrews: @baywet An object with properties _just doesn't work_ with `style: form, explode: true`.  That's part of the point of this whole thing.  You get syntactically invalid cookies that way.

baywet: right... thanks for pointing this out. I was too focused on "comparing the two examples".
And adding a note in there saying the example is different because that "other structure doesn't work" wouldn't add much value in itself.
Thanks for clarifying! Resolving!

baywet: (well, I can't resolve, but please go ahead)

PR #4905, opened 2025-08-21
v3.2: Update RFC1866 to WHATWG-URL for application/x-www-form-urlencoded

handrews:
This updates our reference for `application/x-www-form-urlencoded`.  The question of whether it is strictly compatible is very hard to answer, as RFC1866 paraphrases RFC1738 _incorrectly_, and implementations that I can find that use these don't behave quite like either the paraphrase or the actual requirements.  Which are hard to figure out in the first place.

Given that we are defining percent-encoding in terms of whether you can successfully _decode_ the encoded string, this is compatible.  If we think in terms of exactly what gets encoded, it's not, but also it's probably never been implemented "correctly" because I've never seen a library that percent-encoded _every_ non-alphanumeric character (RFC1866's paraphrase), and the most strict reading of RFC1738 is problematic in terms of its successor, RFC3986, although all in weird corners that would take several more paragraphs to explain.

I think this update is an improvement, as it brings us in line with the current IANA-registered specification for `form-urlencoded`, and is less of a nightmare to try to understand by reading the source specifications.

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -5208,7 +5208,7 @@ For use in URIs, each part is encoded based on the media type (e.g. `text/plain`
 #### Interoperability with Historical Specifications
 
 Prior versions of this specification required [[?RFC1866]] and its use of [[?RFC1738]] percent-encoding rules in place of [[WHATWG-URL]].
-The [[WHATWG-URL]] `form-urlencoded` rules represent the current browser consensus on that media type, and avoid the ambiguity introduce by unclear paraphrasing of RFC1738 in RFC1866.
+The [[WHATWG-URL]] `form-urlencoded` rules represent the current browser consensus on that media type, and avoid the ambiguity introduced by unclear paraphrasing of RFC1738 in RFC1866.
 
 Users needing conformance with RFC1866/RFC1738 are advised to check their tooling and library behavior carefully.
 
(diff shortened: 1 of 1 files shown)


ralfhandl (dismissed):

src/oas.md
```diff
@@ -5196,29 +5195,30 @@ This specification normatively cites the following relevant standards:
...
+For use in URIs, each part is encoded based on the media type (e.g. `text/plain` or `application/json`), and must then be percent-encoded for use in a `form-urlencoded` string (in form-style query strings), or for general URI use in other URL components, unless the media type already incorporates URI percent-encoding.
 
 #### Interoperability with Historical Specifications
 
-In most cases, generating query strings in strict compliance with [[RFC3986]] is sufficient to pass validation (including JSON Schema's `format: "uri"` and `format: "uri-reference"` when `format` validation is enabled), but some `form-urlencoded` implementations still expect the slightly more restrictive [[RFC1738]] rules to be used.
+Prior versions of this specification required [[?RFC1866]] and its use of [[?RFC1738]] percent-encoding rules in place of [[WHATWG-URL]].
+The [[WHATWG-URL]] `form-urlencoded` rules represent the current browser consensus on that media type, and avoid the ambiguity introduce by unclear paraphrasing of RFC1738 in RFC1866.
```
mikekistler:
```suggestion
The [[WHATWG-URL]] `form-urlencoded` rules represent the current browser consensus on that media type, and avoid the ambiguity introduced by unclear paraphrasing of RFC1738 in RFC1866.
```

mikekistler (dismissed):
Looks good. 👍

PR #4907, opened 2025-08-24
v3.1.2 Backport PR #4899 Encoding style default behavior

handrews:
See PR #4899.

Technically no schema changes are _needed_ (validation works correctly as-is), but they are _possible_ to more clearly specify the default.


miqui (approved):

src/oas.md
```diff
@@ -1715,7 +1715,7 @@ See [Appendix B](#appendix-b-data-type-conversion) for a discussion of data type
...
+| <a name="encoding-style"></a>style | `string` | Describes how a specific property value will be serialized depending on its type. See [Parameter Object](#parameter-object) for details on the [`style`](#parameter-style) field. The behavior follows the same values as `query` parameters, including the default value of `"form"` which applies if either `explode` or `allowReserved` are explicitly specified. Note that the initial `?` used in query strings is not used in `application/x-www-form-urlencoded` message bodies, and MUST be removed (if using an RFC6570 implementation) or simply not added (if constructing the string manually). This field SHALL be ignored if the request body media type is not `application/x-www-form-urlencoded` or `multipart/form-data`. If a value is explicitly defined, then the value of [`contentType`](#encoding-content-type) (implicit or explicit) SHALL be ignored. |
```
mikekistler: I apologize for not asking this on the original PR to 3.2, but does this mean that the default value of "form" _does not_ apply if neither `explode` or `allowReserved` are explicitly specified, and if so, then what is the default value in that case? Or maybe this means that the `style` field only applies if either `explode` or `allowReserved` are explicitly specified?

handrews: @mikekistler 

> Or maybe this means that the style field only applies if either explode or allowReserved are explicitly specified?

This one.  LMK if you have better wording and we can improve here and re-port to 3.2.  I think perhaps (wrapped for readability):

```markdown
The behavior follows the same values as `query` parameters,
including the default value of `"form"`. This default applies
if either `explode` or `allowReserved` are explicitly specified.
```

mikekistler: Okay ... I reread the description on #4899 and I understand now how this works. Let me try:

> The behavior follows the same values as `query` parameters, including the default value of `"form"` which applies if either `explode` or `allowReserved` are explicitly specified. When all of `style`, `explode`, and `allowReserved` are absent, these fields are ignored as the behavior is determined by the `contentType` field or its default value if absent.

This is clear to me. Is it too wordy?

handrews: @mikekistler I reworded that a bit in a way that I think condenses it while still getting the point across.  Let me know if I succeeded :-)

handrews:
I'll re-port the changes to 3.2 as part of the section reorganization (separate commits, maybe separate PRs, still working on how to push all of that through, but I don't want to open another 3.2 PR right now b/c that will further delay the reorg).

Issue #4831, opened 2025-08-12
Restructure sections before publication


lornajane:
Could we shuffle some of the sections in the specification? One thing I find frustrating in 3.1.1 is how difficult it is to find the actual fields in the specification (https://spec.openapis.org/oas/v3.1.1.html#openapi-object). All the information we have is needed, but I think the OpenAPI document structure is a logical entry point for many users and it's an extreme version of "below the fold" at the moment.

lornajane:
Ping @handrews and @ralfhandl as being my most likely collaborators on this one, but all comments and suggestions are welcome!

ralfhandl:
What would you like to reshuffle?

I could see sections "4.5 Rich Text Formatting" and sections 4.6 and 4.7 on relative references moved after section "4.8 Schema", and section 4.8 could get a more catchy headline. Unfortunately the natural headline "OpenAPI Description Structure" is already used for describing how to split a Description into Documents.

handrews:
@ralfhandl I think you are working from an outdated structure.  I rearranged the "OpenAPI Description Structure" to encompass all of the mechanics of connecting documents, including URIs (and the related topic of resolving API URLs since they both ultimately can resolve relative to the document).

@lornajane I am skeptical of moving anything that's before §4.7 Schema to after it.  The structure and basics of working with data are prerequisites for implementation.  I'm particularly opposed to burying the reference resolution rules.  People looking for document structure should be directed to the learn site, which should be oriented towards users rather than implementors.

I _do_ think that "Schema" is a horribly confusing name for that section.  I could definitely be convinced to change it (and "OpenAPI Description Structure") to something that works better for each section.  "OpenAPI Description Structure" is not great for what it's doing right now either, TBH.

Here's the current TOC:

```
# OpenAPI Specification
## Version 3.2.0

## Introduction

## Definitions
### OpenAPI Description

### OpenAPI Document

### Schema

### Object

### Path Templating

### Media Types
#### Sequential Media Types
#### Media Type Registry

### HTTP Status Codes

### Case Sensitivity

### Undefined and Implementation-Defined Behavior

## Specification

### Versions

### Deprecation

### Format
#### JSON and YAML Compatibility

### OpenAPI Description Structure
#### Parsing Documents
#### Structural Interoperability
#### Relative References in API Description URIs
##### Establishing the Base URI
##### Resolving URI fragments
##### Relative URI References in CommonMark Fields
#### Relative References in API URLs
[...]

PR #4908, opened 2025-08-26
v3.2: Reorganize sections

handrews:
This is the first step in addressing #4831 (with slight differences from the re-org last posted in that issue).  It _only_ moves sections and adjusts indents, without _any_ other changes to contents or to section names.  All of the more complex changes, including renaming, combining, or splitting sections, will be in subsequent PR(s).

GitHub does not use git's feature to detect moved code in diffs, so ***I do not recommend looking at the normal diff***.  I have provided multiple other ways to look at this.

To prove that all non-heading lines were only moved, you can see screenshots of `git diff --color-moved=dimmed-zebra` [here](https://github.com/handrews/handrews.github.io/tree/main/renderings/oas/diffs).  These are not intended for detailed review, they just show by color what is a move vs a change.

Even in those diffs it's still hard to see what was moved, so I made alternate views of the two commits, one of which does the move and the other of which adjusts heading levels.  I recommend looking at the headings-only diffs in particular:

* Moving sections with minimal level changes
    * [github diff](https://github.com/OAI/OpenAPI-Specification/commit/fc77a7302b536312864d7d01f9703ff5068aec0b)
    * [headings-only diff](https://gist.github.com/handrews/a29cf421b8cf8e87838c4f6595df3b5e#file-oas-reorg-1-sdiff)
* Fixing the heading levels for their new locations
    * [github diff](https://github.com/OAI/OpenAPI-Specification/commit/fb45114032e88d2155d38516c0e7b90690c96ac5)
    * [headings-only diff](https://gist.github.com/handrews/a29cf421b8cf8e87838c4f6595df3b5e#file-oas-reorg-2-sdiff)

Finally, here is the rendered outcome: [OAS 3.2](https://handrews.github.io/renderings/oas/deploy-preview/oas.html)

-----

Two sections, "Object" and "Schema" (both under "Definitions", not the big "Schema" section obviously) have been entirely removed, as the new section names (renamed in a future PR) and organization makes them unnecessary.

Several additional section headings have been removed because they only had subsections (no other content) and the subsections no longer need the grouping: "Specification", "Definitions", "Working With Data".

The resulting top-level sections are:

* Introduction
* Format
* Schema (to be renamed in the next PR)
* Specification Extensions
* Security Considerations
* (appendixes, which I'm not listing as they have not changed in this PR)

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -43,7 +43,7 @@ Behavior described as _implementation-defined_ allows implementations to choose 
 This documents ambiguous requirements that API description authors are RECOMMENDED to avoid in order to maximize interoperability.
 Unlike undefined behavior, it is safe to rely on implementation-defined behavior if _and only if_ it can be guaranteed that all relevant tools support the same behavior.
 
-### Format
+## Format
 
 An OpenAPI Document that conforms to the OpenAPI Specification is itself a JSON object, which may be represented either in [[RFC8259|JSON]] or [[YAML|YAML]] format.
 Examples in this specification will be shown in YAML for brevity.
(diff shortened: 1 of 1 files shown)


ralfhandl (approved):
Thanks for the detailed description of the changes and deletions.

mikekistler (approved):
Looks good. 👍

PR #4909, opened 2025-08-28
OAD schema fix 'parameter object' + in: querystring forbidden fields

notEthan:
https://github.com/OAI/OpenAPI-Specification/pull/4590 introduces parameter object with `in: querystring`, which may not have properties `schema`, `style`, `explode`, or `allowReserved`. The accompanying schema change only forbids an object with _all_ of those four properties, though. This fixes to forbid any of them.

### DIFF
--- a/src/schemas/validation/schema.yaml
+++ b/src/schemas/validation/schema.yaml
@@ -392,12 +392,6 @@ $defs:
         then:
           required:
             - content
-          not:
-            required:
-              - schema
-              - style
-              - explode
-              - allowReserved
     dependentSchemas:
       schema:
         properties:
--- a/tests/schema/fail/parameter-object-content-not-with-style.yaml
+++ b/tests/schema/fail/parameter-object-content-not-with-style.yaml
@@ -1,0 +1,14 @@
+openapi: 3.2.0
+info:
+  title: API
+  version: 1.0.0
+components:
+  parameters:
+    content-not-with-style:
+      in: querystring
+      name: json
+      content:
+        application/json:
+          schema:
+            type: object
+      style: simple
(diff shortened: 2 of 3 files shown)


ralfhandl (changes requested):
I think this change is unnecessary. Adding this `fail` test case with the current schema fails as expected:

~~~yaml
openapi: 3.2.0
info:
  title: API
  version: 1.0.0
components:
  parameters:
    querystring-with-schema-not-allowed:
      in: querystring
      name: json
      content:
        application/json:
          schema:
            type: object
      explode: true
~~~

It also fails with a single one of the other three keywords not allowed with `content`.

Side note: `schema` is not necessary in the not-required list because having both `content` and `schema` is already forbidden by the oneOf-required construct in lines 367-371.

I prefer the current style as it is shorter.

handrews:
@notEthan @ralfhandl actually the entire `not:` clause should be removed, and only `required: [content]` should be kept.  Other parts of the schema handle the mutual exclusion of `content` and the other fields, which is why the test case passes.  The `not:` here is superfluous, so the fact that it is (as @notEthan notes) incorrect doesn't matter.

notEthan:
That example fails due to `unevaluatedProperties: false` in `/$defs/parameter` applying to `explode`. Which is fine and may be sufficient - if so the `not`+`required` subschema this PR changes could instead be removed entirely. But as it is, that part is not doing what it was intended to do, assuming it's intended to forbid any of those four properties. `required` will pass only if all its listed properties are present, and `not`+`required` will only fail if all listed properties are present.

notEthan:
oh, @handrews beat me to it while I was typing

notEthan:
updated to remove this subschema

ralfhandl:
@notEthan Please include the test cases from
* #4910

or equivalents to prove that removing these lines does not break the mutual exclusion.

Or let's wait with this one until #4910 is merged.

Thanks in advance!

notEthan:
rebased against that PR's branch

ralfhandl (approved):

handrews (approved):

Issue #4578, opened 2025-05-14
Registry: ...


PannawatC:
**Which registry do you want to contribute to**
- [x] [Alternative Schema Type Registry](https://spec.openapis.org/registry/alternative-schema)
- [ ] [Draft Features Registry](https://spec.openapis.org/registry/draft-feature)
- [ ] [Specification Extension Registry](https://spec.openapis.org/registry/extension)
- [ ] [Format Registry](https://spec.openapis.org/registry/format)
- [ ] [Extension Namespace Registry](https://spec.openapis.org/registry/namespace)
- [ ] [Tag Kind Registry](https://spec.openapis.org/registry/tag-kind)

**Describe your contribution**
A clear and concise description of what you want to add or change.

**Describe alternatives you've considered**

**Additional context**

ralfhandl:
@PannawatC Closing this as it seems to be an accident.

If you want to open an issue, please fill out all sections.

Thanks in advance!

PR #4912, opened 2025-08-31
"example" and "examples" cannot appear together

karenetheridge:
This affects the places where examples are used: parameter, header, and media-type objects

for #4578 and #4776.


handrews (approved):
Weird- this is handled correctly in the 3.0 schema, but was apparently dropped in 3.1?  Which was definitely an error.

karenetheridge:
This should probably be backported to 3.1.2 then?

handrews:
It should be backported to 3.1.  There are no patch-release-specific schemas.

ralfhandl (changes requested):
Could you please add a "fail" test case that would pass without the additional lines?

PR #4914, opened 2025-08-31
fix: header and parameter objects cannot have examples with "content"

karenetheridge:
The examples instead belong with the media-type object found below.

Also contains a few small no-op optimizations.

### DIFF
--- a/src/schemas/validation/schema.yaml
+++ b/src/schemas/validation/schema.yaml
@@ -369,8 +369,8 @@ $defs:
         - schema
       - required:
         - content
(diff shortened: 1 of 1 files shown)


karenetheridge:
~~This should also be backported to 3.1.2.~~

karenetheridge:
This PR looks to be in error. This was changed in commit 042e70884, which looked like the intent was to explicitly add examples for `content` as well as `schema`, but I've been referencing the 3.1.1 spec as there is no rendered version available of what's in dev for easy querying.

PR #4917, opened 2025-08-31
v3.1: "example" and "examples" cannot appear together

karenetheridge:
This affects the places where examples are used: parameter, header, and media-type objects

for #4598 and #4776, ported from #4912.

### DIFF
--- a/tests/schema/fail/example-examples.yaml
+++ b/tests/schema/fail/example-examples.yaml
@@ -1,0 +1,20 @@
+openapi: 3.1.1
+
+# this example should fail, as example cannot be used together with examples.
+
+info:
+  title: API
+  version: 1.0.0
+components:
+  parameters:
+    animal:
+      name: animal
+      in: header
+      schema: {}
+      example: bear
+      examples:
+        a mammalian example:
+          value: bear
+
+
+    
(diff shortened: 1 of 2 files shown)


handrews (approved):
Thanks!  This is unrelated to the "overriding" thing, btw, which was about either of these fields vs the `example` field in the Schema Object.

ralfhandl (changes requested):
Could you please add a "fail" test case that would pass without the additional lines?

tests/schema/fail/example-examples.yaml
```diff
@@ -0,0 +1,20 @@
...
+    animal:
+      name: animal
+      in: header
+      schema: {}
+      example: bear
+      examples:
+        a mammalian example:
+          dataValue: bear
+
+
```
ralfhandl:
```suggestion
          value: bear
```

mikekistler (dismissed):
+1 to @ralfhandl 's suggestion. the rest looks good.

Approving now so it can merge once Ralf's comment is addressed.

karenetheridge:
ahyes, `dataValue` is new :)

ralfhandl (approved):

ralfhandl:
> Approving now so it can merge once Ralf's comment is addressed.

@mikekistler Unfortunately you have to reapprove.

PR #4926, opened 2025-09-07
uri resolution notes

karenetheridge:
Not intended for merging in this form.

Notes on my initial thoughts on how each uri-reference in the OAD should resolve.

See https://github.com/OAI/OpenAPI-Specification/discussions/4925#discussioncomment-14329850.

### DIFF
--- a/src/schemas/validation/schema.yaml
+++ b/src/schemas/validation/schema.yaml
@@ -10,6 +10,7 @@ properties:
     pattern: '^3\.2\.\d+(-.+)?$'
   $self:
     type: string
+    $comment: resolved against the retrieval uri
     format: uri-reference
     $comment: MUST NOT contain a fragment
     pattern: '^[^#]*$'
@@ -17,9 +18,11 @@ properties:
     $ref: '#/$defs/info'
   jsonSchemaDialect:
     type: string
+    $comment: resolved against the resolved version of $self
     format: uri-reference
     default: 'https://spec.openapis.org/oas/3.2/dialect/WORK-IN-PROGRESS'
   servers:
+    $comment: server urls are resolved against the HTTP request uri itself (template matching ideally happens first, because of url encoding, but matching the template parts separately may produce ambiguous results, so in my implementation I concatenate the server url template with the path template and then match that against the entire uri)
     type: array
     items:
       $ref: '#/$defs/server'
@@ -68,6 +71,7 @@ $defs:
       description:
         type: string
       termsOfService:
+        $comment: resolved against the resolved version of $self
         type: string
         format: uri-reference
       contact:
@@ -89,6 +93,7 @@ $defs:
       name:
         type: string
       url:
+        $comment: resolved against the resolved version of $self
         type: string
         format: uri-reference
       email:
@@ -106,6 +111,7 @@ $defs:
       identifier:
         type: string
       url:
+        $comment: resolved against the resolved version of $self
         type: string
         format: uri-reference
     required:
@@ -225,6 +231,7 @@ $defs:
     type: object
     properties:
       $ref:
+        $comment: resolved against the resolved version of $self
         type: string
         format: uri-reference
       summary:
@@ -327,6 +334,7 @@ $defs:
       description:
         type: string
       url:
+        $comment: resolved against ???
         type: string
         format: uri-reference
     required:
@@ -682,6 +690,7 @@ $defs:
         type: string
       value: true
       externalValue:
+        $comment: resolved against the resolved version of $self
         type: string
         format: uri-reference
     allOf:
@@ -719,6 +728,7 @@ $defs:
     type: object
     properties:
       operationRef:
+        $comment: resolved against the resolved version of $self
         type: string
         format: uri-reference
       operationId:
@@ -825,6 +835,7 @@ $defs:
     type: object
     properties:
       $ref:
+        $comment: resolved against the resolved version of $self
         type: string
         format: uri-reference
       summary:
@@ -929,6 +940,7 @@ $defs:
             flows:
               $ref: '#/$defs/oauth-flows'
             oauth2MetadataUrl:
+              $comment: resolved against the matching and de-templated server url
               type: string
               format: uri-reference
           required:
@@ -944,6 +956,7 @@ $defs:
         then:
           properties:
             openIdConnectUrl:
+              $comment: resolved against the matching and de-templated server url
               type: string
               format: uri-reference
           required:
@@ -977,6 +990,7 @@ $defs:
 
     $defs:
       implicit:
+        $comment: references are resolved against the matching and de-templated server url
         type: object
         properties:
           authorizationUrl:
@@ -994,6 +1008,7 @@ $defs:
         unevaluatedProperties: false
 
       password:
+        $comment: references are resolved against the matching and de-templated server url
         type: object
         properties:
           tokenUrl:
@@ -1011,6 +1026,7 @@ $defs:
         unevaluatedProperties: false
 
       client-credentials:
+        $comment: references are resolved against the matching and de-templated server url
         type: object
         properties:
           tokenUrl:
@@ -1028,6 +1044,7 @@ $defs:
         unevaluatedProperties: false
 
       authorization-code:
+        $comment: references are resolved against the matching and de-templated server url
         type: object
         properties:
           authorizationUrl:
@@ -1049,6 +1066,7 @@ $defs:
         unevaluatedProperties: false
 
       device-authorization:
+        $comment: references are resolved against the matching and de-templated server url
         type: object
         properties:
           deviceAuthorizationUrl:


baywet:
Sharing some notes from the meeting: This is a working document from @karenetheridge with her understanding of the spec today. If some of that is "wrong" it probably means that we need to make the specification clearer. @handrews to provide some feedback to help get to a common understanding and identify the potential for improvements in the specification.

src/schemas/validation/schema.yaml
```diff
@@ -10,16 +10,19 @@ properties:
     pattern: '^3\.2\.\d+(-.+)?$'
   $self:
     type: string
+    $comment: resolved against the retrieval uri
```
handrews, line 13: This is one of several spots where I would not use "retrieval URI."  I tried to make this more clear in 3.2, so it would be good to understand what is not coming across.  I would probably use a phrase like "against the document's base URI," although I don't think that that phrase is used prominently in 3.2.  Maybe that is part of the lack of clarity?

Here's the wording for `$self`:

> This string MUST be in the form of a URI reference as defined by [[RFC3986](https://spec.openapis.org/oas/v3.2.0#bib-rfc3986)] [Section 4.1](https://tools.ietf.org/html/rfc3986#section-4.1). The `$self` field provides the self-assigned URI of this document, which also serves as its base URI in accordance with [[RFC3986](https://spec.openapis.org/oas/v3.2.0#bib-rfc3986)] [Section 5.1.1](https://tools.ietf.org/html/rfc3986#section-5.1.1). Implementations MUST support identifying the targets of [API description URIs](https://spec.openapis.org/oas/v3.2.0#relative-references-in-api-description-uris) using the URI defined by this field when it is present. See [Establishing the Base URI](https://spec.openapis.org/oas/v3.2.0#establishing-the-base-uri) for the base URI behavior when `$self` is absent or relative, and see [Appendix F](https://spec.openapis.org/oas/v3.2.0#appendix-f-examples-of-base-uri-determination-and-reference-resolution) for examples of using $self to resolve references.

Perhaps we could change "See [Establishing the Base URI](https://spec.openapis.org/oas/v3.2.0#establishing-the-base-uri) for the base URI behavior when $self is absent or relative" to "When `$self` is relative, it is resolved against the [document's base URI](https://spec.openapis.org/oas/v3.2.0#establishing-the-base-uri)"?

In section [4.1.2.2.1 Establishing the Base URI](https://spec.openapis.org/oas/v3.2.0#establishing-the-base-uri), the third paragraph starts:

> The most common base URI source that is used in the event of a missing or relative $self (in the [OpenAPI Object](https://spec.openapis.org/oas/v3.2.0#openapi-object)) and (for [Schema Object](https://spec.openapis.org/oas/v3.2.0#schema-object)) $id is the retrieval URI.

My intention with that phrasing was both to note the common behavior and to emphasize that the base URI is _not always_ the retrieval URI.  Does it seem clear enough if the `$self` wording is changed, or is it still unclear?

src/schemas/validation/schema.yaml
```diff
@@ -10,16 +10,19 @@ properties:
...
     pattern: '^[^#]*$'
   info:
     $ref: '#/$defs/info'
   jsonSchemaDialect:
     type: string
+    $comment: resolved against the resolved version of $self
     format: uri-reference
     default: 'https://spec.openapis.org/oas/3.2/dialect/WORK-IN-PROGRESS'
   servers:
+    $comment: server urls are resolved against the HTTP request uri itself (template matching ideally happens first, because of url encoding, but matching the template parts separately may produce ambiguous results, so in my implementation I concatenate the server url template with the path template and then match that against the entire uri)
```
handrews, line 25: Is this "HTTP request uri" the request URI of the OAD document?  This is a place where our URI/URL distinction can get a bit muddled.

In this case, I would call this a URL matters because the document's identity (URI) might be set (by `$self`) to something other than the document's location (which is generally it's retrieval URL).

The tricky part here is that if we are _simulating_ the document's retrieval URL (e.g. because you are testing things and don't want to actually deploy to production, but do want to test as if it were deployed), then it's arguably being a bit more URI-ish than URL-ish? But I still lean towards calling that a URL because the whole point is that you are simulating the _location_, and want it to _behave as a location_.  You are not simulating it to separate identity from location (which is what `$self` does).  You are simulating it to get location-based (not identity-based) behavior.

This simulation use case is discussed in the third paragraph of  [4.1.2.2.1 Establishing the Base URI](https://spec.openapis.org/oas/v3.2.0#establishing-the-base-uri), although that's not linked from the Server Object so maybe that's a concern.

Does all of this make sense, and if so do you have any ideas on how we can clarify it?  I moved the "Relative References in API URLs" section under the Server Object, but might have accidentally dropped explanatory text elsehwere in the process.

src/schemas/validation/schema.yaml
```diff
@@ -68,6 +71,7 @@ $defs:
       description:
         type: string
       termsOfService:
+        $comment: resolved against the resolved version of $self
```
handrews, line 74: FWIW, I had to think about whether Info/License/Contact should be OAD URI-based or API URL-based.  Taking a look at these fields, we do say "URI", and that's what I was expecting to find.  I wonder if we should make "URI" a link to 4.1.2.2 Relative References in API Description URIs for added clarity?

src/schemas/validation/schema.yaml
```diff
@@ -929,6 +940,7 @@ $defs:
             flows:
               $ref: '#/$defs/oauth-flows'
             oauth2MetadataUrl:
+              $comment: resolved against the matching and de-templated server url
```
handrews, line 943: There's an interesting wrinkle here that this is a URL (including possibly a relative URL-reference), but the resolved server URL is expected to be used as a prefix (with resolved Path Templates) rather than a normal base URL.  I'm not sure how much this matters as both behaviors are well-defined, but it means that, given a server URL of `https://example.com/api`,  a Path Template of `/foo`, and a security URL of `/foo`, the resulting API endpoint is `https://example.com/api/foo` but the resulting security URL is `https://example.com/foo` (no `/api`).

Which leads to another observation: The Paths Object requires Path Templates to start with `/`, but the Server Object does _not_ forbid URLs from ending with `/`.  I'm not quite sure what to do with that 😵‍💫

src/schemas/validation/schema.yaml
karenetheridge: > the base URI is not always the retrieval URI

I'm not clear on the difference. Isn't the retrieval URI already defined as "the initial URI we assign to the document when we start reading from it"?  I recall we used the retrieval URI phrasing even if the document is never actually reachable as an URL on the network, but just a thing in application state.

It sounds like we need to come up with a clearer name for whatever this thing is, and use it throughout.

src/schemas/validation/schema.yaml
karenetheridge: > Is this "HTTP request uri" the request URI of the OAD document? 

No, whenever I refer to an HTTP request it is in the server context -- it is a request for one of the APIs described by the document (or perhaps not matching any API, but there is still an attempt to match one of the path-items to it).

When I try to match server urls to the HTTP request, I resolve the url against the retrieval uri (not $self), and then again against the HTTP request's URI, if the retrieval URI was relative.  (We _have_ to have a URI with host and scheme in order to match it against the HTTP request URI, and having a relative retrieval URI would prevent that.

I believe this is correct?

src/schemas/validation/schema.yaml
karenetheridge: > ... but the resulting security URL is https://example.com/foo (no /api)

Yes, that was my observation as well. I assume that this is the best we can do, and it doesn't really make sense for the `api` portion to appear in the resolved form of the server url we use for matching.

It's even more clear when we change the retrieval URL to ".../api.json". That's clearly a filename, not a directory, and we don't expect to see that in any request URLs that we're trying to match.

> The Paths Object requires Path Templates to start with /, but the Server Object does not forbid URLs from ending with /. I'm not quite sure what to do with that.

Yes. I played around with trying to _resolve_ the path template against the server url, instead of appending, but the results were less favourable.

We could say something like "the templated URI resulting from the concatenation of the server url and path template portions should then be normalized", which would also allow for constructs like `../` to be removed, but I feel like there would be a lot of edge cases and surprises here in practice.

The alternative is: you end up with a uri template that contains `//`, which is likely not ever going to match. So don't do that.

src/schemas/validation/schema.yaml
handrews: @karenetheridge the clear name is "base URI."  The less-clear part is that there are several possible base URIs that are searched in order, and if any of them are relative, then they are resolved against the _next_ possible base URI.  Since `$self` is the first place to search, it both _is_ a base URI, and _potentially needs to be resolved against the next base URI in the list._

But all of this is about base URIs.  It is not about retrieval URLs/URIs, except by coincidence.  They happen to be _one of four_ possible base URI sources (and _one of three_ after `$self`).   But it's always about base URIs, and we should always talk in terms of base URIs.

This is for OAD URIs, though, not API URLs, where the situation is different.

src/schemas/validation/schema.yaml
handrews: @karenetheridge I think I am confused.  How can you resolve a relative server URL against an API request, when you need the server URL to make an API request?  That sounds like server-side matching which is, to me, a different concern from URL construction, although they are definitely related.

**I'm glad you are calling attention to server matching.  It had never occurred to me to think through that process, so please bear with me as a try to wrap my head around it.**

By URI/URL construction, I mean _"I'm parsing this OAD and I found some part of an OAD URI or API URL and need to construct the full URI/URL before I can use it."_

By server-side matching, I mean _"I am handling an API request and I am trying to figure out which possible URL construction (as in the previous definition) matches this URL, after which I need to split it back into parts to extract server variables, path template parameters, and handle the query string."_

Most of your comments here seem to be about construction, which is why I am confused.  I think it is important to get the construction parts clear and _separate from the matching guidance._  Because if we aren't constructing correctly, we'll end up attempting to match to an incorrect set of potential URLs.

src/schemas/validation/schema.yaml
handrews: > it doesn't really make sense for the api portion to appear in the resolved form of the server url we use for matching.

In my example, the resolved API endpoint _does_ include `/api` so I am confused as to this statement.  Only the auth endpoint does _not_ include it.  This might be more of me not quite wrapping my head around the server matching use cases.

> We could say something like "the templated URI resulting from the concatenation of the server url and path template portions should then be normalized", which would also allow for constructs like ../ to be removed, but I feel like there would be a lot of edge cases and surprises here in practice.

My sense is that URL normalization is something that does or doesn't happen independently.  I would tend to normalize URLs as much as possible before attempting to compare, athough as you note that could get surprising.  (For those unfamiliar, normalization and comparison is a tricky topic addressed in-depth by [RFC3986 §6](https://www.rfc-editor.org/rfc/rfc3986#section-6).)

> The alternative is: you end up with a uri template that contains //, which is likely not ever going to match. So don't do that.

I want to say that `//` anywhere other than as the authority prefixe (as in `https://example.com`) gets normalized to `/`, but I can't find a reference for that so it might just be a thing that web servers tend to do?  I'm really not sure.

src/schemas/validation/schema.yaml
karenetheridge: Well yes, the key is "which base URI do we use", and that's the thing that needs to be stated clearly. :p

src/schemas/validation/schema.yaml
karenetheridge: > How can you resolve a relative server URL against an API request, when you need the server URL to make an API request?

Those are on different systems though, so we get at the information in a different way.

> Most of your comments here seem to be about construction, which is why I am confused.

I'm confused too, as I'm not sure what has given you that impression.

src/schemas/validation/schema.yaml
handrews: @karenetheridge then I do not understand your question about a clearer name.  "Base URI" is the name.  RFC3986 gives names for each of the four possible sources.  "Retrieval URI" is incorrect except in specific circumstances, being one of the four sources.  So what sort of clearer name are you looking for? It's a multi-step look-up, so there's no one name other than "Base URI".

src/schemas/validation/schema.yaml
handrews: > > How can you resolve a relative server URL against an API request, when you need the server URL to make an API request?
> 
> Those are on different systems though, so we get at the information in a different way.

I don't follow, can you elaborate on this?

Regarding construction, perhaps that is my assumption.  What are you trying to describe here?  Construction, matching, or both?  To me, successful matching requires successful construction first, otherwise you don't know what you are matching.  Does that make sense?  If not, how do you think of it?

Again, I have not thought about this much from the matching side, so I might be getting things really wrong.  I'm trying to understand the perspective so I can understand what your goals are here.

karenetheridge: Constructing the request for the client, and parsing the request on the server, are two totally separate things done by separate systems and generally on different machines.  What I've been talking about regarding resolving the server url against "something" and then matching it plus the path template against the request is all a server-side activity.

The client could construct the request in a number of ways. I have imagined that a client would construct the request based on some static configurations that it has been provided..  basically: "this is the base URL of the server you are talking to, so given the API operation you want to use, find the appropriate server url that matches that (with base URI resolution if needed), append the appropriate path template for this operation to that, and replace the template variables with the values appropriate to this particular request."

But perhaps it constructs the request through its own means and then evaluates the request against the OAD before sending it just to make sure that it's valid, in which case the behaviour would be just like receiving the request on the server side and using the OAD to parse and validate it before further processing.

(Likewise, a server could either construct a response by its own means and then validate it against the OAD, or use the OAD itself as a guide for constructing it, before sending the response back to the client.)

So really, there are just two mechanisms at play, each of which could be running on either a client or server system:
- construct a request URI using information from the path template and server url, providing server and path variables from an outside source, e.g. config files and the application
- parse a request URI against an operation's path template and server url (which might require iterating through a number of path-items if we don't yet know which one matches) to extract its server variables and path template variables (path parameter values).

I've implemented code for the second of these, but not the first.

handrews: @karenetheridge 

> Constructing the request for the client, and parsing the request on the server, are two totally separate things done by separate systems and generally on different machines. 

Ah, now I understand the confusion.  This is not the "construction" I mean.  The "construction" I meant is that _prior to doing anything,_ no matter whether it is on the client or the server, you need to parse the OAD for bits and pieces of API URLs (or URL Templates, by which I mean our weird sort-of-but-not-quite-URI-Template thing split across the Sever Objects and Paths Object) and construct actual full URLs (not relative references or portions to be concatenated) from them.

I agree with everything else you wrote.  So to come back to the original point of confusion:

> server urls are resolved against the HTTP request uri itself (template matching ideally happens first, because of url encoding, but matching the template parts separately may produce ambiguous results, so in my implementation I concatenate the server url template with the path template and then match that against the entire uri)

This I still don't entirely understand.  In your server-side use case where you are parsing a request, I do understand that there is an HTTP request URI in the request.

But my reading of the spec is that you MUST resolve a relative Server Object `url` field against the document's retrieval URI (which may or may not be equal to its base URI), which would preclude using the HTTP requests request URI.

What part of the spec are you reading to get the "resolve against request URI" behavior?  (I am really not as familiar with the Server Object and still might be missing something).

src/schemas/validation/schema.yaml
handrews: @karenetheridge to be clear, I understand and very much sympathize with the desire for there to _be_ a simpler name for all of this.  But having spent a great deal of time with RFC3986 and its predecessors and related specs, I just don't think there's any easy term to use.

It's not unlike the annoying lack of a clear term for "URI with a scheme that MAY have a fragment."  Officially, "URI" is the term for that, but it's also used generically.  RFC3986 is not the greatest for terminology. 😒

src/schemas/validation/schema.yaml
karenetheridge: What happens when the retrieval URI is relative?

> What part of the spec are you reading to get the "resolve against request URI" behavior?

Nothing. But if after resolving the server url against the retrieval URI, and it's still relative, it will never match the request URI. But from a user perspective, it _should_ match. So resolving against the request URI will allow it to do so.

This is the only thing I could think of doing that would make things work, other than requiring the retrieval URI to be absolute (which didn't seem to be an option since sometimes it's not even provided at all).

handrews: The retrieval URI can't be relative, by definition.  It is the URI that _was used, past tense,_ to retrieve the thing.  Even if it is simulated (e.g. you told the tool what it was for an already-in-memory-document rather than the tool actually performing the I/O), then it still has to be a full URI.

There is no way to have a relative retrieval URI.

karenetheridge: > There is no way to have a relative retrieval URI.

I can assure you that if the user doesn't tell me what it is, it defaults to the empty string, which is a relative URI :)

Where in the spec does it say either that providing a retrieval URI is mandatory, or that it must not be relative?

I don't see that I can do anything else here, if there is to be any hope of making any matches at all.  Do you have any suggestions?

handrews: > I can assure you that if the user doesn't tell me what it is, it defaults to the empty string, which is a relative URI :)

RFC3986 does not allow for a default retrieval URI for base URI establishment purposes (RFC 3986 §5.1.3).  Defaulting is handled as an application-specific default base URI, distinct from retrieval, that is discussed in RFC 3986 §5.1.4, and such URIs MUST be non-relative (§4.1.5 uses the term "URI", which in that context means non-relative).

> Where in the spec does it say either that providing a retrieval URI is mandatory, 

OAS v3.2 Section [4.1.2.2.1](https://spec.openapis.org/oas/v3.2.0#establishing-the-base-uri):

> Even if retrieval is supported, it may be impossible due to network configuration or server unavailability (including the server hosting an older version while a new version is in development), or undesirable due to performance impacts. Therefore, all implementations SHOULD allow users to provide documents with their intended retrieval URIs so that references can be resolved as if retrievals were performed.

So it's not mandatory, but it's very strongly recommended.

> or that it must not be relative?

As for a why a retrieval URI cannot be relative, it is inherent in how the term is defined in RFC3986, which comes from this text in [RFC3986 §1.2.2](https://www.rfc-editor.org/rfc/rfc3986#section-1.2.2) which discusses the use of URIs for retrieval:

> When URIs are used within information retrieval systems to identify
   sources of information, the most common form of URI dereference is
   "retrieval": making use of a URI in order to retrieve a
   representation of its associated resource.

Note the use of "URI", as in non-relative, above.

A bit further in that paragraph, we have:

>  Retrieval is
   achieved by a process that might include using the URI as a cache key
   to check for a locally cached representation, resolution of the URI
   to determine an appropriate access mechanism (if any), and
   dereference of the URI for the sake of applying a retrieval
   operation.  

Again, "URI" as in non-relative.

In the next paragraph:

> URI references in information retrieval systems are designed to be
   late-binding: the result of an access is generally determined when it
   is accessed and may vary over time or due to other aspects of the
   interaction.  

This is a bit more obscure, but the "late-binding"-ness here includes the late resolution against a base URI prior to use of the non-relative URI for retrieval.  There are other things that fit there, like URN resolvers that first convert a URN to a URL that can be directly retrieved, but let's not worry about that 😵‍💫 

> I don't see that I can do anything else here, if there is to be any hope of making any matches at all. Do you have any suggestions?

If the document was not retrieved, and the user fails to provide a non-relative retrieval URI, and there is no usable application-specific default base URI ([§5.1.4](https://www.rfc-editor.org/rfc/rfc3986#section-5.1.4), non-relative, so the empty string cannot be used here), then the OAD is not usable.  §5.1.4 addresses this directly:

>    A sender of a representation containing relative references is
   responsible for ensuring that a base URI for those references can be
   established.  Aside from fragment-only references, relative
   references can only be used reliably in situations where the base URI
   is well defined.

So if your only relative references are fragment-only, then you're fine, because literally any base URI (even `about:blank`) will work.  This is more-or-less the use case that §5.1.4 is addressing.  But if you have other sorts of relative references, it's just an error.  Not only can you not do anything about it, it would be a further error to attempt to compensate for such an oversight.

karenetheridge: The process I described above of resolving the server url against the retrieval uri and then again against the HTTP request URI itself (if still relative) is just shorthand for "when attempting to match HTTP requests against path-items, if we don't know the host part of the URI, then just match what we have -- assume that the host part matches".

I believe that performing URI matching on incoming HTTP requests with a variety of hostnames is a valid usecase, and can be accomodated if absolute server urls are relative or not provided, and if the retrieval URI is relative or not provided.  I feel that putting an artificial restriction on this feature is not helpful to the user, and if added now would be a seriously breaking change (suddenly path-items that previously matched requests will no longer be matched).

> §5.1.4 addresses this directly

5.1.4 also says "If none of the conditions described above apply, then the base URI is defined by the context of the application."  The context here is the HTTP request being matched/validated, and there is nothing else I can use.

karenetheridge:
Another reference point -- https://swagger.io/docs/specification/v3_0/api-host-and-base-path/#relative-urls -- says that other URIs in the OAD (e.g. termsOfService, auth) use the server URLs for base URI determination.

handrews:
@karenetheridge that's [correct for 3.0](https://spec.openapis.org/oas/v3.0.4.html#relative-references-in-urls).  But not 3.1 or 3.2.  OAS 3.0.x does not make a distinction between OAD URIs and API URLs.  That was introduced in 3.1.0, and clarfied substantially in 3.1.1.

src/schemas/validation/schema.yaml
handrews: > against the HTTP request URI itself (if still relative)

A request URI cannot be relative, because how would that even work?  You have to know the scheme, host/authority, path, and query to execute the HTTP(S) request.  The only part of the full URI you don't need is the fragment, since it is not sent to the server.

I'm still trying to figure out what you mean here.  Are you talking about something like the [origin form](https://www.rfc-editor.org/rfc/rfc9112.html#section-3.2.1) in the HTTP request line?

```HTTP
GET /where?q=now HTTP/1.1
Host: www.example.org
```

This is not a relative URI, it's just a full URI, `http://www.example.com/where?q=now` or `https://www.example.com/where?q=now` that is written in pieces.  The path and query come from the `request-target` field of the HTTP request line, while the host comes from the mandatory `Host` header.  The scheme (`http` or `https`) comes from the context (as the server, you know whether this came in on port 80 vs some other port, and whether TLS was involved).  Re-assembling this is defined by RFC9112 as it is not a process of relative URI resolution.  It's a process of serialization/parsing of the HTTP protocol.

After you perform the RFC9112 steps, you have a full URI.  There is no relative URI reference plus base URI as RFC3986 defines it.  Just like joining the resolved path template and server template with the OAS: We define that process as concatenation, not RFC3986 reference resolution (although we do use an abbreviated form of RFC3986 relative resolution when the server `url` field itself is relative, which is confusing).

karenetheridge: > A request URI cannot be relative, because how would that even work? You have to know the scheme, host/authority, path, and query to execute the HTTP(S) request. 

Because URIs are not always network-resolveable. Sometimes they're just an identifier.

Practically speaking, URIs _should_ be relative, but what if they're not? I don't think it's practical to outright forbid them, and therefore they need to be handled.

Relative URIs can exist in an HTTP request when the Host header is not provided, and they can exist in OADs when a relative `$id` or `$self` is used, or a relative server url is provided.  Whether or not this practically exists in a real-world web server implementation is not part of my calculation here; I was identifying an ambiguity in how certain scenarios are handled. If the spec ends up saying "this combination of values is unspecified and implementation-defined" that's fine, but I don't think it's correct to flat-out say that they are impossible and therefore don't need to be considered at all.

handrews: > Because URIs are not always network-resolveable. Sometimes they're just an identifier.

But we're talking about HTTP requests being received by a server, right?  If so, then we are exclusively talking about network (specifically HTTP) accessible URLs.  If not, what _are_ we talking about here?

> Relative URIs can exist in an HTTP request when the Host header is not provided

No, per [RFC9112 §3.2](https://www.rfc-editor.org/rfc/rfc9112.html#section-3.2), the Host header is mandatory, and if the request target URI includes an authority the Host header MUST be identical to the host in the authority.

> I don't think it's practical to outright forbid them, and therefore they need to be handled.

I'm not forbidding anything, I'm using RFC9112.  Any HTTP request that violates RFC9112 cannot be expected to work and no other specification should attempt to work around it.  If we tried to provide guidance on everything that violated some other RFC on which we depend, we'd never get anything else done.

-----

> and they can exist in OADs when a relative $id or $self is used, or a relative server url is provided.

This is a completely different case than reading an incoming HTTP request on a server.  Actually, several different cases, as `$id` and `$self` are about OAD URIs and the sever URL is an API URL.  But constructing a full request URI _from an OAD in order to issue an HTTP request_ is a separate process from constructing a URI _on the server from the received request._.

karenetheridge: > But we're talking about HTTP requests being received by a server, right?

Not always -- the OAD itself doesn't have an HTTP request associated with it. It is quite possible it doesn't have an identifier with scheme+host.  The request itself has a scheme+host, which is why I said above that sometimes this is my only choice for resolving an otherwise-relative URI.

handrews: @karenetheridge 

> Not always -- the OAD itself doesn't have an HTTP request associated with it. It is quite possible it doesn't have an identifier with scheme+host. The request itself has a scheme+host, which is why I said above that sometimes this is my only choice for resolving an otherwise-relative URI.

I feel like several things/steps are getting combined (at least in my head while I read this) which is making it hard for us to reach agreement.  I know you understand the various principles involved here, I think it's just the different scenarios are tripping us (or at least me) up.

I'm going to go back to some of your earlier comments and see if I can walk through an example, tying each step to the appropriate RFC or OAS requirement.  I'll post a proper reply once I've done that.  I suspect the exercise will answer several of my confusions and spare you some headache :-)

Issue #4147, opened 2024-10-19
Improve 3.1+ schema `$id` UX


handrews:
Currently, we have the following schema document `$id`s for OAS 3.1 (shown relative to a base URI of `https://https://spec.openapis.org/oas/3.1/`, and without the trailing date segment that all schemas need to support updates:

* `schema`   # Validates everything _except_ the Schema Object, which is only validated against `type: [boolean, object]`
* `schema-base`  # Extends `schema` to validate Schema Objects against `dialect/base` _and_ require both `$schema` and `jsonSchemaDialect` to be set to `schema`'s `$id` if they are present
* `dialect/base` # A dialect equivalent to the default 2020-12 dialect plus the OAS extensions
* `meta/base`  # The vocabulary metaschema for the OAS extensions

`schema-base` was named to indicate that it incorporates `dialect/base`.  No one is certain why the dialect and vocabulary are referred to as "base", but they are called that in the OAS text as well.  I vaguely recall discussing the dialect as a base for further OAS-relevant extensions?  Perhaps that is what the intent was.

AFAICT, the advantage of `schema` is that it's self-contained.  But giving it that name means that it is the schema that most people assume is what they want, yet it really does not produce the behavior they'd expect.

`schema-base` is closer to what people would expect, but is (in my view) overly restrictive in locking `$schema` to `dialect/base`, as it is valid to switch metaschemas if `$schema` appears alongside `$id`, and conforming JSON Schema implementations know to treat `$id`-containing subschemas as embedded resources and check for a changed metaschema.  However, this does ***not*** make `schema-base` wrong:  Users might indeed want to enforce using _only_ `dialect/base`, and this is the correct way to do that.

On the other hand (and contrary to some of my comments earlier on Slack), `schema-base` locking `jsonSchemaDialect` to `dialect/base` is probably correct.  If you want to set `jsonSchemaDialect` to something else, you would want a schema that validates Scheam Objects accordingly by default.

-----

We probably shouldn't change how `schema` behaves, but we should make it more obvious that it is incomplete.  We should also come up with some sort of intuitive name for a schema that is like `schema-base` but does not restrict `$schema`.  This name needs to convey that _most users will want this new schema, and **not** the plain_ `schema`.

[...]

PR #4927, opened 2025-09-08
Oad2

handrews:
Replacement for PR #4918 – please review carefully, I've rewritten this too many times across too many widely separated days which is getting a bit confusing for me.

This further streamlines things and gets rid of "Document" vs "document" by banishing discussion of what had been "documents" to Appendix G, and saying that unless otherwise specified, "document" (regardless of capitalization) refers to documents having either an OpenAPI Object or Schema Object at the root.

I have updated the [rendered page](https://handrews.github.io/renderings/oas/oad-refactor/oas.html) on my personal github pages.

This has two commits (one for moves, one for changes, more or less) as with #4918.  

Please let me know if you want other ways of looking at diffs that I did for that other PR.

- [ ] schema changes are included in this pull request
- [ ] schema changes are needed for this pull request but not done yet
- [X] no schema changes are needed for this pull request

PR #4929, opened 2025-09-08
Make patch optional in openapi field.

handrews:
See discussion #4233 and issue #4147.  Requiring the patch version is creating tremendous amounts of confusion.  While we cannot seem to reach consensus on forbidding it, there has long been a consensus on not requiring it.  Let's at least do that.

Also, the regex in the schema allowed for `-something` in the `openapi` field, which is not part of the OAS field definition at all so I removed it.  We used to do `-rcN` releases, but we have no plan to do so for 3.2 anyway, and we ought not further overload an already confusing mechanism.  If we need to note a pre-release version in the future, let's add a field for it.


lornajane:
I am not in favour of introducing this change for 3.2 and delaying the release again. This is not a review of the change, but a logistical point. Please do not introduce changes for 3.2, and instead target this for 3.3 and we can review it there.

handrews:
@lornajane this is not a delay.  There is a simple change that we had consensus on at one point.  The amount of problems this is causing right now is worth the _completely trivial_ amount of work to merge this PR.

ralfhandl (dismissed):

earth2marsh (dismissed):
I'm in favor of this change, though whether it should be in 3.2 or 3.3 MUST be resolved before this is merged.

karenetheridge:
PLEASE DON'T MERGE this until the discussion on the relevant issues has concluded.

handrews:
@karenetheridge this is still a problem regardless of what happens with schema `$id`s.

handrews:
Rebased and force-pushed because of the merge of PR #4927, but there are _no actual changes_ to the commits/text here (both touched the `openapi` field text but this PR already incorporated the change from 4927).

src/oas.md
```diff
@@ -94,7 +94,7 @@ In addition to the required fields, at least one of the `components`, `paths`, o
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
-| <a name="oas-version"></a>openapi | `string` | **REQUIRED**. This string MUST be the [version number](#versions-and-deprecation) of the OpenAPI Specification that the OpenAPI document uses. The `openapi` field SHOULD be used by tooling to interpret the OpenAPI document. This is _not_ related to the [`info.version`](#info-version) string, which describes the OpenAPI document's version. |
+| <a name="oas-version"></a>openapi | `string` | **REQUIRED**. This string MUST be the major and minor [version number](#versions-and-deprecation) of the OpenAPI Specification that the OpenAPI Document uses (e.g. `"3.2"`), and MAY also contain the patch number (e.g. `"3.2.0"`). The major and minor version numbers from the `openapi` field MUST be used by tooling to interpret the OpenAPI Document, while the patch number SHOULD be ignored. This field is _not_ related to the [`info.version`](#info-version) string, which describes the OpenAPI document's version. |
```
ralfhandl: Why "Document" instead of "document"?

I thought #4927 switched to lowercase and eliminated the uppercase one?

handrews: Oops, lemme fix!  Good catch.

handrews:
@ralfhandl should be fixed now.

ralfhandl (dismissed):

handrews:
@karenetheridge this PR still allows the point release to be present if people want to do something with it.  Without opening another thread here on any other aspects, does that make any difference to you?  Nothing you're doing now with point releases would be impossible after this for any OAD author who wants to specify the point release.  

This PR just allows authors who want to make it very clear that they are using 3.2 in its most clarified/least buggy form, whatever that is, to say so.

karenetheridge (dismissed):

src/oas.md
```diff
@@ -94,7 +94,7 @@ In addition to the required fields, at least one of the `components`, `paths`, o
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
-| <a name="oas-version"></a>openapi | `string` | **REQUIRED**. This string MUST be the [version number](#versions-and-deprecation) of the OpenAPI Specification that the OpenAPI document uses. The `openapi` field SHOULD be used by tooling to interpret the OpenAPI document. This is _not_ related to the [`info.version`](#info-version) string, which describes the OpenAPI document's version. |
+| <a name="oas-version"></a>openapi | `string` | **REQUIRED**. This string MUST be the major and minor [version number](#versions-and-deprecation) of the OpenAPI Specification that the OpenAPI document uses (e.g. `"3.2"`), and MAY also contain the patch number (e.g. `"3.2.0"`). The major and minor version numbers from the `openapi` field MUST be used by tooling to interpret the OpenAPI document, while the patch number SHOULD be ignored. This field is _not_ related to the [`info.version`](#info-version) string, which describes the OpenAPI document's version. |
```
bigwill69, line 97: Your amazing  I'm happy to read your work; :)

bigwill69 (approved):

mikekistler:
We discussed this in the 9/11/25 TDC meeting and could not get consensus on whether this should go in 3.2 or 3.3.

handrews:
I have made my points here and elsewhere.  I will be traveling and will not be able to make next week's TDC call, but I encourage folks to make a decision one way or the other and ship the release in my absence.  I feel like folks have listened sufficiently to what I have to say on this topic, so I don't need to be present for any final call.

Please feel free to edit the PR or replace it with another one if editing this one is inconvenient.  I do not need to approve any edits.

ralfhandl:
I vote for including this in v3.2.

earth2marsh (dismissed):
I'm in favor of accepting this into 3.2, even though we have struggled to reach consensus. (If we need a formal vote on this issue, that's ok, too!)

The version section of the 3.1 spec states that:

> The `major.minor` portion of the version string (for example `3.1`) SHALL designate the OAS feature set.

So as a description document author, the patch should not matter to me (as it only clarifies the guidance in the features). 

Furthermore:

> The patch version SHOULD NOT be considered by tooling, making no distinction between 3.1.0 and 3.1.1 for example.

... since again the `patch` only covered clarifications in the specification itself (outside the feature level). We should discourage tools from caring about the patch (as this PR does).

The last consideration I'm aware of is that by making the patch optional, we run the risk of the value being interpreted as a number. This is a problem mostly for hand-written documents. It doesn't seem to me like that's a compelling enough reason to overcome the benefits described above.

ralfhandl (dismissed):
Resolved conflict with editorial PR changing an adjacent line.

* #4945

earth2marsh (dismissed):

src/schemas/validation/schema.yaml
```diff
@@ -7,7 +7,7 @@ type: object
 properties:
   openapi:
     type: string
-    pattern: '^3\.2\.\d+(-.+)?$'
+    pattern: '^3\.2(\.\d+)?$'
```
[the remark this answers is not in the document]
karenetheridge: Thanks Copilot, but we did that already.

lornajane (changes requested):
Discussion in the Moonwalk and TSC calls this week and we reached consensus that this should be retargeted to the 3.3 release.

ralfhandl (approved):

miqui (approved):

baywet:
From discussing this on the TDC call, this one is locked at this point and needs to go through and @OAI/tsc vote.

src/schemas/validation/schema.yaml
karenetheridge, line 10:
```suggestion
    pattern: '^3\.2(\.[0-9]+)?$'
```

...because "3.2.৪" should not be permitted (BENGALI DIGIT FOUR or U+09EA)

ralfhandl: @karenetheridge I think it would make sense to fix this (and only this) in separate pull requests for 3.3 and 3.2 (and maybe also for 3.1) and don't tie it to the stuck discussion on this PR.

handrews:
I'm closing this because it's been most of a year with no action from the TSC.  If anyone wants to take this forward, please pull the branch and open a new PR.

PR #4942, opened 2025-09-12
3.2 consistent use of "case-(in)sensitive"

ralfhandl:
We used both "case-sensitive" and "case sensitive", switching to the variant with dashes, also for "case-insensitive".

----

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -62,10 +62,10 @@ In order to preserve the ability to round-trip between YAML and JSON formats, YA
 The recommendation in previous versions of this specification to restrict YAML to its "JSON" [schema ruleset](https://yaml.org/spec/1.2/spec.html#id2803231) allowed for the inclusion of certain values that (despite the name) cannot be represented in JSON.
 OAD authors SHOULD NOT rely on any such JSON-incompatible YAML values.
 
-### Case-Sensitivity
+### Case Sensitivity
 
 As most field names and values in the OpenAPI Specification are case-sensitive, this document endeavors to call out any case-insensitive names and values.
-However, the case-sensitivity of field names and values that map directly to HTTP concepts follow the case-sensitivity rules of HTTP, even if this document does not make a note of every concept.
+However, the case sensitivity of field names and values that map directly to HTTP concepts follow the case sensitivity rules of HTTP, even if this document does not make a note of every concept.
 
 ### Rich Text Formatting
 


src/oas.md
```diff
@@ -64,7 +64,7 @@ OAD authors SHOULD NOT rely on any such JSON-incompatible YAML values.
 
 ### Case Sensitivity
 
-As most field names and values in the OpenAPI Specification are case-sensitive, this document endeavors to call out any case-insensitive names and values.
+As most field names and values in the OpenAPI Specification are case sensitive, this document endeavors to call out any case-insensitive names and values.
```
earth2marsh: While I appreciate the push for consistency, it feels weird to me to hypenate "case-insensitive" but not "case sensitive." AFAICT, the more common approach is to hyphenate both. 

WDYT about always hyphenating both variants in the spec?

lornajane: I think hyphenating might be more correct as it's an adjective, so I'm in favour

mikekistler: I'm also in favor of hyphenating both.

ralfhandl: I learned that one has to use dashes in a two-word phrase if it is used as an adjective, here "names are case insensitive" and "case-insensitive names".

Now consistently using dashes in all occurrences.

lornajane (approved):

src/oas.md
```diff
@@ -62,10 +62,10 @@ In order to preserve the ability to round-trip between YAML and JSON formats, YA
 The recommendation in previous versions of this specification to restrict YAML to its "JSON" [schema ruleset](https://yaml.org/spec/1.2/spec.html#id2803231) allowed for the inclusion of certain values that (despite the name) cannot be represented in JSON.
 OAD authors SHOULD NOT rely on any such JSON-incompatible YAML values.
 
-### Case Sensitivity
+### Case-Sensitivity
```
earth2marsh: AFAICT, "case sensitivity" is a compound noun, which should not be hyphenated. "Case-sensitive" is a compound adjective, which is why it should get a hyphen. 

This is a fairly trivial nit since the rest of the doc is already improved by the changes in this PR!

src/oas.md
```diff
@@ -62,10 +62,10 @@ In order to preserve the ability to round-trip between YAML and JSON formats, YA
...
 
 As most field names and values in the OpenAPI Specification are case-sensitive, this document endeavors to call out any case-insensitive names and values.
-However, the case sensitivity of field names and values that map directly to HTTP concepts follow the case sensitivity rules of HTTP, even if this document does not make a note of every concept.
+However, the case-sensitivity of field names and values that map directly to HTTP concepts follow the case-sensitivity rules of HTTP, even if this document does not make a note of every concept.
```
earth2marsh: same comment about compound adjectives vs nouns

src/oas.md
ralfhandl: We live to learn :smile:. Wikipedia does it that way: noun without dashes, adjective with: https://en.wikipedia.org/wiki/Case_sensitivity

mikekistler (approved):
Looks good. 👍

PR #4957, opened 2025-09-15
v3.2 schema edits

karenetheridge:
Several schema improvements for v3.2.  One of them MUST be in the release, as it allows for the use of `in: cookie, style: cookie` that was prohibited before.

I will backport to v3.1 the things that are relevant there.

### DIFF
--- a/src/schemas/validation/schema.yaml
+++ b/src/schemas/validation/schema.yaml
@@ -587,7 +587,6 @@ $defs:
         additionalProperties:
           $ref: '#/$defs/header-or-reference'
       style:
-        default: form
         enum:
           - form
           - spaceDelimited
@@ -596,7 +595,6 @@ $defs:
       explode:
         type: boolean
       allowReserved:
-        default: false
         type: boolean
       encoding:
         type: object
@@ -613,6 +611,20 @@ $defs:
         properties:
           prefixEncoding: false
           itemEncoding: false
+      style:
+        properties:
+          allowReserved:
+            default: false
+      explode:
+        properties:
+          style:
+            default: form
+          allowReserved:
+            default: false
+      allowReserved:
+        properties:
+          style:
+            default: form
     allOf:
       - $ref: '#/$defs/specification-extensions'
       - $ref: '#/$defs/styles-for-form'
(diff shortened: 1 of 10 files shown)


src/schemas/validation/schema.yaml
```diff
@@ -204,7 +204,7 @@ $defs:
         additionalProperties:
           $ref: '#/$defs/media-type-or-reference'
     patternProperties:
-      '^(schemas|responses|parameters|examples|requestBodies|headers|securitySchemes|links|callbacks|pathItems|mediaTypes)$':
+      '^(?:schemas|responses|parameters|examples|requestBodies|headers|securitySchemes|links|callbacks|pathItems|mediaTypes)$':
```
ralfhandl, line 207: Why a non-capturing pattern?

ralfhandl (changes requested):
Test case `pass/path_item_servers_parameters.yaml` fails with these schema changes.

src/schemas/validation/schema.yaml
karenetheridge: It's more efficient, and we use it in our other patterns.

karenetheridge:
Thanks, I've fixed the test failure.

ralfhandl (commented):
Why is this still needed?

https://github.com/karenetheridge/OpenAPI-Specification/blob/39cca6914b9ca9cee416de3049057f65a40a4780/src/schemas/validation/schema.yaml#L1116-L1117

ralfhandl (dismissed):

karenetheridge:
`style` is not required in the parameter object: it's disallowed when combined with `content`, and `style` can be omitted to allow its default value to be used in lieu.

karenetheridge:
I edited the querystring commit to add:

- the same constraint added for parameters in path-item as operation
- disallowed more than one querystring

Because of the restrictions appearing in two places (path-item and operation) I created a new definition for `parameters`.

ralfhandl (approved):

karenetheridge:
I found one more thing to add, but I'll make a new PR for that.

mikekistler (approved):
Looks good! 👍

PR #4958, opened 2025-09-15
v3.1 schema edits

karenetheridge:
Some schema improvements for v3.1.  This is a v3.1 port of the relevant bits from #4957.

### DIFF
--- a/src/schemas/validation/schema.yaml
+++ b/src/schemas/validation/schema.yaml
@@ -494,7 +494,6 @@ $defs:
         additionalProperties:
           $ref: '#/$defs/header-or-reference'
       style:
-        default: form
         enum:
           - form
           - spaceDelimited
@@ -503,8 +502,22 @@ $defs:
       explode:
         type: boolean
       allowReserved:
-        default: false
         type: boolean
+    dependentSchemas:
+      style:
+        properties:
+          allowReserved:
+            default: false
+      explode:
+        properties:
+          style:
+            default: form
+          allowReserved:
+            default: false
+      allowReserved:
+        properties:
+          style:
+            default: form
     allOf:
       - $ref: '#/$defs/specification-extensions'
       - $ref: '#/$defs/styles-for-form'


src/schemas/validation/schema.yaml
```diff
@@ -193,7 +193,7 @@ $defs:
         additionalProperties:
           $ref: '#/$defs/path-item'
     patternProperties:
-      '^(schemas|responses|parameters|examples|requestBodies|headers|securitySchemes|links|callbacks|pathItems)$':
+      '^(?:schemas|responses|parameters|examples|requestBodies|headers|securitySchemes|links|callbacks|pathItems)$':
```
ralfhandl, line 196: Same question: what is the benefit of a non-capturing pattern? I find it confusing.

src/schemas/validation/schema.yaml
```diff
@@ -760,8 +763,6 @@ $defs:
           properties:
             type:
               const: apiKey
-          required:
-            - type
```
ralfhandl, line 764: If `required` is unnecessary here, then why not also clean up https://github.com/karenetheridge/OpenAPI-Specification/blob/72de87b52b771209efcb4c1c89ed5358c2d66879/src/schemas/validation/schema.yaml#L960-L965?

ralfhandl (commented):
Two questions

src/schemas/validation/schema.yaml
karenetheridge: It's more efficient, and we use it in our other patterns.

src/schemas/validation/schema.yaml
karenetheridge: That wouldn't be a valid change, because `style` is not required in the parameter object: it's disallowed when combined with `content`, and `style` can be omitted to allow its default value to be used in lieu.

ralfhandl (approved):

PR #4993, opened 2025-09-24
Fix dialect link from 3.1 to 3.2

valerii15298:

This PR updates link to the new one.
Screenshot 2025-09-24 at 11 58 34 AM

I noticed that in 3.2 spec there is a lot of references to 3.1, more than 20 cases. Shouldn't be they replaced to be `3.2` ?

### DIFF
--- a/versions/3.2.0.md
+++ b/versions/3.2.0.md
@@ -2976,7 +2976,7 @@ Where JSON Schema indicates that behavior is defined by the application (e.g. fo
 
 The OpenAPI Schema Object [dialect](https://www.ietf.org/archive/id/draft-bhutton-json-schema-01.html#section-4.3.3) is defined as requiring the [OAS base vocabulary](#base-vocabulary), in addition to the vocabularies as specified in the JSON Schema Specification Draft 2020-12 [general purpose meta-schema](https://www.ietf.org/archive/id/draft-bhutton-json-schema-01.html#section-8).
 
-The OpenAPI Schema Object dialect for this version of the specification is identified by the URI `https://spec.openapis.org/oas/3.1/dialect/base` (the <a name="dialect-schema-id"></a>"OAS dialect schema id").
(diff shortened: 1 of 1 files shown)


ralfhandl (changes requested):
@valerii15298 Thanks for finding and reporting this!

As explained in https://github.com/OAI/OpenAPI-Specification/blob/main/CONTRIBUTING.md, the published specification files in the main branch are "frozen".

Please open a pull request against the `v3.2-dev` branch with proposed changes to file `src/oas.md`. 

The proposed new text should not contain a concrete date in the dialect URI and rather use a placeholder `YYYY-MM-DD` because schema files are published independently of the spec patch releases.

Thanks in advance!

valerii15298:
@ralfhandl and what about other places in the 3.2 spec where 3.1 is used? Should it be changed to 3.2 also?

As I mentioned in many places 3.1 is used as example version, etc

valerii15298:
@ralfhandl and what about other places in the 3.2 spec where 3.1 is used? Should it be changed to 3.2 also?

As I mentioned in many places 3.1 is used as example version, etc

ralfhandl:
3.1 as an example version in section [2.1 Versions and Depreciation](https://spec.openapis.org/oas/v3.2.0.html#versions-and-deprecation) is ok.

The only problematic place I found is in [4.14.7 Considerations for File Uploads](https://spec.openapis.org/oas/v3.2.0.html#considerations-for-file-uploads):

> In contrast to OAS 3.0, the format keyword has no effect on the content-encoding of the schema in OAS 3.1. 

That should become "in OAS 3.1 and later".

If you find other places problematic, please point to them.

Thanks in advance!

handrews:
> That should become "in OAS 3.1 and later"

I could have sworn I fixed that... might have gotten lost in the switch to the new branching system 🤦

valerii15298:
Created new PR https://github.com/OAI/OpenAPI-Specification/pull/4994

PR #4959, opened 2025-09-15
v3.2: specification: minor clarifications, markup improvements

karenetheridge:
- clarify other prohibited combinations of `query` and `querystring`
- clarify that the styles table is comprehensive: other combinations are prohibited
- markup improvement to the styles table
- removal of json-like formatting in literal yaml examples of `dataValue`s

- [x] no schema changes are needed for this pull request

PR #4998, opened 2025-09-25
v3.1 specification: minor clarifications, markup improvements

karenetheridge:
This is a port of the relevant bits in #4959.

- Literal style names are now rendered in sans-serif as code
- "primitive" is not rendered as code, as it refers to one of string, number, boolean or null, not a literal type name.
-  be explicit that the style table shows ALL valid combinations

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -1192,7 +1192,7 @@ Using `content` with a `text/plain` media type is RECOMMENDED for `in: "cookie"`
 
 ##### Style Values
 
-In order to support common ways of serializing simple parameters, a set of `style` values are defined.
+In order to support common ways of serializing simple parameters, a set of `style` values are defined. Combinations not represented in this table are not permitted.
 
 | `style` | [`type`](#data-types) | `in` | Comments |
 | ---- | ---- | ---- | ---- |


src/oas.md
```diff
@@ -1192,17 +1192,17 @@ Using `content` with a `text/plain` media type is RECOMMENDED for `in: "cookie"`
 
 ##### Style Values
 
-In order to support common ways of serializing simple parameters, a set of `style` values are defined.
+In order to support common ways of serializing simple parameters, a set of `style` values are defined. Combinations not represented in this table are not permitted.
```
ralfhandl, line 1195: That adds a MUST NOT requirement, which we cannot do in a patch release.

ralfhandl (changes requested):

src/oas.md
karenetheridge: This requirement already existed; it is not adding a new one. It was just not mentioned here.

ralfhandl: Where is it mentioned?

karenetheridge: It's the table itself here that sets out the permitted combinations of `style`s and `in`. But this is the way it's been since at least 3.0, and the schemas also set out these restrictions (see the "styles-for-path", "styles-for-header" etc definitions) -- so I believe the very strong intent is that no other combinations are permitted.

karenetheridge: The 3.2 spec contains this language and the same table of combinations of `style`s and `in`, and we did not change any aspect of this spec (other than adding `in: querystring` and `style: cookie`) in 3.2; the schemas are also identical between 3.1 and 3.2. If the restriction had been new in 3.2, it would have been a breaking change, which we do not do in minor releases.

So this is just a clarification, not a change.

ralfhandl: I hear your point and still can't determine whether this is not adding a requirment. Will remove my blocker and abstain.

miqui (approved):

PR #5011, opened 2025-10-06
main: adjust-release-branch: replace release date placeholder with current date

ralfhandl:
When we start a release, we add a line to the history table in Appendix A with the new release number and a "TBD" placeholder instead of the release date.

The `adjust-release-branch.sh` script now replaces the placeholder with the current date and shows the differences. One less thing to remember during a release.

### DIFF
--- a/scripts/adjust-release-branch.sh
+++ b/scripts/adjust-release-branch.sh
@@ -5,6 +5,7 @@
 # Run this script from the root of the repo. It is designed to be run manually in a release branch.
 
 branch=$(git branch --show-current)
+today=$(date +%Y-%m-%d)
 
 if [[ ! $branch =~ ^v[0-9]+\.[0-9]+\.[0-9]+-rel$ ]]; then
   echo "This script is intended to be run from a release branch, e.g. v3.1.2-rel"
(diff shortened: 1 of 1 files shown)


scripts/adjust-release-branch.sh
```diff
@@ -16,7 +17,8 @@ version=${vVersion:1}
 echo Prepare release of $version
 
 cp EDITORS.md versions/$version-editors.md
-mv src/oas.md versions/$version.md
+sed "s/| TBD |/| $today |/g" src/oas.md > versions/$version.md
+diff -w src/oas.md versions/$version.md
```
mikekistler, line 21: What is this diff command for? Could you maybe a comment here to explain?

mikekistler (approved):
Looks good. 👍

lornajane:
@ralfhandl I have the same question as @mikekistler on what's happening in this script.

Could you respond to the question and also update the CONTRIBUTING guide which had a set of manual steps to match the script, but the script is now out of sync.

scripts/adjust-release-branch.sh
ralfhandl: It shows the lines that were changed by the `sed` command, which should only replace the "TBD" placeholder in the history table with the current (release) date. Just to be sure, and give the person preparing the release a hint if for some unfathomable reason the character sequence `| TBD |` appeared outside of the history table and should not have been changed.

ralfhandl:
See
* #5026

PR #5024, opened 2025-10-07
main: create jekyll lander if necessary

ralfhandl:
The spec site needs Markdown lander pages for showing the latest schema iterations with syntax highlighting, for example https://spec.openapis.org/oas/3.2/schema/2025-09-17.html.

Up to now these files had to be manually created when schemas were published the first time for a MAJOR.MINOR spec version, for example via
* #4977

Now these lander pages are created by copying the corresponding lander page from the preceding spec version and adjusting it to match the newly published MAJOR.MINOR spec version.

> [!NOTE]  
> Whenever a new (major) spec version changes the way the schema files are named or structured, lander pages for schema files with new names have to be manually created, and copied lander pages may need to be updated if the purpose of the schema file has changed.

### DIFF
--- a/.github/workflows/validate-markdown.yaml
+++ b/.github/workflows/validate-markdown.yaml
@@ -29,4 +29,8 @@ jobs:
         run: npx --yes markdownlint-cli2 *.md
 
       - name: Check links in markdown files
-        run: npx --yes @umbrelladocs/linkspector check
+        uses: umbrelladocs/action-linkspector@v1
+        with:
+          reporter: github-check
+          fail_level: any
+          filter_mode: nofilter


ralfhandl:
Needs to wait for
* #5025

handrews (approved):

ralfhandl:
Now Linkspector works, but npmjs.org refuses to let it access a package that we mention in the style guide: 

> Cannot reach https://www.npmjs.com/package/markdownlint Status: 403

This might be a temporary problem, after all the same check worked fine with
* #5025 

But I strongly dislike things that randomly fail.

I could change the style guide and reference the packages Github repository https://github.com/DavidAnson/markdownlint instead, hoping that Github will not deny a Github workflow access to one of its repos 😎 

Or I could change the Linkspector configuration and add `403` to the list of [alive status codes](https://github.com/UmbrellaDocs/linkspector?tab=readme-ov-file#alive-status-codes) because `403` kind of means that the page exists and the server just refuses accessing it 😁 

Or I could exclude the style guide from the link check 😢 

Not quite sure which alternative I dislike least 🙄 

@lornajane @handrews Any preferences?

PR #5028, opened 2025-10-09
 Clarified the status of version 3.2 and added information about 3.3 in contributing

baywet:

Since we've released 3.2.0, this clarifies which branch is the upcoming version in the contributing guidelines


baywet:
Thanks for the quick reviews. I'm not sur why the CI failed here (probably because it came from a fork?). @ralfhandl a little help please?
(I don't have contrib on this repo)

ralfhandl:
> I'm not sure why the CI failed here

npmjs.org occasionally responds with 403 instead of 200 when checking links to it. Maybe the link checker is misidentified as an "attacker".

🤷‍♂️

PR #5033, opened 2025-10-13
fix "openapi" field to accept 3.3 versions

karenetheridge:

### DIFF
--- a/src/schemas/validation/schema.yaml
+++ b/src/schemas/validation/schema.yaml
@@ -7,7 +7,7 @@ type: object
 properties:
   openapi:
     type: string
-    pattern: '^3\.2\.\d+(-.+)?$'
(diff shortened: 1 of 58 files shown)


ralfhandl:
The version needs to be bumped in more places in all four schema files. I've got a script for that, included in PR
* #5034

It also initializes the `src/oas.md` file from the last published version (now: 3.2.0), so if you run the in your branch, you'd have to cherry-pick your ["fix a whitespace error" commit](https://github.com/OAI/OpenAPI-Specification/commit/effe24fa6a87019a5645b5a62d33af30a836ebdc) from the `v3.3-dev` branch.

Next time we should run the `start-release.sh` script right after merging the release into `main` and syncing it to the development branches, _before_ we do any work on the subsequent release(s).

PR #5040, opened 2025-10-15
gh-pages: update to Jekyll 4.4

ralfhandl:
Use current Jekyll version instead of the old one that only worked with Ruby 3.2 and older.
- Update Jekyll dependency in Gemfile to allow any patch version of Jekyll minor version 4.4
- Add bigdecimal dependency to avoid warning with Ruby 3.3 and allow update to Ruby 3.4
- Remove warning about maximum Ruby version from README.md and instead point to the development instructions in CONTRIBUTING.md
- Update development instructions in CONTRIBUTING.md: point to the Jekyll theme that we actually use now, improve instructions for local testing

These changes have no visible effect, a preview is available at https://ralfhandl.github.io/OpenAPI-Specification/

----


ralfhandl:
To be recreated in repo https://github.com/OAI/spec.openapis.org

PR #5041, opened 2025-10-15
main: check-restricted-files.yaml uses sync branch names to test for allowed PRs

ralfhandl:
This workflow is triggered if a file in the `versions` folder is changed. This is only allowed for
- release PRs into `main`
- sync PRs into `dev` and `vX.Y-dev`

Sync PRs now use intermediate branches, and the workflow now tests for these sync branch names.
It also now lists the files changed in the `versions` folder.

----

### DIFF
--- a/.github/workflows/check-restricted-files.yaml
+++ b/.github/workflows/check-restricted-files.yaml
@@ -8,28 +8,32 @@ name: check-restricted-files
 on:
   pull_request:
     paths:
-      - 'schemas/**/*.yaml'
-      - 'versions/*.md'
+      - "versions/*.md"
 
 jobs:
   check-files:
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/checkout@v5 # checkout repo content
+        with:
+          fetch-depth: 0
+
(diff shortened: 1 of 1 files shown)


.github/workflows/check-restricted-files.yaml
```diff
@@ -8,28 +8,32 @@ name: check-restricted-files
...
+          fetch-depth: 0
+
       - name: Check changed files
         shell: bash
         run: |
           if [[ "${{ github.event.pull_request.head.repo.full_name }}" == "OAI/OpenAPI-Specification" ]] && \
              [[ "${{ github.event.pull_request.base.repo.full_name }}" == "OAI/OpenAPI-Specification" ]]; then
- 
-            if [[ "${{ github.event.pull_request.head.ref }}" == "main" ]] && \
+
```
miqui, line 26: @ralfhandl just fyi - I have used this:https://github.com/dorny/paths-filter

to detect changes under a specific path. Works well. Not suggesting to change this, just fyi.

PR #5044, opened 2025-10-16
3.3: initialize for minor release 3.3.0

ralfhandl:
- Initialize `src/oas.md` from `versions/3.2.0.md` on the `main` branch with empty history
- Bump version to 3.3.0
- Restore @karenetheridge's change from #5004
- Bump version in schema files
- Bump version in schema tests

This PR's head branch was created with `scripts/start-release.sh` from
- #5034

----

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -1857,7 +1857,7 @@ Using `contentEncoding` for a multipart field is equivalent to specifying an [En
 If `contentEncoding` is used for a multipart field that has an Encoding Object with a `headers` field containing `Content-Transfer-Encoding` with a schema that disallows the value from `contentEncoding`, the result is undefined for serialization and parsing.
 
 Note that as stated in [Working with Binary Data](#working-with-binary-data), if the Encoding Object's `contentType`, whether set explicitly or implicitly through its default value rules, disagrees with the `contentMediaType` in a Schema Object, the `contentMediaType` SHALL be ignored.
-Because of this, and because the Encoding Object's `contentType` defaulting rules do not take the Schema Object's`contentMediaType` into account, the use of `contentMediaType` with an Encoding Object is NOT RECOMMENDED.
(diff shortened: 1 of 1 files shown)


tests/schema/pass/valid_schema_types.yaml
```diff
@@ -1,4 +1,4 @@
-openapi: 3.2.1
+openapi: 3.3.1
```
mikekistler, line 1: Why is this file "3.3.1" but all the others are "3.3.0"?

ralfhandl: This test case also tests a non-zero patch version number in the OpenAPI field, which should pass validation with a 3.3 schema, and it does.

mikekistler: So it is acceptable to use non-existent patch version numbers? So "3.3.987" also validates? That seems odd.

But if that's the case, can we add a comment in the file to make that clear?

ralfhandl: Our schemas traditionally allow any patch version, even ones that haven't been published yet:

https://github.com/OAI/OpenAPI-Specification/blob/de76eb4de13112080f55843c4a22ed32f3468fad/src/schemas/validation/schema.yaml#L4-L10

This test case was copied from `v3.2-dev`, here's the original, referencing a not-yet-published v3.2.1:

https://github.com/OAI/OpenAPI-Specification/blob/55816a2c384b9a5fdb88634f1015ff16a5d40a37/tests/schema/pass/valid_schema_types.yaml#L1

As stated in the PR description

> This PR's head branch was created with `scripts/start-release.sh` from
> - #5034

It is "proof of correctness" for #5034, and I hesitate to "taint" it with additional commits.

Can we improve this copied and auto-adjusted schema test case in a separate PR, or wait with adding to this PR until #5034 is merged?

ralfhandl: To make sure it doesn't get forgotten:

- #5049

PR #5055, opened 2025-10-23
main: update vitest to 4.0.1

ralfhandl:
- update `vitest` to 4.0.1
- remove unnecessary dependency on `@vitest/coverage-v8`

This should solve the dependabot problems with
- #5053
- #5054

----


baywet:
also see #5058 to completely solve for those kind of problems (for vitest)

PR #5059, opened 2025-10-23
ci: adds conditions to avoid jobs which depends on tokens to run on forks

baywet:
this PR adds conditions similar to the Overlay repo for the workflows which rely on a token so they don't run on forks and create noise


lornajane (approved):
Thanks!

PR #5065, opened 2025-10-25
fix: escape meaningful quotes in HTTP status code docs

### LINKED ISSUE #5064: published HTML docs munge quotes
At https://github.com/OAI/OpenAPI-Specification/blob/251d1db061c7b148cc7a9cab49a4c25abd725d68/versions/3.2.0.md?plain=1#L2123, the specific quote marks around `200` are important to the meaning of the documentation.

When published as HTML at https://spec.openapis.org/oas/v3.2.0.html#patterned-fields-0, the quotes have been converted from `"` to `”`:

> “200”

This changes the meaning of the sentence:

> This field MUST be enclosed in quotation marks (for example, "200") for compatibility...

making the example invalid.

Affected versions of the spec likely 3.0.1 through 3.2.0:

https://github.com/search?q=repo%3AOAI%2FOpenAPI-Specification+%22be+enclosed+in+quotation+marks+%22&type=code

alxndrsn:
Closes #5064


src/oas.md
```diff
@@ -2120,7 +2120,7 @@ call.
...
+| <a name="responses-code"></a>[HTTP Status Code](#http-status-codes) | [Response Object](#response-object) \| [Reference Object](#reference-object) | Any [HTTP status code](#http-status-codes) can be used as the property name, but only one property per code, to describe the expected response for that HTTP status code. This field MUST be enclosed in quotation marks (for example, \"200\") for compatibility between JSON and YAML. To define a range of response codes, this field MAY contain the uppercase wildcard character `X`. For example, `2XX` represents all response codes between `200` and `299`. Only the following range definitions are allowed: `1XX`, `2XX`, `3XX`, `4XX`, and `5XX`. If a response is defined using an explicit code, the explicit code definition takes precedence over the range definition for that code. |
```
```suggestion
| <a name="responses-code"></a>[HTTP Status Code](#http-status-codes) | [Response Object](#response-object) \| [Reference Object](#reference-object) | Any [HTTP status code](#http-status-codes) can be used as the property name, but only one property per code, to describe the expected response for that HTTP status code. This field MUST be enclosed in quotation marks (for example, `"200"`) for compatibility between JSON and YAML. To define a range of response codes, this field MAY contain the uppercase wildcard character `X`. For example, `2XX` represents all response codes between `200` and `299`. Only the following range definitions are allowed: `1XX`, `2XX`, `3XX`, `4XX`, and `5XX`. If a response is defined using an explicit code, the explicit code definition takes precedence over the range definition for that code. |
```
I'd format it as code.

alxndrsn: Of course :bulb:

ralfhandl:
Thanks!

mikekistler (approved):
Looks good. 👍

PR #5080, opened 2025-10-28
v3.3: parameter style amendments

karenetheridge:
I have been working through the parameter templating section and filling in some of the styles that I had not previously implemented, and cross-checking against RFC6570, and I found these issues with the "in: path" styles.

Each commit explains the reasoning for the change.

I made the changes against v3.3-dev, but I think they could probably go back to 3.1 and 3.2 as well?

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -907,7 +907,7 @@ Assume a parameter named `color` has one of the following values, where the valu
    object -> { "R": 100, "G": 200, "B": 150 }
 ```
 
-The following table shows serialized examples, as would be shown with the `serializedValue` field of an Example Object, of the different serializations for each value.
+The following table shows serialized examples, as would be shown with the `serializedValue` field of an Example Object, of the different serializations for the value of each supported type.
 
 * The value _empty_ denotes the empty string, and is unrelated to the `allowEmptyValue` field.
 * The behavior of combinations marked _n/a_ is undefined.


src/oas.md
```diff
@@ -519,9 +519,11 @@ Path templating refers to the usage of template expressions, delimited by curly
 Each template expression in the path MUST correspond to a path parameter that is included in the [Path Item](#path-item-object) itself and/or in each of the Path Item's [Operations](#operation-object). An exception is if the path item is empty, for example due to ACL constraints, matching path parameters are not required.
 
 The value for these path parameters MUST NOT contain any unescaped "generic syntax" characters described by [RFC3986](https://tools.ietf.org/html/rfc3986#section-3): forward slashes (`/`), question marks (`?`), or hashes (`#`).
+For example, the template `/foo/{bar}` cannot match the URI path "/foo/alpha/beta" because the value of the template variable "bar" would have to be "alpha/beta".
+
 See [URL Percent-Encoding](#url-percent-encoding) for additional guidance on escaping characters.
 
-The path templating is defined by the following [ABNF](https://tools.ietf.org/html/rfc5234) syntax
+The path templating is defined by the following [ABNF](https://tools.ietf.org/html/rfc5234) syntax.
```
```suggestion
The path templating is defined by the following [ABNF](https://tools.ietf.org/html/rfc5234) syntax:
```

src/oas.md
```diff
@@ -519,9 +519,11 @@ Path templating refers to the usage of template expressions, delimited by curly
 Each template expression in the path MUST correspond to a path parameter that is included in the [Path Item](#path-item-object) itself and/or in each of the Path Item's [Operations](#operation-object). An exception is if the path item is empty, for example due to ACL constraints, matching path parameters are not required.
 
 The value for these path parameters MUST NOT contain any unescaped "generic syntax" characters described by [RFC3986](https://tools.ietf.org/html/rfc3986#section-3): forward slashes (`/`), question marks (`?`), or hashes (`#`).
+For example, the template `/foo/{bar}` cannot match the URI path "/foo/alpha/beta" because the value of the template variable "bar" would have to be "alpha/beta".
```
handrews, line 523: I feel like this is a bit abrupt in jumping from serialization to matching.  How about a sentence like this to transition?

```suggestion
This means that when matching templates to URLs, no values that include a forward slash are matched.
For example, the template `/foo/{bar}` cannot match the URI path "/foo/alpha/beta" because the value of the template variable "bar" would have to be "alpha/beta".
```

src/oas.md
karenetheridge: I used the same formatting as in https://spec.openapis.org/oas/latest#server-variable-object, so perhaps that should be changed as well then.

src/oas.md
karenetheridge: sure. but do we want to be consistent with URI or URL here?

src/oas.md
handrews: Ah, well, if other folks have opinions one way or the other we can go with whatever seems preferred?  Otherwise I'm fine with you matching your existing text elsewhere - if in doubt, defer to whoever wrote what's already there :-)

src/oas.md
handrews: Per the sections on API URLs vs OAD URIs, hese are URLs, because they are API endpoints, so yes, the URI bit should also change to URL. (I also tweaked the new sentence to say "request URLs" instead of just URLs.  All API endpoints (and related locatable tings like OAuth server URLs) should use URL consistently.

```suggestion
This means that when matching templates to request URLs, no values that include a forward slash are matched.
For example, the template `/foo/{bar}` cannot match the URL path "/foo/alpha/beta" because the value of the template variable "bar" would have to be "alpha/beta".
```

src/oas.md
karenetheridge: I'll fix them all :p

karenetheridge:
edits pushed, as discussed.

handrews (approved):

PR #5122, opened 2025-11-15
v3.3: edits to parameter sections

karenetheridge:
Various fixes and wording clarifications to the sections discussing parameters.

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -769,7 +769,7 @@ See [Appendix E](#appendix-e-percent-encoding-and-form-media-types) for a detail
 There are five possible parameter locations specified by the `in` field:
 
 * path - Used together with [Path Templating](#path-templating), where the parameter value is actually part of the operation's URL. This does not include the host or base path of the API. For example, in `/items/{itemId}`, the path parameter is `itemId`.
-* query - Parameters that are appended to the URL. For example, in `/items?id=###`, the query parameter is `id`; MUST NOT appear in the same operation (or in the operation's path-item) as an `in: "querystring"` parameter.
+* query - Parameters that are appended to the URL with the `?` character (or for subsequent query parameters, with the `&` character); MUST NOT appear in the same operation (or in the operation's path-item) as an `in: "querystring"` parameter.
 * querystring - A parameter that treats the entire URL query string as a value which MUST be specified using the `content` field, most often with media type `application/x-www-form-urlencoded` using [Encoding Objects](#encoding-object) in the same way as with request bodies of that media type; MUST NOT appear more than once, and MUST NOT appear in the same operation (or in the operation's path-item) as any `in: "query"` parameters.
 * header - Custom headers that are expected as part of the request. Note that [RFC9110](https://www.rfc-editor.org/rfc/rfc9110.html#section-5.1) states header names are case-insensitive.
 * cookie - Used to pass a specific cookie value to the API.


src/oas.md
```diff
@@ -769,14 +769,15 @@ See [Appendix E](#appendix-e-percent-encoding-and-form-media-types) for a detail
 There are five possible parameter locations specified by the `in` field:
 
 * path - Used together with [Path Templating](#path-templating), where the parameter value is actually part of the operation's URL. This does not include the host or base path of the API. For example, in `/items/{itemId}`, the path parameter is `itemId`.
-* query - Parameters that are appended to the URL. For example, in `/items?id=###`, the query parameter is `id`; MUST NOT appear in the same operation (or in the operation's path-item) as an `in: "querystring"` parameter.
+* query - Parameters that are appended to the URL with the `?` character (or to existing query parameters with the `&` character), as described by [RFC3986 section 3.4](https://datatracker.ietf.org/doc/html/rfc3986#section-3.4);
```
handrews: Why the RFC3986 reference?  That section does define the query component, but only vaguely alludes to key-value pairs and does not mention the `&` character at all.  Technically RFC6570 is the most relevant specification defining the query string _as produced by `in: query`_, although not as produces by `in: querystring`, for which WHATWG URL is more relevant.

src/oas.md
```diff
@@ -769,14 +769,15 @@ See [Appendix E](#appendix-e-percent-encoding-and-form-media-types) for a detail
...
+MUST NOT appear in the same operation (or in the operation's path-item) as an `in: "querystring"` parameter.
 * querystring - A parameter that treats the entire URL query string as a value which MUST be specified using the `content` field, most often with media type `application/x-www-form-urlencoded` using [Encoding Objects](#encoding-object) in the same way as with request bodies of that media type; MUST NOT appear more than once, and MUST NOT appear in the same operation (or in the operation's path-item) as any `in: "query"` parameters.
 * header - Custom headers that are expected as part of the request. Note that [RFC9110](https://www.rfc-editor.org/rfc/rfc9110.html#section-5.1) states header names are case-insensitive.
 * cookie - Used to pass a specific cookie value to the API.
 
 #### Fixed Fields
 
-The rules for serialization of the parameter are specified in one of two ways.
+The rules for serialization and deserialization of the parameter are specified in one of two ways.
```
handrews, line 779: I think in other places we have used "serialization and parsing" or "parsing and serialization."  Do  you think "deserialization" is a better term here?  Is there a distinction between that and "parsing"?  I have no idea what the answer is, I either copied "parsing" or picked it without much thought, so I'm open to changing it as long as we're consistent.

src/oas.md
```diff
@@ -923,15 +924,15 @@ The following table shows serialized examples, as would be shown with the `seria
...
 | simple | false | _empty_ | blue | blue,black,brown | R,100,G,200,B,150 |
 | simple | true | _empty_ | blue | blue,black,brown | R=100,G=200,B=150 |
-| form | false | <span style="white-space: nowrap;">color=</span> | <span style="white-space: nowrap;">color=blue</span> | <span style="white-space: nowrap;">color=blue,black,brown</span> | <span style="white-space: nowrap;">color=R,100,G,200,B,150</span> |
-| form | true | <span style="white-space: nowrap;">color=</span> | <span style="white-space: nowrap;">color=blue</span> | <span style="white-space: nowrap;">color=blue&color=black&color=brown</span> | <span style="white-space: nowrap;">R=100&G=200&B=150</span> |
-| spaceDelimited</span> | false | _n/a_ | _n/a_ | <span style="white-space: nowrap;">color=blue%20black%20brown</span> | <span style="white-space: nowrap;">color=R%20100%20G%20200%20B%20150</span> |
+| form | false | color= | color=blue | color=blue,black,brown | color=R,100,G,200,B,150 |
+| form | true | color= | color=blue | color=blue&color=black&color=brown | R=100&G=200&B=150 |
+| spaceDelimited | false | _n/a_ | _n/a_ | color=blue%20black%20brown | color=R%20100%20G%20200%20B%20150 |
```
handrews, line 928: My recollection is that we need these because otherwise the text will wrap at `=` and/or `&` without this, but I'm not at all certain.  I tried to build locally to check but I'm having problems with that.  Have you verified that it won't wrap without this?

src/oas.md
```diff
@@ -3072,7 +3073,7 @@ Serializing data into such formats requires either examining the schema-validate
 
 When inspecting schemas, given a starting point schema, implementations MUST examine that schema and all schemas that can be reached from it by following only `$ref` and `allOf` keywords.
 These schemas are guaranteed to apply to any instance.
-When searching schemas for `type`, if the `type` keyword's value is a list of types and the serialized value can be successfully parsed as more than one of the types in the list, and no other findable `type` keyword disambiguates the actual required type, the behavior is implementation-defined.
+When searching schemas for `type`, if the `type` keyword's value is a list of types and the serialized value can be successfully parsed as more than one of the types in the list, and no other findable `type` keyword disambiguates the actual required type, the behavior is implementation-defined and must be documented.
```
handrews: There was a debate in the Thursday meeting over documentation requirements like this, which I had put in a lot of places, and the consensus was not to include it.  So I had to take it out, and we can only have language like this if we want to re-open that discussion in the Thursday call.

src/oas.md
```diff
@@ -4964,9 +4965,9 @@ parameters:
     type: object
     additionalProperties:
       type: string
-  explode: true
```
handrews, line 4966: Here I was trying to be unambiguous about the `explode` value rather than leaving it as the default, which can be confusing.  I don't know that we need to keep what I was doing, I'm just explaining why it was there.

src/oas.md
```diff
@@ -4994,16 +4995,14 @@ To do that, we'll add `allowReserved: true` to `formulas`, and change to `style:
...
   schema:
     type: object
     additionalProperties:
       type: string
-  explode: true
-  allowReserved: true
 - name: words
   in: query
   style: spaceDelimited
-  explode: false
```
handrews, line 5005: Likewise here I was trying to show the explicit lack of difference between `explode` values that was clarified in 3.2.  Likewise this is more of an explanation than a real objection.

src/oas.md
```diff
@@ -4909,7 +4910,7 @@ parameters:
     type: string
 ```
 
-This example is equivalent to RFC6570's `{?foo*,bar}`, and **NOT** `{?foo*}{&bar}`. The latter is problematic because if `foo` is not defined, the result will be an invalid URI.
+This example is equivalent to RFC6570's `{?foo*,bar}`, and **NOT** `{?foo*}{&bar}`. The latter is problematic because if `foo` is not defined (or the object has no properties), the result will be an invalid URI.
```
handrews: I remember wrestling with whether to get into this when I wrote this part.  RFC6570's actual requirements are even more complex:

>    A variable defined as a list value is considered undefined if the
   list contains zero members.  A variable defined as an associative
   array of (name, value) pairs is considered undefined if the array
   contains zero members or if all member names in the array are
   associated with undefined values.

I think I decided to not get into it here because of this normative reference in the Parameter Object section:

> The `undefined` column replaces the `empty` column in previous versions of this specification in order to better align with [[RFC6570](https://spec.openapis.org/oas/v3.2.0#bib-rfc6570)] [Section 2.3](https://datatracker.ietf.org/doc/html/rfc6570#section-2.3) terminology, which describes certain values including but not limited to `null` as “undefined” values with special handling; notably, the empty string is _not_ undefined.

Perhaps a more general "see RFC6570 §3.2 for details on what is considered undefined" rather than trying  to enumerate the conditions would be better?  I do like the idea of making some note here that "undefined" is a complex concept.

src/schemas/validation/schema.yaml
```diff
@@ -439,6 +439,8 @@ $defs:
                   const: path
             then:
               properties:
+                name:
+                  pattern: '^[^{}]+$'
```
handrews, line 439: Good catch on the braces.  I don't think we actually forbid the empty string, though, so `*` instead of `+`?  (We probably _should_ forbid it, but I don't think we do).

src/oas.md
karenetheridge: Removed in the new commits, as per https://github.com/OAI/OpenAPI-Specification/issues/5188#issuecomment-3802142333.

src/oas.md
karenetheridge: I searched the document and found that "deserialize" was used more often than "parse". I've standardized on the former in my implementation (after waffling a few times) as I think the symmetry makes more sense. Also I think "parse" is a little more vague -- you parse a string for any number of reasons, but what we're really doing here when we parse a parameter string is deserializing it into the json document model so we can then validate it against a json schema, or pass it to the user application in a format closer to its final form.

src/oas.md
karenetheridge: It doesn't appear that there are any characters in the examples that would cause a wrap now, but a check should be done on the final copy before merging -- what's the process for that? I can look at the rendered document on github but that's not the same processing that our publication process uses for html generation.

src/oas.md
karenetheridge: I would prefer that we state in the spec that implementations should be more rigorous about documenting things like this, but I'm not going to open a TSC debate about it. Removed.

src/oas.md
karenetheridge: I've now gone through this in even more excruciating detail as I've fully rewritten all my style deserialization.. I'm fine with your suggestion.

src/schemas/validation/schema.yaml
karenetheridge: Yes, it's `1*( ... )` in the ABNF. I already had `+` though?

karenetheridge: FWIW the full regex I'm using for path templates is this:

    ^(?:\{[^{}]+\}|%[0-9A-F]{2}|[:@!\$&'()*+,;=A-Za-z0-9._~-]+)+\z

..no one wanted to check my work when I asked last time. :(

src/oas.md
karenetheridge: fair; removed.

karenetheridge:
Force-pushing a new version after checking all commits; more test cases have been added for the  schema changes.

karenetheridge:
As this PR contains some fixes to things that are incorrect, IMO those commits at least should be backported to 3.2 and 3.1.

src/oas.md
karenetheridge: > I searched the document and found that "deserialize" was used more often than "parse".

I doublechecked this assertion and it's false. We talk a lot about "parsing an OAD", and that's definitely the correct language there, but we say both "parse a parameter" and "deserialize a parameter".

I think I still prefer "deserialize" for its extra specificity though, when it is applicable -- an OAD is parsed to interpret its rules, and parts of an HTTP message are deserialized to be applied against the OAD rules.

miqui (dismissed):

src/oas.md
```diff
@@ -4908,7 +4911,7 @@ parameters:
     type: string
 ```
 
-This example is equivalent to RFC6570's `{?foo*,bar}`, and **NOT** `{?foo*}{&bar}`. The latter is problematic because if `foo` is not defined, the result will be an invalid URI.
+This example is equivalent to RFC6570's `{?foo*,bar}`, and **NOT** `{?foo*}{&bar}`. The latter is problematic because if `foo` is not defined (see [RFC6570 §3.2](https://www.rfc-editor.org/rfc/rfc6570#section-3.2) ) for details on what is considered undefined), the result will be an invalid URI.
```
```suggestion
This example is equivalent to RFC6570's `{?foo*,bar}`, and **NOT** `{?foo*}{&bar}`. The latter is problematic because if `foo` is not defined (see [RFC6570 §3.2](https://www.rfc-editor.org/rfc/rfc6570#section-3.2) for details on what is considered undefined), the result will be an invalid URI.
```

tests/schema/pass/parameter-object-examples.yaml
```diff
@@ -9,6 +9,7 @@ paths:
         in: header
         description: token to be passed as a header
         required: true
+        explode: false
```
ralfhandl, line 12: Is there another test case showing that `explode` is not suddenly required here?

In general I would prefer not changing existing "pass" test cases and instead add new ones for testing additional allowed variations.

Refactorings combined with changes to existing test cases make me very uneasy.

ralfhandl (changes requested):

tests/schema/fail/header-object-allowReserved.yaml
```diff
@@ -0,0 +1,12 @@
...
+info:
+  title: allowReserved only permitted with in and style values that percent-encode
+  version: 1.0.0
+components:
+  headers:
+    Style:
+      schema:
+        type: array
+      style: simple
+      explode: true
```
ralfhandl, line 11:
```suggestion
```

I recommend removing this line: just the bare required minimum plus one forbidden thing to make the test case fail.

tests/schema/pass/parameter-object-examples.yaml
karenetheridge: I'll check why I added these; they are not required. I think it was to show that they were permitted (did not cause an error).

ralfhandl: Pass test cases both with and without `explode` would be nice.

tests/schema/fail/header-object-allowReserved.yaml
miqui: +1

tests/schema/pass/parameter-object-examples.yaml
karenetheridge: there are comprehensive tests for all non-required keywords in another commit in this PR.

tests/schema/fail/header-object-allowReserved.yaml
karenetheridge: There are test cases for every non-required keyword in another test in this PR.

karenetheridge: This line needs to stay, in order for the coverage tests to pass.

tests/schema/pass/parameter-object-examples.yaml
karenetheridge: This line needs to stay, in order for the coverage tests to pass.

src/oas.md
handrews: If we're going to switch that terminology, I would rather discuss it in its own GitHub discussion and come to something resembling a consensus, and then make that change across the document.  We have enough consistency problems without doing this piecemeal.  I would feel the same way about changing any existing "deserialize" wording.  Let's make an intentional, group-approved choice of terminology and then implement it.  But in this PR, let's keep the exising terminology.

src/oas.md
handrews: @karenetheridge I pulled your branch and built it and it worked fine.  I'm OK with considering this resolved.

src/oas.md
karenetheridge: It's not changing terminology; it's adding it -- the main intent here is to make clear that parameter declarations are used both in serializing, and in parsing/deserializing.  I am mostly indifferent as to which of these two terms are used, but the point is that there is more going on than just serializing.

tests/schema/pass/parameter-object-examples.yaml
```diff
@@ -55,10 +57,12 @@ paths:
       - in: cookie
         name: my_cookie1
         style: form
+        explode: false
         schema: {}
       - in: cookie
         name: my_cookie2
         style: cookie
+        explode: true
```
ralfhandl, line 65: Please add a cookie parameter without `explode` to "prove" that it still can be omitted.

karenetheridge: That already exists, in tests/schema/pass/style-defaults.yaml.

PR #5177, opened 2026-01-12
Clarify usage of Schema Object in OpenAPI 3.0

Alexrohith:

### DIFF
--- a/versions/3.0.0.md
+++ b/versions/3.0.0.md
@@ -119,7 +119,7 @@ For example, if a field has an array value, the JSON array representation will b
 ```
 All field names in the specification are **case sensitive**.
 
-The schema exposes two types of fields: Fixed fields, which have a declared name, and Patterned fields, which declare a regex pattern for the field name.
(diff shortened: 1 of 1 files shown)


handrews:
@Alexrohith thanks for contributing!  I recommend checking out our [CONTRIBUTING.md](https://github.com/OAI/OpenAPI-Specification/blob/main/CONTRIBUTING.md) guide as it explains our processes and what sort of contributions we accept.

For this PR, I'm afraid we do not make changes to the visible contents of published versions (we do sometimes fix broken links when people change URLs, but that's about it).  The changeable source is in the `src/` directory, and is only editable on certain branches (currently v3.1-dev, v3.2-dev, and v3.3-dev).  There is no current plan to make further edits to 3.0.

handrews:
I also recommend joining the Thursday 9AM Pacific TDC calls, or opening a discussion if you want to propose some general wording changes.  Otherwise, since everyone has their own opinions about what wording is best we'd have too much back-and-forth if we took every wording suggestion.  It's best to build some consensus and also make sure that we use the same words across the spec (I have no idea if this wording appears in other Object descriptions or not, but if we were to change something in one place it should be changed in all).

PR #5183, opened 2026-01-22
Explain that schemas are for mandatory requirements

### LINKED ISSUE #4141: Consider allowing contributions to "linting" schemas
We have previously turned away quite a few suggested schema changes, particularly from @karenetheridge and @jeremyfiel, that attempted to enforce sensible but not mandatory constraints.  These often involve SHOULDs or statements that something will be ignored rather than treated as an error.

We've discussed allowing additional schemas that would `$ref` the validation schemas and add these additional constraints.  Let's make a formal decision whether to do that or not, and decide what would be allowed in such schemas so that folks who want to write them can do so with confidence.

handrews:
We do not host linting schemas here.  Fixes #4141.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -259,11 +259,6 @@ As part of the publishing process, the YAML source files are converted to JSON, 
 
 #### Schemas and OAS Requirements
 
-Not all uses of OAS Objects and fields are necessarily useful.
-However, the OAS does not forbid various questionable combinations, for
-reasons ranging from preserving possible future use with new features to
-maintaining compatibility with past versions that neglected to forbid something.
-
 The schemas published by the OpenAPI Initiative _only_ validate mandatory
 OAS requirements.  This means that a field value or combination with another
 field is only forbidden in a schema if there is a corresponding MUST / MUST NOT /


CONTRIBUTING.md
```diff
@@ -257,6 +257,24 @@ Schemas are updated in and directly published from the `vX.Y-dev` branches.
 
 As part of the publishing process, the YAML source files are converted to JSON, renamed to the relevant last-changed dates, and `WORK-IN-PROGRESS` placeholders are replaced with these dates as appropriate. This is usually done by the [`schema-publish` workflow](https://github.com/OAI/OpenAPI-Specification/blob/main/.github/workflows/schema-publish.yaml) which detects changes on each `vX.Y-dev` branch, which generates a pull request for publishing the new schema iterations to the [spec site](https://spec.openapis.org). The workflow can also be run manually if required.
 
+#### Schemas and OAS Requirements
+
+Not all uses of OAS Objects and fields are necessarily useful.
```
lornajane: Opening paragraph not needed in contributing file

handrews: @lornajane done!

PR #5212, opened 2026-02-23
Fix http QUERY method link text to match its url

valerii15298:
The text of link points to an older version of QUERY method draft(08), while actual url points to newer one(11). Changed text to match the url


src/oas.md
```diff
@@ -607,7 +607,7 @@ The path itself is still exposed to the documentation viewer but they will not k
 | <a name="path-item-head"></a>head | [Operation Object](#operation-object) | A definition of a HEAD operation on this path. |
 | <a name="path-item-patch"></a>patch | [Operation Object](#operation-object) | A definition of a PATCH operation on this path. |
 | <a name="path-item-trace"></a>trace | [Operation Object](#operation-object) | A definition of a TRACE operation on this path. |
-| <a name="path-item-query"></a>query | [Operation Object](#operation-object) | A definition of a QUERY operation, as defined in the most recent IETF draft ([draft-ietf-httpbis-safe-method-w-body-08](https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-11.html) as of this writing) or its RFC successor, on this path. |
+| <a name="path-item-query"></a>query | [Operation Object](#operation-object) | A definition of a QUERY operation, as defined in the most recent IETF draft ([draft-ietf-httpbis-safe-method-w-body-11](https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-11.html) as of this writing) or its RFC successor, on this path. |
```
miqui: there is a -14 now

valerii15298: Updated!

I am curious why not to use this link? https://httpwg.org/http-extensions/draft-ietf-httpbis-safe-method-w-body.html

It seems to point to the latest one

baywet: from the meeting: because that's a working copy, we want to use the ietf links that are snapshots in time from the spec authors.
Also, the PR has been updated to point to draft 14 (the latest as of writing)

PR #5229, opened 2026-02-28
schema-publish: Use commit date instead of author date

ralfhandl:
As described in [CONTRIBUTING, section Schema Iterations](https://github.com/OAI/OpenAPI-Specification/blob/main/CONTRIBUTING.md#schema-iterations)

> As part of the publishing process, the YAML source files are converted to JSON, _renamed to the relevant last-changed dates_, and WORK-IN-PROGRESS placeholders are replaced with these dates as appropriate. 

We currently use the _author date_ of the last commit changing a YAML source file, whereas the Github commit history shows the _commit date_, making it hard to correlate schema iterations with the corresponding changes in the development branch.

As @karenetheridge pointed out, using the author date also risks that the iteration date _goes backwards_ if changes are merged in a different sequence than they were authored. This is avoided by using the commit date, which is _the date the change was merged into the development branch_.

Changes in this PR:

- Use commit date of last change to construct schema file names instead of author date
  - the date a change was merged into the spec version dev branch is safer and easier to track/understand than the date the author made that change in their local repo
- Improve console output of what happened why

----

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,18 +9,18 @@
       "version": "0.0.0",
       "license": "Apache-2.0",
       "dependencies": {
-        "cheerio": "^1.1.2",
+        "cheerio": "^1.2.0",
         "highlight.js": "^11.11.1",
         "markdown-it": "^14.1.0",
-        "respec": "35.6.1",
+        "respec": "35.8.0",
         "yargs": "^18.0.0"
       },
       "devDependencies": {
         "@hyperjump/json-schema-coverage": "^1.2.1",
         "@umbrelladocs/linkspector": "^0.4.7",
         "@vitest/coverage-v8": "^4.0.18",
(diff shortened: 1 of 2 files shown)


karenetheridge:
This should be today's date (the date that the schemas are published), not the date of the most recent merge.

ralfhandl:
> This should be today's date (the date that the schemas are published), not the date of the most recent merge.

There are four schema files, and so far we only create a new iteration of a file if 
- its source changes, or 
- one of its dependencies changes.

Then we use the date of that change as the iteration name.

This minimises the number of iterations:
- If `schema.yaml` changes, only `schema` and `schema-base` get new iterations with the same date.
- If `meta.yaml` changes, all four files get new iterations with the same date.

This PR keeps that logic and only picks a "better" change date - the commit date.

PR #5233, opened 2026-03-05
v3.3: apply the ABNF for header names at rfc9110 §5.1

karenetheridge:
..and the same pattern applies to HTTP method names too

### DIFF
--- a/src/schemas/validation/schema.yaml
+++ b/src/schemas/validation/schema.yaml
@@ -243,7 +243,7 @@ $defs:
           $ref: '#/$defs/operation'
         propertyNames:
           $comment: RFC9110 restricts methods to "1*tchar" in ABNF
-          pattern: "^[a-zA-Z0-9!#$%&'*+.^_`|~-]+$"
+          $ref: '#/$defs/token'
           not:
             enum:
               - GET
@@ -414,6 +414,14 @@ $defs:
         then:
           required:
             - content
+      - if:
+          properties:
+            in:
+              const: header
+        then:
+          properties:
+            name:
+              $ref: '#/$defs/token'
     dependentSchemas:
       schema:
         properties:
@@ -604,6 +612,8 @@ $defs:
         format: media-range
       headers:
         type: object
+        propertyNames:
+          $ref: '#/$defs/token'
         additionalProperties:
           $ref: '#/$defs/header-or-reference'
       style:
@@ -679,6 +689,8 @@ $defs:
         type: string
       headers:
         type: object
+        propertyNames:
+          $ref: '#/$defs/token'
         additionalProperties:
           $ref: '#/$defs/header-or-reference'
       content:
@@ -1149,3 +1161,8 @@ $defs:
       properties:
         explode:
           default: false
+
+  token:
+    $comment: see https://www.rfc-editor.org/rfc/rfc9110.html#section-5.6.2
+    type: string
+    pattern: ^[0-9A-Za-z!#$%&'*+.^_`|~-]+$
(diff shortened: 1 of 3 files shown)


src/schemas/validation/schema.yaml
```diff
@@ -1149,3 +1161,8 @@ $defs:
       properties:
         explode:
           default: false
+
+  token:
+    $comment: see https://www.rfc-editor.org/rfc/rfc9110.html#section-5.6.2
+    type: string
+    pattern: '^[0-9A-Za-z!#$%&\'*+.^_`|~-]+$'
```
ralfhandl: This is not valid YAML: use double quotes or no quotes.


```suggestion
    pattern: "^[a-zA-Z0-9!#$%&'*+.^_`|~-]+$"
```

tests/schema/pass/header-object-name.yaml
```diff
@@ -0,0 +1,8 @@
+openapi: 3.3.0
+info:
+  title: header name has a constrained syntax, but the component name is not the header name
+  version: 1.0.0
+components:
+  headers:
+    'Still=Legal':
```
ralfhandl: The equal sign (`=`) was never legal in names/keys of reusable components, see  [Components Object, subsection Fixed Fields](https://spec.openapis.org/oas/v3.0.0.html#components-object):

> All the fixed fields declared above are objects that MUST use keys that match the regular expression: ^[a-zA-Z0-9\.\-_]+$.

manifested in https://github.com/karenetheridge/OpenAPI-Specification/blob/1f4999c317a468d15aade5c61ea8095d9cf4641e/src/schemas/validation/schema.yaml#L209-L210

ralfhandl (changes requested):

src/schemas/validation/schema.yaml
karenetheridge: Single quotes actually are valid YAML, but if the tools being used here don't allow it, then that's certainly a reason to change.

https://www.yaml.info/learn/quote.html

karenetheridge:
switched to double quotes, and removed a no-longer-useful test case.

src/schemas/validation/schema.yaml
ralfhandl: The catch here seems to be that

- 'to express one single quote, use '' two of them'

Escaping with backslash is only mentioned for double quotes: 

- "the double quote \" needs to be escaped"

Interesting 🙄

src/schemas/validation/schema.yaml
```diff
@@ -1149,3 +1161,8 @@ $defs:
       properties:
         explode:
           default: false
+
+  token:
+    $comment: see https://www.rfc-editor.org/rfc/rfc9110.html#section-5.6.2
+    type: string
+    pattern: '^[0-9A-Za-z!#$%&''*+.^_`|~-]+$'
```
ralfhandl: Using double quotes feels more intuitive than the doubled single quotes within a single-quoted string, and is what was used in the original code. 

And using no quotes is even more readable:
```suggestion
    pattern: ^[0-9A-Za-z!#$%&'*+.^_`|~-]+$
```

Please consider changing.

src/schemas/validation/schema.yaml
karenetheridge: resolved by removing all quotes

tests/schema/pass/header-object-name.yaml
karenetheridge: This has been changed to using the header name under a response object, which is not covered by the component name restriction.

PR #5263, opened 2026-03-25
v3.2: apply the ABNF for path parameter names, OAS 3.2.0 §4.8.2

karenetheridge:
Header parameter names have already been adjusted.

The ABNF for the query component of URIs is at RFC3986 §3.4, but percent-encoding is used, which allows for the use of any character. Additionally, cookies using style=form are also percent-encoded and therefore allow any character.

Moreover, cookie parameter names are percent-encoded when using style=form, and even for style=cookie they are not used in serialization of objects when explode=true, so a restriction on cookie parameter names is not added here; however applications should apply their own restrictions, following the "cookie-name" and "cookie-value" ABNFs at RFC6265 §4.1.1 (and ensure that all disallowed characters are percent-encoded when using style=form).

### DIFF
--- a/src/schemas/validation/schema.yaml
+++ b/src/schemas/validation/schema.yaml
@@ -417,10 +417,20 @@ $defs:
       - if:
           properties:
             in:
+              const: path
+        then:
+          properties:
+            name:
+              $comment: 'see OAS 3.2.0 §4.8.2'
+              pattern: '^[^{}]+$'
+      - if:
+          properties:
+            in:
               const: header
         then:
           properties:
             name:
+              $comment: 'see RFC9110 §5.1'
               $ref: '#/$defs/token'
     dependentSchemas:
       schema:
--- a/tests/schema/fail/parameter-object-path-name.yaml
+++ b/tests/schema/fail/parameter-object-path-name.yaml
@@ -1,0 +1,10 @@
+openapi: 3.2.0
+info:
+  title: path parameter name has a constrained syntax
+  version: 1.0.0
+components:
+  parameters:
+    BadPath:
+      name: 'Bad{Path}'
+      in: path
+      schema: {}


handrews:
@karenetheridge This appears to just have to do with the path parameter, which doesn't seem to align with the PR description.  Was there intended to be more to this PR? (I did remember to check that it's not a draft before commenting this time 🙂 )

karenetheridge:
"path parameter names" - is that more clear?

handrews:
@karenetheridge I did get that it was about path parameter names.  My confusion came from forgetting about `style: matrix`, which is the only style for which the name might end up in the URL.

Is all the stuff about queries and cookies just to say that nothing is needed for those because of percent-encoding?  If so, why is anything needed for path given that percent-encoding also applies to to the path?

I guess this is really just "use the schema to ensure `{` and `}` aren't present in path parameter names"?  That seems fine with me, it just would have been easier to understand if you said that at the start.

handrews:
> it just would have been easier to understand if you said that at the start.

By which I mean that I got confused and thought that the ABNF in question was from RFC3986 as you mention one from RFC3986, but I see that this is about the Path Templating ABNF forbidding curly braces.

miqui (approved):

karenetheridge:
Ah I understand the confusion now.

I originally started out writing the PR with prohibitions for query and cookie parameters as well (which then resulted in the cookie bombshell that I filed as a separate issue), but as I worked through some examples I found that there was no way to bring the restrictions into the schema, given that the explode=true+object combination meant that the parameter name would not actually ever appear.  So I rewrote the commit message to explain why this wasn't possible, but then the focus was more on that, rather than the one change that actually remained - the ABNF for path template names.

sorry for the ambiguity!

handrews:
@karenetheridge makes sense to me now, thanks for confirming!

Looking at the pattern, it does raise the question for me: Do we actually forbid the empty string?  It's a pathological case to be sure, _and I am not advocating for it_, but it could be made to work, and can we consider it forbidden given that it is not called out?  Otherwise this looks good.

karenetheridge:
Yes, we do explicitly forbid the empty string in the ABNF (note the `1*`):

```
template-expression            = "{" template-expression-param-name "}"
template-expression-param-name = 1*( %x00-7A / %x7C / %x7E-10FFFF ) ; every Unicode character except { and }
```

handrews:
Ah, I was looking only at the field definition.  Well, as no one has quibbled that the `1+` in the ABNF constitutes a breaking change, I think we're fine :-)

handrews (approved):

karenetheridge:
> Ah, I was looking only at the field definition.

It might be worth a mention there as well, as it is likely to get noticed more easily than buried in the ABNF :)

karenetheridge:
Maybe add "Note that the use of this value in various parts of the HTTP message may impose restrictions on the syntax of this value." to the end of the language for the "name" field in 4.12.2.1?

PR #5287, opened 2026-04-14
fix Media Type Example Server-Sent Event Streams example (v3.3)

notEthan:
In this example in Media Type Server-Sent Event Streams, the `data` field's schema is `format: int64`, so the data should be an integer

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -1632,7 +1632,7 @@ data: across two lines
 retry: 5
 
 event: addInt64
-data: 1234.5678
+data: 1234
 unknownField: this is ignored
 
 : This is a comment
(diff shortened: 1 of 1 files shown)


ralfhandl (approved):

karenetheridge (approved):

src/oas.md
```diff
@@ -1644,7 +1644,7 @@ To more clearly see how this stream is handled, the following is the equivalent
 
 ```jsonl
 {"event": "addString", "data": "This data is formatted\nacross two lines", "retry": 5}
-{"event": "addInt64", "data": "1234.5678"}
+{"event": "addInt64", "data": "1234"}
```
karenetheridge, line 1647: This is a number, not a string, so no quotes should be used in the JSON form:

```suggestion
{"event": "addInt64", "data": 1234}
```

karenetheridge (changes requested):

miqui (approved):
looks good pending comment resolution

src/oas.md
notEthan: A string is correct per https://github.com/OAI/OpenAPI-Specification/pull/5291#discussion_r3142653500

karenetheridge (approved):

PR #5289, opened 2026-04-14
Correct schema $comment for OAD `content` definition

notEthan:
The `content` object is a property of Parameter, Request Body, Response, and Header. Currently its schema $comment just links to Request Body.

I'm targeting this at 3.2 instead of 3.3 because, in the linked anchors "fixed-fields-{n}", the numbers seem likely to end up different by publication. Not sure what to do about that. All of the other $comment spec links use more unique anchors.

### DIFF
--- a/src/schemas/validation/schema.yaml
+++ b/src/schemas/validation/schema.yaml
@@ -566,7 +566,12 @@ $defs:
       $ref: '#/$defs/request-body'
 
   content:
-    $comment: https://spec.openapis.org/oas/v3.2#fixed-fields-10
+    $comment: |
+      The `content` object is a property of:
+      Parameter Object: https://spec.openapis.org/oas/v3.2#parameter-object
+      Request Body Object: https://spec.openapis.org/oas/v3.2#request-body-object
+      Response Object: https://spec.openapis.org/oas/v3.2#response-object
+      Header Object: https://spec.openapis.org/oas/v3.2#header-object
     type: object
     additionalProperties:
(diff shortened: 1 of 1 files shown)


src/schemas/validation/schema.yaml
```diff
@@ -566,7 +566,11 @@ $defs:
       $ref: '#/$defs/request-body'
 
   content:
-    $comment: https://spec.openapis.org/oas/v3.2#fixed-fields-10
+    $comment: |
+      Parameter Object: https://spec.openapis.org/oas/v3.2#fixed-fields-9
+      Request Body Object: https://spec.openapis.org/oas/v3.2#fixed-fields-10
+      Response Object: https://spec.openapis.org/oas/v3.2#fixed-fields-14
+      Header Object: https://spec.openapis.org/oas/v3.2#fixed-fields-17
```
ralfhandl, line 573: Why not link to the corresponding "Xxx Object" sections instead of the "Fixed Fields" subsections? Those are "close enough", especially for Parameter/Header Object where `content` is way down in the "Fixed Fields for use with `content`" subsection.

ralfhandl (changes requested):

src/schemas/validation/schema.yaml
notEthan: That does seem better, thanks for the suggestion. Pushed.

ralfhandl (approved):

karenetheridge (approved):

PR #5288, opened 2026-04-14
minor wording correction for Example object externalValue (v3.3)

notEthan:
Example object externalValue URI identifies serialized data, does not contain it
- [x] no schema changes are needed for this pull request

PR #5290, opened 2026-04-15
minor wording correction for Example object externalValue (v3.2)

notEthan:
Example object externalValue URI identifies serialized data, does not contain it

related https://github.com/OAI/OpenAPI-Specification/pull/5288


ralfhandl (approved):

src/oas.md
```diff
@@ -2340,7 +2340,7 @@ The various fields and types of examples are explained in more detail under [Wor
...
+| <a name="example-external-value"></a>externalValue | `string` | A URI that identifies the serialized example in a separate document, allowing for values not easily or readably expressed as a Unicode string.  If `dataValue` is present, then this field SHOULD identify a serialization of the given data.  Otherwise, the value SHOULD identify the valid serialization of a data value that itself MUST be valid as described for `dataValue`. If this field is present, `serializedValue` and `value` MUST be absent. See also the rules for resolving [Relative References](#relative-references-in-api-description-uris). |
```
karenetheridge, line 2343: perhaps "reference" rather than identify, as it is more specific as to the nature of the identification?

handrews: @karenetheridge I think "identify" is appropriate as this is a URI and that's what URIs do.  "references" gets caught up in `$ref` so might be better avoided?  (Note:  I don't feel all that strongly about this).

notEthan: 'identify' is used elsewhere in this paragraph, I think staying consistent with the other places it's used is best.

karenetheridge (approved):

PR #5291, opened 2026-04-15
fix Media Type Example Server-Sent Event Streams example (v3.2)

notEthan:
In this example in Media Type Server-Sent Event Streams, the `data` field's schema is `format: int64`, so the data should be an integer

related https://github.com/OAI/OpenAPI-Specification/pull/5287

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -1632,7 +1632,7 @@ data: across two lines
 retry: 5
 
 event: addInt64
-data: 1234.5678
+data: 1234
 unknownField: this is ignored
 
 : This is a comment
@@ -1644,7 +1644,7 @@ To more clearly see how this stream is handled, the following is the equivalent 
 
 ```jsonl
 {"event": "addString", "data": "This data is formatted\nacross two lines", "retry": 5}
-{"event": "addInt64", "data": "1234.5678"}
+{"event": "addInt64", "data": "1234"}
 {"event": "addJSON", "data": "{\"foo\": 42}"}
 ```
 


ralfhandl (approved):

src/oas.md
```diff
@@ -1644,7 +1644,7 @@ To more clearly see how this stream is handled, the following is the equivalent
 
 ```jsonl
 {"event": "addString", "data": "This data is formatted\nacross two lines", "retry": 5}
-{"event": "addInt64", "data": "1234.5678"}
+{"event": "addInt64", "data": "1234"}
```
karenetheridge, line 1647: This is a number, not a string, so no quotes should be used in the JSON form:

```suggestion
{"event": "addInt64", "data": 1234}
```

karenetheridge (changes requested):

src/oas.md
handrews: @karenetheridge String is correct.  Per the SSE specification (and per [our section](https://spec.openapis.org/oas/v3.2.0.html#special-considerations-for-server-sent-events) giving the generic `text/event-stream` schema, `data` is always a string.

handrews (approved):

miqui (approved):

karenetheridge (approved):

src/oas.md
karenetheridge: I went back to see why I was confused here -- the schema has `format: uint64`, _not_ `type: integer` -- as I was originally thinking that the intent was that the deserialized data instance was to have the string coerced to a number during deserialization. But this is just a format, not a type, and string is a valid type for this format.

handrews: @karenetheridge yeah makes sense.  I always approve of blaming things on `format` being confusing 🙃

PR #5301, opened 2026-04-23
v3.3: cleanup: removed json-style format from yaml examples

karenetheridge:
...and fixed a deprecated "value" in an example

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -966,7 +966,7 @@ examples:
     dataValue:
       - 12345678
       - 90099
-    serializedValue: "12345678,90099"
+    serializedValue: 12345678,90099
 ```
 
 
@@ -993,7 +993,7 @@ examples:
     dataValue:
       greeting: Hello%2C%20world!
       code: 42
-    serializedValue: "greeting=Hello%2C%20world!; code=42"
+    serializedValue: greeting=Hello%2C%20world!; code=42
 ```
 
 A cookie parameter relying on the percent-encoding behavior of the default `style: "form"`:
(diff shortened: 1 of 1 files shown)


handrews:
I did all of that on purpose.  I'm not going to fight over it, but it was intentional to make it feel more JSON-like for JSON examples.

src/oas.md
```diff
@@ -1449,22 +1449,20 @@ application/json:
...
-        "color": "White",
-        "gender": "male",
-        "breed": "Persian"
-      }
+      value:
+        name: Fluffy
+        petType: Cat
+        color: White
+        gender: male
+        breed: Persian
```
notEthan, line 1457: The immediately preceding text no longer makes sense...

> Alternatively, since all JSON is valid YAML, the example value can use JSON syntax within a YAML document

... since the example value now does not use JSON syntax

src/oas.md
```diff
@@ -2446,12 +2444,7 @@ content:
           author: A. Writer
           title: An Older Book
           rating: 4.5
-        serializedValue: |
-          {
-            "author": "A. Writer",
-            "title": "An Older Book",
-            "rating": 4.5
-          }
+        serializedValue: {"author":"A. Writer","title":"An Older Book","rating":4.5}
```
notEthan, line 2447: this changes serializedValue from a string to a json object, does not look correct

notEthan:
It's not touched by this PR, but related - I notice that https://spec.openapis.org/oas/v3.2.0.html#json-examples says "When writing in YAML, JSON syntax can be used for `dataValue` (as shown in the `noRating` example)" but then the dataValue for noRating does not use JSON syntax.

karenetheridge:
Since there isn't a consensus that the changes I was going for in this patch are positive, I'm closing.

PR #5302, opened 2026-04-23
v3.2: remove json markup

karenetheridge:
...and fixed a deprecated "value" in an example

v3.2 port of #5301


handrews:
As noted for 3.3, I did all of that on purpose. I'm not going to fight over it, but it was intentional to make it feel more JSON-like for JSON examples.

PR #5324, opened 2026-05-12
v3.3: include an ABNF for contentType

### LINKED ISSUE #4788: v3.x: Does the Encoding Object's contentType allow whitespace around commas?
I expected it would (e.g. `contentType: text/css, text/javascript, image/*`) by analogy with the `Accept` header, while @ralfhandl read it more literally and expects only `contentType: text/css,text/javascript,image/*` would be allowed.  I suppose it depends on your assumptions on the phrase "comma-separated list" which is all we say about it.

karenetheridge:
closes #4788

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -1715,7 +1715,7 @@ These fields MAY be used either with or without the RFC6570-style serialization 
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
-| <a name="encoding-content-type"></a>contentType | `string` | The `Content-Type` for encoding a specific property. The value is a comma-separated list, each element of which is either a specific media type (e.g. `image/png`) or a wildcard media type (e.g. `image/*`). The default value depends on the type as shown in the table below. |
+| <a name="encoding-content-type"></a>contentType | `string` | The `Content-Type` for encoding a specific property. The value is a comma-separated list, each element of which is either a specific media type (e.g. `image/png`) or a [media type range](https://www.rfc-editor.org/rfc/rfc9110.html#appendix-A) (e.g. `image/*`). The default value depends on the type as shown in the table below. |
 | <a name="encoding-headers"></a>headers | Map[`string`, [Header Object](#header-object) \| [Reference Object](#reference-object)] | A map allowing additional information to be provided as headers. `Content-Type` is described separately and SHALL be ignored in this section. This field SHALL be ignored if the media type is not a `multipart`. |
 | <a name="encoding-encoding"></a>encoding | Map[`string`, [Encoding Object](#encoding-object)] | Applies nested Encoding Objects in the same manner as the [Media Type Object](#media-type-object)'s `encoding` field. |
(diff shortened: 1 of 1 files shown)


ralfhandl (dismissed):

src/oas.md
```diff
@@ -1715,7 +1715,7 @@ These fields MAY be used either with or without the RFC6570-style serialization
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
-| <a name="encoding-content-type"></a>contentType | `string` | The `Content-Type` for encoding a specific property. The value is a comma-separated list, each element of which is either a specific media type (e.g. `image/png`) or a wildcard media type (e.g. `image/*`). The default value depends on the type as shown in the table below. |
+| <a name="encoding-content-type"></a>contentType | `string` | The `Content-Type` for encoding a specific property. The value is a comma-separated list, each element of which is either a specific media type (e.g. `image/png`) or a wildcard media type (e.g. `image/*`), with ABNF: `media-range *( "," OWS media-range )`. The default value depends on the type as shown in the table below. |
```
handrews: OK I know I just approved the other one, but it just occurred to me that we may need to say that the OWS rule comes from RFC9110?  Or that the whole rule is imported from there?

I know @char0n tends to check our ABNF rules closely :-)

mikekistler (dismissed):
Looks good. 👍

src/oas.md
char0n: Hi,

@handrews thanks for flagging me ;]

I would possibly align this with [Request Object.content](https://spec.openapis.org/oas/v3.2.0.html#fixed-fields-10) field:

```suggestion
| <a name="encoding-content-type"></a>contentType | `string` | The `Content-Type` for encoding a specific property. The value is a comma-separated list, each element of which is either a specific media type (e.g. `image/png`) or [media type range](https://tools.ietf.org/html/rfc9110#appendix-A) (e.g. `image/*`). The default value depends on the type as shown in the table below. |
```



---

Then If can suggest replacing the following section with:

### Original section:

Determining how to handle a type value of null depends on how null values are being serialized. If null values are entirely omitted, then the contentType is irrelevant. See [Appendix B](https://spec.openapis.org/oas/v3.2.0.html#appendix-b-data-type-conversion) for a discussion of data type conversion options.

### Replaced by:

Determining how to handle a type value of null depends on how null values are being serialized. If null values are entirely omitted, then the contentType is irrelevant. See [Appendix B](https://spec.openapis.org/oas/v3.2.0.html#appendix-b-data-type-conversion) for a discussion of data type conversion options.

The `contentType` field is defined by the following [[ABNF](https://spec.openapis.org/oas/v3.2.0.html#bib-abnf)] syntax:

```abnf
encoding-content-type = media-range *( "," OWS media-range )
```

Here, `media-range` and `OWS` are taken from [[RFC9110](https://spec.openapis.org/oas/v3.3.0.html#bib-rfc9110)].

miqui (approved):

ralfhandl (approved):

char0n:
Great!

PR #5330, opened 2026-05-13
v3.3: be clear which property we are referring to

karenetheridge:
It's not entirely clear whether this refers to a property in the data instance or a `property` keyword in the schema. (v3.1 says that encoding properties must correspond to schema properties; in v3.2, the "Encoding by Position" section says that prefixEncoding entries must correspond to a schema indicating an array.)

Either way, let's clarify this and be consistent with the rules between object data and array data.


handrews (approved):
Great catch, I assume you will backport to 3.2.1?

mikekistler (approved):
Looks good. 👍

ralfhandl (approved):

PR #5331, opened 2026-05-13
v3.3: use extra markup to clarify intent

karenetheridge:
We are specifically talking about the "encoding" keyword here, and the multipart/form-data media-type (most other multipart types decode to arrays).

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -1380,10 +1380,10 @@ Array properties MUST be handled by applying the given Encoding Object to produc
 For all other value types for both top-level non-array properties and for values, including array values, within a top-level array, the Encoding Object MUST be applied to the entire value.
 The order of these name-value pairs in the target media type is implementation-defined.
 
-For `application/x-www-form-urlencoded`, the encoding keys MUST map to parameter names, with the values produced according to the rules of the [Encoding Object](#encoding-object).
+For `application/x-www-form-urlencoded`, the `encoding` keys MUST map to parameter names, with the values produced according to the rules of the [Encoding Object](#encoding-object).
 See [Encoding the `x-www-form-urlencoded` Media Type](#encoding-the-x-www-form-urlencoded-media-type) for guidance and examples, both with and without the `encoding` field.
 
-For `multipart`, the encoding keys MUST map to the [`name` parameter](https://www.rfc-editor.org/rfc/rfc7578#section-4.2) of the `Content-Disposition: form-data` header of each part, as is defined for `multipart/form-data` in [[!RFC7578]].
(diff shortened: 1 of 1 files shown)


src/oas.md
```diff
@@ -1379,10 +1379,10 @@ Array properties MUST be handled by applying the given Encoding Object to produc
...
 The order of these name-value pairs in the target media type is implementation-defined.
 
-For `application/x-www-form-urlencoded`, the encoding keys MUST map to parameter names, with the values produced according to the rules of the [Encoding Object](#encoding-object).
+For `application/x-www-form-urlencoded`, the `encoding` keys MUST map to parameter names, with the values produced according to the rules of the [Encoding Object](#encoding-object).
 See [Encoding the `x-www-form-urlencoded` Media Type](#encoding-the-x-www-form-urlencoded-media-type) for guidance and examples, both with and without the `encoding` field.
 
-For `multipart`, the encoding keys MUST map to the [`name` parameter](https://www.rfc-editor.org/rfc/rfc7578#section-4.2) of the `Content-Disposition: form-data` header of each part, as is defined for `multipart/form-data` in [[!RFC7578]].
+For `multipart/form-data`, the `encoding` keys MUST map to the [`name` parameter](https://www.rfc-editor.org/rfc/rfc7578#section-4.2) of the `Content-Disposition: form-data` header of each part, as is defined for `multipart/form-data` in [[!RFC7578]].
```
handrews: @karenetheridge I agree with you in general here, but there's a bit of history.

The reason this does not say `/form-data` is because of the paragraph further down along the lines of "MAY use a `name` parameter with other `multipart` but most multipart implementations probably won't support it."

I'm unsure what to do here.  That part was added as a possible workaround for lack of proper sequential multipart support, and I somewhat regret it as now it's this weird thing that makes it harder to make clear statements like this.

We could say "For `multipart/form-data` (or any `multipart` format making use of a `name` paraameter)..."?

karenetheridge: How about "For `multipart` types that decode to an object... such as `multipart/form-data` in RFC7578" ?

handrews: I think that works.  In theory there could be some other `multipart` that encodes to an object by some other mechanism, but... I have never heard of such a thing and I've looked, so this should be fine.

mikekistler (dismissed):
This looks fine to me, assuming the resolution of Henry's concerns. 👍

karenetheridge:
I reworded the text about object encodings.

ralfhandl (approved):

ralfhandl (approved):

miqui (approved):

PR #5332, opened 2026-05-13
v3.3: describe stringified response codes without referencing a specific formaat

karenetheridge:
In JSON, property keys are always strings; in YAML they may or not be depending on the parser, but other formats may also be in use (TOML, TOON etc). What matters is the resulting type(s) of the decoded data, not the specific formatting used for the serialized form.


src/oas.md
```diff
@@ -2129,7 +2129,7 @@ call.
...
+| <a name="responses-code"></a>[HTTP Status Code](#http-status-codes) | [Response Object](#response-object) \| [Reference Object](#reference-object) | Any [HTTP status code](#http-status-codes) can be used as the property name, but only one property per code, to describe the expected response for that HTTP status code. This field MUST be represented by a string (for example, `"200"`) for compatibility between JSON and YAML. To define a range of response codes, this field MAY contain the uppercase wildcard character `X`. For example, `2XX` represents all response codes between `200` and `299`. Only the following range definitions are allowed: `1XX`, `2XX`, `3XX`, `4XX`, and `5XX`. If a response is defined using an explicit code, the explicit code definition takes precedence over the range definition for that code. |
```
handrews: @karenetheridge this is definitely more clear about the intent.

Would it make sense to say something about that example being specific to YAML syntax, such as:

```suggestion
| <a name="responses-code"></a>[HTTP Status Code](#http-status-codes) | [Response Object](#response-object) \| [Reference Object](#reference-object) | Any [HTTP status code](#http-status-codes) can be used as the property name, but only one property per code, to describe the expected response for that HTTP status code. This field MUST be represented by a string (for example, `"200"` enclosed in quotes in YAML) for compatibility between JSON and YAML. To define a range of response codes, this field MAY contain the uppercase wildcard character `X`. For example, `2XX` represents all response codes between `200` and `299`. Only the following range definitions are allowed: `1XX`, `2XX`, `3XX`, `4XX`, and `5XX`. If a response is defined using an explicit code, the explicit code definition takes precedence over the range definition for that code. |
```

karenetheridge: sure.

handrews (approved):

mikekistler (approved):
Looks good. 👍

ralfhandl (approved):

PR #5333, opened 2026-05-13
v3.3: remove unnecessary use of "content"

karenetheridge:
Appendix D no longer talks about percent-encoding of headers, as the specification is now clear that percent-encoding of headers is never done (with the exception of the use of style:form in Cookie headers).


handrews (approved):
Oops 🤦 
Thanks!

ralfhandl (approved):

miqui (approved):

PR #5357, opened 2026-06-03
ci: adds browser installation step to unblock workflow

baywet:
based on https://github.com/OAI/Overlay-Specification/pull/343
I don't think we need it anywhere else, but let me know if I missed anything


baywet:
CC @karenetheridge whom I think has a bunch of PRs blocked.

karenetheridge (approved):

miqui (approved):

PR #5365, opened 2026-06-05
AI policy proposal

lornajane:
Following on from our [discussion](https://github.com/OAI/OpenAPI-Specification/discussions/5209) and some of the recent TDC meetings, I've put together a starting point for an AI policy. We're agreed that AI should be permitted, but that there's an important point about the effort ratio between contributions and maintenance tasks.

Claude proposed the outline of the doc after reviewing my preferred AI example policies, and also copyedited my document when finished.

### DIFF
--- a/AI.md
+++ b/AI.md
@@ -1,0 +1,62 @@
+# OpenAPI Initiative AI Policy
+
+The OpenAPI Initiative is a project that brings together experienced industry experts to collaborate on an open, shared standard that benefits us all.
+It is the result of shared ideas, thoughtful discussion, and careful curation over time.
+As the tools landscape changes, the way we work together on that shared vision should not change.
+
+Well-thought-out and meaningful contributions will always be welcome and considered for their merits, however the advent of AI-generated interactions makes this policy a necessary addition to the project.
+We encourage humans to engage and to gain experience with the project and further the mission, but our volunteer maintainers are not to be put in a position of expending more effort on reviewing contributions than was used to create them because the project is unsustainable in that scenario.
+
+TL;DR we do permit use of AI, with disclosure, but low-effort, inaccurate or overly verbose contributions will be dismissed.
+
+## AI and Standards Development
+
+The nature of this project provides a very particular context and any tool (including AI) should be applied with consideration to the context.
+The tools you use day-to-day can feel helpful, but we expect you to consider how they may best add value to each task within the project.
+
+Standards development means devising and developing standards that do not yet exist, with knowledge of the standards landscape and ongoing work in other standards projects.
+Most AI tools are trained on existing or outdated content, making them less useful for standards development than they might be in other applications, so bear that perspective in mind when working with AI on specification improvements or proposals.
+
+## Permitted Use
+
+AI tools may be used to support contributions to OpenAPI Initiative projects, including use of AI for drafting, editing, checking consistency, or exploring ideas.
+However, you are solely responsible for every part of your submission — the content, its accuracy, and its appropriateness for the project.
+You should expect to be asked followup questions about any detail of your contribution and to have already considered your work as deeply as our maintainers will when they review it.
+Using AI can help you to do a good job but we expect you to use these tools intentionally and considerately for the project and audience.
+
+## Disclosure and Accountability
+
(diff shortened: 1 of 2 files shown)


AI.md
```diff
@@ -0,0 +1,62 @@
+# OpenAPI Initiative AI Policy
```
miqui, line 1: @lornajane  -  well-crafted, pragmatic policy that strikes the right balance for a standards body.

The point about AI tools being trained on existing or outdated content is particularly astute for a standards project.

Just a thought:

No ```Assisted-by```: git trailer convention. The broader open source ecosystem is converging on ```Assisted-by```: as a standard commit metadata convention for AI-assisted contributions. Adopting it would make enforcement more tractable and machine-readable, especially for tooling.   

Thanks @lornajane

AI.md
```diff
@@ -0,0 +1,62 @@
...
+If a reviewer asks why a particular design decision was made, "the AI suggested it" is not an acceptable answer.
+Contributions where the submitter cannot demonstrate understanding of the material will be closed without further discussion or explanation.
+
+## Discussion Conduct
+
+Discussions in OpenAPI Initiative projects are for humans.
+Comments that restate or summarise what has already been said, without advancing the conversation, add burden to volunteers and will be removed.
+A useful comment adds new information, identifies a constraint, asks a precise question, or proposes a concrete resolution.
+We expect you to make good-faith contributions that are concise and to the intended point; overly verbose or general contributions will be hidden or removed.
+Users who frequently have comments removed may be blocked from OpenAPI Initiative repositories.
```
ralfhandl:
```suggestion
```
I think this is included in section "Enforcement".

ralfhandl (commented):
Thanks for preparing this!

AI.md
```diff
@@ -0,0 +1,62 @@
...
+
+Do not use AI to generate issue comments, discussion replies, or PR descriptions. These are the primary record of design reasoning for the project; they must reflect genuine human thought.
```
baywet, line 43: The only one that I think we should allow here is the PR description.
I've been notoriously bad a hand writing PR descriptions, and thanks to your help, I *believe* I somewhat have improved.
However, lately I've been using AI to help me write better PR descriptions, based off the issue discussion, and the changes, and that's generally been much better than my poorly handwritten descriptions to set the state for other reviewers.

miqui: > using AI to help me write better PR descriptions

Yeah, been doing this for quite some time. I got tired of writing of good descriptions because I was just going faster and faster.   However, I ALWAYS reviewed the description before pushing.

lornajane: Perhaps the results are good with the tool in your hands, @baywet  but that's not at all been my experience. AI-generated PR descriptions without expert input will at best describe information that is shown in the diff, and at worst both get that wrong and invent an incorrect explanation for what is changed and why.

handrews: @baywet I would also make a distinction between you writing a PR description _with AI assistance_ to tell you if it reads well is distinct from someone generating a PR description by pointing AI at the code (which it may have written).  idk if @lornajane sees a similar distinction, but I don't read your PR descriptions or comments and think "meh, AI wrote that."

baywet: Aaah so the context of this sentence is essentially "don't use AI to write those things, by itself", not "don't use AI at all"? That makes sense. Is there some way we could tweak it to reflect this more closely?

miqui (approved):

AI.md
lornajane: I chose "generate" very intentionally. If the AI is generating the content that is shared: I don't think we want that. Those of you using the tools to augment what you want to say? I think we've already said elsewhere that sort of usage is not a problem.

AI.md
lornajane: I did consider the additional metadata requirements, but I can't see us enforcing it so I didn't include it. It's just one more thing that most contributors will ignore and that we don't have the reviewer bandwidth to get hung up on - and I don't think it helps that much so it's not included. Thanks for prompting me to explain though!

AI.md
baywet: (I can't resolve comments, but I think this addresses the last of my concerns, thanks)

ralfhandl (approved):

PR #5367, opened 2026-06-07
workaround: use old version 0.4.7 of linkspector

ralfhandl:
Linkspector 0.5.3 randomly fails for valid links in `src/oas.md`.

Moving back to Linkspector 0.4.7 fixes that. Unfortunately the older version runs into 403 responses for valid links in other markdown documents (`GOVERNANCE.md` and `style-guide.md`), so two link patterns are added to the `ignorePatterns` list.

The linkspector action is hard-coded to the most recent version, so it is replaced by a direct call to the "locally" installed linkspector.

### DIFF
--- a/.github/workflows/validate-markdown.yaml
+++ b/.github/workflows/validate-markdown.yaml
@@ -19,7 +19,7 @@ jobs:
 
       - uses: actions/setup-node@v6 # setup Node.js
         with:
-          node-version: "20.x"
+          node-version: "24.x"
 
       - name: Lint work-in-progress spec
         run: npx --yes markdownlint-cli2 --config spec.markdownlint.yaml src/oas.md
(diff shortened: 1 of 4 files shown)


.github/workflows/validate-markdown.yaml
```diff
@@ -36,8 +36,12 @@ jobs:
           echo "PUPPETEER_EXECUTABLE_PATH=/usr/bin/google-chrome" >> $GITHUB_ENV
 
       - name: Check links in markdown files
-        uses: umbrelladocs/action-linkspector@v1
-        with:
-          reporter: github-check
-          fail_level: any
-          filter_mode: file
+        # uses: umbrelladocs/action-linkspector@v1
```
lornajane, line 39: Suggest we remove rather than comment out. Also do we still need the workaround that was added on line 30 ish?

miqui: Good to have as ver 20 is very old

miqui (approved):

.github/workflows/validate-markdown.yaml
ralfhandl: > Suggest we remove rather than comment out

We want to switch back to the action once Linkspector is fixed.

> do we still need the workaround that was added on line 30 ish?

Yes, the fix is also needed for the old Linkspector version.

PR #5368, opened 2026-06-08
ci: pass linkspector config to workflow

baywet:
## Summary
- pass the repository Linkspector config file explicitly to the GitHub Action
- ensure the action uses the existing URL suppression rules from .linkspector.yml

## Validation
- 
pm test
- 
pm run validate-markdown *(fails locally in Linkspector startup with Network.enable timed out)*

### DIFF
--- a/.github/workflows/validate-markdown.yaml
+++ b/.github/workflows/validate-markdown.yaml
@@ -38,6 +38,7 @@ jobs:
       - name: Check links in markdown files
         uses: umbrelladocs/action-linkspector@v1
         with:
+          config_file: .linkspector.yml
           reporter: github-check
           fail_level: any
           filter_mode: file


miqui (dismissed):

miqui (approved):

baywet:
@miqui I don't have merge permissions, so you'll need to perform the merge. Thanks.

ralfhandl:
I am hesitant to merge this as `umbrelladocs/action-linkspector@v1` uses Linkspector v0.5.3 which randomly flags correct links as broken in `src/oas.md` in the development branches.

baywet:
@ralfhandl I think among all of us, you have the most experience with that tooling. I was just trying to unblock work on both this repository and the Overlays one. If you're satisfied with the current approach (pinning the version, using an npm script), I'll replicate that to overlays. Just let me know.

ralfhandl:
The target is using the action and the latest Linkspector version. If that works for Overlays, just keep it.

I don't know why Linkspector 0.5.3 behaves so erratic for the OAS markdown source file - maybe it is "too large"?

This needs some experimenting, but "unfortunately" I can easily find more interesting things to do 😎

baywet:
A couple of calls ago @handrews was making a compelling argument that in order to support SAFs, and potentially additional SIGs, we'll need to align those automations across repositories (and potentially move it to an "engineering systems" repo), so it doesn't become a nightmare to manage/use.

I'm not married to any given solution for this specific problem, but I *do think* this should be aligned across repositories.

What do you think?

ralfhandl:
We dropped the idea of having a common tools repo for the respec and schema publishing stuff for Arazzo, OpenAPI, and Overlays because the repos are structured differently, for example how branches are used for different versions. Can't find the issue/discussion where we documented that decision.

If all SAF/SIG repos are structured identically, a common tools repo might work for them.

One question of course would be: should the new repos use the simple Overlays approach with version files in the `main` branch, or the more complicated Arazzo/OpenAPI approach with version branches?

With differently structured repos, plus more or less subtle differences in the respec processing, "aligning" means "copy, paste, and adjust".

ralfhandl:
Linkspector issue https://github.com/UmbrellaDocs/linkspector/issues/177 is fixed with v0.5.5 and `action-linkspector` uses that version.

Issue #568, opened 2016-02-17
Is OpenAPI an Open World or a Closed World Contract?


darrelmiller:
Forgive me for the nebulous issue, but as we head towards a v.Next document, I believe that understanding the scope of what OpenAPI is trying to define will help in making the many decisions that need to be made to get to v.Next.  

The terms [Open World](https://en.wikipedia.org/wiki/Open-world_assumption) and [Closed World](https://en.wikipedia.org/wiki/Closed-world_assumption) seem to have reasonable definitions on Wikipedia.  Let me try and apply the concepts to Open API.

Consider any arbitrary Open API document that describes an API.  It is assumed that the document accurately describes the API.  We sometimes use the word contract to define the role that the Open API document plays between the client and the server.

Is the OpenAPI document still valid if 
- there exists a resource in the API that is not described?
- a resource accepts a parameter that is not described?
- a resource supports a representation format that is not described?
- a resource returns a status code that is not described?

In a closed world, the API Description describes exactly what is supported, if it is not described, then it must not exist.  In an open world we cannot make assumptions about what is not described.  There may be other resources, other supported formats, other parameters.

I get the feeling from the conversations I have had with people who use OpenAPI that many perceive it to be a closed world description.  Certainly I believe that many people would say that returning an unspecified status code would be a contract violation.

Which brings us to the challenge.  Closed world systems work well when the entire system is under the control of the creator.  However, in the world of network based HTTP systems, many parts of the system are not under the control of the API designer.  It is quite common for intermediary components to return status codes and media types that the API designer may have no knowledge of.  It is quite common for APIs to change in ways that are additive and non-breaking and should not be considered a contract violation.

For this reason I do not believe that OpenAPI can be a truly closed world specification.  My fear is that we continue to pretend that it is a closed world specification and client developers continue building clients based on a false assumption that the only thing that is true is what is described in the specification document.  I am convinced this is a common cause of buggy client applications.

Another less than ideal outcome would be if we try and hold some middle ground where some parts of the specification must be exhaustively described, whereas other parts only describe some subset of what is possible.  This is going to further challenge us when it comes to identifying what should be considered a breaking change and therefore require a version change.

I would like to see wording in the Open API specification that states that what is being described is what can happen, but the behavior of the API is not limited by the description document.  It would be reasonable to state that behaviour that is described will not be removed without a corresponding version change.  Obviously no such guarantee is made about the availability of undescribed behavior!

[...]

PR #5391, opened 2026-06-12
v3.3: describe behaviour when deserializing a multipart with a Content-Type header

karenetheridge:
This is not really a breaking change, since I think what's described here is intuitively the right thing to do anyway, but as it's a new directive I'd still only add it in 3.3 and not backport it to 3.2.1.

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -1872,7 +1872,7 @@ Note that there are significant restrictions on what headers can be used with `m
 
 ##### Handling Multiple `contentType` Values
 
-When multiple values are provided for `contentType`, deserializing remains straightforward as the part's actual `Content-Type` SHOULD be included as a header of that part and SHOULD match one of the provided `contentType` values.
+When multiple values are provided for `contentType`, deserializing remains straightforward as, per [[RFC 2046]] [Section 5.1](https://datatracker.ietf.org/doc/html/rfc2046#section-5.1) each part will include a `Content-Type` header, or can reasonably be assumed to be `text/plain; charset=US-ASCII`.  To ensure interoperability, OAD authors SHOULD ensure that all reasonably expected media types are provided in `contentType`.
 
 For encoding and serialization, implementations MUST provide a mechanism for applications to indicate which media type is intended.
 Implementations MAY choose to offer media type sniffing ([[SNIFF]]) as an alternative, but this MUST NOT be the default behavior due to the security risks inherent in the process.
--- a/src/oas.md
+++ b/src/oas.md
@@ -1734,6 +1734,9 @@ This table is based on the value to which the Encoding Object is being applied a
 Note that in the case of [Encoding By Name](#encoding-by-name), this value is the array item for properties of type `"array"`, and the entire value for all other types.
 Therefore the `array` row in this table applies only to array values inside of a top-level array when encoding by name.
 
+When deserializing a multipart message, if a `Content-Type` header is present in the specific part, its value SHALL be used instead of these default values. Behavior is undefined when both this header and `contentType` are defined but with different values; it is RECOMMENDED to use the value of `Content-Type`. However, this behavior may produce unexpected results due to the possibility that the `schema` is not structured for the received `Content-Type`.
+
+
 | `type` | `contentEncoding` | Default `contentType` |
 | ---- | ---- | ---- |
 | [_absent_](#working-with-binary-data) | _n/a_ | `application/octet-stream` |
@@ -1869,7 +1872,7 @@ Note that there are significant restrictions on what headers can be used with `m
 
 ##### Handling Multiple `contentType` Values
 
-When multiple values are provided for `contentType`, parsing remains straightforward as the part's actual `Content-Type` is included in the document.
+When multiple values are provided for `contentType`, deserializing remains straightforward as the part's actual `Content-Type` SHOULD be included as a header of that part and SHOULD match one of the provided `contentType` values.
 
 For encoding and serialization, implementations MUST provide a mechanism for applications to indicate which media type is intended.
 Implementations MAY choose to offer media type sniffing ([[SNIFF]]) as an alternative, but this MUST NOT be the default behavior due to the security risks inherent in the process.


src/oas.md
```diff
@@ -1729,6 +1729,8 @@ This table is based on the value to which the Encoding Object is being applied a
 Note that in the case of [Encoding By Name](#encoding-by-name), this value is the array item for properties of type `"array"`, and the entire value for all other types.
 Therefore the `array` row in this table applies only to array values inside of a top-level array when encoding by name.
 
+When deserializing a multipart message, if a `Content-Type` header is present in the specific part, its value shall be used instead of these values. Behavior is undefined when both this header and `contentType` are defined but with different values.
+
```
handrews: @karenetheridge if I am reading this right, the only potential change here is in the case where `contentType` is absent and therefore assumed to have a default value, but the per-part `Content-Type` type contradicts that default value?

I think we can make this fit the compatibility guidelines more cleanly by switching the order of the sentences and changing "shall" to "SHOULD" (or "RECOMMENDED" which means the same thing but fit better in the sentence I'm proposing):

Something like this?  But less awkward with the last part?

```suggestion
When deserializing a multipart message, behavior is undefined when the `Content-Type` header and the `contentType` field have different values.  If the `contentType` field is absent but `Content-Type` is present, it is RECOMMENDED to use the value of `Content-Type`.  However, this behavior is still technically undefined due to the possibility that the `schema` is not structured for the received `Content-Type`.
```

src/oas.md
```diff
@@ -1864,7 +1866,7 @@ Note that there are significant restrictions on what headers can be used with `m
 
 ##### Handling Multiple `contentType` Values
 
-When multiple values are provided for `contentType`, parsing remains straightforward as the part's actual `Content-Type` is included in the document.
+When multiple values are provided for `contentType`, parsing remains straightforward as the part's actual `Content-Type` is included in the document; it SHOULD match one of the media types in `contentType`, if it is present.
```
handrews: I think that with the re-ordered sentences above, this is not necessary as it is simply a consequence of the other statement.  It's not entirely clear to me that we can put a SHOULD on the API behavior, though.  I think we can only put SHOULDs on our side of things.  We should maybe look and see if we have other SHOULDs that impact API behavior rather than OAS tool behavior — If we already do, then I'm perfectly happy for us to add another.

src/oas.md
karenetheridge: > The only potential change here is in the case where contentType is absent and therefore assumed to have a default value, but the per-part Content-Type type contradicts that default value?

No, I am not presuming that the default value would be considered here at all -- I think the Content-Type header in the message part should be used in preference to that: a value being underspecified means that we do the best with whatever information we do have, and in this case we have a header telling us what type the message sender intends it to be.

Your edit sounds fine, except I would remove the last sentence: the schema may be just fine with the Content-Type (consider the case where encoding/contentType says "text/plain; charset=UTF-8" but the Content-Type header says "text/plain; charset=Shift_JIS"... or application/yaml instead of application/json).

The logic I'm using is:


  - if there is style/explode/allowReserved present, use those;
  - otherwise, check for encoding/contentType AND a Content-Type header:
    - if we have just one or the other, use it;
    - else if we have both, they must match, else generate an error, but still decode with the Content-Type header (because we can presume the message sender knows what they're doing);
    - else (we have neither): do we have a schema for this data location?
      - yes, examine the schema to see what type this should be, and select the media-type on that basis; if it's "text/plain", also apply the charset from any adjacent `_charset_` message part.
      - no: do nothing; leave the payload as a string (and now we start unwinding our recursion as there is nothing more we can do with this part of the messge).

However, while looking at my test cases a little more closely, I'm wondering if it might be more sensible to simply prefer Content-Type over encoding/contentType _without erroring at all_ when the values conflict.  contentType is useful as a recipe for serializing a message, and a hint for how to deserialize when insufficient information is present in the message itself, but when Content-Type _is right there in the message next to the data_, surely we should just go ahead and use it? That's what an application would do in the absence of an OAD telling them what the message should look like.  If the data decodes to the desired result, why should this be an error at all?

\[...and after experimenting with tests, I am convinced this is correct.. it would not be useful to generate an error when we're still able to successfully deserialize the message. The schema is still there as a verification step to catch issues with the deserialized content, and there's a good chance that it came out correctly anyway even if the media-type didn't match what the OAD said the serialization process ought to be using. If a media-type was used that we don't know how to decode, then we can error on _that_ instead.]

karenetheridge: (edited to fix some errors; read in web rather than in email)

src/oas.md
karenetheridge: > It's not entirely clear to me that we can put a SHOULD on the API behavior, though. I think we can only put SHOULDs on our side of things

I'm not sure what you mean? What is "our side"?

karenetheridge: but anyway, yes, I agree with removing this addition, as it is unneeded; of course we want the Content-Type header to match encoding/contentType, but if they don't there's nothing useful the deserializer can do about it now (as above, generating an error about this is not helpful assuming we were able to deserialize anyway).

(Also, I feel good that the original statement here is consistent with my assumption that the Content-Type header is the value we should prefer over anything else) :)

src/oas.md

> I'm wondering if it might be more sensible to simply prefer Content-Type over encoding/contentType without erroring at all when the values conflict.

It might be more sensible (in some cases I'm pretty sure it is more sensible, and it definitely fits more with the architecture of the web where you can have advance hints, but the runtime behavior is always what's real), but it's not the way the OAS is written (or more broadly, the way it is marketed as a way to define an API interface contract).

Our general paradigm is that if the runtime behavior doesn't match the OAD, then that's an error ***[EDIT: except I keep forgetting the "open world" approach, so maybe it's not?  See below]***.  Whether it's an error in the API or the OAD depends on the circumstances, but the general idea is that you should be able to tell when your API and its design contract fail to match.  In theory, whatever the OAD says that contradicts the `Content-Type` is an assertion that the application depends on, so if the OpenAPI tooling to ignores that mismatch, it is likely to cause problems downstream — precisely the sort of problem that a contract is supposed to prevent.

_(note: I just like em-dashes and know how to use them.  No AI was used on this comment 😛 )_

This is a philosophical point that would be worth revisiting in a 4.0.

src/oas.md
handrews: "our side" -> OAS tooling, the stuff that implements our requirements
"not our side" -> includes API behavior, although honestly I'm not sure I agree with my earlier point given the overall perspective of an OAD-as-contract.  There's a subtle distinction in that with OAD-as-contract it's the _mismatch_ that causes the error, but the error could be in either the contract or the API.  So _technically_ the OAS (by way of the OAD) is _not_ imposing a requirement on the API.  It's documenting an expectation, which could be erroneous either because the API is wrong or the expectation is wrong.

As you can tell, I'm thinking through this as I'm typing, so don't take it as any sort of settled law here.

src/oas.md
karenetheridge: What should we do about this? Should we add in 3.3 some stricter language that says that `contentType` (or one of the values, if comma-separated) MUST match the part's Content-Type, only to potentially remove it in 4.0? Or just leave it vague for now?

In my testing I also flagged another edge case: `contentType` says `text/plain`, but the Content-Type header says `text/plain; charset=Shift_JIS`. IMO this is not a mismatch; the header is simply being specific about the character encoding, and `contentType` is saying "any text/plain type will do".

src/oas.md
karenetheridge: I think there's definitely room for a philosophical discussion here as to what we consider our scope and what we expect the OAD to actually prescribe.  I've been coming at things mostly from a "deserialize incoming requests and outgoing responses, and validate as much as we can" perspective, as that aligned with my previous JSON Schema work, but there's a lot more than that (and once I exhaust myself with deserializing, I'll be pivoting to thinking more about serializing and client generation).. but other tools have done a lot more various things with the OAD and perhaps made some wildly different interpretations on some points.

karenetheridge: one more thought: I do think we should call out the mismatch, but maybe an actual error (which in my tooling would cause a request to be rejected with an HTTP 400 error) might be too severe?  It's an interesting distinction between "we were able to deserialize this message completely using the data provided in the message, AND its deserialized form validated against the schema" and "something about the message's structure didn't match what the OAD said it should do".

src/oas.md
handrews: You're definitely correct about the `text/plain` not being a mismatch.

Regarding the rest of it... I have to apologize for shifting my views a bit again.  I'm even frustrating myself with this, and it's a good bit of evidence for why we should really nail down the philosophy of the OAS more clearly (which various folks here have attempted to do over the years but we've never quite managed it).

-----

This scenario is nearly analogous to matching (or not) the keys under the `content` field of a Response Object (as that one specifically allows media ranges).  We don't consider failure to match there an error, because many things in OpenAPI take an "open world" view (which is what I was forgetting before): an OAD describes things that can happen, and _if_ they happen, what they should look like.  But other things can also happen, and are not necessarily an error.  So if you get back `application/xml` but don't have a `content` key for that in the Response Object, it's not automatically an error.  It's just a thing you have no information about.

Or at least, this is how it should work in an "open world" view.  But we're not very clear about that being the view, I think.

It's not quite perfectly analogous because at the Response Object level the `content` keys can be related to the `Accept` request header, and there's no way to specify individual part media types in `Accept: multipart/whatever` AFAIK.  And the server is allowed to disregard `Accept` anyway.

So if we look at `contentType` as analogous to the Response Object (or Request Body Object)'s `content` keys, I guess a failed match is not an error?  It's just a thing we don't know much about?

I feel like I've gotten myself confused at this point 😵‍💫

src/oas.md
miqui: How about 409 ?

src/oas.md
karenetheridge: > So if you get back application/xml but don't have a content key for that in the Response Object, it's not automatically an error. It's just a thing you have no information about.

Interesting, that hasn't been my interpretation -- in my implementation, the way to get past a "wrong Content-Type" error is to explicitly provide a media-type entry for "\*/\*" (which matches anything that didn't match anything else).

I also error if there is no defined decoding for the media-type that matched under `content`, unless the schema is empty (`{}` or `true`) or missing entirely -- in these cases the string value of the content is returned without alteration.

handrews: > Interesting, that hasn't been my interpretation

Yeah, we're not very clear about it- obviously I forgot in this thread!  But I found [the big conversation about this from a while ago](https://github.com/OAI/OpenAPI-Specification/issues/568#issuecomment-186677860) which is interesting.

> is to explicitly provide a media-type entry for "/"

Wouldn't that need to be `*/*` ?  I didn't thin `/` on its own was a valid media type or media range?

> I also error if...

The thing is, we're not all that clear on how this is supposed to work.  That could be a good or bad thing.  It's a good thing in that tools can be set up for a variety of use cases.  "Fail-fast" is good for some use cases, while "if you can more-or-less make it work, keep going" is good for others.

I don't know what we should do here- let's talk about it in the Thursday meeting.

karenetheridge: that was `*/*` - markdown ate the \*.

karenetheridge:
I've pushed another commit that corrects the wording.

src/oas.md
```diff
@@ -1864,8 +1869,6 @@ Note that there are significant restrictions on what headers can be used with `m
 
 ##### Handling Multiple `contentType` Values
 
-When multiple values are provided for `contentType`, parsing remains straightforward as the part's actual `Content-Type` is included in the document.
-
```
handrews, line 1873: Note from today's TDC call:  Let's make sure we're preservint the guidance that if `contentType` has multiple values and `Content-Type` matches _one of those values_, then the deserialization (@karenetheridge 's better term than "parsing" here) is straightforward.

karenetheridge:
rebased

src/oas.md
```diff
@@ -1731,6 +1731,10 @@ This table is based on the value to which the Encoding Object is being applied a
 Note that in the case of [Encoding By Name](#encoding-by-name), this value is the array item for properties of type `"array"`, and the entire value for all other types.
 Therefore the `array` row in this table applies only to array values inside of a top-level array when encoding by name.
 
+When deserializing a multipart message, if a `Content-Type` header is present in the specific part, its value shall be used instead of these values. Behavior is undefined when both this header and `contentType` are defined but with different values.
```
handrews: The "shall" should either be a SHALL (or MUST) if we think we can mandate this within 3.2->3.3 compatibility rules, or it should be a SHOULD, noting that the only accepted reason for disregarding the SHOULD would be compatibility.

handrews (commented):
One more minor thing about normative language.  I assume since the last comment on this PR indicated just a rebase, that the "Let's make sure we're preserving the guidance" comment has not yet been addressed?

src/oas.md
karenetheridge: I think SHALL is fine here.

karenetheridge:
>  I assume since the last comment on this PR indicated just a rebase, that the "Let's make sure we're preserving the guidance" comment has not yet been addressed?

I looked, and it seemed to me at the time that we were already preserving the guidance, so there was no additional change needed.

src/oas.md
karenetheridge: However the line that comes after it is somewhat redundant and could even be contradictory.  So a bit more revision is needed here.

handrews:
@karenetheridge I don't see anything that specifically addresses multiple values _in_ the `contentType` field and using `Content-Type` to disambiguate that, but maybe it's in a non-changed block that I haven't noticed or I am otherwise overlooking it?  Or do you think it is just sufficiently obvious from the field definition as to not be worth calling out?

karenetheridge:
This line:

> if a `Content-Type` header is present in the specific part, its value SHALL be used instead of these values

Is this intended to only refer to `contentType`'s default value (calculated from the schema type)? I've never interpreted it that way, instead seeing it as referring to the `contentType`'s explicit value(s) as well as the default.

handrews:
Oh, I see the confusion now.  I read "these values" as applying to the defaults only because that line immediately follows the explanation of default behavior.

But if "these values" includes explicit `contentType` values, I don't understand the "Behavior is undefined when both this header and `contentType` are defined but with different values." line.  That line (correctly) treats an outright conflict as undefined.

So that leaves:

* default values, including a possible contradiction - I read the first line as treating `Content-Type` as an override.  Which now that I think of it, probably ought to be a SHOULD as a compliant 3.2 implementation could have implemented other behavior such as an error.  I'll defer to overall TSC opinion there, though.  It's in the murky area, but in 3.2 we usually, and possibly always, used SHOULDs for those.
* explicit value, not a contradiction (although the `Content-Type` might be more specific than a matching media range).  This is the part I was concernd was not being addressed, as it ought to just note that `Content-Type` disambiguates if `contenType` holds multiple values.

At least I think. I've gotten turned around on this one so much I don't entirely trust myself at this point.

But if I'm correct, I think all we need is:

* Clarify that the override behavior applies to defaults (if you agree with that) by using something other than "these values" in that line
* Clarify that `Content-Type` disambiguates multiple values when at least one is a match (since it would no longer be covered by the current "these values" line, if you agree with me here)
* The existing line that declares `Content-Type` not matching any explicit `contentType` values is fine as it is
* (TSC decision pending) maybe demote the SHALL to a SHOULD

karenetheridge:
I just pushed a change to bring back the bit that I had thought was redundant, and tweaked the wording:

```
-When multiple values are provided for `contentType`, parsing remains straightforward as the part's actual `Content-Type` is included in the document.
+When multiple values are provided for `contentType`, deserializing remains straightforward as the part's actual `Content-Type` SHOULD be included as a header of that part.
```

karenetheridge:
> At least I think. I've gotten turned around on this one so much I don't entirely trust myself at this point.

:) me too!

After re-reading several times I think your initial inclination to see that line as only applying to the default value is correct.

And, I think we're clear enough now about what to do in all cases:

- neither `contentType` nor a Content-Type header is provided: use schema inference to determine a media-type to use for decoding
- no `contentType`, but we have a Content-Type header: use that, and maybe the schema will be invalid due to a weird deserialization, oh well
- one `contentType`, no Content-Type header: use that
- more than one `contentType` provided, no Content-Type: might as well just try each one in turn and go with whatever can be deserialized ~~correctly~~ without errors (this is what I'll do when the new schema is released and I can implement multiple contentTypes)
- one or more `contentType`s provided, and a Content-Type header: use Content-Type header (I'm not going to generate an error for this, but it might be a runtime warning or lint or something, if the Content-Type isn't contained in the set of `contentType`s)

handrews:
@karenetheridge I think this point:

> one or more contentTypes provided, and a Content-Type header: use Content-Type header (I'm not going to generate an error for this, but it might be a runtime warning or lint or something, if the Content-Type isn't contained in the set of contentTypes)

is actually two cases:

* one or more `contentTypes` provided, and a `Content-Type` header that matches at least one media range in `contentTypes`: Use `Content-Type` (and if relevant somehow, note which media range was matched- idk if that ever matters, though)
* one or more `contentTypes` provided, and a `Content-Type` header that *does not match any of them*:  Use the `Content-Type` but with the same caveats as "no `contentType`, but we have a `Content-Type` header" as these are kind-of the same (just implicit default vs explicit `contentType`

the important distinction is that one of these cases is well-defined, and the other is not.  That's the guidance that I have been trying to make sure does not get dropped.

miqui (dismissed):
ok with the normative language changes, but lets see about the remaining unresolved comment.

src/oas.md
```diff
@@ -1734,6 +1734,10 @@ This table is based on the value to which the Encoding Object is being applied a
 Note that in the case of [Encoding By Name](#encoding-by-name), this value is the array item for properties of type `"array"`, and the entire value for all other types.
 Therefore the `array` row in this table applies only to array values inside of a top-level array when encoding by name.
 
+When deserializing a multipart message, if a `Content-Type` header is present in the specific part, its value SHALL be used instead of these values. Behavior is undefined when both this header and `contentType` are defined but with different values.
```
mikekistler: In the first sentence, does "these values" mean the default values described in the previous paragraph, or the explicit or default values of the fields of the encoding object? I'm guessing it is only the defaults, since otherwise the first and second sentences seem inconsistent. But I think it would be best to make that clear, e.g. "these values" -> "these default values".

src/oas.md
```diff
@@ -1734,6 +1734,10 @@ This table is based on the value to which the Encoding Object is being applied a
...
+If the `contentType` field is absent but `Content-Type` is present, it is RECOMMENDED to use the value of `Content-Type`.  However, this behavior may produce unexpected results due to the possibility that the `schema` is not structured for the received `Content-Type`.
```
mikekistler: And here I'm confused because if `Content-Type` is present, I think we already said one previous line that its value "SHALL" be used. So why here is it only "RECOMMENDED"? Isn't this inconsistent?

src/oas.md
karenetheridge: Yes, it's referring to the defaults, as the case of when `contentType` has an explicit value is covered separately. That's a good edit; thanks! (and applied)

src/oas.md
karenetheridge: nice catch. we can just delete that whole second sentence.

karenetheridge: I'll merge parts of that second sentence with the first, as we ought to be describing the pitfalls of using the Content-Type header in preference to the `contentType` encoding field.

karenetheridge: much better. now:

> +When deserializing a multipart message, if a `Content-Type` header is present in the specific part, its value SHALL be used instead of these default values. Behavior is undefined when both this header and `contentType` are defined but with different values; it is RECOMMENDED to use the value of `Content-Type`. However, this behavior may produce unexpected results due to the possibility that the `schema` is not structured for the received `Content-Type`.

src/oas.md
```diff
@@ -1869,7 +1872,7 @@ Note that there are significant restrictions on what headers can be used with `m
 
 ##### Handling Multiple `contentType` Values
 
-When multiple values are provided for `contentType`, parsing remains straightforward as the part's actual `Content-Type` is included in the document.
+When multiple values are provided for `contentType`, deserializing remains straightforward as the part's actual `Content-Type` SHOULD be included as a header of that part and SHOULD match one of the provided `contentType` values.
```
handrews: One tiny question regarding "SHOULD be included..." this "SHOULD" comes from the `multipart` specification, correct?  In which case we wouldn't use normative language as that is reserved for requirements we enforce, but would instead cite the relevant RFC.

For "SHOULD match one of the provided `contentType` values", this is obviously related to the OAS but we're talking about what the API sends, and I'm not sure we can put normative requirements on that either.  We put normative requirements on OAD authors and OAS implementors, so maybe:


```suggestion
When multiple values are provided for `contentType`, deserializing remains straightforward as, per [[RFC 2046]] [Section 5.1](https://datatracker.ietf.org/doc/html/rfc2046#section-5.1) each part will include a `Content-Type` header, or can reasonably be assumed to be `text/plain; charset=US-ASCII`.  To ensure interoperability, OAD authors SHOULD ensure that all reasonably expected media types are provided in `contentType`.
```

I think this says essentially the same thing, but defers to the `multipart` RFC for its behavior, and places the normative requirement on the OAD author rather than the API implementation, which I think is the place where we can require things.

karenetheridge: Okay, we normally just defer to the RFCs when applicable and don't go into the gory details, but I don't mind being explicit here. Let's just get this done. ;)

handrews (approved):
The only reason I included some "gory details" here is that the RFC is very old and its wording on the default assumptions when `Content-Type` is absent is weird, and I wanted to call attention to that.  It's not just that the default MUST be `text/plain...`, sadly.

miqui (approved):

PR #5395, opened 2026-06-17
Update .gitignore to include Claude AI config files

miqui:
Add rules to ignore Claude AI agent configuration files.

The config can define hooks that could be malicious.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,6 @@ deploy-preview/
 coverage/
 _site/
 Gemfile.lock
+# ignore local claude AI agent config
+.claude/settings.local.json
+.claude/*.local.json


lornajane:
I suggest that we don't try to go adding exclusions for every file pattern used by every tool out there. Please use a global gitignore config on your own system for files generated by your IDE or other tools that should be ignored.

lornajane:
(was closing it a bit harsh? Re-open if you want to keep the discussion going!)

miqui:
@lornajane 

>I suggest that we don't try to go adding exclusions for every file pattern used by every tool out there. Please use a global gitignore config on your own system for files generated by your IDE or other tools that should be ignored.

This assumes a user who clones the repo will follow the practice of managing their own .gitignore. As you know, not always the case.   This change is particularly important because I am trying to avoid someone enabling potentially dangerous hooks.   This goes beyond IDE settings, etc...etc.

(AI Search) below:
Why AI hooks are risky
Security research (e.g., “IDEsaster” and “TrustFall”) has shown that AI coding tools and CLIs can be driven to execute arbitrary commands or exfiltrate data via prompt injection plus auto-approved tool/hook execution.

Malicious repos and config files (CLAUDE.md, .claude/settings.json, .github/hooks/*.json, Gemini/Copilot hook configs, etc.) can plant or mutate hooks so they run with no or minimal user interaction.

Once an AI tool has write access to your repo and can trigger hooks, it becomes part of your supply chain and CI/CD attack surface, not just “autocomplete in the terminal.”

lornajane:
I think we're talking about two different things here. The .gitignore file is useful for avoiding mistakes with adding files such as built assets, log files that should not be committed to the repo, whatever tools people use they can configure their own ignores. Gitignore doesn't prevent matching files from being added, it just stops it from happening without clear intent. I don't think this change is going to protect us against malicious files in our repository that will cause an agent to run riot - our limited number of careful committers and diligent reviewers are the guard against that.

PR #5407, opened 2026-06-29
v3.3: provide additional guidance on usage of "style" with multipart/form-data

karenetheridge:
see also https://github.com/OAI/OpenAPI-Specification/discussions/5406

I'm assuming this change is too great to be backported to 3.2.1.

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -4893,9 +4893,14 @@ Implementations of this specification MAY use an implementation of RFC6570 to pe
 
 Note that when using `style: "form"` RFC6570 expansion to produce an `application/x-www-form-urlencoded` HTTP message body, it is necessary to remove the `?` prefix that is produced to satisfy the URI query string syntax.
 
-When using `style` and similar keywords to produce a `multipart/form-data` body, the query string names are placed in the `name` parameter of the `Content-Disposition` part header, and the values are placed in the corresponding part body; the `?`, `=`, and `&` characters are not used, and URI percent encoding is not applied, regardless of the value of `allowReserved`.
+When using `style` and/or `explode` to produce a `multipart/form-data` body, the query string name is placed in the `name` parameter of the `Content-Disposition` part header, and the value is placed in the corresponding part body; the `?`, `=`, and `&` characters are not used, and URI percent encoding is not applied, regardless of the value of `allowReserved`.
 Note that while [RFC7578](https://datatracker.ietf.org/doc/html/rfc7578) allows using [[RFC3986]] percent-encoding in "file names", it does not otherwise address the use of percent-encoding within the format.
 Users are expected to provide names and data with any escaping necessary for conformance with RFC7578 already applied.
+
+Due to these restrictions, the only `style` and `explode` combinations that can be used in `multipart/form-data` are:
+- `style: form` with any primitive type
+- `style: form` and `explode: false` with arrays and objects (note that due to the omission of the query string name, this serialization is equivalent to `style: simple`)
+- `style: spaceDelimited` or `style: pipeDelimited` and `explode: false` with arrays and objects
 
 Note also that not all RFC6570 implementations support all four levels of operators, all of which are needed to fully support the OpenAPI Specification's usage.
 Using an implementation with a lower level of support will require additional manual construction of URI Templates to work around the limitations.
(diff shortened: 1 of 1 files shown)


miqui:
hi @karenetheridge I went through #5406.  I did several passes to get a more solid understanding. However, I am still a bit confused. Is ```explode``` to be treated as a workaround or fix? Was it simply missing from the language in the spec?

karenetheridge:
@miqui I don't understand your question, so let me summarize what I'm seeing:

- issue #2018 was created because a gap in the spec was noticed in 3.0, that the equivalent of `collectionFormat: csv` was no longer available when serializing multipart/form-data messages
- PR #2066 was created to resolve this issue, by adding "or multipart/form-data" to the section of the spec that says "This field SHALL be ignored if the request body media type is not application/x-www-form-urlencoded" (ref. https://spec.openapis.org/oas/v3.0.4.html#fixed-fields-for-rfc6570-style-serialization); by doing so, it was then possible to use `type: array` or `type: object` with `style: form` and `explode: false` to create CSV values in multipart requests.
- however, no additional guidance was added about how the other style and explode combinations would work. Notably, arrays and objects with explode: true cannot be serialized or deserialized given the way the encoding object is constructed right now (encoding objects apply to multipart _parts_, not the message as a whole).

Therefore, I propose that we explicitly mention these non-functional combinations, and in a version of the spec that allows breaking backwards compatibility we prohibit them outright.

miqui (dismissed):

src/oas.md
```diff
@@ -4879,10 +4879,15 @@ Implementations of this specification MAY use an implementation of RFC6570 to pe
...
+When using `style` and/or `explode` to produce a `multipart/form-data` body, the query string name is placed in the `name` parameter of the `Content-Disposition` part header, and the value is placed in the corresponding part body; the `?`, `=`, and `&` characters are not used, and URI percent encoding is not applied, regardless of the value of `allowReserved`.
 Note that while [RFC7578](https://datatracker.ietf.org/doc/html/rfc7578) allows using [[RFC3986]] percent-encoding in "file names", it does not otherwise address the use of percent-encoding within the format.
 Users are expected to provide names and data with any escaping necessary for conformance with RFC7578 already applied.
 
+Due to these restrictions, the only `style` and `explode` combinations that can be used in `multipart/form-data` are:
+- `style: form` with any primitive type
+- `style: form` and `explode: false` with arrays and objects (note that due to the omission of the query string name, this serialization is equivalent to `style: simple`)
+- `style: spaceDelimited` or `style: pipeDelimited` and `explode: false` with arrays and objects
+
```
handrews: @karenetheridge does this modification make sense?

```suggestion
Due to these restrictions, the only `style` and `explode` combinations that can safely be used in `multipart/form-data` are:
- `style: form` with any primitive type
- `style: form` and `explode: false` with arrays and objects (note that due to the omission of the query string name, this serialization is equivalent to `style: simple`)
- `style: spaceDelimited` or `style: pipeDelimited` and `explode: false` with arrays and objects

The behavior of other combinations with `multipart/form-data` is undefined.
```

karenetheridge:
rebased, and suggested change accepted.

handrews (dismissed):

src/oas.md
```diff
@@ -4881,10 +4881,16 @@ Implementations of this specification MAY use an implementation of RFC6570 to pe
...
+When using `style` and/or `explode` to produce a `multipart/form-data` body, the query string name is placed in the `name` parameter of the `Content-Disposition` part header, and the value is placed in the corresponding part body; the `?`, `=`, and `&` characters are not used, and URI percent encoding is not applied, regardless of the value of `allowReserved`.
 Note that while [RFC7578](https://datatracker.ietf.org/doc/html/rfc7578) allows using [[RFC3986]] percent-encoding in "file names", it does not otherwise address the use of percent-encoding within the format.
 Users are expected to provide names and data with any escaping necessary for conformance with RFC7578 already applied.
 
+Due to these restrictions, the only `style` and `explode` combinations that can safely be used in `multipart/form-data` are:
+- `style: form` with any primitive type
```
handrews, line 4902: I believe this will fix the latest linting error:

```suggestion
Due to these restrictions, the only `style` and `explode` combinations that can safely be used in `multipart/form-data` are:

- `style: form` with any primitive type
```

handrews (approved):

handrews:
@miqui you'll need to re-approve this b/c of the latest linter fixes.

karenetheridge:
side note - I'm assuming this change is too great to be backported to 3.2.1.

miqui (approved):

miqui:
@karenetheridge > side note - I'm assuming this change is too great to be backported to 3.2.1.

I tend to agree with you.  Interested to see what others feel.

PR #5412, opened 2026-07-02
SAML 2.0 Bearer flow as alternative to Authorization Code flow

HeikoTheissen:
SAML 2.0 Bearer flow is an OAuth authentication flow with principal propagation which, unlike Authorization Code flow, does not prompt the user for consent. It is typically used for integration between business systems where trust has been set up centrally by administrators and individual users (employees) need not consent.

This pull request introduces SAML 2.0 Bearer flow as a new option into the OpenAPI Security Scheme Object.


handrews:
@HeikoTheissen  We will definitely look at this for 3.3, but there are two major things in flight that will need to settle first (just so you understand we're not ignoring you):

* We are moving security features out into their own specification in a separate repository so that we can make changes in this area more rapidly.  So at some point, this work will need to be moved over to the appropriate repo, which is not yet set up.
* We are also considering some larger changes to how we handle security, which might impact the approach here.  While there is a GitHub discussion on this, it is lengthy and there is a new proposal being written that will more directly summarize the intention.  This could lead to an entirely new security approach alongside the existing one, or it could be (like this PR) an enhancement of what we have.  We don't really know yet.

Things move rather slowly during the summer so it might be a while before we get to this, but we will respond once these other things are more clear.

PR #5404, opened 2026-06-25
v3.3: fix error: encoding objects also now apply to array items

karenetheridge:
- [x] no schema changes are needed for this pull request

PR #5403, opened 2026-06-25
v3.3: define a mechanism for preserving value/part order in forms

karenetheridge:
This change is accompanied by an edit to the "Forms" page in the Media-Type Registry (https://github.com/OAI/spec.openapis.org/pull/81)

- [x] no schema changes are needed for this pull request

PR #5420, opened 2026-07-02
v3.2: ports from 3.3 (June/July 2026)

karenetheridge:
Backports from 3.3 to 3.2 of recent spec edits.

Will continue populating this list until all 3.3 PRs are merged.

* #5403 a5edad7c 2026-06-03  encoding headers are not needed for serializing an array
* #5403 ed1d645d 2026-06-09  some media-types, e.g. multipart/form-data, define an explicit ordering
* #5403 92e84052 2026-06-24  define a mechanism for preserving value/part order in forms
* #5404 2f7b55b2 2026-06-03  fix error: encoding objects also now apply to array items
* #5390 3fb284b0 2026-05-30  set the type here, to ensure the body content is deserialized as an object
* #5386 03344fc9 2026-05-13  fix error: these keywords deal with items, not properties
* #5331 1432e75e 2026-05-13  use extra markup to clarify intent
* #5324 d1eb6f55 2026-05-12  include an ABNF for contentType (amended)
* #5330 7ce1fec3 2026-05-13  be clear which property we are referring to
* #5333 1189d3a4 2026-05-13  remove unnecessary use of "content"
* #5332 f1c2dcf2 2026-05-13  describe stringified response codes without referencing a specific format

I think all of these are suitable for inclusion into 3.2.1, as they are all clarifications rather than introducing new capabilities.


karenetheridge:
All the source PRs from v3.3 have been merged now, so this is ready to go. I edited each commit to indicate the original PR it came from.

karenetheridge:
..and in comparing the diff of this branch vs v3.3-dev, it looks like nothing was missed -- the only differences from 3.3 are in the new security changes, and my two PRs that tweak the handling of multipart messages (#5407 and #5391).

handrews:
@karenetheridge Were there any merge conflicts when porting these that you had to resolve?

karenetheridge:
No conflicts.

miqui (dismissed):

src/oas.md
```diff
@@ -1392,16 +1394,18 @@ See [Encoding `multipart` Media Types](#encoding-multipart-media-types) for furt
...
 As with `prefixItems`, it is _not_ an error if the instance array is shorter than the `prefixEncoding` array; the additional Encoding Objects SHALL be ignored.
 
 The `itemEncoding` field can also be used with `itemSchema` to support streaming `multipart` content.
 
 ##### Additional Encoding Approaches
 
-The `prefixEncoding` field can be used with any `multipart` content to require a fixed part order.
-This includes `multipart/form-data`, for which the Encoding Object's `headers` field MUST be used to provide the `Content-Disposition` and part name, as no property names exist to provide the names automatically.
+The `prefixEncoding` and/or `itemEncoding` fields, and/or usage of a [`schema` indicating an array type](#non-json-data) (or `itemSchema`), can be used with any `multipart` content to require a fixed part order.
+This includes `multipart/form-data`, in which the property name of each item is used to populate the `Content-Disposition` headers with the part name.
```
handrews: Was this intended to change the requirement around the Encoding Object's headers field when the data structure is an array? Actually I'm realizing now that I don't know what "property name" would be used with an array data structure?

handrews (commented):
One thing which might just be me misunderstanding/misremembering.

src/oas.md
karenetheridge: multipart/form-data, when presented as an array, looks like:

```json
[
  { name: value },
  { name: value },
  ...
]
```

This is also described at https://spec.openapis.org/registry/media-type/forms.

There is no new requirement here; it's just being more correct about how the serialization process uses the names and values from the data structure to construct the multipart body. There is no need to create a `header` object for `Content-Disposition`, as we already know what the part name should be from the data to be serialized itself.

handrews: My recollection is that this array form is new in 3.3.  The reason for the language around using `header` for `Content-Disposition` was that we were specifically talking about `[value, value, ...]` arrays.  I think the name-value pair array is great for 3.3 but it feels like a new requirement that can't go in 3.2.1.

It _could_ still be in the media type registry which makes it more opt-in?  I'm honeslty not sure what the policy ought to be here, so don't take this as an outright rejection.

karenetheridge: I've removed the commit that contains this change.

handrews (approved):

miqui (approved):

PR #5424, opened 2026-07-04
Migrate to new shared build infrastructure.

handrews:
This Codex-authored draft pull request is the companion to OAI/build-infra#1 and MUST NOT be merged until after that PR has been merged, and until after we have proven that the new system works with OAI/sig-lifecycle#1.  I am posting it here purely to illustrate the impact so the complete approach can be considered.  See OAI/build-infra#1 for full context and a discussion of the use of AI and how to review these changes.

This might try to run CI which may or may not work given the state of OAI/build-infra.  I'll look into any failures and either explain them or fix them prior to marking this ready to review.

This new system pushes all dependency management into the build-infra repository, eliminating the continual flood of dependabot updates and branch syncs we are currently experiencing.  It also makes it trivial to set up the SIG repositories for specificaiton publishing.


handrews:
Some failures were due to a platform-specific dependency installation problem.  This has been fixed, and the issue has been documented in the build-infra repository with the instructions on setting up repos to use the shared infrastructure.

The last failure was due to linkspector checking a link to a build-infra markdown file that is not yet on main, so I added it to the exclusion list temporarily.  I'll remove that commit from this branch before taking this out of draft, assuming we decide this PR is worth reviewing in its current form.

package.json
```diff
@@ -13,27 +13,17 @@
...
-  },
-  "devDependencies": {
-    "@hyperjump/json-schema-coverage": "^1.2.1",
-    "@umbrelladocs/linkspector": "0.5.5",
-    "@vitest/coverage-v8": "^4.1.9",
-    "c8": "^11.0.0",
-    "markdownlint-cli2": "^0.22.1",
-    "vitest": "^4.0.1",
-    "yaml": "^2.9.0"
+    "@oai/build-infra": "git+https://github.com/OAI/build-infra.git#main"
```
baywet, line 26: I think git installs are going to be disabled by default with npm 12 for security reasons. Maybe some we want to account for in the workflows here?

handrews: From what I understand, we will just need to pass `--allow-git=all`
We definitely do not want to publish this package to npm.

baywet: probably `--allow-git=root` would be better since it'd allow any git package at the root, but not transitive dependencies. That's been a common attack vector over the last year or so, hence why npm is disabling by default and adding those switches

handrews: @baywet is this something we can/should do now (and where)?  Or do we need to wait until npm 12 is out?

handrews:
Force-pushed (twice because I messed it up the first time) for rebase (no changes as package conflicts were just over having deleted pretty much everything in favor of the git repo).

Also one new commit to add `--allow-git=root` to all `npm ci` workflow installs.

package.json
handrews: @baywet I added it to the workflows.

.github/workflows/respec.yaml
```diff
@@ -36,7 +36,7 @@ jobs:
           node-version: "24.x"
 
       - name: Install dependencies
-        run: npm ci
+        run: npm ci --allow-git=top
```
baywet:
```suggestion
        run: npm ci --allow-git=root
```

I don't think "top" is a valid option, is it? npm ci --help on my local machine gives me this

```
[--allow-git <all|none|root>] [--allow-remote <all|none|root>]
```

handrews: @baywet I added a commit fixing this in all four places it appears.

handrews:
regarding the CI failure: I think this does not properly handle `main` (without `src/oas.md`) vs other branches.  I will look into it but it might be a few days.  setting back to draft for now.

handrews:
I'm going to close and re-open this to try to get CI unstuck, it seems to be hung up on something even with a few re-runs.

handrews:
The only remaining CI failure is due to the broken link fixed in PR #5471

handrews:
CI is passing!  This is ready for review — the most recent problems were purely linkspector issues, nothing actually in this PR has changed other than the `--allow-git=root` update in response to @baywet , plus the most recent update to which commit in build-infra we are pointing.

lornajane (approved):
Thanks, this is an experimental merge as discussed

PR #5440, opened 2026-07-18
add more cross-references for how to determine the json schema dialect

karenetheridge:
v3.2.0 §4.24.7 does talk about how `jsonSchemaDialect` and `$schema` interact and the effects of changing them, but this adds more references from where `jsonSchemaDialect` and the default dialect are introduced.


3huss99-ui (approved):

miqui (approved):

baywet (approved):

lornajane (approved):
Thanks

PR #5441, opened 2026-07-20
v3.3: remove invalid mention of prefixEncoding, itemEncoding here

karenetheridge:
In section 4.14.5.2 (Encoding by Position) we already said: "To use the prefixEncoding and/or itemEncoding fields, either itemSchema or a schema indicating an array type MUST be present.", however there is no parallel guidance in 4.14.5.1 (Encoding by Name) that says that the encoding field MUST also have a schema indicating an object type present.

(I already edited this sentence earlier in this release cycle to add mention of prefixEncoding alongside itemEncoding, but this entire sentence is inconsistent with the previous section, so we should either remove this sentence entirely, or soften the previous statement about prefixEncoding and itemEncoding requiring an array schema.)

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -1405,7 +1405,7 @@ The `itemEncoding` field can also be used with `itemSchema` to support streaming
 
 ##### Additional Encoding Approaches
 
-The `prefixEncoding` and/or `itemEncoding` fields, and/or usage of a [`schema` indicating an array type](#non-json-data) (or `itemSchema`), can be used with any `multipart` content to require a fixed part order.
+Usage of a [`schema` indicating an array type](#non-json-data) (or `itemSchema`), can be used with any `multipart` content to produce a fixed part order.
 This includes `multipart/form-data`, in which the property name of each item is used to populate the `Content-Disposition` headers with the part name.
 
 Prior versions of this specification advised using the [`name` parameter](https://www.rfc-editor.org/rfc/rfc7578#section-4.2) of the `Content-Disposition: form-data` header of each part with `multipart` media types other than `multipart/form-data` in order to work around the limitations of the `encoding` field.
--- a/src/oas.md
+++ b/src/oas.md
@@ -1397,7 +1397,8 @@ Data for these media types are modeled as an array, with one item per part, in o
 
 For applications that wish to preserve part order, `multipart/form-data` content may also be modelled as an array, with one item per part, in order (where each item consists of an object containing the name/value pair); the `schema` SHALL be used to determine whether deserializing to an `object` or an `array` is preferred. See examples in [Media Type Registry: Forms](https://spec.openapis.org/registry/media-type/forms).
 
-To use the `prefixEncoding` and/or `itemEncoding` fields, either `itemSchema` or a [`schema` indicating an array type](#non-json-data) MUST be present.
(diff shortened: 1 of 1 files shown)


src/oas.md
```diff
@@ -1405,7 +1405,7 @@ The `itemEncoding` field can also be used with `itemSchema` to support streaming
 
 ##### Additional Encoding Approaches
 
-The `prefixEncoding` and/or `itemEncoding` fields, and/or usage of a [`schema` indicating an array type](#non-json-data) (or `itemSchema`), can be used with any `multipart` content to require a fixed part order.
+Usage of a [`schema` indicating an array type](#non-json-data) (or `itemSchema`), can be used with any `multipart` content to produce a fixed part order.
```
karenetheridge, line 1409: I don't much like this statement either, because it is redundant for multipart content that is not form-data: other multipart types can only be deserialized to an array, as they have no part names, so there should be no need for a schema (or an encoding object, for that matter, if the part values are all ordinary strings, and even if they're serialized values, e.g. application/json, we don't need an `encoding` object to make use of `Content-Type` on the parts).

src/oas.md
```diff
@@ -1397,15 +1397,16 @@ Data for these media types are modeled as an array, with one item per part, in o
 
 For applications that wish to preserve part order, `multipart/form-data` content may also be modelled as an array, with one item per part, in order (where each item consists of an object containing the name/value pair); the `schema` SHALL be used to determine whether deserializing to an `object` or an `array` is preferred. See examples in [Media Type Registry: Forms](https://spec.openapis.org/registry/media-type/forms).
 
-To use the `prefixEncoding` and/or `itemEncoding` fields, either `itemSchema` or a [`schema` indicating an array type](#non-json-data) MUST be present.
+To use the `prefixEncoding` and/or `itemEncoding` fields, each item referenced by the field MUST exist in the data instance; `prefixEncoding` and `itemEncoding` entries with no corresponding item SHALL be ignored.
```
karenetheridge, line 1400: This is mostly obvious, since if there is no data to operate on, what would the encoding object even do? but it is consistent with Encoding by Name.

The part that is removed also brings consistency with Encoding by Name, which does not require that `type: object` must exist in the schema, nor even does it require that there is a `properties` entry for each individual part name. Both multipart/form-data and multipart/mixed can be successfully serialized from their original object or array representations without either a schema or an encoding object, and deep objects can be deserialized successfully as well if `Content-Type` headers are included.

karenetheridge (commented):
some extra commentary here as I keep re-reading various sections of the spec and noticing inconsistencies between them, or things that we mandate without a clear reason for doing so.

src/oas.md
baywet, line 1400:
```suggestion
To use the `prefixEncoding` and/or `itemEncoding` fields, each item referenced by the field MUST exist in the data instance; `prefixEncoding` and `itemEncoding` entries with no corresponding `items` SHALL be ignored.
```

miqui: Resolving... looks good to me

karenetheridge:
I would like to determine if this can be backported to 3.2.1. I would argue it should, as it is a simplification in both understanding and implementation, and not an imposition of new requirements.  (I am also not aware of anyone having implemented prefixEncoding/itemEncoding yet, either, so fixing this now makes everything nicer).

src/oas.md
karenetheridge: I don't think this a valid change, because we are referring to the data itself, not the keyword called `item`.

handrews (dismissed):

handrews (approved):

miqui (approved):

PR #5443, opened 2026-07-23
Clarify local validation commands in CONTRIBUTING.md

miqui:
List the individual npm scripts (validate-markdown, format-markdown, build-src, test) so contributors can pick the fast lint/link-check loop while editing versus the full build/test pass before opening a PR.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -179,16 +179,23 @@ Draft pull requests can still be reviewed while in draft state.
 > [!NOTE]
 > `npm run build-src` calls bash scripts. Use [Git Bash](https://gitforwindows.org/) on Windows, or use the Windows Subsystem for Linux (WSL).
 
-The markdown source files are converted to HTML before publishing.
-To do this locally, please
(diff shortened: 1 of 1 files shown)


handrews:
@miqui please make sure this won't conflict with #5424 which updates this documentation based on changes to the scripts due to shared build infrastructure.  It might be easier to update that part after getting the build system updated.

miqui:
@handrews good call, will check, thanks

miqui:
hi @handrews - just checked your PR and there is no conflict.  In ```package.json``` it keeps the exact same npm script names I  documented in my PR: validate-markdown, format-markdown, build-src, test. Your PR is only swapping their internal implementation from inline npx/bash calls to thin wrappers around a new @oai/build-infra CLI (oai-spec-validate-markdown, etc.).

We are good.

miqui:
@handrews hi, sorry for the ping. I was hoping you can take a look at this before the TDC call, which I am not sure I will be able to make it. Not sure yet. Thanks buddy!

handrews (approved):

PR #5446, opened 2026-07-25
v3.3: Add a global parameters field

### LINKED ISSUE #5320: v3.3: Global parameters
I'm breaking this out from discussion #5314 which covers allowing various fields at multiple levels.

This issue is to consider the implementation challenges of a global (OpenAPI Object in the entry document) `parameters` field, and ultiimately decide whether to add such a field.

TL;DR:
* `in: path` — keep the existing mandatory `required: true` and make it an error any time an `in: path` parameter does not match a path template variable; this limits the utility of global `in: path` parameters, but that's fine
* `in: querystring` — this can only be used once per operation, so setting it globally is unlikely to be useful, but we needn't forbid it; this is probably rarely used even at the Path Item level for the same reason
* `in: header` — it's already possible to specify headers at the Path Item level that don't make sense for every operation, so it doesn't matter if it's also possible at the global level; we don't worry about it now, and we don't need to start worrying about it

I can't think of any concerns with `in: query` or `in: cookie` parameters that don't already exist with the Path Item plus Operation levels.

More details and options below for those interested:

## global `in: path` parameters

Use case:  Every item path template looks like `/foo/{id}`, so there is a desire to specify the `id` path parameter globally.

Problem: Not every path template is an item template, and `in: path` parameters MUST state `required: true`

* If other path templates have an `id` path parameter with different syntax or semantics, then defining one globally is just a user error, and we need not concern ourselves with it.  
* If (as is likely), some path templates don't have an `id` path parameter at all, what happens?
    * If it is an error (retaining the `required: true` requirement), we are making a useful use case impossible.
    * If we allow `required: false` on path parameters, we could say that that means that it can be ignored if there is no matching path template variable; however this behavior is more complex and no longer just adding another level

My thought is that for 3.3, we retain `required: true`, so ending up with an `in: path` parameter that does not match a path template variable is _always an error_.  This is the simplest option, and we can add more functionality in a future release if there is strong enough demand.

## global `in: querystring` parameters

Since you can't have two of these on the same operation, regardless of where they were defined, nor can you combine them with any `in: query` parameters, this will probably not be used globally.  And probably isn't often used at the Path Item level.

This restriction mostly exists to keep request matching complexity to a manageable level.  Note that there is a weird exception in that an API Key query parameter presumably gets tacked on regardless, as we don't address it.  We should probably clarify that.

My thought for 3.3 is to not do anything special about this.  No one has complained about it AFAIK, so let's keep going.  API Keys are a SAF and by definition cause exceptions to normal rules, so it being an exception already is fine.

## global `in: header` parameters

Similarly, some request headers are broadly but not universally useful.  However, unlike path template variables, there is no `required: true` requirement.

Specifying an irrelevant header parameter is possible now, and we don't worry about it.  For example:

```yaml
/foos:
  parameters:
  - in: header
    name: Prefer
    schema:
      type: string
      pattern: ^return=(minimal|representation)
  get: {...}
  post: {...}
```

Applying this preference to the GET request doesn't really make sense, as GET by definition returns the representation.  A `return=minimal` GET request would effectively be a HEAD request.

And for that matter, as written above, the `Prefer` header parameter would apply to a HEAD operation (even though none is explicitly described), which also doesn't make any sense.

My thought for 3.3 is that we don't worry about this now, and I suggest we continue to not worry about it.  It might not make sense, but irrelevant HTTP fields are generally silently ignored, so even if someone sent this field (and note that it is not required), it is unlikely it would cause a problem.

handrews:
Fixes #5320 : Add global `parameters` field without changing any of the ways in which parameters behave at different levels.

This is the simplest expansion, and should support known use cases like universal request headers or cookies.

This feature does not interact productively with `in: querystring`, and that should be fine for now.  We can add complexity there if and when clear use cases emerge.  It is not even entirely clear how common a global `in: query` parameter (which would prevent use of `in: querystring` everywhere) would be.

Global path parameters would also be odd, but _could_ be used somewhat like Server Variables.  The oddness of them does not seem sufficient reason to forbid them, although I could be persuaded given that they would not be clearly associated with any one path template.


karenetheridge (commented):
LGTM. Perhaps add a failing test that shows that query+querystring are prohibited at the global level also (should be able to just copy an existing test)?

handrews:
The rebase was a clean rebase with no changes.
The subsequent commit addresses @karenetheridge 's request for additional schema tests.

karenetheridge (approved):

miqui (approved):

src/oas.md
```diff
@@ -103,6 +103,7 @@ In addition to the required fields, at least one of the `components`, `paths`, o
...
+| <a name="path-item-parameters"></a>parameters | [[Parameter Object](#parameter-object) \| [Reference Object](#reference-object)] | A list of parameters that are applicable for all operations in this API. These parameters can be overridden at the path item or operation levels, but cannot be removed there. The list MUST NOT include duplicated parameters. A unique parameter is defined by a combination of a [name](#parameter-name) and [location](#parameter-in). The list can use the [Reference Object](#reference-object) to link to parameters that are defined in the [OpenAPI Object's `components.parameters`](#components-parameters). |
```
mnahkies, line 106: How does this interact with definitions split across multiple files? The restriction

> cannot be removed there

Makes me worried that this will not be useful outside of very narrow use cases, though if it is scoped to the file it is defined in then it might be fine (though bring it's own complexities).

Eg: I could imagine using this for authentication related headers, but often you might have a handful of authentication related routes (for login, token refresh, password reset) that don't want those headers.

The other main use case I picture is flow control/rate limit related headers which might be ok without any scoping.

By multiple files, I mean I normally structure things like
```
openapi.yaml <- all routes exposed by the service using $ref to the other files
docs/auth.yaml <- auth related routes
docs/foo.yaml <- routes/schemas for the foo domain
...
docs/common.yaml <- cross cutting schemas like error responses
```

I'd wonder if being able to specify `(name, in, schema: null)` or similar on a route/operation level to remove the parameter would be worthwhile.

miqui: Good point.  Can we consider the overriding mechanism as a way to "remove" by nulling it out down stream?

mikekistler: I shudder at the idea of `schema: null`. But I think `schema: false` should probably do just what you want.

handrews: @mikekistler is correct.  `null` is not a valid schema but `false` is.

@mnahkies We expect this to have a fairly narrow use case, and probably more for request headers or cookies than query params (or certainly path params).

Document vs API scope is a mess, currently, and there's really no way to fix it until 4.0.  Things at the root level (except the Info Object) are all scoped at API level (`security` and `servers`), so `parameters` should probably follow that rather than create yet another usage pattern (as @baywet has noted, we have too many already).

karenetheridge: > How does this interact with definitions split across multiple files?

I'm hoping to clarify some of this a bit in 3.3. But currently I would say that global `parameters` applies in exactly the same way as global `servers`, `securitySchemes` and `tags` do. I have my own personal interpretation of how this works _today_, but this may differ from others' interpretations, which is why I hope to clarify the current state and then use that to move forward with more features to allow more flexibility in how APIs are defined with multiple documents.

mnahkies: That all makes sense. 

To be clear my concern on the narrow use case is that even for headers and cookies, it may often not be useable if there is no way to remove it from a few specific operations.

`schema: false` would work for me.

_(I can think of ways to hack around being unable to remove parameters from operations but I don't particularly like them and think it would muddy the conversation)_

handrews: @mnahkies While technically the spec does say that you can't remove a parameter at the Operation Object level (and therefore with this new addition, also at the Path Item Object level), setting `schema: false` effectively does this.  The UX isn't great because documentation tools will still show it, but a `false` schema forbids all values so even if it is there it is not usable.

karenetheridge: Overlays can remove things, can't it? so that's also an option to remove a parameter.

handrews: @karenetheridge With overlays, you'd have to remove the global and distribute it to every Path Item where you wanted it, and just leave it out of those you don't.

src/oas.md
karenetheridge, line 686:
```suggestion
| <a name="operation-parameters"></a>parameters | [[Parameter Object](#parameter-object) \| [Reference Object](#reference-object)] | A list of parameters that are applicable for this operation. If a parameter is already defined at the [root object](#openapi-object) level or in [Path Item](#path-item-parameters), the new definition will override it but can never remove it. The list MUST NOT include duplicated parameters. A unique parameter is defined by a combination of a [name](#parameter-name) and [location](#parameter-in). The list can use the [Reference Object](#reference-object) to link to parameters that are defined in the [OpenAPI Object's `components.parameters`](#components-parameters). |
```

karenetheridge, line 615:
```suggestion
| <a name="path-item-parameters"></a>parameters | [[Parameter Object](#parameter-object) \| [Reference Object](#reference-object)] | A list of parameters that are applicable for all the operations described under this path. These parameters can be overridden at the operation level, but cannot be removed there. If a parameter is already defined at the [root object](#openapi-object) level, the new definition will override it but can never remove it. The list MUST NOT include duplicated parameters. A unique parameter is defined by a combination of a [name](#parameter-name) and [location](#parameter-in). The list can use the [Reference Object](#reference-object) to link to parameters that are defined in the [OpenAPI Object's `components.parameters`](#components-parameters). |
```

karenetheridge (changes requested):

src/oas.md
karenetheridge: These two changes may need to be applied manually. I did edit the correct line in the editor, but it seems to insist upon making the change to the following line instead of the same line.  In each case, a clause was inserted into a sentence.

src/oas.md
baywet, line 106: what does it mean to have a path parameter in this collection if the Path Item path template does not have a match for it?

karenetheridge: [v3.2.0 §4.12.2.1](https://spec.openapis.org/oas/latest#common-fixed-fields): "If [in](https://spec.openapis.org/oas/latest#parameter-in) is "path", the name field MUST correspond to a single template expression occurring within the [path](https://spec.openapis.org/oas/latest#paths-path) field in the [Paths Object](https://spec.openapis.org/oas/latest#paths-object)."

baywet: Yes, what I'm worried about is that it'll be easy for people to produce invalid documents. But maybe I'm over-indexing?

PR #5447, opened 2026-07-25
v3.3: First pass at explaining SAFs

handrews:
This PR is to get conversation started on how we want to explain SAFs in the OAS.  I'm not particularly attached to it, but as we get the SIG/SAF repositories up and running, we should figure out how to explain this to readers.

### DIFF
--- a/src/oas.md
+++ b/src/oas.md
@@ -28,6 +28,25 @@ At this time, such elements are expected to remain part of the OAS until the nex
 
 Occasionally, non-backwards compatible changes may be made in `minor` versions of the OAS where impact is believed to be low relative to the benefit provided.
 
+### Standardized API Features
+
+Many API features have behavior that is difficult or impossible to infer from
+descriptions of paths, parameters, headers, or other operation elements that
+this specification supports.
+
+We refer to such features that are standardized by other standards bodies
+as Standardized API Features, or SAFs.  These features benefit from special
+handling, such as how the `Authorization` header is managed through the
+Security Scheme Object rather than the Parameter Object.  Security features
+also illustrate another aspect of SAFs, which is that they often evolve
+at a more rapid pace than the core of HTTP.
+
+To better support security features and expand into other SAFs, the OpenAPI
+Initiative is introducing companion specifications grouping independently
+versioned SAFs.  These specifications describe opt-in rules that establish
+SAF-specific Objects as well as exception rules such as those regarding
+the `Authorization` header and Security Schemes.
+
 ### Undefined and Implementation-Defined Behavior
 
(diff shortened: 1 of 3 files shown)


3huss99-ui (dismissed):

handrews:
Clean rebase force-push, no changes to commits.  This is ready for review.

miqui (approved):
looks good to me.  Perhaps a few more examples?

src/oas.md
```diff
@@ -95,6 +114,7 @@ In addition to the required fields, at least one of the `components`, `paths`, o
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
 | <a name="oas-version"></a>openapi | `string` | **REQUIRED**. This string MUST be the [version number](#versions-and-deprecation) of the OpenAPI Specification that the OpenAPI document uses. The `openapi` field SHOULD be used by tooling to interpret the OpenAPI document. This is _not_ related to the [`info.version`](#info-version) string, which describes the OpenAPI document's version. |
+| <a name="using-features"></a>usingFeatures | Map[`string`, `string`] | A map of [SAF](#standardized-api-features) names, as defined in various companion specifications, to a valid version string for the named SAF.  Implementations MUST apply the SAF's rules, such as treating header parameters that are managed by the SAF as illegal in Parameter Objects,  while processing the current document.  The default is `{"legacySecurity": "3.3.0"}`. |
```
ralfhandl, line 117: What does this default value mean, and where is that meaning described?

handrews: @ralfhandl oh, I don't think I meant to include this yet as it references something we have not decided, but the idea is:

* We should move the existing Security Scheme, OAuth Flows, and OAuth Flow Objects to the SIG-Security specification and treat them as a SAF, which I was calling `legacySecurity`
* For compatibility with OAS 3.2, if you don't explicitly set `usingFeatures` it assumes you are using the legacy security, as otherwise everyone's OAS 3.2 (or 3.1)-style security would break.
* The `3.3.0` was me thinking that since the legacy security was coming from the original OAS specification revision stream, it should keep that numbering (implicitly, the security in past OAS versions could be considered a SAF with the matching version... why we'd want to do that idk, I'm just explaining my thought process).
* The Security Requirement Object _probably_ stays in the main OAS, although I think that will become clear one way or another as we work with stuff.

I am also happy to take this part (and anything else that suggests we've made a decision on security compatibility) out of this PR.  We can add it back when we make a decision there.

ralfhandl: Thanks for the explanation, makes sense. Security evolves at a different speed than the core API description language.

I'd keep this out until it can be added with references to the new document describing security features.

PR #5464, opened 2026-08-03
v3.1: fix schema error in contentType

karenetheridge:
a contentType may be a comma-separated list of media-ranges, not just a single media-range

backport of #5283. This is a fix, not an improvement, so it should go into 3.1 as well.


handrews (approved):
Good catch!

miqui (approved):

PR #5466, opened 2026-08-05
There is no src dir on main

handrews:
This silences a noisy but otherwise harmless linkspector error showing up in the CI log for #5424 (although it is not the real error- that is being fixed in OAI/build-infra#16).

***The auto-merge of this will need to be closed, not merged!!!.***

### DIFF
--- a/.linkspector.yml
+++ b/.linkspector.yml
@@ -1,5 +1,4 @@
 files:
-  - src/oas.md
   - CONTRIBUTING.md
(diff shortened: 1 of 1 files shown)


handrews:
Because of our auto-merge strategy, this would be hard to keep out of the dev branches.  It's a harmless error, so we will leave it for now and maybe think about how to handle this another way later.

PR #5470, opened 2026-08-06
Another CI debugging PR

handrews:
Another attempt to get CI unstuck after today's outage- the original PR is apparently in a completely borked state for now, so see if this one will run CI.


handrews:
Meh, outage is still a problem, never mind.

PR #5476, opened 2026-08-08
Fix slack link, stop inspecting it

handrews:
Slack returns 403 for linkspector, it will never work.

Also, used wrong invite somehow, just re-copied this from the web site.  But the other one worked, it was just linkspector couldn't access it in CI.  This one is no different, it just changes who it says the invitation came from.


karenetheridge (approved):
disclaimer: I am doing this blindly to unlock the process

PR #5494, opened 2026-08-13
ci: negates path spec and neutralizes non-zero exit codes

baywet:
follow up to #5492 
I forgot two things:
- the negate operator to actually EXCLUDE the file
- the || echo to negate non zero exit code, which happens whens nothing is selected (from main typically)

unblocks #5493 which should also have replicated the mjs file

### DIFF
--- a/.github/workflows/sync-dev-to-vX.Y-dev.yaml
+++ b/.github/workflows/sync-dev-to-vX.Y-dev.yaml
@@ -46,6 +46,8 @@ jobs:
             git checkout -b $SYNC origin/$SYNC || git checkout -b $SYNC origin/$BASE
             git merge origin/$HEAD -m "Merge $HEAD into $SYNC"
             git checkout origin/$BASE src/
+            # exclude mjs files because we want to sync from the head
+            # neutralizes non-zero exit code because an empty selection result returns 1
             git checkout origin/$BASE -- tests/schema/ ':!*.mjs' || echo ""
             git commit -m "Restored src/ and tests/schema/" || echo ""
             git push -u origin $SYNC
--- a/.github/workflows/sync-dev-to-vX.Y-dev.yaml
+++ b/.github/workflows/sync-dev-to-vX.Y-dev.yaml
@@ -46,7 +46,7 @@ jobs:
             git checkout -b $SYNC origin/$SYNC || git checkout -b $SYNC origin/$BASE
             git merge origin/$HEAD -m "Merge $HEAD into $SYNC"
             git checkout origin/$BASE src/
-            git checkout origin/$BASE -- tests/schema/ ':*.mjs'
+            git checkout origin/$BASE -- tests/schema/ ':!*.mjs' || echo ""
             git commit -m "Restored src/ and tests/schema/" || echo ""
             git push -u origin $SYNC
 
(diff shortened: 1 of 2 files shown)


handrews:
@baywet thanks!  Can you put a comment in explaining this rather cryptic pipeline?  Pretty much what you just said here.

miqui (dismissed):

.github/workflows/sync-dev-to-vX.Y-dev.yaml
```diff
@@ -46,7 +46,7 @@ jobs:
             git checkout -b $SYNC origin/$SYNC || git checkout -b $SYNC origin/$BASE
             git merge origin/$HEAD -m "Merge $HEAD into $SYNC"
             git checkout origin/$BASE src/
-            git checkout origin/$BASE -- tests/schema/ ':*.mjs'
+            git checkout origin/$BASE -- tests/schema/ ':!*.mjs' || echo ""
```
baywet, line 51:
```suggestion
            # exclude mjs files because we want to sync from the head
            # neutralizes non-zero exit code because an empty selection result returns 1
            git checkout origin/$BASE -- tests/schema/ ':!*.mjs' || echo ""
```

.github/workflows/sync-main-to-dev.yaml
```diff
@@ -42,7 +42,7 @@ jobs:
           git checkout -b $SYNC origin/$SYNC || git checkout -b $SYNC origin/$BASE
           git merge origin/$HEAD -m "Merge $HEAD into $SYNC"
           git checkout origin/$BASE src/
-          git checkout origin/$BASE -- tests/schema/ ':*.mjs'
+          git checkout origin/$BASE -- tests/schema/ ':!*.mjs' || echo ""
```
baywet, line 47:
```suggestion
          # exclude mjs files because we want to sync from the head
          # neutralizes non-zero exit code because an empty selection result returns 1
          git checkout origin/$BASE -- tests/schema/ ':!*.mjs' || echo ""
```

.github/workflows/sync-main-to-dev.yaml
```diff
@@ -42,7 +42,9 @@ jobs:
           git checkout -b $SYNC origin/$SYNC || git checkout -b $SYNC origin/$BASE
           git merge origin/$HEAD -m "Merge $HEAD into $SYNC"
           git checkout origin/$BASE src/
-          git checkout origin/$BASE -- tests/schema/ ':*.mjs'
+          # exclude mjs files because we want to sync from the head
+          # neutralizes non-zero exit code because an empty selection result returns 1
+          git checkout origin/$BASE -- tests/schema/ ':!*.mjs' || echo ""
```
handrews, line 47: OK I'm starting to question the larger context here.  Why do we commit a merge, then add a 2nd commit extracting things?  Why not just edit before committing the merge?

Also, I'm trying to figure out what to do about `tests/schema/schema.test.mjs` which only exists on `dev` and the 3.x dev branches, and I think will have a probably have a problem with the new infrastructure.  I can't tell if we need to except it here or not.

baywet: Probably because it's more difficult to do an unattended selective merge in a single command?

baywet: Ideally we'd move this file in the infrastructure repository if we can. We have similar patterns of testing schemas in all the specification repositories.

handrews: OK we'll follow up on those (or at least the `schema.test.mjs` part) later.  This will be an improvement so I'll approve and merge and we'll see how it goes.

handrews (approved):

PR #5496, opened 2026-08-13
Remove stray duplicate schema test on the 3.1 branch.

handrews:
fail/parameter-object-path-allowReserved.yaml already exists with the same commit message at the same time.  I suspect this was a copy-paste error that got swept up in the commit.

It is only on v3.1-dev AFAICT.


karenetheridge (approved):

handrews:
Unchanged rebase force-push to avoid conflicts.

miqui (approved):

PR #5499, opened 2026-08-13
Use build-infra schema test helpers on dev branches

handrews:
This was missed in the main shared infrastructure PR because the file does not exist on `main`.  It has changes on the `v3.x-dev` branches so we might see some conflicts.

All this does is parallel the changes to `oas-schema.mjs` (currently only on `main` because of auto-merge issues separate from this issue) to configure through our build package rather than directly.  There are complex JavaScript import issues that make this necessary; it is explained in the build-infra repo.


miqui (approved):

lornajane (approved):
Looks good

PR #5528, opened 2026-08-28
OAS v3.2.1 release

handrews:
Release OAS v3.2.1.


handrews:
I think the`check-restricted-files` script might be failing because I did this from a fork rather than just a branch within this repo?  I can re-do this as a branch.

handrews:
I'm going to close this b/c of the failed check and also b/c we need to create the branch and publish it on the same day or else the release history date is wrong.  So probably better to do on the call.

PR #5529, opened 2026-08-30
v3.2: fix regex for openapi version

karenetheridge:
The "pattern" keyword is defined as following unicode semantics, where \d will match digits outside of the ascii range. But this is not likely to be treated as a valid version, so we must restrict to ascii digits.

### DIFF
--- a/src/schemas/validation/schema.yaml
+++ b/src/schemas/validation/schema.yaml
@@ -7,7 +7,7 @@ type: object
 properties:
   openapi:
     type: string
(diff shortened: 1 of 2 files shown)


miqui (approved):

ralfhandl (approved):

mikekistler (approved):
Looks good. 👍 

Had to think about this one a little bit, but since OpenAPI uses [semver](https://semver.org/) and semver only allows ASCII digits in the patch version (according to its ABNF), I think this is a proper change.

PR #5536, opened 2026-09-07
v3.3: define an explicit default for the top-level "security" object

karenetheridge:
Similar to the "servers" default, this does not change any functionality; it only adds clarity.

(I'm on the fence as to whether this should/could go into v3.2.)


miqui (dismissed):

src/oas.md
```diff
@@ -102,7 +102,8 @@ In addition to the required fields, at least one of the `components`, `paths`, o
...
-| <a name="oas-security"></a>security | [[Security Requirement Object](#security-requirement-object)] | A declaration of which security mechanisms can be used across the API. The list of values includes alternative Security Requirement Objects that can be used. Only one of the Security Requirement Objects need to be satisfied to authorize a request. Individual operations can override this definition. The list can be incomplete, up to being empty or absent. To make security explicitly optional, an empty security requirement (`{}`) can be included in the array. |
+| <a name="oas-security"></a>security | [[Security Requirement Object](#security-requirement-object)] | A declaration of which security mechanisms can be used across the API. The list of values includes alternative Security Requirement Objects that can be used. Only one of the Security Requirement Objects need to be satisfied to authorize a request. Individual operations can override this definition.
+If the `security` field is not provided, or is an empty array, the default value would be an array consisting of a single Security Requirement Object with no properties (`[{}]`), which indicates that security is optional. |
```
mikekistler: I am uncomfortable with this change, as it seems to change the meaning of an OpenAPI description that does not have a `security` field in an "Open World" interpretation of OpenAPI (which I think is how we want OpenAPI descriptions to be interpreted).

From [[Is OpenAPI an Open World or a Closed World Contract?](https://github.com/OAI/OpenAPI-Specification/issues/568#top)](https://github.com/OAI/OpenAPI-Specification/issues/568):
> In an open world we cannot make assumptions about what is not described. 

So if `security` is not provideed/described, we cannot make assumptions about it -- including about whether it is optional or required. It may very well be required and simply not described.

mikekistler (changes requested):
I have concerns about this change.

src/oas.md
karenetheridge: Isn't it already described that "no security" is the default, in the absence of any `security` specification? That is: _how would an implementation's behaviour change_?

miqui: @mikekistler good point, but we cannot make any assumptions either way.  Which is why I rely on what the spec stipulates today. 

>isn't it already described that "no security" is the default, in the absence of any security specification?

Let's discuss in the TDC

mikekistler: Where in the spec does it state that "no security" is the default? I did try to find this myself but was unsuccessful.

karenetheridge: In an "Open World" interpretation of OpenAPI, what would be the security requirements if there are no `security` keywords anywhere in the document?

miqui: 100%

lornajane:
Discussion from the meeting: not documented is different to being optional. We agree we should close this PR.

karenetheridge:
Example: an API has some sort of security requirement (that is not "allow all access"), but it is not documented as a `security` requirement. In that case, the `default` coming into effect is not accurate and would contradict the actual behaviour, forcing the API maintainer to document the security requirement.

Since we do not want to force users to document _everything_ about the API, having this default present in the schema is not desirable. A lack of `security` does not necesarily mean "allow all access", but rather "security requirements are unknown".

PR #5540, opened 2026-09-09
v3.3: define defaults for "nodeType" in the XML Object

karenetheridge:
see https://spec.openapis.org/oas/latest#xml-object

### DIFF
--- a/src/schemas/validation/meta.yaml
+++ b/src/schemas/validation/meta.yaml
@@ -21,6 +21,34 @@ properties:
     $ref: '#/$defs/external-docs'
   xml:
     $ref: '#/$defs/xml'
+if:
+  type: object
+  required:
+    - xml
+then:
+  if:
+    anyOf:
+      - required:
+          - $ref
+      - required:
+          - $dynamicRef
+      - required:
+          - type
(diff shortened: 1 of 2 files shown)


karenetheridge:
I don't see why the coverage test is failing, as as far as I can tell, schema-base.yaml is used for pass/ tests, and there are sufficient properties in tests/schema/pass/media-type-examples.yaml to cover both sides of the new conditional.

karenetheridge:
coverage test failure is fixed

miqui (approved):

src/schemas/validation/meta.yaml
```diff
@@ -21,6 +21,34 @@ properties:
     $ref: '#/$defs/external-docs'
   xml:
     $ref: '#/$defs/xml'
+if:
```
kanoru3101, line 24: Could this be simplified to avoid the nested `if`? The outer `if` looks redundant, so a single `if`/`then`/`else` should be enough here.

karenetheridge: No, because the inner 'else' must also apply to the outer 'if'.

